PR #758, opened 2017-05-30
Add bindings for using sockets

Duhemm:


muxanick:
Thanks!

nativelib/src/main/resources/posix/sys/socket.c
```diff
@@ -0,0 +1,123 @@
...
+
+struct scalanative_sockaddr {
+    char sa_data[14];
+    sa_family_t sa_family;
+};
+
+void scalanative_convert_sockaddr(struct scalanative_sockaddr *in,
+                                  struct sockaddr *out) {
+    out->sa_family = in->sa_family;
+    void *ignored = memcpy(out->sa_data, in->sa_data, 14);
```
rwhaling: I'm not sure there's a better way to do this for now, but I think this will clip off some data if sockaddr is actually a pointer to a larger struct like sockaddr_in6 or sockaddr_storage.

nativelib/src/main/scala/scala/scalanative/posix/inttypes.scala
```diff
@@ -0,0 +1,12 @@
+package scala.scalanative
+package posix
+
+import native.{CUnsignedChar, CUnsignedInt, CUnsignedShort}
+
+object inttypes {
+
+  type uint8_t  = CUnsignedChar
+  type uint16_t = CUnsignedShort
+  type uint32_t = CUnsignedInt
```
asoltysik: Is that true for a 64-bit processor?

asoltysik:
@Duhemm getaddrinfo and getnameinfo seem to be missing.

nativelib/src/main/resources/posix/sys/socket.c
Duhemm: [The posix spec for `struct sockaddr`](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/sys/socket.h.html#tag_13_61) says that `sockaddr.sa_data` has a variable length. However, [its length has been `14` since the kernel uses git (at least)](https://github.com/torvalds/linux/blob/master/include/linux/socket.h#L29-L32).

[The length also appears to be `14` in Mac OS](https://github.com/opensource-apple/xnu/blob/10.12/bsd/sys/socket.h#L591-L595), though they say that it is actually larger. I'm not sure how to read that comment, but I've been using that code on both MacOS and Linux without any issue.

Indeed once we touch a `sockaddr_storage` or `sockaddr_in6` we'll lose some data. Another solution would be to define, in Scala Native, `sockaddr` as `CStruct0`, have it completely opaque to Scala code, and define functions in C to extract information from that `CStruct0`. I think that's the way to go.

Thanks for pointing out these issues!

Duhemm:
@asoltysik These two are defined in `<netdb.h>` which I didn't implement in this PR (I don't need it for the work on the test interface at the moment).

nativelib/src/main/scala/scala/scalanative/posix/inttypes.scala
```diff
@@ -0,0 +1,12 @@
+package scala.scalanative
+package posix
+
+import native.{CUnsignedChar, CUnsignedInt, CUnsignedShort}
+
+object inttypes {
+
+  type uint8_t  = CUnsignedChar
```
densh: type uint8_t = UByte

nativelib/src/main/scala/scala/scalanative/posix/inttypes.scala
```diff
@@ -0,0 +1,12 @@
...
+  type uint16_t = CUnsignedShort
```
densh: type uint16_t = UShort

nativelib/src/main/scala/scala/scalanative/posix/inttypes.scala
densh: type uint32_t = UInt

nativelib/src/main/scala/scala/scalanative/posix/inttypes.scala
```diff
@@ -0,0 +1,12 @@
...
+  type uint32_t = CUnsignedInt
+
```
densh: `type uint64_t = ULong`

nativelib/src/main/resources/posix/sys/socket.c
densh: I agree with @rwhaling, this approach doesn't seem to work here. Lets try `CStruct0` option.

asoltysik:
@Duhemm cool, I will do bindings for them then

Duhemm:
@densh I addressed your comments, this one should be ready for review. I tested it with the test interface on Linux and macOS, both work.

@rwhaling If you have additional comments, they would be greatly appreciated!

densh:
@Duhemm This PR needs some love from `scalafmt`.

Duhemm:
@densh Done

densh:
LGTM!

PR #759, opened 2017-05-30
java.net.URI and java.util.StringTokenizer

asoltysik:
All the code and one test ported from Apache Harmony, the rest of the tests ported from Scala.js. @densh please let me know what to include in the files because of that.

### DIFF
--- a/javalib/src/main/scala/java/net/URI.scala
+++ b/javalib/src/main/scala/java/net/URI.scala
@@ -755,7 +755,7 @@ final class URI private () extends Comparable[URI] with Serializable {
     var index: Int     = first.indexOf('%', previndex)
     while (index != -1 && second.indexOf('%', previndex) == index) {
       var `match`: Boolean = first.substring(previndex, index) == second
-        .substring(previndex, index)
+          .substring(previndex, index)
       if (!`match`) {
         return false
       }
--- a/javalib/src/main/scala/java/net/URI.scala
+++ b/javalib/src/main/scala/java/net/URI.scala
@@ -8,15 +8,15 @@ import java.io.UnsupportedEncodingException
 import java.util.StringTokenizer
 
 object URI {
-  val unreserved: String = "_-!.~\'()*" 
+  val unreserved: String = "_-!.~\'()*"
 
-  val punct: String = ",;:$&+=" 
+  val punct: String = ",;:$&+="
 
-  val reserved: String = punct + "?/[]@" 
+  val reserved: String = punct + "?/[]@"
 
   val someLegal: String = unreserved + punct
 
-  val queryLegal: String = unreserved + reserved + "\\\"" 
+  val queryLegal: String = unreserved + reserved + "\\\""
 
   val allLegal: String = unreserved + reserved
 
@@ -28,7 +28,7 @@ object URI {
 
 }
 
-final class URI private() extends Comparable[URI] with Serializable {
+final class URI private () extends Comparable[URI] with Serializable {
 
   import URI._
 
@@ -792,7 +792,8 @@ final class URI private() extends Comparable[URI] with Serializable {
       }
     }
     if (uri.opaque && opaque) {
-      return equalsHexCaseInsensitive(uri.schemespecificpart, schemespecificpart)
+      return equalsHexCaseInsensitive(uri.schemespecificpart,
+                                      schemespecificpart)
     } else if (!uri.opaque && !opaque) {
       if (!equalsHexCaseInsensitive(path, uri.path)) {
         return false
--- a/javalib/src/main/scala/java/net/URI.scala
+++ b/javalib/src/main/scala/java/net/URI.scala
@@ -52,7 +52,7 @@ final class URI private() extends Comparable[URI] with Serializable {
 
   def this(str: String) = {
     this()
-    new Helper().parseURI(str, false)
+    Helper.parseURI(str, false)
   }
 
   def this(scheme: String, ssp: String, fragment: String) = {
@@ -69,7 +69,7 @@ final class URI private() extends Comparable[URI] with Serializable {
       uri.append('#')
       uri.append(quoteComponent(fragment, allLegal))
     }
-    new Helper().parseURI(uri.toString, false)
+    Helper.parseURI(uri.toString, false)
   }
 
   def this(scheme: String,
@@ -126,7 +126,7 @@ final class URI private() extends Comparable[URI] with Serializable {
         uri.append('#')
         uri.append(quoteComponent(fragment, allLegal))
       }
-      new Helper().parseURI(uri.toString, true)
+      Helper.parseURI(uri.toString, true)
     }
   }
 
@@ -162,10 +162,10 @@ final class URI private() extends Comparable[URI] with Serializable {
       uri.append('#')
       uri.append(quoteComponent(fragment, allLegal))
     }
-    new Helper().parseURI(uri.toString, false)
+    Helper.parseURI(uri.toString, false)
   }
 
-  private class Helper {
+  private object Helper {
 
     def parseURI(uri: String, forceServer: Boolean): Unit = {
       var temp: String = uri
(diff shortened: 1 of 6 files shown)


Duhemm:
Usually we add the following header to classes that we port from Apache Harmony:

```scala
// Ported from Apache Harmony
```

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
-    query
+  val queryLegal: String = unreserved + reserved + "\\\"" //$NON-NLS-1$
 
-  def getScheme(): String =
-    scheme
+  val allLegal: String = unreserved + reserved
 
-  def getUserInfo(): String =
-    userInfo
+  /**
```
Duhemm: We usually don't include javadoc as it will not be visible to anyone. If you have comments about the logic in your code, we use `// comments` to describe what's going on close the the code.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
-    extends Serializable {
+import java.io.IOException
+import java.io.Serializable
+import java.io.UnsupportedEncodingException
+import java.util.StringTokenizer
 
-  def getFragment(): String =
-    fragment
+object URI {
+  val unreserved: String = "_-!.~\'()*" //$NON-NLS-1$
```
Duhemm: These values don't appear in the Javadoc, so I think they should be made `private`

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
+    if (fragment != null) {
+      uri.append('#')
+      uri.append(quoteComponent(fragment, allLegal))
+    }
+    new Helper().parseURI(uri.toString, false)
+  }
+
+  private class Helper {
+
+    @throws(classOf[URISyntaxException])
```
Duhemm: The `@throws` annotations are only affecting the scaladoc, so you should get rid of them too.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
+      uri.append(quoteComponent(query, allLegal))
+    }
+    if (fragment != null) {
+      uri.append('#')
+      uri.append(quoteComponent(fragment, allLegal))
+    }
+    new Helper().parseURI(uri.toString, false)
+  }
+
+  private class Helper {
```
Duhemm: I think this could be an object. I doesn't keep any state.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
+
+  /**
+   * Creates a new URI instance according to the given string {@code uri}.
+   *
+   * @param uri
+   * the textual URI representation to be parsed into a URI object.
+   * @throws URISyntaxException
+   * if the given string { @code uri} doesn't fit to the
+   *                             specification RFC2396 or could not be parsed correctly.
+   */
```
Duhemm: It looks like there's an implementation missing here. Did you copy/paste the javadoc twice or did you forget a method?

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
+  private def quoteComponent(component: String, legalset: String): String = {
+    try {
+      URIEncoderDecoder.quoteIllegal(component, legalset);
+    } catch {
+      case e: UnsupportedOperationException =>
+        throw new RuntimeException(e.toString)
+    }
+  }
+
+  def compareTo(uri: URI): Int = {
```
Duhemm: I'm pretty sure there's something wrong with this method. You're probably missing some `return`s. The body of the two conditions at lines 744 and 746 won'y return `-1` or `1` if there evaluate to `true`. The value will just be discarded. I guess that's not expected. Could you refactor or fix that code?

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
+
+  /*
+   * Takes a string that may contain hex sequences like %F1 or %2b and
+   * converts the hex values following the '%' to lowercase
+   */
+
+  private def convertHexToLowerCase(s: String): String = {
+    val result: StringBuilder = new StringBuilder("")
+    if (s.indexOf('%') == -1) {
+      s
```
Duhemm: You probably want a `return` in here. Otherwise the value of `s` is just going to be ignored, and execution will continue.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -1,33 +1,1415 @@
...
+  /*
+   * Takes two strings that may contain hex sequences like %F1 or %2b and
+   * compares them, ignoring case for the hex values. Hex values must always
+   * occur in pairs as above
+   */
+
+  private def equalsHexCaseInsensitive(first: String,
+                                       second: String): Boolean = {
+    if (first.indexOf('%') != second.indexOf('%')) {
+      first == second
```
Duhemm: You're probably missing a `return` here too.

javalib/src/main/scala/java/util/StringTokenizer.scala
```diff
@@ -0,0 +1,100 @@
...
+    }
+    throw new NoSuchElementException()
+  }
+
+  def nextToken(delims: String): String = {
+    delimiters = delims
+    nextToken()
+  }
+
+}
```
Duhemm: I have an implementation of `StringTokenizer` coming from Apache Harmony as well, including tests. You may want to cherry pick it or just the tests from: https://github.com/Duhemm/scala-native/commit/922caa8e58dd8b82eb3b65809b53365bebdb886e

unit-tests/src/main/scala/java/net/URISuite.scala
```diff
@@ -0,0 +1,169 @@
+package java.net
+
+// Tests ported from Scala.js project
```
Duhemm: We've been using:

```scala
// Ported from Scala.js
```

I think it's better to stick to it.

unit-tests/src/main/scala/java/net/URISuite.scala
```diff
@@ -0,0 +1,169 @@
...
+      "1scheme://reg/", // non alpha char as 1st char
+      "asche\u00dfme:ssp", // unicode char , not USASCII
+      "asc%20heme:ssp"
+    )
+
+    for (uri <- badURIs) {
+      assertThrows[URISyntaxException] { new URI(uri) }
+    }
+  }
+
```
Duhemm: Are there tests in Apache Harmony too? Why not port them? Why are you porting `java.net.URI` from Apache Harmony and the tests from Scala.js? Why not port `java.net.URI` from Scala.js?

asoltysik: Porting URI from Scala.js was the first try, which failed because it's based on regular expressions and their regexps are not compatible with re2 (they use lookaheds which are not supported in re2). So I ported Apache Harmony's version instead. 
Yes, there are a lot of tests in Apache Harmony. I used some of them in the test "should fail on bad uris" but I can add more.  I decided to leave Scala.js tests because why not ;)

javalib/src/main/scala/java/net/URI.scala
asoltysik: Yup, when porting I forgot about the "return" difference in Java and Scala. I have corrected most of it, but I have missed some cases. I will fix them and double check.

javalib/src/main/scala/java/util/StringTokenizer.scala
asoltysik: Thanks, I will get the tests ;)

javalib/src/main/scala/java/net/URI.scala
asoltysik: It was the former, fixed in the next commit

asoltysik:
@Duhemm I think code is ready for another review pass

Issue #765, opened 2017-05-31
FileInputStream constructor should throw FileNotFoundException on invalid file names


mccartney:
A difference in behavior between JVM and Scala-Native (follows #763)

# Steps to reproduce

Try to open a file which doesn't exist:
```
  		val fis = new java.io.FileInputStream("/this/path/has/no/chance/to/exist")
  		println(fis.read)
  		fis.close
```

# Observed in Scala-Native
```java.io.IOException: couldn't read from the file```
happening during the first read

# Expected (JVM behavior):
```java.io.FileNotFoundException```
happening already on the construction call
as specified by [FileInputStream javadoc](https://docs.oracle.com/javase/8/docs/api/java/io/FileInputStream.html#FileInputStream-java.lang.String-)

mccartney:
Moreover, it seems the JVM behavior is to throw exception already on creating a new stream: `new java.io.FileInputStream("/this/path/has/no/chance/to/exist")` without even reading from it.  Scala-Native fails only on reading.

Duhemm:
Fixed in #769 / #835

PR #767, opened 2017-05-31
Adding a failing test for new FileInputStream("nonexisting") (#765)

mccartney:
A test case for #765

### DIFF
--- a/unit-tests/src/main/scala/java/io/FileInputStreamSuite.scala
+++ b/unit-tests/src/main/scala/java/io/FileInputStreamSuite.scala
@@ -62,4 +62,12 @@ object FileInputStreamSuite extends tests.Suite {
     assert(fis.read() == 0xFF)
     assert(fis.read() == -1)
   }
+
+  testFails("throws proper FileNotFoundException on invalid file names", 765) {
+    assertThrows[FileNotFoundException] { 
+      val fis = new FileInputStream("/the/path/does/not/exist/for/sure")
+      fis.read
+    }
+  }
+ 
 }


unit-tests/src/main/scala/java/io/FileInputStreamSuite.scala
```diff
@@ -62,4 +62,12 @@ object FileInputStreamSuite extends tests.Suite {
     assert(fis.read() == 0xFF)
     assert(fis.read() == -1)
   }
+
+  testFails("throws proper FileNotFoundException on invalid file names", 765) {
+    assertThrows[FileNotFoundException] { 
+      val fis = new FileInputStream("/the/path/does/not/exist/for/sure")
+      fis.read
```
Duhemm: You don't need this line. On the JVM, it will throw a `FileNotFoundException` in the constructor.

mccartney: You're right. I have removed the line and amended the test name.

unit-tests/src/main/scala/java/io/FileInputStreamSuite.scala
```diff
@@ -62,4 +62,10 @@ object FileInputStreamSuite extends tests.Suite {
     assert(fis.read() == 0xFF)
     assert(fis.read() == -1)
   }
+
+  testFails("throws FileNotFoundException already on creating new FileInputStream", 765) {
```
Duhemm: Could you change the name of the test to say that it should throw if the file doesn't exist? Currently it's not very clear why it should throw an exception by reading the name of the test.

PR #769, opened 2017-05-31
Extracting logic for opening file descriptors for read to one place

mccartney:
When opening file descriptors for reading, now:
- done in one place (companion object of `FileDescriptor`)
- honors `-1` special value returned from the native `open` call and throws `FileNotFoundException` then

Fixes #765

PR #772, opened 2017-05-31
Testing result from open call in FileWriter (fix #770)

### LINKED ISSUE #770: FileWriter ignores -1 result from open
`FileWriter` suffers from similar problems as described in #765 

The following:
```  	new java.io.FileWriter("/does/not/exist")```

- fails on JVM with `java.io.FileNotFoundException: /does/not/exist (No such file or directory)`
- succeeds in Scala-Native

mccartney:
Similar to #769, this time for FileWriter:
- when the `open` call fails and returns -1, we throw `FileNotFoundException`
- added `FileWriterSuite` to cover this


mccartney:
Thanks for the relevant comments. I will take care of it in the next few days.

densh:
@Duhemm I think we should leave that for the next PR, can you please file an issue with description of the problem that remains?

PR #773, opened 2017-05-31
Cache Coursier and Ivy artifacts downloaded by Travis

### LINKED ISSUE #757: Add caching for artifacts downloaded from llvm.org in CI
http://lists.llvm.org/pipermail/llvm-dev/2017-May/113269.html

We're partly responsible for this. There is no caching in our travis build for artifacts downloaded from `apt.llvm.org` and it only got worse recently with new matrix builds over GC implementation.

jonas:
- Install Linux packages using apt addon
 - Cache Coursier and Ivy artifacts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,3 +65,14 @@ script:
   - if [ -n "$CLANG_FORMAT_PATH" ]; then bin/clangfmt --test; fi
   - bin/scalafmt --test
   - bash sbt -J-Xmx2G rebuild "set scriptedBufferLog in sbtScalaNative := false" test-all
+
+before_cache:
+  - find $HOME/.sbt -name "*.lock" | xargs rm
+  - find $HOME/.ivy2 -name "ivydata-*.properties" | xargs rm
+  - rm -rf $HOME/.ivy2/{cache,local}/org.scala-native/
+
+cache:
+  directories:
+    - $HOME/.coursier
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot/


.travis.yml
```diff
@@ -32,6 +32,21 @@ matrix:
...
+    packages:
+    - clang++-3.7
+    - llvm-3.7
+    - llvm-3.7-dev
+    - llvm-3.7-runtime
+    - llvm-3.7-tools
+    - libgc-dev
+    - libunwind7-dev
+    - make
+
```
densh: One of the giant benefits of having unified travis setup script is that we can also share it to simplify 3-rd party builds too. Can we do caching without using `addons: apt: ...` (e.g. by using `before_cache`) ?

jonas: There's quite a few things to fix before the travis setup script can be used, such as making the LLVM version configurable. I've removed the `addons: apt` syntax, not sure if `cache: apt` will still work. It should if it caches the `.deb` files.

.travis.yml
```diff
@@ -65,3 +65,14 @@ script:
   - if [ -n "$CLANG_FORMAT_PATH" ]; then bin/clangfmt --test; fi
   - bin/scalafmt --test
   - bash sbt -J-Xmx2G rebuild "set scriptedBufferLog in sbtScalaNative := false" test-all
+
+before_cache:
+  - find $HOME/.sbt -name "*.lock" | xargs rm
+  - find $HOME/.ivy2 -name "ivydata-*.properties" | xargs rm
+
+cache:
+  apt: true
```
densh: This doesn't seem to be sufficient, it still downloads things from LLVM repo:

https://travis-ci.org/scala-native/scala-native/jobs/239922054#L242

jonas: It looks like Travis only supported APT caching of private repositories (https://github.com/travis-ci/travis-ci/issues/5876 and https://github.com/travis-ci/docs-travis-ci-com/commit/bd8b56d501dbb47f76d139831c99f55ac69d1314).

Would you accept this PR if I rephrase the message to not fix #757 and remove this line?

densh: Sure, absolutely.

.travis.yml
```diff
@@ -65,3 +65,14 @@ script:
   - if [ -n "$CLANG_FORMAT_PATH" ]; then bin/clangfmt --test; fi
   - bin/scalafmt --test
   - bash sbt -J-Xmx2G rebuild "set scriptedBufferLog in sbtScalaNative := false" test-all
+
+before_cache:
+  - find $HOME/.sbt -name "*.lock" | xargs rm
+  - find $HOME/.ivy2 -name "ivydata-*.properties" | xargs rm
```
densh: Please exclude `.ivy2/cache/org.scala-native/` and `.ivy2/local/org.scala-native/` here. Those should always be rebuild from source and never cached.

densh:
LGTM, thanks @jonas !

PR #774, opened 2017-06-01
added posix errno bindings (Fix #771)

### LINKED ISSUE #771: Posix errno bindings
There's currently support for the 3 mandatory C99 errno values, but it would be quite helpful to have all the standard POSIX errno values available as well.  If no one else is working on this, I can do the gruntwork, unless the clang Bindgen tools will provide it in the very near future.

rwhaling:
Strictly gruntwork here.  I worked from the 2016 posix spec here: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html

In terms of organization: the 3 ISO C errno values have helper functions in wrap.c, and all the rest in posix.c.  Since the posix spec does in fact include those 3 codes, however, I've included bindings to them in posix/errno.scala.  

I moved `scalanative_eexist` from wrap.c to posix.c according to the scheme described above.

Thought: providing a `Map[CInt,String]` of error numbers to names, or even a `Map[CInt,(String,String)]` to include standard descriptions, would be really useful.  Does that belong in user code, or is that worth adding now?

### DIFF
--- a/nativelib/src/main/resources/posix.c
+++ b/nativelib/src/main/resources/posix.c
@@ -11,6 +11,159 @@ extern char **environ;
 // can not expand C macros, and that's the easiest way to
 // get the values out of those in a portable manner.
 
+// Omitting EDOM EILSEQ and ERANGE since they are duplicated in wrap.c
+int scalanative_e2big() { return E2BIG; }
+
+int scalanative_eacces() { return EACCES; }
+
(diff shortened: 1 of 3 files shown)


densh:
> Thought: providing a Map[CInt,String] of error numbers to names, or even a Map[CInt,(String,String)] to include standard descriptions, would be really useful. Does that belong in user code, or is that worth adding now?

I think it's probably better left for end-user code for now.

Issue #135, opened 2016-05-26
Better templating


densh:
We currently use `project/gyb.py` script (that we borrowed from Swift) for textual template instantiation in some places. Currently it needs to be called manually to expand the templates.
- [ ] Auto-expand gyb files in sbt
- [ ] Fix absolute paths in sourcelocation comments

Alternatively we could get rid of it altogether and replace it with something better. I'd personally love to use scala.meta here but we're currently blocked by https://github.com/scalameta/scalameta/issues/295.

densh:
It would also be great to get rid of it by replacing with something less ad-hoc.

densh:
Manual gyb invocations has been working well enough for us and this is not really an issue.

PR #778, opened 2017-06-01
Use sbt-boilerplate for templating

jonas:
Convert .gyb files which nicely maps to the templates supported by
sbt-boilerplate.

References #135

---

Doesn't convert `Arrays.scala.gyb`.


densh:
Hi @jonas, thanks for trying out sbt-boilerplate on our repo! It does look more integrated for simple things but at the same time it falls apart on more advanced cases like Arrays. Having two code generation systems at once is in fact more complicated than having just one that's less integrated. I think we need to keep looking but `sbt-boilerplate` doesn't seem to be the solution.

Issue #786, opened 2017-06-05
Upgrade scalafmt to 0.6.8


densh:
As requested by @muxanick.

muxanick:
Committed to #691 PR, but I can split it to separate PR

densh:
Fixed #787

PR #787, opened 2017-06-05
Upgrade to scalafmt 0.6.8

muxanick:
Issue logged as #786


tools/src/main/scala/scala/scalanative/nir/parser/Base.scala
```diff
@@ -72,8 +72,10 @@ object Base {
   val Int: P[Int]         = neg((HexNum | DecNum).!) map (_.toInt)
   val Long: P[Long]       = neg((HexNum | DecNum).!) map (_.toLong)
   val Infinity: P[String] = P("Infinityf".!) map (_.init)
-  val Float: P[Float]     = neg(Infinity | Literals.Literals.Float.!) map (_.toFloat)
-  val Double: P[Double]   = neg(Infinity | Literals.Literals.Float.!) map (_.toDouble)
+  val Float
+    : P[Float] = neg(Infinity | Literals.Literals.Float.!) map (_.toFloat)
+  val Double
+    : P[Double] = neg(Infinity | Literals.Literals.Float.!) map (_.toDouble)
```
densh: @olafurpg That's an unexpected one. I'd expect for type annotation to stay on the same line. i.e.:
```
val Double: P[Double] = 
  neg(Infinity | Literals.Literals.Float.!) map (_.toDouble)
```

olafurpg: Hmmm 🤔 the penalty for breaking before `:` is very high, but in this case it's not allowed to break after `=` because  `map` is an infix operator. It doesn't break on `=` when the rhs body is an infix expression because I hit on funky routing/akka DSLs where it clearly wasn't supposed to break after `=`. There is a rewrite http://scalameta.org/scalafmt/#AvoidInfix that converts misuse of infix into regular application, that should fix this particular case. 

Ideally, scalafmt should take way more control over infix operators, but I never managed to get it right for corner cases.

PR #788, opened 2017-06-06
Add pointer substraction method to the .gyb file

jonas:
Commit 13afe9c2c90184ab7aec3ddff1c35f296adc5da4 only updated the
generated file.

### DIFF
--- a/javalib/src/main/scala/java/net/URI.scala
+++ b/javalib/src/main/scala/java/net/URI.scala
@@ -755,7 +755,7 @@ final class URI private () extends Comparable[URI] with Serializable {
     var index: Int     = first.indexOf('%', previndex)
     while (index != -1 && second.indexOf('%', previndex) == index) {
       var `match`: Boolean = first.substring(previndex, index) == second
-          .substring(previndex, index)
+        .substring(previndex, index)
       if (!`match`) {
         return false
       }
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/net/URI.scala
```diff
@@ -755,7 +755,7 @@ final class URI private () extends Comparable[URI] with Serializable {
     var index: Int     = first.indexOf('%', previndex)
     while (index != -1 && second.indexOf('%', previndex) == index) {
       var `match`: Boolean = first.substring(previndex, index) == second
-          .substring(previndex, index)
+        .substring(previndex, index)
```
jonas: Updated by `./bin/scalafmt` with the new version.

nativelib/src/main/scala/scala/scalanative/native/Ptr.scala.gyb
```diff
@@ -18,6 +18,9 @@ final abstract class Ptr[T] {
   /** Compute a derived pointer by subtracting given offset. */
   def -(offset: Word)(implicit tag: Tag[T]): Ptr[T] = undefined
 
+  /** Compute the offset between two pointers. */
+  def -(ptr: Ptr[T])(implicit tag: Tag[T]): CPtrDiff = undefined
+
```
densh: Oooops! Nice catch, @jonas !

PR #790, opened 2017-06-06
Try caching apt cache on travis

densh:


densh:
[It's not that simple it seems](https://travis-ci.org/scala-native/scala-native/jobs/240072764#L171).

PR #791, opened 2017-06-07
Fix #757: Switch CI to use Ubuntu Trusty and provided LLVM 3.8

### LINKED ISSUE #757: Add caching for artifacts downloaded from llvm.org in CI

jonas:
Trusty is still marked as BETA but since it contains LLVM 3.8 it allows
us to avoid fetching from apt.llvm.org. Also improves the Travis setup
script to read the LLVM version from an environment variable and use
Bash strict mode so failures bubble up.

### DIFF
--- a/bin/travis_setup.sh
+++ b/bin/travis_setup.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# Enable strict mode and fail the script on non-zero exit code,
+# unresolved variable or pipe failure.
+set -euo pipefail
+IFS=$'\n\t'
+
 if [ "$(uname)" == "Darwin" ]; then
 
     brew update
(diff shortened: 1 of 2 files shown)


jonas:
References #757

bin/travis_setup.sh
```diff
@@ -13,26 +18,20 @@ if [ "$(uname)" == "Darwin" ]; then
 
 else
 
-    # Install LLVM/Clang 3.7, Boehm GC, libunwind
-    sudo apt-get -qq update
-    sudo sh -c "echo 'deb http://apt.llvm.org/precise/ llvm-toolchain-precise-3.7 main' >> /etc/apt/sources.list"
-    sudo sh -c "echo 'deb http://apt.llvm.org/precise/ llvm-toolchain-precise main' >> /etc/apt/sources.list"
-    wget -O - http://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+    export LLVM_VERSION=${LLVM_VERSION:-3.8}
```
densh: Why do we need this export? It doesn't seem to add much value.

bin/travis_setup.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# Strict mode
+set -euo pipefail
+ORIG_IFS="$IFS"
+IFS=$'\n\t'
```
densh: Do we really need those?

jonas: The build should fail if one of the commands in this script fails. This is not currently the case and `set -e` fixes that. As for the other flags I invite you to read [this](http://redsymbol.net/articles/unofficial-bash-strict-mode/), the gist is that it hardens the script.

bin/travis_setup.sh
jonas: I can remove it, but I thought the idea was to make this script usable by external projects, such as https://github.com/scala-native/scala-native.g8.

jonas:
I can remove the `LLVM_VERSION` code, but the strict mode is important to ensure that setup failures fails the build.

densh:
@jonas Sounds good, but please leave a short comment why that extra code is there.

jonas:
One failing mac build looks unrelated.

densh:
Mac failure is unrelated indeed, LGTM!

PR #795, opened 2017-06-07
Fix #794: Fixes type of java.util.Arrays.asList

### LINKED ISSUE #794: Signature of Arrays.asList is wrong
```
import java.util.Arrays

object Test {

    println(Arrays.asList(Array[String]("asdfsd")))
    println(Arrays.asList(new Array[String](2)))
  }
}
```

This code works in plain scala & inside the javalib subproject.
But in a standalone scala native projects it can't be linked.

```
[error] cannot link: @java.util.Arrays$::asList_scala.scalanative.runtime.ObjectArray_java.util.List
[error] unable to link
[error] (sandbox/compile:nativeLinkNIR) unable to link
``` 

If the `asList` function inside javalib is modified so that the argument is `Array[String]` instead of `Array[T]` the function can be linked and the code in this use case is executed properly.

`def asList[T](a: Array[T]): List[T]`

So i assume that the Linker has a Bug regarding generic Arrays.

hewie:


densh:
Please reformat your code with `bin/scalafmt`. Otherwise this looks good to me!

densh:

Issue #761, opened 2017-05-31
Add support for java.util.ArrayList


Duhemm:
We still don't have an implementation of `ArrayList`, but it could be ported from Apache Harmony.

https://github.com/apache/harmony/blob/java6/classlib/modules/luni/src/main/java/java/util/ArrayList.java

Issue #375, opened 2016-11-08
Linker requires dummy forwarders to implement Java APIs 


densh:
This has been discovered by @soc in #373. The issue can be clearly seen in the implementation of `ArrayList`:

```
package java.util

class ArrayList[E] ... {
  ...

  override def subList(fromIndex: Int, toIndex: Int): List[E] =
    super.subList(fromIndex, toIndex)

  override def listIterator(): ListIterator[E] = super.listIterator(0)

  override def listIterator(index: Int): ListIterator[E] =
    super.listIterator(index)
}
```

The last three methods should not have been necessary.

PR #796, opened 2017-06-08
Implement java.util.Arraylist

matil019:
Implemented the essential methods using Array[AnyRef] as its underlying storage.

Methods which would only contribute to performance boost or are introduced by Java 8 are left unimplemented.

All of the code was written from scratch, and then compared with that of scala.js to tidy up.
Signed Scala CLA.

Related to: #761

### DIFF
--- a/javalib/src/main/scala/java/util/AbstractList.scala
+++ b/javalib/src/main/scala/java/util/AbstractList.scala
@@ -11,7 +11,7 @@ abstract class AbstractList[E] protected ()
     true
   }
 
-  // tests/compile:nativeLinkNIR fails without this re-declaration for some reason
+  // tests/compile:nativeLinkNIR fails without this re-declaration (issue: #375)
   // cannot link: @java.util.AbstractList::get_i32_java.lang.Object
   def get(index: Int): E
 
--- a/javalib/src/main/scala/java/util/AbstractList.scala
+++ b/javalib/src/main/scala/java/util/AbstractList.scala
@@ -11,6 +11,10 @@ abstract class AbstractList[E] protected ()
     true
   }
 
+  // tests/compile:nativeLinkNIR fails without this re-declaration for some reason
+  // cannot link: @java.util.AbstractList::get_i32_java.lang.Object
+  def get(index: Int): E
+
   def set(index: Int, element: E): E =
     throw new UnsupportedOperationException
 
--- a/javalib/src/main/scala/java/util/ArrayList.scala
+++ b/javalib/src/main/scala/java/util/ArrayList.scala
@@ -5,10 +5,8 @@ import java.io.Serializable
 // Added extra private constructors to handle all of the overloads.
 // To preserve method signatures, we cannot take ClassTag via implicit parameters.
 // We use an Array[AnyRef] for underlying storage and box/unbox AnyVals where needed as the JDK class would do.
-/**
- * @param inner The underlying array
- * @param _size The effective size of the underlying array. a.k.a. end index exclusive
- */
+// inner: The underlying array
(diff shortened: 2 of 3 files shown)


javalib/src/main/scala/java/util/AbstractList.scala
```diff
@@ -11,6 +11,10 @@ abstract class AbstractList[E] protected ()
     true
   }
 
+  // tests/compile:nativeLinkNIR fails without this re-declaration for some reason
+  // cannot link: @java.util.AbstractList::get_i32_java.lang.Object
```
densh: This probably another instance of https://github.com/scala-native/scala-native/issues/375

Please cite this issue in the comment.

javalib/src/main/scala/java/util/ArrayList.scala
```diff
@@ -0,0 +1,164 @@
...
+
+import java.io.Serializable
+
+// Added extra private constructors to handle all of the overloads.
+// To preserve method signatures, we cannot take ClassTag via implicit parameters.
+// We use an Array[AnyRef] for underlying storage and box/unbox AnyVals where needed as the JDK class would do.
+/**
+ * @param inner The underlying array
+ * @param _size The effective size of the underlying array. a.k.a. end index exclusive
+ */
```
densh: We don't do doc comments for the java library code. The only docs are the reference docs from the JDK.

javalib/src/main/scala/java/util/ArrayList.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  override def clear(): Unit = {
+    // fill the content of inner by null so that the elements can be garbage collected
+    for (i <- (0 until _size)) {
+      inner(i) = null
+    }
+    _size = 0
+  }
+
+  // define for better performance
```
densh: `// TODO: define ...`

javalib/src/main/scala/java/util/ArrayList.scala
```diff
@@ -0,0 +1,164 @@
...
+  //override def addAll(index: Index, c: Collection[_ <: E]): Boolean =
+  //override def removeRange(fromIndex: Int, toIndex: Int): Boolean =
+  //override def removeAll(c: Collection[_ <: AnyRef]): Boolean =
+  //override def retainAll(c: Collection[_ <: AnyRef]): Boolean =
+  //override def listIterator(index: Int): ListIterator[E] =
+  //override def listIterator(): ListIterator[E] =
+  //override def iterator(): Iterator[E] =
+  //override def subList(fromIndex: Int, toIndex: Int): List[E] =
+
+  // JDK 1.8
```
densh: `// TODO: JDK 1.8`

densh:
CI failure seems unrelated. We are really hitting travis time budget now, so it fails every now and then due to a timeout.

javalib/src/main/scala/java/util/ArrayList.scala
matil019: Do you mean that it should be written in `//` format, or the entire comment (including L. 5-8) be removed? I think I'm not getting your point, but the documented constructor is a glue code added by me and there is no corresponding one in JDK.

densh: Doc comments in `/** ... */` format should be avoided for java library code. Regular `//` comments are fine as long as they clarify the code.

matil019: All right. Thank you for explanation!

densh:
LGTM, congrats on first contribution @matil019 ! 🍰

Issue #667, opened 2017-04-25
java.util.HashMap.KeySet::contains doesn't work


Duhemm:
This snippets produces a segfault:

```scala
object Main {
  def main(args: Array[String]): Unit = {
    val map = new java.util.HashMap[Int, Int]
    map.put(1, 2)
    val ks = map.keySet()
    ks.contains(1)
  }
}
```

Backtrace:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x0000000100000f48 hashmap-out`Main$::main_class.ssnr.ObjectArray_unit + 152
    frame #2: 0x0000000100001021 hashmap-out`main + 97
    frame #3: 0x00007fff9c4a4235 libdyld.dylib`start + 1

```

Surprisingly, `ks.size()` works as expected.

densh:
`EXC_BAD_ACCESS` on `address=0x0` is our current equivalent of NullPointerException.

densh:
It's a bug in trait method dispatch implementation. The null here is the result of incorrect method resolution.

densh:
Fixed in #797

PR #797, opened 2017-06-09
Fix indirect trait method dispatch

### LINKED ISSUE #780: sys.env throws RuntimeException
`sys.env` throws a `java.lang.RuntimeException`.
It seems to break when trying to convert a Java Map to a Scala Map.

Here's `sys.env` from scala/src/library/scala/sys/package.scala:
```scala
def env: immutable.Map[String, String] = immutable.Map(System.getenv().asScala.toSeq: _*)
```

Recreating that:
```scala
val envMap = System.getenv()
println(envMap.get("HOME")) // works

val scalaMap = envMap.asScala
println(scalaMap.get("HOME")) // works

println(scalaMap) //java.lang.RuntimeException: Nonzero exit code: 139
println(scalaMap.toSeq) //java.lang.RuntimeException: Nonzero exit code: 139
```

The `.asScala` doesn't throw an exception, but as soon as we try and access all of its contents we get a RuntimeException.
`.asScala` appears to work as expected on Java Maps I manually construct though:

```scala
val manualMap: java.util.Map[String, String] = new java.util.HashMap[String, String]
manualMap.put("HOME", "works")
println(manualMap) //works
val res = immutable.Map(manualMap.asScala.toSeq: _*) //(replicating the sys.env treatment)
println(res) //works
```

densh:
Previously we would not discover trait method implementations in cases
when a method that implements a trait is inherited from the parent class
(see IndirectTraitReachabilitySuite for an example.)

Fixes #780, #667

### DIFF
--- a/nativelib/src/main/resources/gc/none/gc.c
+++ b/nativelib/src/main/resources/gc/none/gc.c
@@ -25,7 +25,6 @@ void scalanative_init() {
     current = mmap(NULL, CHUNK, DUMMY_GC_PROT, DUMMY_GC_FLAGS, DUMMY_GC_FD,
                    DUMMY_GC_FD_OFFSET);
     end = current + CHUNK;
-    scalanative_safepoint_init();
 }
 
 void *scalanative_alloc(void *info, size_t size) {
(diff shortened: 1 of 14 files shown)


DarkDimius:
potential testfile: https://github.com/lampepfl/dotty/blob/master/tests/run/mixins.scala

densh:
@DarkDimius Thanks! Added the test case to the PR.

PR #804, opened 2017-06-10
Port java.util.Formatter from Apache Harmony

matil019:
Port of `java.util.Formatter` was done, but several classes which it depends on had to be added as well.
The methods of those are unimplemented.

This doesn't require regex, unlike the other attempt #529.

- [x] Port tests from Harmony (some fail)
- [x] Add tests that test only `Locale.US` (some fail)
- [x] Implement classes on which `java.util.Formatter` depends (common uses, `Locale.US` only)

PR #808, opened 2017-06-11
FileOutputStream must throw FileNotFoundException if open() fails

matil019:
Discovered while porting test cases for `java.util.Formatter` from Harmony. (#804)


densh:
@Duhemm Given that we don't run that combination any longer, it's safe to merge.

PR #811, opened 2017-06-11
Fix #809

### LINKED ISSUE #809: Assigning/casting a `null` into `AnyRef` causes `##` to segfault
Demonstration:

```sandbox/Test.scala
object Test {
    println(null)                         // prints "null"
    println(null.asInstanceOf[AnyRef])    // prints "null"
    println(null.##)                      // prints "0"
    println(null.asInstanceOf[AnyRef].##) // should print "0", but segfaults

    val o = null.asInstanceOf[AnyRef]
    println(o)                            // prints "null"
    println(o.asInstanceOf[Null].##)      // prints "0"
    println(o.##)                         // should print "0", but segfaults
  }
}
```

`println` is not necesarry to reproduce this. `null.asInstanceOf[AnyRef].##` is enough.

densh:
Another method that has non-throwing semantics on the JVM for nulls.

### DIFF
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirCodeGen.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirCodeGen.scala
@@ -1449,9 +1449,17 @@ abstract class NirCodeGen
     }
 
     def genHashCode(argp: Tree, focus: Focus) = {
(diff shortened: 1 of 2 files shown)


DarkDimius:
@densh, hashcode does throw on jvm. We intercept it only in Scala code

PR #828, opened 2017-06-13
Alternative Regular Expressions for Scala-Native

muxanick:
These are alternative classes for regular expressions built on top of c++11 regex code.
It supports missing features like back tracing and requires no additional libraries on the machine. Full support for Unicode etc.
It's not replacement for re2 code but co-existing in nativelib project for mostly internal usage, but nothing stops you to use it from another project.
It allows us to simplify many algorithms like url parser, ip address parser, formatter etc. It also allows to reuse regex from scala.js because of high compatibility with JavaScript regex (I would say 99.9% compatible).
Primary goals for code were runtime performance and scala-style interfaces, including `Scope.scala` created by @densh which allows us to use memory more effective without zombi allocations.

Please check `RegexSuite.scala` for examples.


muxanick:
Travis said that I haven't signed CLA, but I did it before, I don't know why it asks again, so I signed CLA again.

muxanick:
CI breaks are not related to the change

densh:
@muxanick Thanks a lot for the suggestion but I'm afraid we can't maintain more than one implementation of regular expressions in the core. For C++11 regular expressions to be accepted they have to replace RE2 completely.

densh:
Two options we have going forward with this PR:

1. Make it a stand-alone library.

2. Make C++11-based regex supersede RE2-based ones by replacing the implementation of `java.util.regex.*`.

bin/travis_setup.sh
```diff
@@ -22,9 +22,11 @@ else
     sudo add-apt-repository --yes ppa:ubuntu-toolchain-r/test
     sudo apt-get -qq update
     sudo apt-get install -y -qq \
+      gcc-5 g++-5 \
```
densh: Adding more dependencies to the core is really hard to justify. Every single additional dependency makes it harder to setup environment on older Linux distributions.

nativelib/src/main/scala/scala/scalanative/misc/CRegex.scala
```diff
@@ -0,0 +1,106 @@
+package scala.scalanative.misc
+package regex
```
densh: We've been struggling really hard not to introduce new APIs outside in core with the exception of those necessary for interop. The reason being is that as soon as you use any of the non-standard APIs your code becomes completely unportable wrt JVM and JS targets.

nativelib/src/main/scala/scala/scalanative/misc/Scope.scala
```diff
@@ -0,0 +1,63 @@
...
+
+@annotation.implicitNotFound(msg = "Resource acquisition requires a scope.")
+trait Scope {
+
+  /** Push resource onto the resource stack. */
+  def acquire[R <: Resource](res: R): R
+
+  /** Clean up all the resources in FIFO order. */
+  def close(): Unit
+}
```
densh: These two are extremely valuable and deserve a separate PR. Given that scope/resource API is platform independent we can cross-publish it to be available on JVM and JS.

muxanick: I'm just copied it from tools, it's your file :)

densh: Sure! This one deserves it's own module which is cross-published for JVM and JS. Tools can just depend on it.

bin/travis_setup.sh
muxanick: It is only for linux plarform.
GCC 4.9.0,	April 22, 2014
I agreed with you that it's very risky.
I found that scala-native doesn't work with java 1.7, it requires to upgrade to java 1.8 which is not supported on Ubuntu 14.04 for instance (there is third party package but it's not stable, shows errors)
I wasn't expected you accept this PR :) I just can't work on separate repository until scala-native becomes more mature.

nativelib/src/main/scala/scala/scalanative/misc/Scope.scala
muxanick: I would not make this module myself, it seems not right to make a module from file which is written by other person.

densh: I mean a separate subproject in our sbt build inside of the main repo. It's perfectly fine to move stuff around as part of a PR. :)

muxanick: When you said `These two are`, which is second?

densh: Scope and Resource. Btw Resource can just be an alias to AutoCloseable.

muxanick: AutoCloseable crashes the compiler, pretty much anything from other libs is crashing the compiler

xuwei-k: > AutoCloseable crashes the compiler

https://github.com/scala-native/scala-native/pull/853 fixed?

muxanick: I will check, thanks

PR #837, opened 2017-06-15
Add support for `testOnly`

Duhemm:
It allows to run a single test with Scala Native. Currently, `testOnly` will run on the JVM. This commit fixes that.


sjrd:
What about `testQuick`?

How come you're not using the existing `Test` configuration anyway?

Duhemm:
@sjrd Thanks, I missed it.

The tasks that run tests for Scala Native are defined in the `NativeTest` configuration. We do it that way because we need to generate sources based on `test:definedTests` and `test:loadedTestFrameworks`, thus we can't define a `sourceGenerator` in the `Test` configuration that uses them without introducing a cyclic dependency.

sjrd:
Ah I see. OK.

PR #838, opened 2017-06-15
Update for release 0.3.0

ekrich:
The docker image will need to be created and pushed once the release is ready.


densh:
@ekrich The plan is to slowly phase out docker image. It's not being maintained any longer.

PR #845, opened 2017-06-16
Fixes for the test interface

Duhemm:
This fixes 2 issues with test interface:
 - A test framework may report a stack trace that has `null` as filename for elements of the stack trace, in which case the serialisation fails.
 - Tests contained inside a class were not instantiated correctly.

The second bullet was reported by @xuwei-k : https://github.com/rickynils/scalacheck/pull/337#issuecomment-308925196


densh:
LGTM, thanks for a quick fix! 👍

PR #857, opened 2017-06-19
Color console for tests/test

muxanick:
It's much convenient to find failed test if it's different color:
 
image


ekrich:
👍 Great Idea. Looks that a pretty good addition to me.

Issue #849, opened 2017-06-16
Random without a seed is deterministic


mccartney:
### Steps to reproduce
Run the following program several times:
```
     val r= new java.util.Random()
     println(r.nextInt)
```

### Observed
I keep getting the same results.
Obviously, I might be totally (un)lucky.

```1206531168``` the whole time

### Expected
Different results each time I call it as per [Random() constructor javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#Random--):
> This constructor sets the seed of the random number generator to a value very likely to be distinct from any other invocation of this constructor.

**Comment by mccartney** [comment]
Which begs for a reference to: [https://xkcd.com/221/](https://xkcd.com/221/)

**Comment by mccartney** [comment]
When I looked at it. the `stdlib.rand` yields the same results for me every time I call it:
```
println(scala.scalanative.native.stdlib.rand)
```
And there I am not fully sure how to troubleshoot further.

**Comment by ekrich** [comment]
When running this bit of code:
```
for (i <- 1 to 10) println(scala.scalanative.native.stdlib.rand)
```
I get the following:
```
16807
282475249
1622650073
984943658
1144108930
470211272
101027544
1457850878
1458777923
2007237709
```

ekrich:
Looking at the default `Random` ctor. The seed should probably be the time in Millis or something?

densh:
Keep filing those, @mccartney ! 👍

ekrich:
We should initialize srand first before calling rand. See https://www.tutorialspoint.com/c_standard_library/c_function_srand.htm

PR #868, opened 2017-06-23
Fix for #849 - Random without a seed is deterministic

ekrich:


javalib/src/main/scala/java/util/Random.scala
```diff
@@ -16,7 +17,11 @@ class Random(seed_in: Long) extends AnyRef with java.io.Serializable {
 
   setSeed(seed_in)
 
-  def this() = this(Random.randomSeed())
+  def this() = {
+    this(0) // required but don't use
```
muxanick: Why not just this? it will prevent double initialization
```
this(System.currentTimeMillis() + hashCode)
```

ekrich: Apparently the `hashCode` is not set until after the main constructor completes so the test case I added will not pass. The `hashCode` guarantees the seed is different even if `Random`s get created in the same millisecond.

muxanick: that makes sense, thanks

ekrich: I updated the code to make things more clear.

muxanick:
what if you make seen lazy and then you can use hashcode there? then you don't need this(0) anymore

ekrich:
Funny you should say this but `lazy var` are not allowed and there was no combination of call-by-name that I could make work either. That is why I made some changes. It also follows the "boy scout rule".

muxanick:
btw, this change will fix some performance issues, like `createTempDir` which takes a lot of time because it tries to create same folder every time and fails until finds new one

muxanick:
I actually thought about `lazy val` inside pimpl (Private Implementation), it will keep immutable state which helps when you do multi-thread programing and don't need mutex for setSeed

unit-tests/src/test/scala/java/util/RandomSuite.scala
```diff
@@ -223,4 +223,11 @@ object RandomSuite extends tests.Suite {
     assert(random.nextGaussian() == 0.3990565555091522)
     assert(random.nextGaussian() == 2.0051627385915154)
   }
+  test("no seed") {
```
densh: Needs a blank line above.

Issue #848, opened 2017-06-16
FileWriter fails to create a new file


mccartney:
As observed by @Duhemm in #772, when you ask the `FileWriter` to create a new file, it fails:

### Steps to reproduce 
```
       val fw = new FileWriter("/tmp/new.file")
       fw.write("something")
       fw.close()
```

mccartney:
Moreover. It's not enough to specify {{O_CREAT}} flag. As per documentation for `open`:
>  mode  specifies  the  mode to use in case a new file is created.  This argument must be supplied when O_CREAT or O_TMPFILE is specified in flags;

matil019:
I think there is already a working piece of code in `FileOutputStream`. Perhaps delegating the implementation of `FileWriter` to `FileOutputStream` is a good idea.

PR #869, opened 2017-06-24
Fix #818 and #848

### LINKED ISSUE #818: InputStreamReader/OutputStreamWriter should throw NPE if null is passed to constructor
Demonstration:

```sandbox/Test.scala
object Test {
    new java.io.InputStreamReader(null)
    // JVM:
    // java.lang.NullPointerException
    // at java.io.Reader.<init>(Reader.java:78)
    // at java.io.InputStreamReader.<init>(InputStreamReader.java:72)
    // ... 32 elided
    println(":(")

    new java.io.OutputStreamWriter(null)
    // JVM:
    // java.lang.NullPointerException
    // at java.io.Writer.<init>(Writer.java:88)
    // at java.io.OutputStreamWriter.<init>(OutputStreamWriter.java:109)
    // ... 32 elided
    println(":(")

    // Scala Native so far:
    // :(
    // :(
  }
}
```

As implied by the stacktrace on JVM, the true cause of this problem is lack of support for locking mechanism of `Reader`/`Writer` base classes.

allanrenucci:


densh:
Congrats on your first contribution @allanrenucci! 👍

PR #872, opened 2017-06-25
Fix #870: Make nativeMode work without scoping

### LINKED ISSUE #870: Performance regression in 0.3.0
Below code snippet shows 2.5 times performance regression on 0.3.0 in comparison with 0.2.0.

By the way, scala-native is 3-10 times slower than Java8. Is that intended?

build.sbt:
    scalaVersion := "2.11.11"
    nativeMode := "release"

Code:

    object Main {
        println("Hello, world!")
        println(Seq(1, 2, 3))
        var t = System.nanoTime
        var n = 0
        var k = 0
        while (true) {
          val seq = Seq(1, 2, 3)
          seq.foreach(k += _)
          n += 1
          if (System.nanoTime - t > 1000 * 1e6) {
            println(n)
            n = 0
            t = System.nanoTime
          }
        }
      }
    }

matil019:
This commit moves `nativeCompileOptions` to `scopedOptions` so that `nativeMode` is respected.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -145,19 +145,6 @@ object ScalaNativePluginInternal {
       checkThatClangIsRecentEnough(clang)
       clang
     },
-    nativeCompileOptions := {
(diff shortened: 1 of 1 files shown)


densh:
The approach taken here is reasonable quickfix but it doesn't solve the general problem. 

#880 attempts to be the definitive resolution to challenges wrt sbt scoping.

Issue #863, opened 2017-06-22
Unit tests fail due to new test harness in Travis CI


ekrich:
This failure was while compiling for Mac.
https://travis-ci.org/scala-native/scala-native/jobs/245773193
```
[info] Compiling 1 Scala source to /Users/travis/build/scala-native/scala-native/unit-tests/target/scala-2.11/test-classes...
[info] Linking (6775 ms)
[info] Discovered 3023 classes and 17652 methods
[info] Optimizing (9606 ms)
[info] Generating intermediate code (2383 ms)
[info] Produced 62 files
[info] Compiling to native code (3555 ms)
[info] Linking native code (663 ms)
[info] Starting process '/Users/travis/build/scala-native/scala-native/unit-tests/target/scala-2.11/tests-out' on port '9736'.
java.lang.Exception: Couldn't communicate with remote runner.
	at scala.scalanative.sbtplugin.testinterface.ComRunner.getSocket(ComRunner.scala:96)
	at scala.scalanative.sbtplugin.testinterface.ComRunner.<init>(ComRunner.scala:44)
	at scala.scalanative.sbtplugin.testinterface.ScalaNativeRunner.createRemoteRunner(ScalaNativeRunner.scala:49)
	at scala.scalanative.sbtplugin.testinterface.ScalaNativeRunner.<init>(ScalaNativeRunner.scala:26)
	at scala.scalanative.sbtplugin.testinterface.ScalaNativeFramework.runner(ScalaNativeFramework.scala:33)
	at sbt.Defaults$$anonfun$createTestRunners$1.apply(Defaults.scala:562)
	at sbt.Defaults$$anonfun$createTestRunners$1.apply(Defaults.scala:556)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
	at scala.collection.immutable.Map$Map1.foreach(Map.scala:109)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
	at scala.collection.AbstractTraversable.map(Traversable.scala:105)
	at sbt.Defaults$.createTestRunners(Defaults.scala:556)
	at sbt.Defaults$.allTestGroupsTask(Defaults.scala:576)
	at sbt.Defaults$$anonfun$testTasks$7.apply(Defaults.scala:439)
[...]

PR #874, opened 2017-06-27
Test interface: Start the server on the sbt side

Duhemm:
Previously, sbt would tell the distant program to start a server on a
given port. Unfortunately, this would prevent tests from running if the
randomly allocated port was already used, and couldn't recover from this
situation.

This commit moves the server part to the sbt plugin. Given a complete
implementation of `java.net`, it's easier to select a free port and
make sure that we can listen on this port. Once we're sure everything is
fine, we can start the distant program and instruct it to connect to the
right port.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
@@ -46,7 +46,10 @@ class ComRunner(bin: File,
       case _: SocketTimeoutException =>
         throw new MessageOnlyException(
           "The test program never connected to sbt.")
-    } finally serverSocket.close()
+    } finally {
+      // We can close it immediately, since we won't receive another connection.
+      serverSocket.close()
+    }
 
   private[this] val in = new DataInputStream(
     new BufferedInputStream(socket.getInputStream))
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
@@ -9,7 +9,7 @@ import sbt.{Logger, MessageOnlyException, Process}
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration.Duration
 import scala.scalanative.testinterface.serialization._
-import java.net.{ConnectException, Socket, SocketTimeoutException}
+import java.net.{ServerSocket, SocketTimeoutException}
 
 import scala.scalanative.testinterface.serialization.Log.Level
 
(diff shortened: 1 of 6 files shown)


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
```diff
@@ -34,14 +34,18 @@ class ComRunner(bin: File,
       logger.info(s"Starting process '$bin' on port '$port'.")
       Process(bin.toString +: port.toString +: args, None, envVars.toSeq: _*) ! logger
       running = false
+      socket.close()
+      serverSocket.close()
```
densh: You must always `close` in finally clause. Otherwise if exception happens, you leak the resource.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
```diff
@@ -24,8 +24,8 @@ class ComRunner(bin: File,
                 args: Seq[String],
                 logger: Logger) {
 
-  /** Port over which we communicate with the distant program */
-  val port: Int = scala.util.Random.nextInt(1000) + 9000
+  private[this] val serverSocket = new ServerSocket(0)
+  private[this] val port         = serverSocket.getLocalPort
```
densh: It's better to make this socket a var, and open it right next to close to make it clear when resource is acquired/released.

densh:
This fixes #863

Duhemm:
@densh Done!

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
```diff
@@ -24,27 +24,29 @@ class ComRunner(bin: File,
...
 
-  serverSocket.setSoTimeout(30 * 1000)
-  private[this] val socket = serverSocket.accept()
+      serverSocket.setSoTimeout(30 * 1000)
+      serverSocket.accept()
+    } catch {
+      case _: SocketTimeoutException =>
+        throw new MessageOnlyException(
+          "The test program never connected to sbt.")
+    } finally serverSocket.close()
```
densh: Finally clause is run every time, even if the exception is not thrown. Are you closing the socket immediately after the connection has been established on purpose?

densh: Server socket should probably be closed in `def close()` together with all the other resources.

Duhemm: Yes I'm closing it immediately on purpose. After someone connected, we don't need to keep the server socket open because nobody will ever connect again to this very socket, so we can close it.

I'm happy to change it if you prefer, but we don't need to keep this socket open any longer.

densh: Sounds reasonable. Lets document it with a comment that explains why we close it so soon.

PR #877, opened 2017-06-27
Fix #850: ArrayList should throw on negative capacity

### LINKED ISSUE #850: ArrayList should throw exception with negative initial capacity
A minor issue, yet a violation from Java spec.

### Steps to reproduce
```new java.util.ArrayList(-1)```

### Observed
works, creates a list with size = 0

as per [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ArrayList-int-)

densh:


densh:
I find the code easier to read this way, but that's just personal preference. Both ways it's fine as long as scalafmt allows it.

PR #880, opened 2017-06-28
Fix #870: Improve scope handling in the sbtplugin

### LINKED ISSUE #870: Performance regression in 0.3.0

densh:


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -167,48 +118,33 @@ object ScalaNativePluginInternal {
...
+    nativeWarnOldJVM := {
+      val logger = nativeLogger.value
+      Try(Class.forName("java.util.function.Function")).toOption match {
+        case None =>
+          logger.warn("Scala Native is only supported on Java 8 or newer.")
+        case Some(_) =>
+          ()
+      }
+    },
+    nativeLogger := streams.value.log,
```
sjrd: This `nativeLogger` is bullshit. You must directly use `streams.value.log` in tasks that require a logger. Otherwise `last` will not work for those tasks.

`streams` is magical: it gives you a task-specific `Logger` depending on which task requests it. Right now you are bypassing that mechanism, and all logs are associated with the `nativeLogger` task instead of the relevant tasks.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -167,48 +118,33 @@ object ScalaNativePluginInternal {
...
-              (tf,
-               new ScalaNativeFramework(f, id, logger, testBinary, envVars))
-          }
-        },
-        definedTests := (definedTests in Test).value
-      )
-
-  lazy val scopedSettings = Seq(
+  lazy val scalaNativeConfigSettings: Seq[Setting[_]] = Seq(
+    nativeWarnOldJVM := {
```
sjrd: I would keep this one in the base settings. Better, I would put it in `globalSettings` (which ends up being `in Global`).

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -125,39 +81,34 @@ object ScalaNativePluginInternal {
...
+  lazy val scalaNativeBaseSettings: Seq[Setting[_]] = Seq(
+    crossVersion := ScalaNativeCrossVersion.binary,
+    crossPlatform := NativePlatform,
     nativeClang := {
       val clang = discover("clang", clangVersions)
       checkThatClangIsRecentEnough(clang)
       clang
     },
+    nativeClang in Compile := nativeClang.value,
+    nativeClang in Test := nativeClang.value,
```
densh: It's true that those aliases are useless for `Test` and `Config` but we still need some of them for `NativeTest` to work.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
densh: Fixed.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
densh: Fixed.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
sjrd: Are you sure? `NativeTest` extends `Test`, so it should receive those for free.

densh: Yeah, I double checked just now. It's really strange.

PR #884, opened 2017-06-29
bin/travis_setup.sh now supports Debian.

### LINKED ISSUE #883: Debian support in bin/travis_setup.sh
ditto

frgomes:
Fix #883


densh:
Hi @frgomes, thanks for the suggestion! To best of my knowledge Travis doesn't support Debian. `travis_setup.sh` isn't meant to be a general-purpuose everything-included installer script, it's very much a special purpose and should stay as simple as possible.

PR #889, opened 2017-07-01
Fix #640: Implement C-like escape sequences for c"..." literals

### LINKED ISSUE #640: cstring constant are not escaped
Scala-Native (0.1.0)

The first example produce an invalid llvm output

```
[error] /home/gui/hands-on-scala-native/target/scala-2.11/ll/__const.ll:763:48: error: constant expression type mismatch
[error] @"__const:: 353358909.10" = constant [17 x i8] c"%s \\t %.2f %s/s\00"
[error]                                                ^
[error] 1 error generated.

...

[error] clang-3.9: error: no such file or directory: '/home/gui/hands-on-scala-native/target/scala-2.11/ll/__const.ll.o'
```

```scala
def mvwprintw(window: Ptr[Window], y: CInt, x: CInt, fmt: CString, args: CVararg*): CInt = extern

// fails
mvwprintw(
  window,
  line,
  1,
  c"%s \\t %.2f %s/s",
  toCString(label),
  rate,
  toCString(unit)
)

// pass
mvwprintw(
  window,
  line,
  1,
  c"foo",
  toCString(label),
  rate,
  toCString(unit)
)

// pass
mvwprintw(
  window,
  line,
  1,
  toCString("%s \\t %.2f %s/s"),
  toCString(label),
  rate,
  toCString(unit)
)
```

matil019:
Fix: #640 

This PR contains questions. Please see `TODO` in the changed file.

This PR includes two changes:
* Count bytes of CString constants when they contain escape sequences
* Format CString constants into proper LLVM IR string constants  
  e.g. `\t` -> `\09`

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Vals.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Vals.scala
@@ -3,6 +3,7 @@ package nir
 
 import java.lang.Float.floatToRawIntBits
 import java.lang.Double.doubleToRawLongBits
+import scala.annotation.tailrec
 
 sealed abstract class Val {
   final def ty: Type = this match {
@@ -36,7 +37,7 @@ sealed abstract class Val {
 
     // Subtracts from the length of the bytes for each escape sequence
     // uses String, not Seq[Byte], but should be okay since we handle ASCII only
-    def uncountEscapes(from: Int, accum: Int): Int =
+    @tailrec def uncountEscapes(from: Int, accum: Int): Int =
       s.indexOf('\\', from) match {
         case -1 => accum
         case idx if idx == s.length - 1 =>
--- a/nir/src/main/scala/scala/scalanative/nir/Vals.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Vals.scala
@@ -19,13 +19,54 @@ sealed abstract class Val {
     case Val.Double(_)          => Type.Double
     case Val.Struct(name, vals) => Type.Struct(name, vals.map(_.ty))
     case Val.Array(ty, vals)    => Type.Array(ty, vals.length)
-    case Val.Chars(s)           => Type.Array(Type.Byte, s.getBytes.length + 1)
+    case Val.Chars(s)           => Type.Array(Type.Byte, countBytes(s))
     case Val.Local(_, ty)       => ty
     case Val.Global(_, ty)      => ty
 
     case Val.Unit      => Type.Unit
     case Val.Const(_)  => Type.Ptr
     case Val.String(_) => Rt.String
+  }
+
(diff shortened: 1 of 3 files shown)


nir/src/main/scala/scala/scalanative/nir/Vals.scala
```diff
@@ -28,6 +28,47 @@ sealed abstract class Val {
     case Val.String(_) => Rt.String
   }
 
+  private def countBytes(s: String): Int = {
+    import Character.isDigit
+
+    // TODO: should be a compilation error, how to report?
+    def malformed() = ???
```
densh: `throw new IllegalArgumentException("malforrmed C string: " + string)` is the way to crash here.

nir/src/main/scala/scala/scalanative/nir/Vals.scala
```diff
@@ -28,6 +28,47 @@ sealed abstract class Val {
...
+
+    // Subtracts from the length of the bytes for each escape sequence
+    // uses String, not Seq[Byte], but should be okay since we handle ASCII only
+    def uncountEscapes(from: Int, accum: Int): Int =
```
densh: Needs `@tailrec` annotation.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -394,6 +392,56 @@ object CodeGen {
         unsupported(v)
     }
 
+    def genChars(value: String): Unit = {
+      // `value` should contain a content of a CString literal as is in its source file
+      // malformed literals are assumed absent
+      str("c\"")
+      def loop(from: Int): Unit =
```
densh: `@tailrec`.

matil019:
Where do `nir` and/or `tools` go into? `tools/src/test/scala/scala/scalanative/FrameworkTest.scala` or similar classes?

densh:
@matil019 `unit-tests` tests should work for this, as it wouldn't compile if something went wrong.

matil019:
@densh Can't get `nir` and `tools` to go into the classpath for `unit-tests`.

densh:
@matil019 The idea would be to just have a few tests that use c strings with escape sequences in the test source code. You do not need nir and/or tools to be on the classpath.

matil019:
@densh Ah! I didn't have that idea. Thanks, I'll try that.

matil019:
One point: `fromCString(c"\"")` is rejected by the Scala compiler (before compiled into LLVM) for some reason. Seeing that `s"\""` is rejected as well, I suspect it comes from the restriction of Scala's String Interpolation.

densh:
LGTM, thanks @matil019 ! 👍

PR #894, opened 2017-07-04
[WIP] Initial port of google/re2j

### LINKED ISSUE #873: Reimplement regular expressions in Scala
This will let us address two problems we currently have with RE2:

1. People find it rather hard to setup the environment in cases when RE2 is not available via system package manager.

2. RE2 supports only a subset of the regular expressions supported by JDK. 

Alternative is to use another C/C++ library but I'm afraid this is not going to address the first problem.

densh:
This is a WIP of a port of RE2 regular expression engine in Scala.

Fixes #873


javalib/src/main/scala/java/util/regex/UnicodeTables.scala
```diff
@@ -0,0 +1,5051 @@
+// AUTOGENERATED by make_unicode_tables.awk from the output of
+// go/src/pkg/unicode/maketables.go.  Yes it's awful, but frankly
+// it's quicker than porting 1300 more lines of Go.
+
+package java.util.regex
+
+import java.util.HashMap
+import java.util.Map
+
+object UnicodeTables {
```
densh: At the moment our plugin can't handle this as constructor of the module is essentially a single 5k LOC method. We've not seen methods this big before.

densh: Still blocked by this, but there is a a fix in the works now.

densh:
I'm closing this for now. We've fixed most of the re2-related issues so this is not as urgent as it used to be. We should revisit this port in the future.

PR #896, opened 2017-07-06
Fix infinite recursion in `java.util.regex.Matcher`.

atty303:
It looks like a typo.


densh:
LGTM, thanks for the fix @atty303 and congrats on first contribution! 👍

PR #898, opened 2017-07-06
[WIP] One million green threads

muxanick:
I was able to make skeleton for GTs which works on all major platforms.
It's very primitive at the moment, but it gives a sense of GTs.
I think we need to give them some fancy name like other languages did :)


densh:
Please rebase your PR on top of latest master due to CI changes, thanks!

densh:
Given recent progress by @valdisxp1 on multi-threading, I suggest we postpone this for now.

Issue #841, opened 2017-06-15
Interop: Mixing CFunctionPtr fails to accept a function argument; crashes the compiler


rcano:
This seems to work:
```scala
  val a: CFunctionPtr0[Any] = () => 4
```
This piece of code will crash the compiler:
```scala
  def t(a: => Any): Unit = {
    val b: CFunctionPtr0[Any] = () => a
  }
```
And so does this:
```scala
  def t(a: () => Any): Unit = {
    val b: CFunctionPtr0[Any] = a
  }
```

the stack trace looks like
```
scala.MatchError: native.this.CFunctionPtr.fromFunction0(a) (of class scala.reflect.internal.Trees$ApplyImplicitView)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase.genFunPtrOp(NirCodeGen.scala:1699)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase.genPrimitiveOp(NirCodeGen.scala:1154)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase.genApply(NirCodeGen.scala:1096)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase.genExpr(NirCodeGen.scala:568)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase.genExpr(NirCodeGen.scala:536)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase$$anonfun$35.apply(NirCodeGen.scala:915)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase$$anonfun$35.apply(NirCodeGen.scala:915)
        at scala.scalanative.nir.Focus$$anonfun$sequenced$1.apply(Focus.scala:91)
        at scala.scalanative.nir.Focus$$anonfun$sequenced$1.apply(Focus.scala:90)
        at scala.collection.LinearSeqOptimized$class.foldLeft(LinearSeqOptimized.scala:124)
        at scala.collection.immutable.List.foldLeft(List.scala:84)
        at scala.scalanative.nir.Focus$.sequenced(Focus.scala:90)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase.genBlock(NirCodeGen.scala:915)
        at scala.scalanative.nscplugin.NirCodeGen$NirCodePhase.genExpr(NirCodeGen.scala:621)
[...]

Issue #781, opened 2017-06-03
Compiler doesn't create a symbol for CFunctionPtr.fromFunction1


muxanick:
I spend two days on trying to fix it or find workaround, but I don't have enough experience with scala native yet. It's blocking me with many tasks. I'm calling for help.
I simplified the code and it still crashes if I put it in `nativelib` folder:
```
import scala.scalanative.native._

object CallbackFail {

    def runCallback(callId:CInt):Unit = {
        println(s"Callback #$callId")
    }

    // compiler crash: (GenICode.scala:740) function's symbol is null
    val runCallBackPtr:CFunctionPtr1[CInt, Unit] = CFunctionPtr.fromFunction1(runCallback _)
}
```

Compilation command is:
`;nativelib/clean;nativelib/compile`

Error is:
`[error] (nativelib/compile:compileIncremental) java.lang.NullPointerException`

Callstack is:
```
java.lang.NullPointerException
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.genLoadApply6$1(GenICode.scala:740)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.scala$tools$nsc$backend$icode$GenICode$ICodePhase$$genLoad(GenICode.scala:809)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:123)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:71)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:148)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:98)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:71)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:89)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:71)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:89)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.gen(GenICode.scala:67)
        at scala.tools.nsc.backend.icode.GenICode$ICodePhase.apply(GenICode.scala:63)
[...]

PR #905, opened 2017-07-13
[WIP] java.util.{Thread, ThreadGroup, ThreadLocal} ported from Harmony

Korf74:
This is a work in progress but it should be easier to discuss Threads over a PR so that I can attend to changes easier.

The PR includes commits from my two other PRs (pthreads and atomics) since I needed them (Thread is done on top of pthread).

It compiles but cannot work for now, I wasn't able to test Thread because of #841 / #781 that prevent me from using FromFunctionPtr method.

Methods should be mostly implemented with the exception of deprecated methods and several ones I was not sure / cannot implement yet

@muxanick we discussed about it a bit, I tried to use ThreadLocal to implement Thread.currentThread() but it seems not to be possible with Harmony's implementation, there are circular dependencies if I do so.

EDIT: That's a lot of commits (because of pthread and atomics) but there should be less of them in the final work

### DIFF
--- a/javalib/src/main/scala/java/lang/Thread.scala
+++ b/javalib/src/main/scala/java/lang/Thread.scala
@@ -4,11 +4,11 @@ import java.util
 import java.lang.Thread._
 
 import scala.scalanative.native.{
-CFunctionPtr,
-CFunctionPtr1,
-CInt,
-Ptr,
-stackalloc
+  CFunctionPtr,
+  CFunctionPtr1,
+  CInt,
+  Ptr,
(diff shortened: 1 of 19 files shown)


muxanick:
Great work!
it's little bit imperative style everywhere o_O,
but functionality is more important

densh:
@Korf74, can you please reopen this PR against `topic/multithreading`, it'll be a staging area for multithreading support before we're confident it's stable functionality that's ready to be included in master.

PR #754, opened 2017-05-29
support for a subset of std atomic operations

Korf74:
Preliminary work for #753 based on http://en.cppreference.com/w/cpp/atomic

PR #912, opened 2017-07-21
[WIP] Part of java.util.concurent.atomic

Korf74:
Mostly the same remarks as for #905 except for compilation that seems to work. 

I did not implement Atomic Field Updater classes, are they heavily used classes ?

This PR is mostly based on #754.

Once again that's a lot of commits, I'll most certainly create another PR with squashed commits once it's final.


densh:
@Korf74, can you please reopen this PR against `topic/multithreading`, it'll be a staging area for multithreading support before we're confident it's stable functionality that's ready to be included in master.

PR #919, opened 2017-08-07
Support for basic low level atomic operations

Korf74:
As discussed in #754 this is a PR containing some low level atomic operations. For this I created CAtomicX classes which will be the basis for java.util.concurrent.atomic.*


densh:
Builds fails due to scalafmt.

Korf74:
Seems like CI doesn't work in two of OSX builds

densh:
LGTM, thanks @Korf74 !

PR #929, opened 2017-08-12
Exposing FFI for libc's time.h

schrepfler:


asoltysik:
Anyone contributing to this project is required to sign the Scala CLA: https://www.lightbend.com/contribute/cla/scala

schrepfler:
Signed the Scala CLA.

asoltysik:
Although I am not a maintainer, I thought I can add my two cents.
If users wanted to alloc this `CStruct9` in the code, they would need to its definition. It is useful to provide a type alias for this struct:
``` scala
type ourstruct = CStruct9[Int, Int, ...]
```
so when they alloc it they can just do
``` scala
val ptrToOurStruct = stackalloc[ourstruct]
```
Also it is nice to write
``` scala
ourstruct.fieldName
```
instead of
``` scala
!ourstruct._1
``` 
we can do that by creating an implicit class. An example of both type alias and implicit class can be found here: https://github.com/scala-native/scala-native/blob/master/nativelib/src/main/scala/scala/scalanative/posix/netdb.scala
Since 0.4 we will no longer have to write these implicit classes, but it's a way to go now. Hope it helps.

schrepfler:
Don't merge this yet, I'll amend the PR with this suggestion and try to add some tests as well.

schrepfler:
I've gone over the whole time.h definitions. I've tried to follow some common sense in defining the signatures and try things out until they compiled and didn't segfault but I have a working example of only one function. Would like to go over all of them but I didn't have the time really.

ekrich:
@schrepfler You could just provide the function you have with the tests and ops object. You don't have to implement the entire API.

nativelib/src/main/scala/scala/scalanative/native/time.scala
```diff
@@ -0,0 +1,412 @@
+package scala.scalanative.native
+package native
```
muxanick: Here you define double native, I'm not sure if you want to do this.
I think it should be either:
`package scala.scalanative.native`
or
```
package scala.scalanative
package native
```

nativelib/src/main/scala/scala/scalanative/native/time.scala
```diff
@@ -0,0 +1,412 @@
...
+  /**
+    * Computes difference between two calendar times as time_t objects (time_end - time_beg) in seconds.
+    * If time_end refers to time point before time_beg then the result is negative.
+    *
+    * @param time_end End time to compare
+    * @param time_beg Beginning time to compare
+    * @return Difference between two times in seconds. On POSIX systems, time_t is measured in seconds, and difftime
+    *         is equivalent to arithmetic subtraction, but C and C++ allow fractional units for time_t.
+    */
+  def difftime(time_end: CLong, time_beg: CLong): CLong = extern
```
muxanick: should it be `time_t` instead of CLong?

nativelib/src/main/scala/scala/scalanative/native/time.scala
schrepfler: I must have extended the import somehow, maybe optimise imports in my IDE.

muxanick:
@schrepfler time.scala is mis-formatted. You might want to run `./bin/scalafmt` from the root of the repo.

schrepfler:
I've added some basic tests over the most important functions, feel free to augment/fix if needed. Please squash commits if merging.

densh:
Please rebase your PR on top of latest master due to CI changes, thanks!

nativelib/src/main/scala/scala/scalanative/native/time.scala
```diff
@@ -0,0 +1,405 @@
...
+
+@extern
+object time {
+
+  // Types
+  /** Arithmetic (until C11) Real (since C11) type capable of representing times.
+   * Although not defined by the C standard, this is almost always an integral value holding the number of seconds
+   * (not counting leap seconds) since 00:00, Jan 1 1970 UTC, corresponding to POSIX time
+   *
+   */
```
densh: We don't leave comments in the bindings for C headers.

unit-tests/src/test/scala/scala/scalanative/native/TimeSuite.scala
```diff
@@ -0,0 +1,138 @@
...
+    timePtr.tm_min = 0
+    timePtr.tm_hour = 0
+    timePtr.tm_mday = 0
+    timePtr.tm_mon = 0
+    timePtr.tm_year = 0
+    timePtr.tm_wday = 0
+    timePtr.tm_yday = 0
+    timePtr.tm_isdst = 0
+
+    // TODO there's no fromWideChart helper function yet in scala-native
```
densh: We avoid TODOs in the code. They often get stale and it's better to have everything in one central place (the issue tracker).

unit-tests/src/test/scala/scala/scalanative/native/TimeSuite.scala
```diff
@@ -0,0 +1,138 @@
...
+    anno_zero_ptr.tm_isdst = 0
+
+    val theTime: time_t = mktime(anno_zero_ptr)
+    val time_ptr = stackalloc[time_t]
+    !time_ptr = theTime
+    val time: Ptr[tm] = localtime(time_ptr)
+    val cstr: CString = asctime(time)
+    val str: String = fromCString(cstr)
+
+//    assert("Thu Jan  1 00:59:59 1970\n".equals(str))
```
densh: Please remove all commented out code.

nativelib/src/main/scala/scala/scalanative/native/time.scala
densh: Please remove all comments in this file.

schrepfler: :''(
Taking them out but I don't understand why, wouldn't that help with scaladoc tooltips ala intellisense?

densh: This is a good point. Lets leave them out for now and open a new issue for adding docs to all the bindings we ship. We need to agree upon what's the detail of documentation we want to include there. Also we should use a consistent source for the docs everywhere.

densh:
[CI fails due to mis-formatted code](https://travis-ci.org/scala-native/scala-native/jobs/308670472#L1004). A quick pass of scalafmt and it's good to go!

densh:
Thanks @schrepfler, congrats on first contribution! 🍰

PR #938, opened 2017-08-16
Port `BufferedOutputStream` from Apache Harmony

Duhemm:
We already had an implementation, but unfortunately it turns out that it
had several bugs (for instance, it would flush its whole buffer instead
of only what has been written.)

I preferred to port it from Harmony rather than trying to fix the current bugs and introduce new bugs.


densh:
@Duhemm Needs a pass of scalafmt.

PR #941, opened 2017-08-17
Deterministic linking output

densh:
After introduction of the parallel optimizer we stopped guaranteeing
deterministicly equivalent output between subsequent compilations on
the same codebase. This PR fixes this problem.


ekrich:
Does this essentially use numbering via `Fresh` to make sure everything is consistent with no collisions?

densh:
@ekrich It does following changes:

1. Make sure all output definitions are sorted by name in codegen. 
2. Makes sure that `Fresh` instance is not shared across parallel threads in the optimizer. We used to have a single global fresh instance, now we have one per method. This ensures that new ids are generated in the same order.
3. Make sure that constants are lowered in the same order.  This is not easy to guarantee in the optimizer, so it's moved to codegen instead.

nir/src/main/scala/scala/scalanative/nir/Fresh.scala
```diff
@@ -3,13 +3,31 @@ package nir
...
+      case Inst.Label(local, params) =>
+        max = Math.max(max, local.id)
+        params.foreach { param =>
+          max = Math.max(max, param.name.id)
+        }
+      case _ =>
+        ()
+    }
+    new Fresh(max)
+  }
```
Duhemm: My understanding is that:
 - We do this to be able to deterministically generate an instance of `Fresh` for a given  `Define`.
 - Because things are run in a non-deterministic order, we can't easily give a unique, deterministic number to each instance of `Fresh`.
 - This introduces a full pass over all the instructions of all `Define`s at the beginning of every pass.

If I've understood that correctly, couldn't we derive a unique, deterministic prefix from the `Define`? Or is the cost of this traversal negligible?

nir/src/main/scala/scala/scalanative/nir/Fresh.scala
```diff
@@ -3,13 +3,31 @@ package nir
...
-final class Fresh(scope: String) {
-  private var i = new AtomicInteger(0)
-  def apply() = {
-    val value = i.getAndIncrement()
-    Local(scope, value)
+final class Fresh private (private var start: Int) {
+  def apply(): Local = {
+    start += 1
+    val value = start
+    Local(value)
```
densh: At the moment we don't guarantee binary compatibility of the tools, only of the libraries.

nir/src/main/scala/scala/scalanative/nir/Fresh.scala
densh: It seems like performance overhead is negligible: 

optimizeNIR on unit-tests, before:

```

[info] Optimizing (2814 ms)
[info] Optimizing (2056 ms)
[info] Optimizing (1648 ms)
[info] Optimizing (1729 ms)
[info] Optimizing (2055 ms)
```

after:
```
[info] Optimizing (2801 ms)
[info] Optimizing (2059 ms)
[info] Optimizing (1763 ms)
[info] Optimizing (1686 ms)
[info] Optimizing (1682 ms)
```

densh: Prefixes seems to be an overdesign, we don't need more than an integer to identify a local within a method.

PR #943, opened 2017-08-18
Disable Travis CI on macOS

densh:
The builds are consistently getting timed out due to travis limits.
We really need to start looking for a replacement.


densh:
It looks like Jenkins works on Mac, we should try running that on our own hardware instead.

PR #947, opened 2017-08-19
Add implementation for java.lang.{Thread, ThreadLocal, ThreadGroup}

Korf74:
This PR contains a first implementation for Threads.

It is not yet usable because start() cannot be implemented yet but it compiles and basic tests with static methods work.


densh:
Nice! 🎉

PR #950, opened 2017-08-20
Add implementation for a subset of java.util.concurrent used in ForkJoin

Korf74:
This PR contains several classes for java.util.concurrent that are needed in ForkJoinPool/ForkJoinTask

This is ported from Harmony

This PR depends on all the previous ones


densh:
Thanks! 👍

hs_err_pid23425.log
```diff
@@ -0,0 +1,86 @@
+#
```
hexsel: included by accident, I assume?

densh: Yep, `topic/multithreading` is a staging area for our experiments on multithreading, we're going to remove this before merging to master.

PR #963, opened 2017-08-21
Cache stack trace elements

### LINKED ISSUE #920: Stack trace generation is slow
Currently we spent a lot of time on conversion from C strings to Scala Strings in fillInStackTrace. It is a performance bottleneck on applications that use exceptions for control-flow (e.g. non-local returns.)

Possible solution: we can do it lazily.

/cc @darkdimius

Duhemm:
Generating stack traces is slow because we convert C strings to Scala
strings. This commits improves the situation by caching stack trace
elements and lazily converting strings.

Fixes #920

### DIFF
--- a/javalib/src/main/scala/java/lang/StackTraceElement.scala
+++ b/javalib/src/main/scala/java/lang/StackTraceElement.scala
@@ -1,15 +1,92 @@
 package java.lang
 
-final class StackTraceElement(val getClassName: String,
-                              val getMethodName: String,
-                              val getFileName: String,
-                              val getLineNumber: Int) {
+import scala.scalanative.native.{
+  fromCString,
+  stackalloc,
+  string,
+  CChar,
+  CString,
+  CUnsignedLong,
+  Ptr
+}
+import scala.scalanative.runtime.{ByteArray, unwind}
 
-  if (getClassName == null)
-    throw new NullPointerException("Declaring class is null")
+final class StackTraceElement private (_symbol: CString) {
+  var symbol: Array[Byte] = _
 
-  if (getMethodName == null)
-    throw new NullPointerException("Method name is null")
+  if (_symbol != null) {
+    symbol = new Array[Byte](256)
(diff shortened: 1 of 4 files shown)


javalib/src/main/scala/java/lang/StackTraceElement.scala
```diff
@@ -37,21 +113,18 @@ final class StackTraceElement(val getClassName: String,
...
+    collection.mutable.HashMap.empty[CUnsignedLong, StackTraceElement]
+  private def makeStackTraceElement(
+      cursor: Ptr[scala.Byte]): StackTraceElement = {
+    val name   = stackalloc[CChar](256)
+    val offset = stackalloc[scala.Byte](8)
+
+    unwind.get_proc_name(cursor, name, 256, offset)
+    new StackTraceElement(name)
   }
+  def cached(cursor: Ptr[scala.Byte],
```
densh: `private[lang]`

javalib/src/main/scala/java/lang/StackTraceElement.scala
```diff
@@ -1,15 +1,91 @@
...
+          }
+      }
 
-  if (getMethodName == null)
-    throw new NullPointerException("Method name is null")
+    initDone = true
+    _className = className
+    _methodName = methodName
+    _fileName = null
+    _lineNumber = 0
```
densh: As soon as you complete the initialization you can set `symbol` field to null as it's never going to be used again.

javalib/src/main/scala/java/lang/StackTraceElement.scala
densh: Lets also add a comment that explains why and how are stack trace elements cached.

Duhemm:
@densh Done!

densh:
Needs some from scalafmt, otherwise LGTM!

densh:
Needs some love from scalafmt, otherwise LGTM!

Duhemm:
@densh Done!

densh:
LGTM, thanks!

PR #965, opened 2017-08-21
Allow configuring scala/scala repo location

Duhemm:
Currently we always clone the repo from https://github.com/scala/scala.
Being able to configure the location of the scala repository is useful
for when GitHub is down, or working without Internet access for
instance.

Example:

    SCALAREPO="file:///home/martin/scala/scala" sbt rebuild shell

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -349,9 +349,13 @@ lazy val scalalib =
 
         val s      = streams.value
         val trgDir = target.value / "scalaSources" / scalaVersion.value
+        val scalaRepo = sys.env
+          .get("SCALANATIVE_SCALAREPO")
+          .getOrElse("https://github.com/scala/scala.git")
 
         if (!trgDir.exists) {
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -347,19 +347,20 @@ lazy val scalalib =
       assembleScalaLibrary := {
         import org.eclipse.jgit.api._
 
-        val s      = streams.value
-        val trgDir = target.value / "scalaSources" / scalaVersion.value
+        val s         = streams.value
+        val trgDir    = target.value / "scalaSources" / scalaVersion.value
+        val scalaRepo = sys.env.get("SCALAREPO").getOrElse("https://github.com/scala/scala.git")
```
densh: We use `SCALANATIVE_` prefix for environment variables that influence our builds. That makes it a bit more explicit.

densh:
LGTM!

PR #968, opened 2017-08-22
Add javalib stubs we might support in the future

densh:


ekrich:
Where are the RMI stubs used or needed?

densh:
[Scala compiler needs those](https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/Definitions.scala#L380).

ekrich:
Super, thanks.

sjrd:
Why would you want these merged? You're going to degrade the experience for normal users. Instead of linking errors, they'll get run-time errors. You don't wanna do that.

densh:
@sjrd 

1. I expect many of those we're going to support in near future.
2. As just discussed offline, we can mark those specially (i.e. `@stub`), and don't link the by default.

PR #974, opened 2017-08-27
[WIP] Support for ForkJoin

Korf74:
This PR is a WIP since it crashes at compilation (NullPointerException), and I didn't find out why yet.

It contains a porting from [scala 2.11 ForkJoin](https://github.com/scala/scala/tree/2.11.x/src/forkjoin/scala/concurrent/forkjoin), that should permit, once working, to use scala parallel collections.

This PR depends on the previous ones.


densh:
Thanks for the initial port! 🍰

PR #976, opened 2017-08-28
Fix #866: Support scala.Enumeration

### LINKED ISSUE #866: Add support for scala.Enumeration
It uses runtime reflection under the hood to enumerate all possible values. It should be doable to make it use compile-time reflection instead.

Duhemm:
Port the phase `PrepJSInterop` from ScalaJS and remove all JS-specific
transformations and checks. All that is left is the rewriting of calls
to `scala.Enumeration.Value`.

Fixes #866


densh:
LGTM, thanks @Duhemm ! 👍

PR #982, opened 2017-08-30
Short-circuit virtual dispatch for == and ##

densh:
Previously we would do two virtual calls to perform ==/##
on reference types: one to dispatch to scala_==/scala_##
method on java.lang.Object and second one to find concrete
equals/hashCode implementation.

Instead, we short-circuit the scala_== and scala_## to immediately
pick the implementation for the equals and hashCode.

### DIFF
--- a/nativelib/src/main/scala/java/lang/Object.scala
+++ b/nativelib/src/main/scala/java/lang/Object.scala
@@ -32,11 +32,19 @@ class _Object {
   @inline def __wait(timeout: scala.Long, nanos: Int): Unit =
     runtime.getMonitor(this)._wait(timeout, nanos)
 
-  @inline def __scala_==(other: _Object): scala.Boolean =
+  @inline def __scala_==(other: _Object): scala.Boolean = {
+    // This implementation is never called as we short-circuit
+    // the virtual dispatch to go directly to the equals implementation
+    // for the reference types in the optimizer. We provide it here for clarity.
     __equals(other)
+  }
 
-  @inline def __scala_## : scala.Int =
+  @inline def __scala_## : scala.Int = {
+    // This implementation is never called as we short-circuit
+    // the virtual dispatch to go directly to the hashCode implementation
+    // for the reference types in the optimizer. We provide it here for clarity.
     __hashCode
+  }
 
   protected def __clone(): _Object = {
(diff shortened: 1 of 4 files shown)


densh:
**EDIT**: updated the PR to also short-circuit the ##.

nativelib/src/main/scala/java/lang/Object.scala
```diff
@@ -32,11 +32,19 @@ class _Object {
   @inline def __wait(timeout: scala.Long, nanos: Int): Unit =
     runtime.getMonitor(this)._wait(timeout, nanos)
 
-  @inline def __scala_==(other: _Object): scala.Boolean =
+  @inline def __scala_==(other: _Object): scala.Boolean = {
+    // This impelementation is never called as we short-circuit
```
Duhemm: typo: implementation

nativelib/src/main/scala/java/lang/Object.scala
```diff
@@ -32,11 +32,19 @@ class _Object {
...
+    // the virtual dispatch to go directly to equals for the reference
+    // types in the optimizer. We provide it here for clarity.
     __equals(other)
+  }
 
-  @inline def __scala_## : scala.Int =
+  @inline def __scala_## : scala.Int = {
+    // This impelementation is never called as we short-circuit
```
Duhemm: typo: implementation

nativelib/src/main/scala/java/lang/Object.scala
```diff
@@ -32,11 +32,19 @@ class _Object {
...
+    // the virtual dispatch to go directly to equals for the reference
```
Duhemm: typo: ~equals~ hashcode

nativelib/src/main/scala/java/lang/Object.scala
densh: Fixed.

nativelib/src/main/scala/java/lang/Object.scala
densh: Fixed.

nativelib/src/main/scala/java/lang/Object.scala
densh: Fixed.

PR #984, opened 2017-08-30
Cannot build if a parent directory is called `gc` Fixes #977

### LINKED ISSUE #977: Cannot build if a parent directory is called `gc`
Scala Native will fail to link all `.o` files if sbt is run from a directory that is named `gc` or has a directory named `gc` as parent.

In `nativeCompileLib`, we select what are the `.c` and `.cpp` files that need to be compiled using `def include(path: Path): Boolean`:
https://github.com/scala-native/scala-native/blob/7ac1b4afab79a6e01f95d1c94d9ab457a6f44fa8/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L235-L249

The issue is that we'll exclude sources that do not belong to the implementation of the gc (because the  ALL the paths contain `sep + gc + sep`. Therefore, we'll exclude almost everything from the nativelib, and we'll fail during linking.

Potential solution:
 - Set `val prefix = crossTarget.value + sep + "native" + sep + "lib"`
 - Check if the path starts with `prefix + "gc"`.
 - Do the same for the other comparisons in that function

@densh Any better idea?

ekrich:


densh:
Thanks for a quick fix! 👍

densh:
By the way this is 500-th PR closed, wooohoo! ✨ 🍰 ✨

ekrich:
Thanks, cake sounds good! 👍

PR #985, opened 2017-08-31
Fix #867: Support `scala.util.Properties`

### LINKED ISSUE #867: Add support for scala.util.Properties
The standard library version uses `getResourceAsStream` to find a properties file in `scala-library.jar` that specifies four key-value combinations. It should be easy to reimplement it without this call.

Duhemm:
`scala.util.Properties` works by reading the resource
`/library.properties` using `getResourceAsStream`, which we don't
support. To be able to use `scala.util.Properties`, we rewrite the lhs
of `lazy val scalaProps` (where the resource is read), so that we return
a pre-populated `java.util.Properties` instance if the resource to read
is `/library.properties`. The original implementation is used otherwise.

The content of `scala.util.Properties` that is returned is the same as
the one that was available when the `scalalib` was compiled.

Fixes #867

### DIFF
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/PrepNativeInterop.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/PrepNativeInterop.scala
@@ -7,11 +7,9 @@ import nsc._
 import scala.collection.immutable.ListMap
 import scala.collection.mutable.Buffer
 
-// Partially
-// Ported from ScalaJS
 /**
  * This phase does:
- * - Rewrite calls to scala.Enumeration.Value (include name string)
+ * - Rewrite calls to scala.Enumeration.Value (include name string) (Ported from ScalaJS)
  * - Rewrite the body `scala.util.PropertiesTrait.scalaProps` to
  *   avoid calls to `getResourceByStream` if the resource to read
  *   is `/library.properties`.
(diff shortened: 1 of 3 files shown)


nscplugin/src/main/scala/scala/scalanative/nscplugin/PrepNativeInterop.scala
```diff
@@ -252,6 +266,57 @@ abstract class PrepNativeInterop
...
   }
 
+  /**
+   * Rewrite the rhs of `lazy val scalaProps` in trait `PropertiesTrait` to return a pre-populated
+   * `java.util.Properties` if the resource to read is `/library.properties`.
+   * @param original  The original `DefDef`
+   * @param freshName A function that generates a fresh name
+   * @return The new (typed) rhs of the given `DefDef`.
+   */
+  private def ShortCircuitLibraryProperties(
```
densh: Why is this method capitalized? Lets make it lower-case for consistency.

nscplugin/src/main/scala/scala/scalanative/nscplugin/PrepNativeInterop.scala
```diff
@@ -5,12 +5,16 @@ import scala.tools.nsc
 import nsc._
 
 import scala.collection.immutable.ListMap
-import scala.collection.mutable
+import scala.collection.mutable.Buffer
 
+// Partially
```
densh: This comment is not very clear any longer. (i.e. what does partially mean here exactly?)

Lets put "ported from Scala.js" down to the list item `Rewrite calls to scala.Enumeration.Value`.

nscplugin/src/main/scala/scala/scalanative/nscplugin/PrepNativeInterop.scala
Duhemm: To follow the same convention as `private def ScalaEnumValName` that is defined a few lines before. Should I still change it?

densh: Lets switch to lower-case for the start of the name of the all methods in there.

Duhemm:
@densh Comments addressed!

densh:
LGTM!

PR #991, opened 2017-08-31
Unconditionally replace values of `PropertiesTrait.scalaProps`

Duhemm:
We would previously check whether the resource to be read was
"/library.properties", and return the pre-computed value iff it matched.
We get rid of the conditional and always return the pre-computed values.


densh:
Needs a pass of scalafmt.

Issue #939, opened 2017-08-17
Can't cast from Class to Ptr during nativeOptimizeNIR


jokade:
In the following example, calling `fooWithSeq( Seq(FooBar(1)) )` compiles and links fine, but calling `fooWithCArray( Seq(FooBar(1)) )` throws an exception during `nativeOptimizeNIR`:
```
scala.scalanative.util.UnsupportedException: can't cast from Class(Top(java.lang.Object)) to Ptr (class java.lang.String)
```

```scala
@extern
object lib {
  type id = Ptr[Byte]
  def foo(foos: Ptr[id], count: CInt): CInt = extern
}

object bar {
  trait Foo
  case class FooBar(i: Int) extends Foo

  implicit def fooToId(foo: Foo): id = foo.cast[id]

  // works fine when called with fooWithSeq( Seq(FooBar(1)) )
  def fooWithSeq[T<:Foo](foos: Seq[T]): CInt = Zone { implicit z =>
    val count = foos.size
    val array = stackalloc[id]( sizeof[id] * count)
    for(i<-0 until count)
      !(array + i) = foos(i)
    lib.foo(array,count)
  }

  def withCArray[T<:Foo, R](seq: Seq[T])(f: (Ptr[id],Int)=>R): R = Zone { implicit z =>
    val count = seq.size
    val array = stackalloc[id]( sizeof[id] * count)
    for(i<-0 until count)
      !(array + i) = seq(i)
    f(array,count)
  }

  // compiles when used, but exception during optimizeNIR phase
  def fooWithCArray[T<:Foo](foos: Seq[T]): CInt = withCArray(foos){ (array,count) =>
    lib.foo(array,count)
  }
}

**Comment by matil019** [comment]
I think I've found another case. The following piece of code causes the same error at `nativeOptimizeNIR`:

```scala
  private[regex] def toCStringN(str: String,
                                charset: Charset = Charset.defaultCharset())(
      implicit z: Zone): (CString, Int) = {
    val bytes = str.getBytes(charset)
    val cstr  = z.alloc(bytes.length + 1)

    var c = 0
    while (c < bytes.length) {
      !(cstr + c) = bytes(c)
      c += 1
    }

    !(cstr + c) = 0.toByte

    (cstr, bytes.length)
  }
```

Perhaps putting `Ptr` into a tuple causes the error. (FYI: `type CString = Ptr[CChar]`)

densh:
[...]

Issue #367, opened 2016-10-29
Ptr[T] must box when it escapes to generic context


tindzk:
`env` and `env2` should be semantically equivalent, but the latter leads to a segmentation fault (Linux, x86_64):

``` scala
import scalanative.native._

object Main {
  def env(s: String): Option[String] =
    stdlib.getenv(toCString(s)) match {
      case null  => None
      case value => Some(fromCString(value))
    }

  def env2(s: String): Option[String] =
    Option(stdlib.getenv(toCString(s))).map(fromCString(_))

  def main(args: Array[String]): Unit = {
    println(env("HOME"))
    println(env("test"))

    println(env2("HOME"))
    println(env2("test"))
  }
}
```

densh:
Thanks for the bug report!

densh:
Looks like another incarnation of #178. Lets keep this ticket open to fix pointer boxing, and constrain the other one to structs.

jonas:
I think I ran into this while trying to read `Ptr[_]` inside `while` condition:

```scala
// The following segfaults
var size = 0
var ptr: Ptr[CString] = ???
while (ptr(0) == null) {
  size += 1
  ptr += 1
}

// workaround:
def isDefined(ptr: Ptr[CString]): Boolean = {
  val s: CString = ptr(0)
  s != null
}

var size = 0
var ptr: Ptr[CString] = ???
while (isDefined(ptr)) {
  size += 1
  ptr += 1
}
```

densh:
@jonas Thanks for another test case here!

MasseGuillaume:
I followed this advice: https://github.com/scala-native/scala-native/issues/178#issuecomment-225882988

A workaround is to manually box/unbox ptr

for example:

```scala
  type Ifaddrs = CStruct7[
    Ptr[CInt],     // [1] ifa_next
    Ptr[CChar],    // [2] ifa_name
    CInt,          // [3] ifa_flags
    Ptr[SockAddr], // [4] ifa_addr  
    Ptr[SockAddr], // [5] ifa_netmask
    Ptr[SockAddr], // [6] ifa_ifu
    Ptr[Byte]      // [7] ifa_data
  ]

  implicit class IfaddrsOps(val ptr: Ptr[Ifaddrs]) extends AnyVal {
    def next: Ptr[Ifaddrs] = (!(ptr._1)).cast[Ptr[Ifaddrs]]
    def name: Ptr[CChar] = !(ptr._2)
[...]

PR #992, opened 2017-08-31
Fix #817: Maintain a fixed-size cache of compiled RE2 patterns to replace the finalize method

### LINKED ISSUE #817: Pattern.scala potentially leaks a native memory
There is a class Pattern:
`scala-native/javalib/src/main/scala/java/util/regex/Pattern.scala`
and there is a method:
```
override protected def finalize(): Unit = {
    super.finalize()
    cre2.delete(_regex)
  }
```
You can search `cre2.delete` over codebase, there is no other place where it called. Seems like memory for native part of regex never released.

There are 4 more places where finalize method is declared:
```
scala-native/javalib/src/main/scala/java/io/FileInputStream.scala
scala-native/javalib/src/main/scala/java/io/FileOutputStream.scala
scala-native/javalib/src/main/scala/niocharset/ISO_8859_1_And_US_ASCII_Common.scala
scala-native/javalib/src/main/scala/niocharset/UTF_8.scala
```

matil019:
Fix: #817 

The singleton object `java.util.regex.Pattern` gets a new "Store" which keeps a fixed-size, FIFO circular storage to cache compiled RE2 patterns. The Store automatically deletes older patterns when needed, eliminating the need for `Pattern.finalize`.

### DIFF
--- a/javalib/src/main/scala/java/util/regex/Pattern.scala
+++ b/javalib/src/main/scala/java/util/regex/Pattern.scala
@@ -61,12 +61,13 @@ object Pattern {
         }
     }
 
-    def release(regex: String, flags: Int): Unit = synchronized {
+    def release(regex: String, flags: Int): Int = synchronized {
       val idx = cache.view
         .indexWhere(entry =>
           entry != null && entry.regex == regex && entry.flags == flags)
       if (idx >= 0)
         cache(idx) = null
+      idx
     }
 
     def put(regex: String, flags: Int, re2: RE2RegExpOps): Unit = synchronized {
@@ -77,6 +78,14 @@ object Pattern {
       cache(next) = Entry(regex, flags, re2)
       next = if (next + 1 >= cache.size) 0 else next + 1
     }
+
+    def put(regex: String, flags: Int, re2: RE2RegExpOps, at: Int): Unit =
+      synchronized {
+        if (cache(at) == null)
+          cache(at) = Entry(regex, flags, re2)
+        else
+          put(regex, flags, re2)
+      }
 
     def doCompile(regex: String, flags: Int): RE2RegExpOps = Zone {
       implicit z =>
@@ -175,12 +184,12 @@ final class Pattern private[regex] (
     // get the compiled regex (called "re2" here) from the store's cache (or re-compiled if necessary).
     val re2 = Store.get(_pattern, _flags)
     // temporarily unmanage the re2 until the callback returns
-    Store.release(_pattern, _flags)
+    val idx = Store.release(_pattern, _flags)
     try {
       f(re2)
     } finally {
       // put back the re2 to the store so that it gets deleted when it becomes old
-      Store.put(_pattern, _flags, re2)
+      Store.put(_pattern, _flags, re2, idx)
     }
   }
 
--- a/javalib/src/main/scala/java/util/regex/Pattern.scala
+++ b/javalib/src/main/scala/java/util/regex/Pattern.scala
@@ -19,10 +19,7 @@ object Pattern {
 
   def compile(regex: String): Pattern = compile(regex, 0)
 
-  def compile(regex: String, flags: Int): Pattern =
-    compile(regex, 0, adapt = true)
-
-  def compile(regex: String, flags: Int, adapt: Boolean): Pattern = Zone {
+  def compile(regex: String, flags: Int): Pattern = Zone {
     implicit z =>
       def notSupported(flag: Int, flagName: String): Unit = {
         if ((flags & flag) == flag) {
@@ -117,10 +114,6 @@ object Pattern {
       val res = fromRE2String(quoted)
       res
     }
-
-  def adaptPatternToRe2(regex: String): String = {
-    regex
-  }
 }
 
 final class Pattern private[regex] (
--- a/javalib/src/main/scala/java/util/regex/Pattern.scala
+++ b/javalib/src/main/scala/java/util/regex/Pattern.scala
@@ -51,9 +51,10 @@ object Pattern {
 
     def get(regex: String, flags: Int): RE2RegExpOps = synchronized {
       cache.view
-        .find(entry => entry != null && entry.regex == regex && entry.flags == flags)
+        .find(entry =>
+          entry != null && entry.regex == regex && entry.flags == flags)
         .map(_.re2)
-        .getOrElse{
+        .getOrElse {
           val re2 = doCompile(regex, flags)
           put(regex, flags, re2)
           re2
@@ -62,7 +63,8 @@ object Pattern {
 
     def release(regex: String, flags: Int): Unit = synchronized {
       val idx = cache.view
-        .indexWhere(entry => entry != null && entry.regex == regex && entry.flags == flags)
+        .indexWhere(entry =>
+          entry != null && entry.regex == regex && entry.flags == flags)
       if (idx >= 0)
         cache(idx) = null
     }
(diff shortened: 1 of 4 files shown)


javalib/src/main/scala/java/util/regex/cre2h.scala
```diff
@@ -57,4 +57,8 @@ object cre2h {
   final val ERROR_BAD_UTF8           = 12
   final val ERROR_BAD_NAMED_CAPTURE  = 13
   final val ERROR_PATTERN_TOO_LARGE  = 14
+
+  // this trivial wrapper class is needed to avoid compiler crash at nativeOptimizeNIR
+  // a possible instance of #367 or #939
```
densh: It's definitely #367, #939 is a duplicate.

unit-tests/src/test/scala/java/util/regex/PatternSuite.scala
```diff
@@ -372,6 +372,18 @@ object PatternSuite extends tests.Suite {
...
+  if (runStressTests) {
+    test("compile should not leak memory") {
+      println("===== AN ENDLESS STRESS TEST INVOKED =====")
+      while (true) {
+        for {
+          pattern <- (1 to 500).map(_.toString)
+        } Pattern.compile(pattern)
+      }
+    }
+  }
```
densh: If we can't run this test in CI, we should not include it as it's likely to get out of date in long-term.

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,157 @@ object Pattern {
       res
     }
 
-  def adaptPatternToRe2(regex: String): String = {
-    regex
+  private object CompiledPatternStore {
+    private case class Entry(regex: String, flags: Int, re2: RE2RegExpOps)
+
+    private val cache =
+      scala.collection.mutable.IndexedSeq.fill[Entry](100)(null)
```
densh: Given that the memory for compiled regular expressions is never going to be freed, the overhead of storing them permanently is negligible. Lets use normal (non-fixed-size) hash map for the cache instead.

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,157 @@ object Pattern {
       res
     }
 
-  def adaptPatternToRe2(regex: String): String = {
-    regex
+  private object CompiledPatternStore {
+    private case class Entry(regex: String, flags: Int, re2: RE2RegExpOps)
```
densh: We don't expose our RE2 as a public API, so flags are always constant, no need to remember them in the cache. It can be as simple as a mapping from regex string to a compiled regex instance.

matil019: The overload `compile(String regex, int flags)` does exist in JDK. `Entry.flags` is meant to cache the second parameter (not RE2's).

javalib/src/main/scala/java/util/regex/Pattern.scala
matil019: Compiled regexes can be and actually are freed in`put` (l. 76). That said, I agree with you that a normal map is appropriate. How about using (Scala's) `LinkedHashMap` with a separate `size` field? This way, we can keep cache with a Map while limiting the memory usage by keeping `tail`s.

javalib/src/main/scala/java/util/regex/Pattern.scala
densh: Good point.

javalib/src/main/scala/java/util/regex/Pattern.scala
densh: Lets try the following then:

```
final case class Key(regex: String, flags: Int)
final class Node(var key: Key, var value: RE2RegExpOps, var rc: Int, var next: Node)

var current: Node = ???
var map: java.util.HashMap[Key, Node] = ???
```

1. Nodes form a ringbuffer of some fixed length (lets say 128).
   Every node stores a key that identifies the value it contains,
   precomplied regular expression and a reference count of current
   users of given regular expression.

2. Apart from the ringbuffer there is also a map from keys to 
   nodes that provides direct mapping to already precompiled regular expressions.

3. Whenever a key is requested we do the following:

   1. Check if it's in the map, if it is, lookup the regex through the map
      and increase the reference count for the node while it's being used. 

   2. If it's not, start traversing the ring buffer until a node with rc=0
      is found. We dispose of the old regular exression and remove the key
      from the map. Now the node can be re-used to store a new key/value pair.

   3. If while traversing we get back to the location where we started, 
      we need to grow a ringbuffer by 1 more node, and then we can do the same as ii.

sjrd: You realize that your algorithm never decrements `rc`? Obviously it can't, since you don't know when someone does not use a regex anymore.

Just use an LRU and stop worrying.

densh: @sjrd The only access to the store is through withRE2Regex so it does know when to decrement and increment.

densh: It also scales trivially to multithreaded equivalent with CAS and atomic RC in presence of different threads using different regexes at their own pace.

matil019: @densh Implemented your suggestion.

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,177 @@ object Pattern {
...
+          n
+        else
+          f(freshNode(n), num - 1)
+      val last = freshNode(null)
+      last.next = f(last, 128)
+      last
+    }
+
+    // Used to quickly look up a Node from a Key.
+    val map = scala.collection.mutable.HashMap.empty[Key, Node]
```
densh: `private`

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,177 @@ object Pattern {
...
+                     var rc: Int,
+                     var next: Node)
+
+    private def freshNode(next: Node) =
+      new Node(null, new RE2RegExpOps(null), 0, next)
+
+    import scala.annotation.tailrec
+
+    // The tip of Nodes. The Nodes form a ring buffer of some length.
+    var last: Node = {
```
densh: `private`

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,177 @@ object Pattern {
...
+        map += reused.key -> reused
+        // advance `last` so that it points to the next node (which is likely the least recently used one)
+        last = reused.next
+        reused
+      }
+    }
+
+    def withRE2Regex[A](regex: String, flags: Int)(f: RE2RegExpOps => A): A = {
+      // increase the refcount of the selected node while in use to prevent it from deleted
+      val node = synchronized {
```
densh: `synchronized` here is probably not the best way to make this thread-safe. It's safe to assume single-threaded-only for now as we don't support threads yet.

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,177 @@ object Pattern {
...
+    final case class Key(regex: String, flags: Int)
+    final class Node(var key: Key,
+                     var value: RE2RegExpOps,
+                     var rc: Int,
+                     var next: Node)
+
+    private def freshNode(next: Node) =
+      new Node(null, new RE2RegExpOps(null), 0, next)
+
+    import scala.annotation.tailrec
```
densh: This import should go to the top of the file.

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,177 @@ object Pattern {
...
+      val node = synchronized {
+        val n = selectNode(regex, flags)
+        n.rc += 1
+        n
+      }
+      try f(node.value)
+      finally node.rc -= 1
+    }
+
+    def doCompile(regex: String, flags: Int): RE2RegExpOps = Zone {
```
densh: Is this used anywhere outside this file? If not lets make it `private`.

javalib/src/main/scala/java/util/regex/Pattern.scala
```diff
@@ -118,18 +41,177 @@ object Pattern {
...
+            newnode
+          } else if (n.rc <= 0)
+            n
+          else
+            findUnused(n.next)
+        val reused =
+          if (last.rc <= 0)
+            last
+          else
+            findUnused(last.next)
```
densh: It's good to keep curly braces if `if` is multi-line, here and above.

densh:
LGTM, nicely done @matil019 !

PR #993, opened 2017-08-31
Build fails if there are no sources in src/main/scala Fixes #911

### LINKED ISSUE #911: Build fails if there are no sources in src/main/scala
If you have a pure testing project with source only in `src/test/scala` then the linking phase fails. Work around is to just add a source file in `src/main/scala`.
```
[debug] running
[debug] /usr/bin/clang
[debug] 	-S
[debug] 	-xc
[debug] 	-emit-llvm
[debug] 	-o
[debug] 	/Users/eric/workspace/scala-native-ml/unit-tests/target/scala-2.11/native/target/ll.probe
[debug] 	/Users/eric/workspace/scala-native-ml/unit-tests/target/scala-2.11/native/target/c.probe
[trace] Stack trace suppressed: run last tests/nativetest:nativeLinkNIR for the full output.
[error] (tests/nativetest:nativeLinkNIR) java.lang.UnsupportedOperationException: Neither a jar, nor a directory: /Users/eric/workspace/scala-native-ml/unit-tests/target/scala-2.11/classes
[error] Total time: 5 s, completed Jul 19, 2017 9:58:44 AM
```

ekrich:
**sbt** automatically adds `target/scala-2.11/classes` to the classpath so this fix filters out classpath files that don't exist like when source is empty. Otherwise, the classpath goes into `Config` which goes to `Linker` and ends up in `VirtualDirectory` where it errors. Then the link fails and thus any tests. There is an `EmptyDirectory` but it is not used in this case when constructing a `VirtualDirectory`. That could be an alternative perhaps but it is a much more intrusive change.

The second change is removing a deprecation.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -179,7 +179,7 @@ object ScalaNativePluginInternal {
       val mainClass = selectMainClass.value.getOrElse(
         throw new MessageOnlyException("No main class detected.")
       )
-      val classpath = fullClasspath.value.map(_.data)
+      val classpath = fullClasspath.value.map(_.data).filter(_.exists)
       val entry     = nir.Global.Top(mainClass.toString + "$")
       val cwd       = nativeWorkdir.value
 
@@ -424,7 +424,7 @@ object ScalaNativePluginInternal {
         if (exitCode == 0) None
         else Some("Nonzero exit code: " + exitCode)
 
-      Defaults.toError(message)
+      message.foreach(sys.error)
     },
     nativeMissingDependencies := {
       (nativeExternalDependencies.value.toSet --


sjrd:
FWIW Scala.js has [exactly the same test](https://github.com/scala-js/scala-js/blob/19742d5e2b720934c75ec5089bb3c5260d8ed9e9/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L261), added in https://github.com/scala-js/scala-js/commit/0b2828e9c166bd60f1d9314efaa44869bd56c5eb to fix https://github.com/scala-js/scala-js/issues/2198.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -179,7 +179,7 @@ object ScalaNativePluginInternal {
       val mainClass = selectMainClass.value.getOrElse(
         throw new MessageOnlyException("No main class detected.")
       )
-      val classpath = fullClasspath.value.map(_.data)
```
densh: Nice and easy fix! 👍

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -424,7 +424,7 @@ object ScalaNativePluginInternal {
         if (exitCode == 0) None
         else Some("Nonzero exit code: " + exitCode)
 
-      Defaults.toError(message)
```
densh: Why this change?

Duhemm: That's to fix the deprecation warning:
https://github.com/sbt/sbt-zero-thirteen/blob/3d50ca2f6ab531ac0c5ca23b1ff2b40cda84300a/main/src/main/scala/sbt/Defaults.scala#L2200-L2201

densh: Sounds good then!

densh:
LGTM, thanks @ekrich ! 👍

ekrich:
Thanks for the quick review, @duhemm and @densh :-)

PR #995, opened 2017-09-01
Fix #979: Caching of stack trace elements on MacOS

### LINKED ISSUE #979: Test `java.lang.StackTraceElementSuite.getMethodName` fails on MacOS
We didn't catch it until now because the CI is disabled for MacOS.

This is the failing assertion:
https://github.com/scala-native/scala-native/blob/dd9aa215368338e76e099bb24599a061f56c120a/unit-tests/src/test/scala/java/lang/StackTraceElementSuite.scala#L29

We have `dummy2.getMethodName == "dummy1"`.

This was certainly caused by #963.

Duhemm:
To avoid re-computing all stack trace elements for every exception, we
cache the stack traces. Previously, we were using the address of the
first instruction of a function as a key to retrieve the corresponding
stack trace element. It turns out, however, that retrieving this
information doesn't work as expected on MacOS, which was the cause for
issue #979.

Fortunately, we can also use the current instruction pointer inside each
stack frame as key to get the same effect. This information can be
queried using libunwind and gives consistent results on MacOS and on
Linux.

Fixes #979

### DIFF
--- a/javalib/src/main/scala/java/lang/Throwables.scala
+++ b/javalib/src/main/scala/java/lang/Throwables.scala
@@ -29,14 +29,14 @@ class Throwable(s: String, private var e: Throwable)
   def fillInStackTrace(): Throwable = {
     val cursor  = stackalloc[scala.Byte](2048)
     val context = stackalloc[scala.Byte](2048)
-    val startIp = stackalloc[CUnsignedLong](1)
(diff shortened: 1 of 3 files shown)


Duhemm:
Unfortunately we don't have CI for MacOS at the moment, but it works on my machine :)

PR #1000, opened 2017-09-04
Cache primitive numeric boxes

densh:
JavaDoc for valueOf specifies that primitive numeric types must
be boxed in ranges of -128 to 127 for signed primitives and up
to at least 127 for characters.

Review @sjrd

### DIFF
--- a/javalib/src/main/scala/java/lang/Byte.scala
+++ b/javalib/src/main/scala/java/lang/Byte.scala
@@ -225,12 +225,27 @@ object Byte {
   @inline def toUnsignedLong(x: scala.Byte): scala.Long =
     byteToULong(x)
 
-  @inline def valueOf(byteValue: scala.Byte): Byte =
-    new Byte(byteValue)
+  import ByteCache.cache
+
+  @inline def valueOf(byteValue: scala.Byte): Byte = {
+    val idx    = byteValue - MIN_VALUE
+    val cached = cache(idx)
+    if (cached != null) {
+      cached
+    } else {
+      val newbyte = new Byte(byteValue)
+      cache(idx) = newbyte
+      newbyte
+    }
+  }
 
   @inline def valueOf(s: String): Byte =
     valueOf(parseByte(s))
 
   @inline def valueOf(s: String, radix: scala.Int): Byte =
     valueOf(parseByte(s, radix))
+}
+
+private[lang] object ByteCache {
+  private[lang] val cache = new Array[java.lang.Byte](256)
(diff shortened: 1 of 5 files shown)


javalib/src/main/scala/java/lang/Byte.scala
```diff
@@ -225,12 +225,27 @@ object Byte {
...
 
   @inline def valueOf(s: String): Byte =
     valueOf(parseByte(s))
 
   @inline def valueOf(s: String, radix: scala.Int): Byte =
     valueOf(parseByte(s, radix))
 }
+
+private[lang] object ByteCache {
+  private[lang] val cache = new Array[java.lang.Byte](256)
```
sjrd: I suggest you fill in that cache immediately, so that you avoid the branching in `valueOf` altogether.

Same for the other caches, except of course for `Short`, `Int` and `Long` you still have the branch for whether it's small enough.

javalib/src/main/scala/java/lang/Character.scala
```diff
@@ -1111,7 +1111,23 @@ object Character {
 
   def hashCode(value: scala.Char): scala.Int = value
 
-  def valueOf(charValue: scala.Char): Character = new Character(charValue)
+  import CharacterCache.cache
+
+  def valueOf(charValue: scala.Char): Character = {
+    if (charValue >= 127) {
```
sjrd: `>`

javalib/src/main/scala/java/lang/Integer.scala
```diff
@@ -564,8 +564,23 @@ object Integer {
   @inline def toUnsignedLong(x: scala.Int): scala.Long =
     intToULong(x)
 
-  @inline def valueOf(i: scala.Int): Integer =
-    new Integer(i)
+  import IntegerCache.cache
+
+  @inline def valueOf(intValue: scala.Int): Integer = {
+    if (intValue < -128 || intValue > 127) {
```
sjrd: You can do this test with only one branch as `(intValue.toByte.toInt != intValue)`. Same for the `Short` and `Long` cases.

densh:
@sjrd All comments addressed an ready for next pass.

densh:
@sjrd Eager initialization of the cache array seems to have negative effect on the benchmarks results. It's likely caused by memory locality effects. In any case having one extra null check doesn't seem to cost us much, so I reverted back to lazy initialization.

sjrd:
OK

PR #944, opened 2017-08-18
Implementation for java.util.concurrent.lock.*

Korf74:
This PR contains java.util.concurrent.lock classes, it was ported from harmony.

It contains some minor changes to CAtomics as well

PR #1002, opened 2017-09-07
Revert #944

densh:
We can not remove methods in a point release, due
to binary compatibility constraints. We'll need to reapply
them when 0.4.0 happens.

Review @Duhemm 
/cc @matil019

### DIFF
--- a/javalib/src/main/scala/java/io/FileInputStream.scala
+++ b/javalib/src/main/scala/java/io/FileInputStream.scala
@@ -19,6 +19,9 @@ class FileInputStream(fd: FileDescriptor) extends InputStream {
   override def close(): Unit =
     fcntl.close(fd.fd)
 
+  override protected def finalize(): Unit =
+    close()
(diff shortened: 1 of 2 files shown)


Duhemm:
What about the removal of `finalise` in `java.util.regex.Pattern`? https://github.com/scala-native/scala-native/pull/992/files#diff-180072e950a6b3db11ffd0d53d383c7fL183

densh:
`Pattern` is a final class, so there is no way to call that finalize from 3-rd party code.

matil019:
I got it. @densh How did you detect binary incompatibility?

densh:
@matil019 We should start enforcing it automatically with MIMA, but we don't do this yet. Rule of thumb is not to change any publicly visible APIs (i.e. don't remove methods or change their signatures.)

matil019:
@densh All right. I'll keep that in mind.

PR #1004, opened 2017-09-08
Fix topic/mutithreading build

valdisxp1:
I noticed the Travis builds are failing on `topic/multithreading`. I identified 2 issues so far.
- [x] Conflicted symbols in javalib `java.langThread` https://travis-ci.org/scala-native/scala-native/jobs/272416452#L1376 
- [x] ~~NullPointerException in nativelib https://travis-ci.org/scala-native/scala-native/jobs/272416843#L1289~~ Fixed by #1020 
- [x] Removed conflicting `@stub` definitions in javalib `java.langThread`

### DIFF
--- a/javalib/src/main/scala/java/lang/Thread.scala
+++ b/javalib/src/main/scala/java/lang/Thread.scala
@@ -2,7 +2,7 @@ package java.lang
 
 import java.util
 import java.lang.Thread._
-import scala.scalanative.native.stub
+
 import scala.scalanative.runtime.NativeThread
 import scala.scalanative.native.{
   CFunctionPtr,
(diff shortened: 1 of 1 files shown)


densh:
This might be related to the second problem: https://github.com/scala-native/scala-native/issues/781

valdisxp1:
@densh I moved the the BoxedUnit definitions similar to what ScalaJS did, but it did not work.
It fails with `scala.reflect.internal.MissingRequirementError: object scala.scalanative.native.Ptr`
I also tried moving the corresponding files over as well, but it looks like it is asking for all the types.

valdisxp1:
An alternative would be to just comment out or rewrite the offending code, which some places is already broken. For example an infinite loop in `ForkJoinPool.scala`:
```scala
  final def firstComplete: CountedCompleter[_] = {
    var c: Int = 0
    while(true) {
      c = pending
      if(c == 0)
        this
      else if(pending.compareAndSwapStrong(c, c - 1))
        null
    }
    null
  }
```

densh:
@valdisxp1 If there is a simple workaround to resolve the crash, lets use that for now. But please make sure that there is an issue open with minimal example that causes the crash.

valdisxp1:
@densh Now it builds successfully with the workaround. I also included your changes for Travis setup.
I will now create minimized examples of the crashes

valdisxp1:
I created issues for the 3 distinct cases I had to workaround. #1016 #1017 #1018

densh:
@valdisxp1 #1020 fixes the compiler crash you discovered. Lets wait for it to get merged first to avoid workarounds.

valdisxp1:
That probably leaves this PR with the first commit (removing duplicate `yield` definitions). The last commit is #1010 cherry-picked.

densh:
I've just rebased `topic/multithreading` on top of latest master. We only need the first commit from this PR now. Please rebase it on top of latest head of the branch.

valdisxp1:
The rebase left some unmerged changes in `java.lang.Thread`. Fixed that as well.

densh:
LGTM, thanks!

PR #1010, opened 2017-09-11
Fix travis build breakage due to latest update

densh:


densh:
2017-09-06 update of travis environment broke our build due to
duplicate symbols for libunwind. Now, travis pre-install libunwind
that comes with clang and it conflicts with system-provided libunwind.

To fix this we force remove pre-bundled libunwind, before we install
system-provided one.

Issue #1013, opened 2017-09-13
Running sbt test with no tests throws an error


ekrich:
This is for a multi-project where the top level and one component has no tests.
```
> test
... some updating omitted
[info] Compiling 1 Scala source to /Users/eric/workspace/scala-native-ml/blas/target/scala-2.11/test-classes...
[info] Compiling 1 Scala source to /Users/eric/workspace/scala-native-ml/target/scala-2.11/test-classes...
[info] Linking (857 ms)
[info] Discovered 1496 classes and 10819 methods
[info] Optimizing (1571 ms)
[info] Generating intermediate code (295 ms)
[info] Produced 47 files
[info] Compiling to native code (1284 ms)
[info] Linking native code (254 ms)
[trace] Stack trace suppressed: run last blas/nativetest:nativeLinkNIR for the full output.
[error] (blas/nativetest:nativeLinkNIR) java.nio.file.FileSystemAlreadyExistsException
[error] Total time: 6 s, completed Sep 13, 2017 9:49:57 AM
> 
```
Workaround: create an alias to tests in `build.sbt`.
`addCommandAlias("test", ";tests/test")`

densh:
Can your provide a link to a project that reproduces this?

ekrich:
Sure, I was going to look into it because I don't think it should be making this error. Also, Scala Native produces `TestMain` for every project whether there are any test classes or not. https://github.com/ekrich/scala-native-ml

ekrich:
Related #532.

Issue #532, opened 2017-03-03
can't execute nativeMissingDependencies task in parallel


xuwei-k:
https://github.com/xuwei-k/nativeMissingDependencies-parallel

```
java.nio.file.FileSystemAlreadyExistsException
	at com.sun.nio.zipfs.ZipFileSystemProvider.newFileSystem(ZipFileSystemProvider.java:113)
	at java.nio.file.FileSystems.newFileSystem(FileSystems.java:326)
	at java.nio.file.FileSystems.newFileSystem(FileSystems.java:276)
	at scala.scalanative.io.VirtualDirectory$JarDirectory.<init>(VirtualDirectory.scala:151)
	at scala.scalanative.io.VirtualDirectory$.jar(VirtualDirectory.scala:57)
	at scala.scalanative.io.VirtualDirectory$.real(VirtualDirectory.scala:65)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$availableDependenciesTask$1$$anonfun$apply$4$$anonfun$17.apply(ScalaNativePluginInternal.scala:255)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$availableDependenciesTask$1$$anonfun$apply$4$$anonfun$17.apply(ScalaNativePluginInternal.scala:254)
	at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:251)
	at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:251)
	at scala.collection.immutable.List.foreach(List.scala:318)
	at scala.collection.TraversableLike$class.flatMap(TraversableLike.scala:251)
	at scala.collection.AbstractTraversable.flatMap(Traversable.scala:105)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$availableDependenciesTask$1$$anonfun$apply$4.apply(ScalaNativePluginInternal.scala:254)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$availableDependenciesTask$1$$anonfun$apply$4.apply(ScalaNativePluginInternal.scala:251)
	at scala.scalanative.util.Scope$.apply(Scope.scala:30)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$availableDependenciesTask$1.apply(ScalaNativePluginInternal.scala:251)
[...]

PR #1022, opened 2017-09-20
Fix for #1013, error running sbt test

ekrich:


util/src/main/scala/scala/scalanative/io/VirtualDirectory.scala
```diff
@@ -98,9 +99,18 @@ object VirtualDirectory {
...
+        Try(FileSystems.newFileSystem(uri, Map("create" -> "false").asJava)) match {
+          case Success(s) => s
+          case Failure(e) =>
+            e match {
+              case e: FileSystemAlreadyExistsException =>
+                FileSystems.getFileSystem(uri)
+              case _ => throw e
+            }
+        }
+      }
```
densh: `scala.uil.Try` might be an overkill here, normal `try catch` would do the job just as well.

densh:
Also #532

ekrich:
It is really strange when tests run in parallel as the output messages get interleaved.

densh:
LGTM, thanks for another nice contribution @ekrich ! 👍

PR #1024, opened 2017-09-22
Fix #892: Files.scala doesn't close files and so keeps files open

### LINKED ISSUE #892: Files.scala doesn't close files and so keeps files open
Methods don't close file stream and so keeps file open:
- [ ] `Files.lines`
- [ ] `Files.newByteChannel`
- [ ] `Files.copy`

Daemonhost:
Files.copy(source: Path, target: Path, options: Array[CopyOption]) didn't close the source file because the close method of the InputStream returned by FileSystemProvider.newInputStream was empty. I've fixed that issue by overriding it.

Files.lines now closes the file as soon as the end of the returned stream is reached.

I'm not sure why Files.newByteChannel was mentioned in #892. The close method of the returned FileChannel works correctly.

### DIFF
--- a/javalib/src/main/scala/java/io/BufferedReader.scala
+++ b/javalib/src/main/scala/java/io/BufferedReader.scala
@@ -125,7 +125,10 @@ class BufferedReader(in: Reader, sz: Int) extends Reader {
   }
 
   def lines(): Stream[String] =
-    new WrappedScalaStream(toScalaStream(), None)
+    lines(false)
+
+  private[java] def lines(closeAtEnd: Boolean): Stream[String] =
+    new WrappedScalaStream(toScalaStream(closeAtEnd), None)
 
   /** Prepare the buffer for reading. Returns false if EOF */
   private def prepareRead(): Boolean =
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/io/BufferedReader.scala
```diff
@@ -159,10 +162,17 @@ class BufferedReader(in: Reader, sz: Int) extends Reader {
...
+
+  private[this] def toScalaStream(closeAtEnd: Boolean): SStream[String] = {
     Option(readLine()) match {
-      case None       => SStream.empty
-      case Some(line) => line #:: toScalaStream()
+      case None => {
+        if (closeAtEnd)
+          close()
+        SStream.empty
+      }
```
densh: We don't put curly braces for bodies of the case clauses.

javalib/src/main/scala/java/io/BufferedReader.scala
```diff
@@ -159,10 +162,17 @@ class BufferedReader(in: Reader, sz: Int) extends Reader {
...
+  private def toScalaStream(): SStream[String] =
+    toScalaStream(false)
+
+  private[this] def toScalaStream(closeAtEnd: Boolean): SStream[String] = {
     Option(readLine()) match {
-      case None       => SStream.empty
-      case Some(line) => line #:: toScalaStream()
+      case None => {
+        if (closeAtEnd)
+          close()
```
densh: This if clause would look a bit better with curlies around the close call.

Daemonhost:
@densh Comments addressed.

densh:
LGTM, congrats on first contribution! 🎉

ekrich:
Nice 👍

PR #1026, opened 2017-09-22
Fix #1025 System.getProperty("user.home") returns null

### LINKED ISSUE #1025: Add support for System.getProperty("user.home")
On the JVM, `System.getProperty("user.home")` returns something like `/Users/foo`.  On Scala Native, it just returns `null`.

ekrich:


densh:
LGTM! We can improve robustness in the future, using `$HOME` should be sufficient for most uses.

javalib/src/main/scala/java/lang/System.scala
```diff
@@ -91,6 +92,10 @@ object System {
   def nanoTime(): scala.Long          = time.scalanative_nano_time
   def currentTimeMillis(): scala.Long = time.scalanative_current_time_millis
 
+  /**
+   * POSIX environment variables
+   * http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html
+   */
```
densh: We don't do doc comments for JDK APIs as they are never published.

densh:
LGTM! 👍

PR #1027, opened 2017-09-23
Implementation and test of List.contains

cquiroz:
Minimal implementation of `List.contains` as needed by #1023 

This is my first attempt at a PR for scala-native. Please let me know any changes needed. I ran scalafmt on this

### DIFF
--- a/javalib/src/main/scala/java/util/AbstractList.scala
+++ b/javalib/src/main/scala/java/util/AbstractList.scala
@@ -45,8 +45,6 @@ abstract class AbstractList[E] protected ()
     findIndex(listIterator(size))
   }
 
-  override def contains(o: Any): Boolean = indexOf(o) >= 0
-
   override def clear(): Unit =
     removeRange(0, size)
 
--- a/javalib/src/main/scala/java/util/AbstractList.scala
+++ b/javalib/src/main/scala/java/util/AbstractList.scala
@@ -45,6 +45,8 @@ abstract class AbstractList[E] protected ()
     findIndex(listIterator(size))
   }
 
+  override def contains(o: Any): Boolean = indexOf(o) >= 0
+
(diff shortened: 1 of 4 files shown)


cquiroz:
i'm not sure how travis work. I did tests this way:

```
sbt
publishLocal
tests/test
```

Daemonhost:
contains method is already implemented in java.util.AbstractCollection (ArrayList extends AbstractList which extends AbstractCollection) as follows.
def contains(o: Any): Boolean =
    iterator.asScala.exists(o === _)

where asScala returns JIteratorWrapper defined in scala.collection.convert.Wrappers, and the exists method is implemented in scala.collection.Iterator
Implementing it again in java.util.AbstactList seems unnecessary.

However, linking of the following code does, indeed, fail.
val al = new ArrayList[Int](Seq(1, 2, 3, 2).asJava)
println(al.contains(1))
sbt log:
[info] Compiling 1 Scala source to /home/vic/projects/scala/scala-native-test/target/scala-2.11/classes...
[info] Linking (1479 ms)
[error] cannot link: @java.util.ArrayList::contains_java.lang.Object_bool
[error] unable to link
[error] (compile:nativeLinkNIR) unable to link
[error] Total time: 2 s, completed 23.09.2017 20:15:01

I'm not yet sure why, or how to address this issue.

cquiroz:
That's right, I noted that. It is implemented on `AbstractCollection` but it doesn't get linked. With these changes the linking works

Just let me know if there are better alternatives

Daemonhost:
I think that it's best to add
`override def contains(o: Any): Boolean = super.contains(o)`
to `ArrayList` since such solution is already used for other methods of `ArrayList`, e.g. `addAll` and `iterator`.

cquiroz:
I updated it as suggested

PR #1041, opened 2017-09-29
Fix #1040: Allow easy switching of GC setting at sbt command line

### LINKED ISSUE #1040: Allow easy switching of GC setting at sbt command line
Allow at sbt prompt `set nativeGC in ThisBuild := "immix"`

ekrich:
Allows the following at the sbt prompt. `set nativeGC in ThisBuild := "immix"`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,7 +34,8 @@ addCommandAlias(
     "javalib/publishLocal",
     "auxlib/publishLocal",
     "scalalib/publishLocal",
-    "publishLocal"
+    "sbtScalaNative/publishLocal",
+    "testInterface/publishLocal"
   ).mkString(";", ";", "")
 )
 
@@ -46,7 +47,7 @@ addCommandAlias(
     "tests/test",
     "tools/test",
     "benchmarks/run --test",
-    "scripted"
+    "sbtScalaNative/scripted"
   ).mkString(";", ";", "")
 )
 
@@ -190,14 +191,13 @@ lazy val libSettings =
     scalacOptions ++= Seq("-encoding", "utf8")
   )
 
-lazy val gcSettings =
-  if (!System.getenv.containsKey("SCALANATIVE_GC")) {
-    println("Using default gc")
-    Seq.empty
-  } else {
-    val gc = System.getenv.get("SCALANATIVE_GC")
-    println(s"Using gc based on SCALANATIVE_GC=$gc")
-    Seq(nativeGC := gc)
+lazy val gcSetting =
+  nativeGC := {
+    val log     = sLog.value
+    val default = (nativeGC in ThisBuild).value
+    val gc      = Option(System.getenv.get("SCALANATIVE_GC")).getOrElse(default)
+    log.info(s"Using $gc gc")
+    gc
   }
 
 lazy val projectSettings =
@@ -205,7 +205,38 @@ lazy val projectSettings =
     scalaVersion := libScalaVersion,
     resolvers := Nil,
     scalacOptions ++= Seq("-target:jvm-1.8")
-  ) ++ gcSettings
+  )
+
+lazy val scalaNative =
+  project
+    .in(file("."))
+    .aggregate(
+      util,
+      nir,
+      tools,
+      nscplugin,
+      sbtScalaNative,
+      nativelib,
+      javalib,
+      auxlib,
+      scalalib,
+      testingCompilerInterface,
+      testingCompiler,
+      testInterface,
+      testInterfaceSerialization,
+      testInterfaceSbtDefs,
+      sandbox,
+      demoNative,
+      tests,
+      demoJVM,
+      benchmarks
+    )
+    .settings(
+      name := "scala-native",
+      gcSetting,
+      aggregate in publishLocal := false
+    )
+    .settings(noPublishSettings)
 
 lazy val util =
   project
--- a/build.sbt
+++ b/build.sbt
@@ -554,7 +554,10 @@ lazy val testInterface =
     .settings(
       name := "test-interface",
       libraryDependencies += "org.scala-sbt"    % "test-interface"   % "1.0",
-      libraryDependencies -= "org.scala-native" %%% "test-interface" % version.value % Test
+      libraryDependencies -= "org.scala-native" %%% "test-interface" % version.value % Test,
+      publishLocal := publishLocal
+        .dependsOn(publishLocal in testInterfaceSerialization)
+        .value
     )
     .enablePlugins(ScalaNativePlugin)
     .dependsOn(testInterfaceSerialization)
@@ -567,7 +570,10 @@ lazy val testInterfaceSerialization =
     .in(file("test-interface-serialization"))
     .settings(
       name := "test-interface-serialization",
-      libraryDependencies -= "org.scala-native" %%% "test-interface" % version.value % Test
+      libraryDependencies -= "org.scala-native" %%% "test-interface" % version.value % Test,
+      publishLocal := publishLocal
+        .dependsOn(publishLocal in testInterfaceSbtDefs)
+        .value
     )
     .dependsOn(testInterfaceSbtDefs)
     .enablePlugins(ScalaNativePlugin)
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -240,8 +240,6 @@ object ScalaNativePluginInternal {
       val cpppaths  = (cwd ** "*.cpp").get.map(_.abs)
       val paths     = cpaths ++ cpppaths
 
-      logger.info(s"Using $gc gc")
-
       // predicate to check if given file path shall be compiled
       // we only include sources of the current gc and exclude
       // all optional dependencies if they are not necessary
@@ -400,7 +398,7 @@ object ScalaNativePluginInternal {
       val paths     = apppaths.map(_.abs) ++ opaths
       val compile   = clangpp.abs +: (flags ++ paths)
 
-      logger.time("Linking native code") {
+      logger.time(s"Linking native code ($gc GC)") {
         logger.running(compile)
         Process(compile, cwd) ! logger
       }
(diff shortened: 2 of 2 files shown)


ekrich:
It prints `Using immix gc` or `Using boehm gc` three time as it is used by three projects.

build.sbt
```diff
@@ -184,21 +184,24 @@ lazy val libSettings =
...
   )
 
-lazy val gcSettings =
+lazy val gcSetting =
   if (!System.getenv.containsKey("SCALANATIVE_GC")) {
-    println("Using default gc")
-    Seq.empty
+    nativeGC := {
+      val gc = nativeGC.value
+      println(s"Using $gc gc")
```
Duhemm: It would be better to log using `sLog.info(s"Using $gc gc")`

https://github.com/sbt/sbt/blob/bcd21ba9b64cf89023bc9626bcb5129c155ba90b/main/src/main/scala/sbt/Keys.scala#L107

build.sbt
```diff
@@ -184,21 +184,24 @@ lazy val libSettings =
...
+      gc
+    }
   } else {
     val gc = System.getenv.get("SCALANATIVE_GC")
     println(s"Using gc based on SCALANATIVE_GC=$gc")
-    Seq(nativeGC := gc)
+    nativeGC := gc
```
Duhemm: Same here, it would be better to use `sLog`. Wouldn't it be clearer if written like this?

```scala
nativeGC := {
  val default = nativeGC.value
  val log = sLog.value
  val gc = Option(System.getenv.get("SCALANATIVE_GC")).getOrElse(default)
  log.info(s"Using $gc gc")
}
     
```

build.sbt
ekrich: Didn't know about that. Thank-you.

build.sbt
ekrich: That looks really nice, ready now. 👍

ekrich:
I'm not seeing your comment, sorry. You want me to remove the `gc` log in the plugin `nativeCompileLib` or change `gcSetting` back to `gcSettings`? The only other weird thing is the printing 3 times, one for each use in subprojects. This requires a `root` project and when I tried it - tons of stuff broke - that was in my first commit.

build.sbt
```diff
@@ -184,21 +184,20 @@ lazy val libSettings =
...
-    val gc = System.getenv.get("SCALANATIVE_GC")
-    println(s"Using gc based on SCALANATIVE_GC=$gc")
-    Seq(nativeGC := gc)
+lazy val gcSetting =
+  nativeGC := {
+    val log     = sLog.value
+    val default = nativeGC.value
+    val gc      = Option(System.getenv.get("SCALANATIVE_GC")).getOrElse(default)
+    log.info(s"Using $gc gc")
+    gc
```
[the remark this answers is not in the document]
ekrich: Sounds like a good idea, I'll give it a try.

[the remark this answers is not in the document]
ekrich: I added this line. Do you think it is okay?
https://github.com/ekrich/scala-native/blob/topic/build-gc/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L239

```
[info] Compiling 1 Scala source to /Users/eric/workspace/scala-native/unit-tests/target/scala-2.11/test-classes...
[info] Linking (6851 ms)
[info] Discovered 4246 classes and 24348 methods
[info] Using boehm gc
[info] Optimizing (7485 ms)
[info] Generating intermediate code (2001 ms)
[info] Produced 66 files
[info] Compiling to native code (3157 ms)
[info] Linking native code (643 ms)
```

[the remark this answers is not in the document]
Duhemm: On the one hand it’s a bit redundant, but on the other hand this task actually produces very different result based on the value of `nativeGC`... I’m fine with it, but I’ll let @densh decide.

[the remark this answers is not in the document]
ekrich: That is why I was asking about the log stuff as he is very disciplined about details. I tried your suggestion and it prints only once vs 3 times on startup but `session save` actually changes the build file so this is not good. I was getting help from Dale while doing this PR and he suggested making the setting global in the plugin and adding a root project which only printed the `Using gc...` once but that busted tons of stuff. This change helps a ton as you can actually change the GC at the command line and everything else still works except macOS with immix which is broken for another reason.

ekrich:
@Duhemm I reintroduced the root project which has some benefits and leaves some concerns.

Benefits.

1. There is no longer a default `scala-native` (or what ever your top level directory is named) project published.
2. The GC setting only logs once during startup.
3. The following projects do not get published locally a second time. `nscplugin, nativelib, javalib, auxlib, and scalalib`

Potential concerns. 

1. For what ever reason, the default `publishLocal` somehow went about doing a `publishLocal` for all the projects based on the project dependencies, I am guessing. Now I had to introduce some additional `publishLocal` dependencies to make sure all the dependencies are published.
2.  I am not sure the `dirtymoney` plugin is doing the job as a `rebuild` no longer seems to refetch Scala and force as much work. These are the `cleanCache` and `cleanLocal` targets although they look ok on `inspect`ion.
3. I am not sure the roll of `testingCompiler` and `testingCompilerInterface` so they were left alone.

ekrich:
@Duhemm This is much better now. I am able to get `clean`, `cleanLocal`, and `cleanCache` to work now again so I am ready for review. In the long run we can probably streamline the build even more using the top project as a driver.

build.sbt
```diff
@@ -184,21 +185,51 @@ lazy val libSettings =
...
+      tests,
+      demoJVM,
+      benchmarks
+    )
+    .settings(
+      name := "scala-native",
+      gcSetting,
+      aggregate in publishLocal := false
+    )
+    .settings(noPublishSettings)
```
densh: Lets *not* use aggregates. They are confusing in a number of ways. One should think of Scala Native a collection of smaller projects, not a single mega-project.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -236,6 +236,8 @@ object ScalaNativePluginInternal {
       val cpppaths  = (cwd ** "*.cpp").get.map(_.abs)
       val paths     = cpaths ++ cpppaths
 
+      logger.info(s"Using $gc gc")
```
densh: I'm not completely sold on printing this in every project that uses Scala Native. It's useful for developing Scala Native but should not be shown by default.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -135,7 +133,9 @@ object ScalaNativePluginInternal {
         case Some(_) =>
           ()
       }
-    }
+    },
+    nativeGC := "boehm",
+    nativeGC in NativeTest := (nativeGC in Test).value
```
densh: `nativeGC` should be in base settings, not global settings. Global settings are not scoped separately for compile and test and should generally be avoided.

dwijnand: What's the concern with global settings not being scoped separately for compile and test?

I believe the intent here is to define the default value of for the key `nativeGC` (introduced by this plugin), which is one of the use cases for which global settings _should_ be used (http://www.scala-sbt.org/1.0/docs/Plugins.html#projectSettings+and+buildSettings).

The benefit of doing this is you if you want to redefine the value of `nativeGC` for all the projects in your build you can do so by defining `nativeGC in ThisBuild := "immix"` (either in your build code, or in the shell with `set`).

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
dwijnand: Perhaps it should be a debug level log message instead. (sorry, for the intrusion in your PR)

build.sbt
ekrich: I agree that aggregates are a bit messy looking but a `default` main project is like "aggregate all" by default and it publishes a default project in the current build. Also, if I put the gcSetting in project settings rather than the root project the setting gets set three times and repeats the message.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
ekrich: I was not able to get a behaviour where I could change the `nativeGC` setting for ScalaNative or a user project that would apply across all the native projects so this is what led me to the global key along with help from Dale. The original intent of this PR was just to be able to `set nativeGC in ThisBuild := "immix"` and all the other changes are somewhat incidental to making that happen. I really tried to make as few changes as possible.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
ekrich: This was mainly used as a debug tool so I could guarantee that each Scala Native link was using the correct setting so debug probably make more sense. I wasn't completely sold but it does tell what is happening. Maybe "Using" could be " Linking" or something as well?

ekrich:
Overall, since the attempt to rebase was not going to work well, I could go ahead and try again at this as I have more experience and the notes from the first time. From what I can tell, we really need the global key. Even to do something like this in a user project,
```
    nativeOptimizerReporter := scala.scalanative.optimizer.Reporter
      .toDirectory(file("/tmp"))
```
requires that I import `import scalanative.sbtplugin.ScalaNativePluginInternal._` in my build.

Even if I just got the changes redone on top of master, at least it could be looked at more closely.
What do you think?

build.sbt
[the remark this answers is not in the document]
ekrich: Just to document this, when I put the `nativeGC` setting raw in the build. sbt saveSettings would overwrite this value in the actual build file - when they say sbt is interactive, they really mean it. so this approach would not work.

ekrich:
@densh I am not happy with this solution either at this point. I did find a place to put in the output where we can see the GC getting used that I think will be ok but maybe with a different wording or format.
```
[info] Linking (968 ms)
[info] Discovered 1278 classes and 9472 methods
[info] Optimizing (1670 ms)
[info] Generating intermediate code (492 ms)
[info] Produced 38 files
[info] Compiling to native code (1017 ms)
[info] Linking native code (immix GC) (191 ms)
```
The current implementation in this PR not working if I set the env var `SCALANATIVE_GC` but works if set via `set nativeGC in ThisBuild := "immix"`.

I will start over and try and minimize changes but would like to leave this PR open for reference for the time being if that is OK?

ekrich:
Opened a new PR. https://github.com/scala-native/scala-native/pull/1117

PR #1042, opened 2017-10-01
Thread.start() basic thread support

valdisxp1:
Calling `pthread_create` with a generic function(`callRun`) and provide the `Thread` object as the argument.

I had to use seemingly unsafe casts between `Ptr[scala.Byte]` and `Ptr[Thread]`. However, `pthread_create` definition has pointers to void and it works fine. 

```c
int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
                  void *(*start_routine) (void *), void *arg);
```


valdisxp1:
Changes included in #1068

PR #1044, opened 2017-10-01
Make InetAddress a concrete class instead of abstract

asoltysik:
Although InetAdress is only intended as a base class for `Inet4Address` and `Inet6Address`, it's supposed to be a concrete class but it's abstract in Scala Native.


densh:
LGTM, thanks @asoltysik ! 👍

PR #1047, opened 2017-10-02
Fix #1046: Double.parseDouble and and Float.parseFloat throw `NumberFormatException` when given bad data

### LINKED ISSUE #1046: parseFloat and parseDouble return 0 when parsing invalid data
When `Double.parseDouble` and `Float.parseFloat` are given non-numeric data, e.g. `potato`, they return `0.0` instead of throwing `NumberFormatException`.

PR with fix impending.

longshorej:
Fixes #1046 

`strtod` and `strtof` don't set `errno.errno` when given invalid data. Instead, the following behavior (quoted below) is performed. This PR adjusts the implementations accordingly so that when calling e.g. `Double.parseDouble("potato")` an exception is raised instead of the value `0.0` being returned.

> If the subject sequence is empty or does not have the expected form, no conversion shall be performed; the value of str is stored in the object pointed to by endptr, provided that endptr is not a null pointer.

Reference: http://pubs.opengroup.org/onlinepubs/009695399/functions/strtod.html

p.s. Thanks a bunch for Scala Native! We're using it on a CLI project @ Lightbend and have been impressed with it so far!

### DIFF
--- a/javalib/src/main/scala/java/lang/Double.scala
+++ b/javalib/src/main/scala/java/lang/Double.scala
@@ -244,7 +244,7 @@ object Double {
       errno.errno = 0
       val res = stdlib.strtod(cstr, end)
 
-      if (errno.errno == 0) res
+      if (errno.errno == 0 && cstr != !end && string.strlen(!end) == 0) res
       else throw new NumberFormatException(s)
     }
 
--- a/unit-tests/src/test/scala/java/lang/DoubleSuite.scala
+++ b/unit-tests/src/test/scala/java/lang/DoubleSuite.scala
@@ -194,6 +194,10 @@ object DoubleSuite extends tests.Suite {
     assert(Double.parseDouble("Infinity") == Double.POSITIVE_INFINITY)
     assert(Double.parseDouble("-Infinity") == Double.NEGATIVE_INFINITY)
     assert(Double.isNaN(Double.parseDouble("NaN")))
+    assertThrows[NumberFormatException](Double.parseDouble(""))
+    assertThrows[NumberFormatException](Double.parseDouble("potato"))
+    assertThrows[NumberFormatException](Double.parseDouble("0.0potato"))
+    assertThrows[NumberFormatException](Double.parseDouble("0.potato"))
   }
 
   // Not fully JVM compliant yet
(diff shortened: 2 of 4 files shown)


ekrich:
Great that Lightbend is using Scala Native. Do you need the extra check in the C example? `*endp == '\0'` or `!endp == '\0'` in this case.

longshorej:
Good question - I don't think so. There might be other non-numeric data in `endp` which would cause `!endp` to not be equal to `\0` but still parse successfully. Per the spec, if `endp`  equals `str` that (and only that) means no successful parsing was performed.

ekrich:
Thanks for the explanation and the contribution.

javalib/src/main/scala/java/lang/Double.scala
```diff
@@ -244,7 +244,7 @@ object Double {
       errno.errno = 0
       val res = stdlib.strtod(cstr, end)
 
-      if (errno.errno == 0) res
+      if (errno.errno == 0 && cstr != !end) res
```
densh: This is much better than what we had before, and it indeed fixes `parseDouble("foobar")` to throw as expected. But the problem still remains for the `parseFloat("0.0foobar")`. `strtod` would happily eat `0.0` and then stop.

We can fix it by: 
1. Finding pointer to the end of the string by skipping all characters until we see `'\0'`. 
2. And then checking that `end` as returned by `strtod` is the end of the string.

unit-tests/src/test/scala/java/lang/DoubleSuite.scala
```diff
@@ -194,6 +194,7 @@ object DoubleSuite extends tests.Suite {
     assert(Double.parseDouble("Infinity") == Double.POSITIVE_INFINITY)
     assert(Double.parseDouble("-Infinity") == Double.NEGATIVE_INFINITY)
     assert(Double.isNaN(Double.parseDouble("NaN")))
+    assertThrows[NumberFormatException](Double.parseDouble("potato"))
```
densh: Needs another test case to check that `Double.parseDouble("0.0potato")` throws as well.

javalib/src/main/scala/java/lang/Double.scala
longshorej: Thanks - I'll get this fixed. Yeah, I think that'll work @ekrich - thanks for the suggestion.

densh:

PR #1048, opened 2017-10-02
Fix the slow execution of the CfChainsSimplification pass

hkapp:
This PR fixes the execution speed issues with the CfChainsSimplification pass. This is done by computing the UseDef only once per method, instead of once per Jump instruction encountered (!)

### DIFF
--- a/tools/src/main/scala/scala/scalanative/optimizer/pass/CfChainsSimplification.scala
+++ b/tools/src/main/scala/scala/scalanative/optimizer/pass/CfChainsSimplification.scala
@@ -14,20 +14,23 @@ class CfChainsSimplification(implicit top: Top) extends Pass {
   import CfChainsSimplification._
 
   override def onInsts(insts: Seq[Inst]): Seq[Inst] = {
-    val cfg = ControlFlow.Graph(insts)
+    val cfg    = ControlFlow.Graph(insts)
+    val usedef = UseDef(cfg)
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/optimizer/pass/CfChainsSimplification.scala
```diff
@@ -14,20 +14,22 @@ class CfChainsSimplification(implicit top: Top) extends Pass {
   import CfChainsSimplification._
 
   override def onInsts(insts: Seq[Inst]): Seq[Inst] = {
-    val cfg = ControlFlow.Graph(insts)
+    val cfg    = ControlFlow.Graph(insts)
+    val usedef = UseDef(cfg)
+    val method = MethodInfo(cfg, usedef)
```
densh: The code would be a bit more clear if you make `MethodInfo` an implicit argument. This way you don't have to manually pass it around in recursive calls down the chain.

hkapp: I changed it, is that what you had in mind ?

densh: Yep, it's much better now! 👍

densh:
LGTM, thanks for the fix @hkapp ! 👍

PR #1051, opened 2017-10-04
Fix #1030: Automatically detect LLVM 5.0 binaries

### LINKED ISSUE #1030: Automatically detect LLVM 5.0 binaries
Currently we only detect up to LLVM 4.0, which should be easy to fix in https://github.com/scala-native/scala-native/blob/master/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/Utilities.scala#L51

ekrich:


densh:
LGTM, thanks @ekrich ! 👍

PR #1056, opened 2017-10-14
Improved unit test runner

valdisxp1:
- Logging the runner's exit code
- Catching segmentation faults and dumping the relevant stack trace
- Catching arithmetic faults, like division by zero.
The catching can be disabled by specifying `--no-catch-fault`
- Added the ability to run tests directly from:
 `unit-tests/target/scala-2.11/tests-out run-test <fully qualified names>`

The function I used to reproduce segfaults:
```scala
  private def segfault: Unit = {
    val ptr: Ptr[CSignedChar] = null.asInstanceOf[Ptr[CSignedChar]]
    Console.out.println(!ptr)
  }
```

And division by zero:
```scala
  private def aritFail: Unit = {
    def x = 0
    Console.out.println(1000 / x)
  }
```

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
@@ -29,7 +29,10 @@ class ComRunner(bin: File,
       import sbt.Process._
       val port = serverSocket.getLocalPort
       logger.info(s"Starting process '$bin' on port '$port'.")
-      Process(bin.toString +: port.toString +: args, None, envVars.toSeq: _*) ! logger
+      val retCode = Process(bin.toString +: port.toString +: args,
+                            None,
+                            envVars.toSeq: _*) ! logger
+      logger.info(s"Process '$bin' on port '$port' exited with code $retCode.")
     }
   }
 
--- a/test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
+++ b/test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
@@ -26,9 +26,47 @@ abstract class TestMainBase {
 
   /** Actual main method of the test runner. */
   def testMain(args: Array[String]): Unit = {
-    val serverPort   = args.head.toInt
-    val clientSocket = new Socket("127.0.0.1", serverPort)
-    testRunner(Array.empty, null, clientSocket)
+    args.toList match {
+      case "run-test" :: className :: Nil =>
+        runSingleTest(className)
+      case serverPort :: _ =>
+        val clientSocket = new Socket("127.0.0.1", serverPort.toInt)
+        testRunner(Array.empty, null, clientSocket)
+    }
+  }
+
+  private def runSingleTest(value: String): Unit = {
+    val runner = frameworks(0).runner(Array.empty,
+                                      Array.empty,
+                                      new PreloadedClassLoader(tests))
+    val taskDef = new TaskDef(
+      value,
+      DeserializedSubclassFingerprint(isModule = true,
+                                      "tests.Suite",
+                                      requireNoArgConstructor = false),
+      false,
+      Array(new SuiteSelector))
+    val Array(task: Task) = runner.tasks(Array(taskDef))
+    val logger = new Logger {
+      def debug(msg: String): Unit = Console.err.println("DEBUG: " + msg)
+
+      def error(msg: String): Unit = Console.err.println("ERROR: " + msg)
+
+      val ansiCodesSupported = true
+
+      def warn(msg: String): Unit = Console.err.println("WARN: " + msg)
+
+      def trace(t: Throwable): Unit = {
+        Console.err.println("TRACE:")
+        t.printStackTrace()
+      }
+
+      def info(msg: String): Unit = Console.err.println("INFO: " + msg)
+    }
+    val eventHandler = new EventHandler {
+      def handle(event: SbtEvent): Unit = {}
+    }
+    task.execute(eventHandler, Array(logger))
   }
 
   /** Test runner loop.
--- a/test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
+++ b/test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
@@ -35,12 +35,32 @@ abstract class TestMainBase {
       signal.signal(signal.SIGFPE, TestMainBase.faultHandler)
     }
     values match {
-      case "run-test" :: className :: Nil =>
-        runSingleTest(className)
+      case "run-test" :: classNames =>
+        classNames.foreach(runSingleTest)
       case serverPort :: _ =>
         val clientSocket = new Socket("127.0.0.1", serverPort.toInt)
         testRunner(Array.empty, null, clientSocket)
     }
+  }
+
+  val singleTestLogger = new Logger {
+    def debug(msg: String): Unit = Console.err.println("DEBUG: " + msg)
+
+    def error(msg: String): Unit = Console.err.println("ERROR: " + msg)
+
+    val ansiCodesSupported = true
+
+    def warn(msg: String): Unit = Console.err.println("WARN: " + msg)
+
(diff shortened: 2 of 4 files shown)


test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
```diff
@@ -26,9 +26,55 @@ abstract class TestMainBase {
...
+        Console.err.println("TRACE:")
+        t.printStackTrace()
+      }
+
+      def info(msg: String): Unit = Console.err.println("INFO: " + msg)
+    }
+    val eventHandler = new EventHandler {
+      def handle(event: SbtEvent): Unit = {}
+    }
+    task.execute(eventHandler, Array(logger))
```
Duhemm: The goal is to run a single test directly using the binary that sbt generates with `test:nativeLink`? Unfortunately, this will only work with our own tests, even though the option will be available to everyone.

I was thinking about the following approach:
 - Add a `lazy val nativeDebug = settingKey[Boolean]("Enable debugging of tests with LLDB")` to the sbt plugin
 - When running the tests:
    - if `nativeDebug in Test` is `false`, the tests are run exactly as they are today.
    - if `nativeDebug in Test` is `true`, then sbt **does not** start the test process itself. Instead, it prints out a command line that users should copy-paste in their terminal and waits for the user to press `return`:
> Copy-paste the following command to open the tests in LLDB, start the program, and then press `return` in this window to continue:
> $ lldb target/scala-2.11/classes/test-out 63559

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
```diff
@@ -148,3 +194,29 @@ abstract class TestMainBase {
...
+        Console.err.println("Erroneous Arithmetic Operation")
+        136
+      case _ =>
+        Console.err.println("Unknown fault")
+        -1
+    }
+    // the handler will run on the same thread that caused the segmentation fault
+    new Throwable().printStackTrace()
+    // segfault return code
+    System.exit(retCode)
```
Duhemm: Would it be feasible to fail the current test but continue the execution after a `SIGSEGV` or `SIGFPE` (and maybe other non-fatal signals)?

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
```diff
@@ -148,3 +194,29 @@ abstract class TestMainBase {
...
+    val retCode = id match {
+      case `SIGSEGV` =>
+        Console.err.println("Segmentation fault")
+        139
+      case `SIGFPE` =>
+        Console.err.println("Erroneous Arithmetic Operation")
+        136
+      case _ =>
+        Console.err.println("Unknown fault")
+        -1
```
Duhemm: Shouldn't it return `128 + id`? Otherwise, sbt will just say it exited with `-1`, which is wrong.

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
```diff
@@ -148,3 +194,29 @@ abstract class TestMainBase {
...
+    val SIGSEGV = signal.SIGSEGV
+    val SIGFPE  = signal.SIGFPE
+
+    val retCode = id match {
+      case `SIGSEGV` =>
+        Console.err.println("Segmentation fault")
+        139
+      case `SIGFPE` =>
+        Console.err.println("Erroneous Arithmetic Operation")
+        136
```
Duhemm: I don't think these values are enforced: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html

My understanding is that the exit code should be `1 << 7 + id`. The first bit is set to indicate an abnormal exit, the following 7 bits are represent the exit code.

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
Duhemm: It should probably say that another debugger can be used, but the first argument to the test program should be the given port number.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
```diff
@@ -29,7 +29,10 @@ class ComRunner(bin: File,
       import sbt.Process._
       val port = serverSocket.getLocalPort
       logger.info(s"Starting process '$bin' on port '$port'.")
-      Process(bin.toString +: port.toString +: args, None, envVars.toSeq: _*) ! logger
+      val retCode = Process(bin.toString +: port.toString +: args,
+                            None,
+                            envVars.toSeq: _*) ! logger
+      logger.info(s"Process '$bin' on port '$port' exited with code $retCode.")
```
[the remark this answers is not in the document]
valdisxp1: Makes sense, will add logging to `error` for non zero exit codes. Why  `error`? Something obviously broke, so that would be an error.

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
valdisxp1: I did not know that. Will change it to the formula.

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
valdisxp1: Same here.

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
ekrich: Yes, that was what I was experimenting with too. A port between `1024` and `65535`, otherwise you get an error.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/testinterface/ComRunner.scala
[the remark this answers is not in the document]
ekrich: I really think we need to have more stuff at `info` level but also have `debug` level for more detailed info. If you look and standard make and C/C++ compiler, quite a lot is logged, like the commands.

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
valdisxp1: Hmm. Maybe this would work:
```scala
throw new Error(...)
//instead of
System.exit(...)
```
Need to test that.

test-interface/src/main/scala/scala/scalanative/testinterface/TestMainBase.scala
valdisxp1: 1. Sockets will likely cause problems when using breakpoints in the debugger. The debugger stops at a break point for a while. The JVM part is waiting to receive a message, then timeouts. All further tests fail with `SocketClosed`. It should be better to provide a command for running a single test directly. Currently I have some parameters hardcoded, but those could also be passed via command line.
2. A direct command would be more user friendly than a flag. For example, `testCommands` could take the same arguments as `testOnly` and produce a list of commands to run - 1 for each test suite. 
```bash
sbt "; set nativeDebug := true; testOnly java.lang.ThreadSuite"
```
vs
```bash
sbt testCommands java.lang.ThreadSuite
```

valdisxp1:
Close+reopen to make Travis rebuild.

valdisxp1:
@Duhemm Throwing an exception so it fails only the current test and the runner continues to work.

densh:
Hi @valdisxp1, this PR motivation is clear and it's indeed something we should improve upon. Current state where segfaults in tests remain opaque and is indeed quite hard to debug. Unfortunately, the changes suggested here are going to complicate the build tool interface (having to have multiple ways to run tests will have to be duplicated for every build tool), and I don't believe that's the correct solution to the problem.

Instead, I suggest we catch segfaults using a catch-all segfault handler and print the status to the stderr. This would not require any changes in the build, and would provide a good amount of extra information for debugging. Here is a ticket for this: #1262.

PR #1059, opened 2017-10-16
Add ability to set env vars where Java can see the changes

ekrich:
This is the result so far of trying to remove the setting of env vars in the build file. This allows the Test to become DRY as well. The disadvantage here is that in order to see changes in the env created in the `C` realm, we have to refresh the `java.lang.System` env vars before use.

There are all sorts of hacky code out there in the JVM world to change environment variables out in the wild. Ideally it would be nice to add `System.setenv` methods if that is possible but that would mean that if people used this in Scala Native, to support JVM they would have to use something like the following.

``` scala
if (System.getProperty("java.vm.name").equals("Scala Native") System.setenv ...
else doJvmHackyStuff

```
If we could add the `setenv` methods then we could update the `envVar` map when the method succeeded to avoid always refreshing the environment.

Edit: Couldn't we provide a shim for JVM compilation so the above code would compile on both platforms?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -439,14 +439,7 @@ lazy val tests =
       // nativeOptimizerReporter := OptimizerReporter.toDirectory(
       //   crossTarget.value),
       libraryDependencies += "org.scala-native" %%% "test-interface" % nativeVersion,
-      testFrameworks += new TestFramework("tests.NativeFramework"),
-      envVars in (Test, test) ++= Map(
-        "USER"                           -> "scala-native",
-        "HOME"                           -> baseDirectory.value.getAbsolutePath,
-        "SCALA_NATIVE_ENV_WITH_EQUALS"   -> "1+1=2",
-        "SCALA_NATIVE_ENV_WITHOUT_VALUE" -> "",
-        "SCALA_NATIVE_ENV_WITH_UNICODE"  -> 0x2192.toChar.toString
-      )
+      testFrameworks += new TestFramework("tests.NativeFramework")
     )
     .enablePlugins(ScalaNativePlugin)
 
--- a/nativelib/src/main/scala/scala/scalanative/native/system.scala
+++ b/nativelib/src/main/scala/scala/scalanative/native/system.scala
@@ -1,0 +1,39 @@
+package scala.scalanative.native
+
+object system {
+
+  /**
+   * Sets an environment variable.
+   * @param name of the variable - must not be null, empty, or contain the = character
+   * @param value of the variable
+   * @param overwrite if overwriting the existing value is desired
+   * @return true if successful, false if not enough memory in the environment
+   */
+  def setenv(name: String,
+             value: String,
+             overwrite: Boolean = true): Boolean = {
+    checkName(name)
+    val ow = if (overwrite) 1 else 0
+    Zone { implicit z =>
+      val res = stdlib.setenv(toCString(name), toCString(value), ow)
+      if (res == 0) true else false
+    }
+  }
+
+  /**
+   * Unset the environment variable.
+   * @param name of the variable - must not be null, empty, or contain the = character
+   */
+  def unsetenv(name: String): Unit = {
+    checkName(name)
+    Zone { implicit z =>
+      stdlib.unsetenv(toCString(name))
+    }
+  }
+
+  private[system] def checkName(name: String): Unit = {
+    if (name == null || name.isEmpty || name.contains('='))
+      throw new IllegalArgumentException(s"name is invalid '$name'")
+  }
+
+}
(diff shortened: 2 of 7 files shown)


wookietreiber:
Yes, please at least add `setenv` and `unsetenv`, unrelated to how javalib handles this. I think, these are two separate issues.

build.sbt
```diff
@@ -439,14 +439,7 @@ lazy val tests =
...
       //   crossTarget.value),
       libraryDependencies += "org.scala-native" %%% "test-interface" % nativeVersion,
-      testFrameworks += new TestFramework("tests.NativeFramework"),
-      envVars in (Test, test) ++= Map(
-        "USER"                           -> "scala-native",
-        "HOME"                           -> baseDirectory.value.getAbsolutePath,
-        "SCALA_NATIVE_ENV_WITH_EQUALS"   -> "1+1=2",
-        "SCALA_NATIVE_ENV_WITHOUT_VALUE" -> "",
-        "SCALA_NATIVE_ENV_WITH_UNICODE"  -> 0x2192.toChar.toString
-      )
```
densh: The reason we set those environment variables in the build is that they have to be external to the testing process. If we set those environment variables from within tests, we don't know if getenv/setenv actually interacts with the outer world, or just updates a process-local hash map.

nativelib/src/main/scala/scala/scalanative/native/system.scala
```diff
@@ -0,0 +1,39 @@
+package scala.scalanative.native
+
+object system {
```
densh: We already have C API for setenv/getenv and Java API, I don't think that adding yet another one is a good idea here.

build.sbt
ekrich: The reason I was trying to move this to the test was that it lack encapsulation and it is not DRY this way. I was also trying to move stuff out of the build so we are less dependent on the sbt build. 

Also, the way an env var is set on the JVM is via reflection on a private member to modify the env map. https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java

Overall, there are quite a few things going on here so it may be better if I break this PR into multiple pieces since I can't demonstrate the problem that easily without some changes.

Typically, the environment in C can be changed for the running process and then is inherited by any child processes. It appears we may be inheriting the env from the Java process as described in the SOF above so it is masking the underlying problem.

nativelib/src/main/scala/scala/scalanative/native/system.scala
ekrich: The idea here was to provide some wrapper so we could setenv without a Zone etc. I know this may be a bigger issue but I think at some point we may need to support common functions via easy to use Scala interfaces.

build.sbt
densh: We used to have a test that checks that environment variables set from outside are visible in native. After this change it's no longer testing this, it's testing that environment variables set from inside native are visible inside native which is not the same thing. This is a regression.

nativelib/src/main/scala/scala/scalanative/native/system.scala
densh: It's not pulling it's weight unfortunately, especially given that there is a Java API for this.

build.sbt
ekrich: I looked into this in `sbt` and it seems that the `envVars` are passed to the spawned process for the tests. I think since there is no way to change the `env` of a running Java process, this was the only way to actually test this at all. If we add a native `setenv`, this updates the current process `env` so we can test this way.

If you are conceptually ok with moving the data to the test area to reduce the dependency on `sbt` and to make the tests DRY then I have started another PR to add the `setenv` and I can open another one to adjust the tests.
See https://github.com/scala-native/scala-native/pull/1151

nativelib/src/main/scala/scala/scalanative/native/system.scala
ekrich: Ok, this is not the best code and creating an API is a big deal - it was just added to eliminate using `Zone` in the test. Not sure to which Java API you are referring.

ekrich:
I think this make sense to address as separate issues.

1. Ability to set an env variable in Scala Native.   Addressed here - https://github.com/scala-native/scala-native/pull/1151
2. Make Java see env changes when the above occurs.
3. Do we need the env vars passed from sbt to the tests? If so, can we share the test code from the JVM side to the native side to avoid code duplication? Otherwise, we can move all the code into the tests.
4. Will Scala Native support Scala wrappers for commonly used native C functions? They would handle `Zone`s, exceptions, and `errno`s.

We would need issues for 2 - 4 above.

densh:
@ekrich Feel free to open more fine-grain issues.  

For point 3, we need to keep env variables set on the sbt sides as discussed earlier. Otherwise we're not testing that externally-set variables are visible inside SN. 

For point 4, we only expose 1-to-1 bindings over C library and POSIX. The only higher-level APIs build on top are reimplementation of `java.*`, the rest should go to a 3-rd party library space.

PR #1084, opened 2017-11-21
Devirtualize trait methods with 1 or less implementations

densh:
Previously, we would only statically devirtualize class methods. This PR addresses the discrepancy.


densh:
A nice side effect of this is that it decreases size of the dispatch table. This can cause noticeable binary size improvements for large apps.

PR #1094, opened 2017-11-25
clang-format update 5.0 and add docs for brew install

adam-singer:
* Updated the required version to 5.0 as its default to `brew install` on mac osx
* Reformatted c code with `./bin/clangfmt`

### DIFF
--- a/docs/user/setup.rst
+++ b/docs/user/setup.rst
@@ -56,6 +56,7 @@ Native has been used with:
 
     $ brew install llvm
     $ brew install bdw-gc re2 # optional
+    $ brew install clang-format # required if using ./bin/clangfmt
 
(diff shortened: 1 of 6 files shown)


docs/user/setup.rst
```diff
@@ -56,6 +56,7 @@ Native has been used with:
 
     $ brew install llvm
     $ brew install bdw-gc re2 # optional
+    $ brew install clang-format # required if using ./bin/clangfmt
```
densh: User guide mentions things necessary to use the latest released version of Scala Native, not build and develop it from source. Putting clang format in there might be confusing as it's not necessary from the end user point of view.

densh:

PR #1097, opened 2017-11-28
Implement DatagramSocket and related classes

RadoslawDryzner:
UDP blocking sockets for IPv4 implemented.
Complete classes from java.net:
* DatagramPacket
* DatagramSocket
* InterfaceAddress
* NetworkInterface

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -7,3 +7,6 @@ sbt "project nscplugin" ++2.11.11 publishSigned
 sbt "project nscplugin" ++2.11.8 publishSigned
 sbt "project sbtScalaNative" ^^0.13.16 publishSigned
 sbt "project sbtScalaNative" ^^1.0.4 publishSigned
+sbt "project nir" ^^1.0.4 publishSigned
+sbt "project tools" ^^1.0.4 publishSigned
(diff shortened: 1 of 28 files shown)


RadoslawDryzner:
To reproduce the bug, you can use this code:
```
for (i <- 0 until 1000) {
    NetworkInterface.getNetworkInterfaces()
}
```

RadoslawDryzner:
Signed the CLA.

jvican:
I'd like to experiment with this over at bloop. We currently have a python script that acts as the nailgun client, but python poses on us several limitations on how we read/write to the socket (it currently does it per line, but we'd like to read as much as there's in the buffer). On top of this, we'd like to have a cross-platform implementation of the plugin so that we can use the JVM one in JMH benchmarks and the native one for our actual users.
  
I'm interested in this getting merged so that we can dip our toes into Scala Native.

densh:
Hi @RadoslawDryzner, thanks for the contribution, but unfortunately we can't accept linux-specific code that doesn't run on macOS. Feel free to reopen the PR if you address this issue.

PR #1099, opened 2017-11-29
[WIP] Native bindings for termios

adam-singer:
Adding support for [termios](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/termios.h.html). Still left is to generate a useful example and align the types for the definitions, `CInt` might not be the right type when the specification says some values are `unsigned int` or `unsigned char`.


densh:
CI fails due to scalafmt.

densh:
Supersided by #1196.

PR #1045, opened 2017-10-02
Jenkins setup

Duhemm:

PR #1101, opened 2017-12-01
Fix #1032: Simplify stack trace element caching

### LINKED ISSUE #1032: Unit tests fail on MacOS with GC=immix
This regression was introduced in #963. The test runner crashes, for a reason that is still unknown to me. See http://lampmac7.epfl.ch:8080/blue/organizations/jenkins/scala-native-5/detail/PR-10/36/pipeline/5#step-89-log-1

Ref #995

densh:
Stack trace elements used to lazily convert from CString for the underlying native symbol to the corresponding class/method name. Now it's done eagerly, similarly to the original implementation before stack trace caching.

### DIFF
--- a/javalib/src/main/scala/java/lang/StackTraceElement.scala
+++ b/javalib/src/main/scala/java/lang/StackTraceElement.scala
@@ -1,92 +1,15 @@
 package java.lang
 
-import scala.scalanative.native.{
-  fromCString,
-  stackalloc,
-  string,
-  CChar,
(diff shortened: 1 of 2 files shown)


ekrich:
The code looks great. This is the fix for #1032? Were you able to test since we don't have macOS CI?

densh:
@ekrich I've run everything locally with immix and it works. New Mac CI is coming as part of #1045.

PR #1102, opened 2017-12-01
Fix #855: Add support for @volatile annotation

### LINKED ISSUE #855: Add support for @volatile annotation

pbatko:
This is still in a rough shape. I'm creating this PR to see if I'm going in the right direction.

1. In nscplugin
a. Add `JavaVolatile` to `@volatile` fields
b. Set `isAtomic` to load and stores from `@volatile` fields 

2. In codegen
a. emit stores and loads with `atomic seq_cst` and explicit alignment
`%_20 = load atomic i8, i8* %_28 seq_cst, align 8 `
`store atomic i8 %_21, i8* %_37 seq_cst, align 8`

3. A new pass
LLVM requires that atomic loads and stores operate on types with bit width 8 or higher and because booleans are represented as `i1` a new pass `scala.scalanative.optimizer.pass.VolatileBooleanToIntegerLowering`:
a. converts all `@volatile` boolean fields into byte fields,
b. fixs types of all loads and stores from/to `@volatile` boolean to operate on bytes,
c. inserts casts to/from booleans immediately after/prior loads/stores from/to byte fields.  

- [x] Write tests

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Attrs.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Attrs.scala
@@ -20,6 +20,7 @@ object Attr {
 
   final case object Pure                  extends Attr
   final case object Extern                extends Attr
+  final case object JavaVolatile          extends Attr
   final case class Override(name: Global) extends Attr
 
   final case class Align(value: Int) extends Attr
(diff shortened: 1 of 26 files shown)


densh:
Hi @pbatko, can you please rebase this PR on top of latest master?

pbatko:
@densh I'll do it during the weekend.
Do you think this feature is something we can include into scala-native release in the near future?
If so, is there anything we should add/change for this feature to be releasable?

densh:
@pbatko I don't see any reason not to include it, as long as we're careful about NIR format versioning and stay compatible with older NIR files (I'll comment on the details in the next review pass).

pbatko:
@densh rebased, and 3/4 ci builds succeded, 4th one seems to be hanged

WojciechMazur:
Outdated, @pbatko  thank you for your work in that area, your PR was used a reference for more current implementation.

PR #1106, opened 2017-12-04
Fix #1078: Fix '%' overflow on division by -1

### LINKED ISSUE #1078: Array(-2147483648).map(_ % -1) crash
- https://travis-ci.org/xuwei-k/scala-native-1078/builds/300520927
- https://github.com/xuwei-k/scala-native-1078/commit/272c45821646289a2eabacfcfffae2788c6847d3

### build.sbt

```scala
scalaVersion := "2.11.11"

```

### project/plugin.sbt

```scala
addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.3.3")
```

### Main.scala

```scala
package example

object Main {
    println(Array(-2147483648).map(_ % -1))
  }
}
```

### project/build.properties

```
sbt.version = 0.13.16
```

### result

```
[info] Compiling 1 Scala source to /home/travis/build/xuwei-k/scala-native-1078/target/scala-2.11/classes...
[info] 'compiler-interface' not yet compiled for Scala 2.11.11. Compiling...
[info]   Compilation completed in 15.097 s
[info] Linking (2377 ms)
[info] Discovered 1268 classes and 9348 methods
[info] Optimizing (6315 ms)
[info] Generating intermediate code (1788 ms)
[info] Produced 40 files
[info] Compiling to native code (3906 ms)
[info] Linking native code (120 ms)
java.lang.RuntimeException: Nonzero exit code: 136
	at scala.sys.package$.error(package.scala:27)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$scalaNativeConfigSettings$14$$anonfun$apply$16$$anonfun$apply$17$$anonfun$apply$18.apply(ScalaNativePluginInternal.scala:430)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$scalaNativeConfigSettings$14$$anonfun$apply$16$$anonfun$apply$17$$anonfun$apply$18.apply(ScalaNativePluginInternal.scala:430)
	at scala.Option.foreach(Option.scala:236)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$scalaNativeConfigSettings$14$$anonfun$apply$16$$anonfun$apply$17.apply(ScalaNativePluginInternal.scala:430)
	at scala.scalanative.sbtplugin.ScalaNativePluginInternal$$anonfun$scalaNativeConfigSettings$14$$anonfun$apply$16$$anonfun$apply$17.apply(ScalaNativePluginInternal.scala:415)
	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
[trace] Stack trace suppressed: run last compile:run for the full output.
[error] (compile:run) Nonzero exit code: 136
```

pbatko:
Inserts check for divisor being -1 and in such case changes it to 1.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/optimizer/pass/SRemOverflow.scala
+++ b/tools/src/main/scala/scala/scalanative/optimizer/pass/SRemOverflow.scala
@@ -1,0 +1,57 @@
+package scala.scalanative.optimizer.pass
+
+import scala.scalanative.nir
+import scala.scalanative.nir._
+import scala.scalanative.optimizer.analysis.ClassHierarchy.Top
+import scala.scalanative.optimizer.{Pass, PassCompanion}
+import scala.scalanative.tools.Config
+
+/**
+ * Detects taking remainder for division by -1 and replaces it by division by 1 which can't overflow.
+ *
+ *
+ * We implement '%' (remainder) with LLVM's 'srem' and it can overflow for cases:
+ * Int.MinValue % -1
+ * Long.MinValue % -1
+ * E.g. On x86_64 'srem' might get translated to 'idiv' which computes both quotient and remainder at once
+ * and quotient can overflow.
+ */
+class SRemOverflow(implicit top: Top) extends Pass {
+
+  override def onInsts(insts: Seq[Inst]): Seq[Inst] = {
+    val buf = new nir.Buffer
+    import buf._
+
+    insts.foreach {
+      case Inst.Let(name,
+                    sremBin @ Op.Bin(Bin.Srem, intType: Type.I, _, divisor))
+          if intType.width == 32 || intType.width == 64 =>
+        val safeDivisor         = Val.Local(fresh(), intType)
+        val thenL, elseL, contL = fresh()
+
+        val isPossibleOverflow =
+          let(Op.Comp(Comp.Ieq, intType, divisor, Val.Int(-1)))
+        branch(isPossibleOverflow, Next(thenL), Next(elseL))
+
+        label(thenL)
+        jump(contL, Seq(Val.Int(1)))
+
+        label(elseL)
+        jump(contL, Seq(divisor))
+
+        label(contL, Seq(safeDivisor))
+
+        let(name, sremBin.copy(r = safeDivisor))
+
+      case other => buf += other
+    }
+
+    buf.toSeq
+  }
+}
+
+object SRemOverflow extends PassCompanion {
+
+  def apply(config: Config, top: Top): Pass = new SRemOverflow()(top)
+
+}
--- a/unit-tests/src/test/scala/scala/PrimitiveSuite.scala
+++ b/unit-tests/src/test/scala/scala/PrimitiveSuite.scala
@@ -50,4 +50,43 @@ object PrimitiveSuite extends tests.Suite {
     val y: Long = 33
     assert((x << y) == 6)
   }
+
+  test(" x % y doesn't overflow (noinline)") {
+    assert(intRemByNegOne(4) == 0)
+    assert(intRemByNegOne(Int.MinValue) == 0)
+    assert(longRemByNegOne(4) == 0L)
+    assert(longRemByNegOne(Long.MinValue) == 0L)
+
+    // prevent partial evaluation
+    @noinline
+    def intRemByNegOne(i: Int): Int = {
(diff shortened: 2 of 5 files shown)


tools/src/main/scala/scala/scalanative/optimizer/pass/SRemOverflow.scala
```diff
@@ -0,0 +1,68 @@
...
+    def matchSrem(IntType: Type,
+                  makeVal: Int => Val): PartialFunction[Inst, Unit] = {
+
+      case Inst.Let(name, sremBin @ Op.Bin(Bin.Srem, IntType, _, divisor)) =>
+        val safeDivisor         = Val.Local(fresh(), IntType)
+        val thenL, elseL, contL = fresh()
+
+        val isPossibleOverflow = let(
+          Op.Comp(Comp.Ieq, IntType, divisor, makeVal(-1)))
+        branch(isPossibleOverflow, Next(thenL), Next(elseL))
```
sjrd: @densh Are there conditional moves in your IR?

If yest, this should definitely be a conditional move instead of a conditional branch.

densh: LLVM doesn't expose conditional moves explicitly. AFAIK instruction selection does pick cmov if it feels like it.

pbatko: In the release mode we are indeed getting a conditional move:

```
  def foo(value: Long, divisor: Long): Long = {
    value % divisor
  }
```

```
	.type	"Test$::foo_i64_i64_i64",@function
"Test$::foo_i64_i64_i64":               # @"Test$::foo_i64_i64_i64"
# BB#0:                                 # %_4.0
	cmp	rdx, -1
	mov	ecx, 1
	cmovne	rcx, rdx
	mov	rax, rsi
	cqo
	idiv	rcx
	mov	rax, rdx
	ret
```

unit-tests/src/test/scala/scala/PrimitiveSuite.scala
```diff
@@ -50,4 +50,22 @@ object PrimitiveSuite extends tests.Suite {
...
+
+  // use method to prevent partial evaluation
+  def intRemByNegOne(i: Int): Int = {
+    i % -1
+  }
+
+  // use method to prevent partial evaluation
+  def longRemByNegOne(l: Long): Long = {
+    l % -1L
+  }
```
densh: Actually we want to test both the case when partial evaluation happens and the case when it doesn't. To do this you can define two duplicate definitions with the same body, one marked as `@inline` and another one as `@noinline`.

densh: Those small methods should probably be defined within `test("...") { ... }` as well.

tools/src/main/scala/scala/scalanative/optimizer/pass/SRemOverflow.scala
```diff
@@ -0,0 +1,68 @@
...
+class SRemOverflow(implicit top: Top) extends Pass {
+
+  override def onInsts(insts: Seq[Inst]): Seq[Inst] = {
+    val buf = new nir.Buffer
+    import buf._
+
+    def matchSrem(IntType: Type,
+                  makeVal: Int => Val): PartialFunction[Inst, Unit] = {
+
+      case Inst.Let(name, sremBin @ Op.Bin(Bin.Srem, IntType, _, divisor)) =>
```
densh: Instead of taking `IntType` as an argument, you can match using `case ... Type.I(width) ... if width == 32 || width == 64`. The whole helper function can then be hand inlined into foreach.

pbatko: Sure, done.

I'm still always constructing `Int.Val` and I was worried it wouldn't work for longs, but turns out it doesn't seem to matter.

unit-tests/src/test/scala/scala/PrimitiveSuite.scala
pbatko: Done

pbatko:
@densh Can we merge it?

densh:
LGTM!

PR #1111, opened 2017-12-06
Add emit assembly files setting

pbatko:


pbatko:
@ekrich That's the option from Clang 6 which hasn't been released yet. :(
Latest release is 5.0.1 but the docs keep time outing for me http://releases.llvm.org/5.0.1/tools/clang/docs/UsersManual.html

Fyi, I'm on Fedora 27 (~2 monts old) and have clang 4.0.1

densh:
Hi @pbatko, thanks for the feature suggestion! 

Since this PR, the sbt plugin has gone significant refactoring to modularize most of the functionality into build tools API that doesn't depend on sbt itself. It's meant to be used by other build tools. This implies that we only expose *core* functionality, and don't include nice-to-have helpers as we used to before (they would have to be duplicated in all build tools for consistency). 

Such helpers should go into 3-rd party sbt plugins instead. Sbt plugins can depend on other plugins and/or build API to reuse code.

pbatko:
@densh Do you think we could have such feature only within scala-native project? 
I made this PR because I felt I need this feature when I working on some scala-native issue.

densh:
@pbatko It would make sense to define it within our build (i.e. wihtin project/ folder), as long as we don't distribute and officially support this functionality. The same applies for #1228, it would be much better to have it as feature purely for the development of Scala Native itself.

PR #1112, opened 2017-12-08
user.home must be set independently of user.language

densh:


ekrich:
Good catch 👍

densh:
This got caught by our new (still slightly WIP) Jenkins CI.

PR #1114, opened 2017-12-11
Keep using unailableBlockCount in ShouldGrow

densh:
PR #1085 changed growth heuristic not to account for
unavailableBlockCount. This regresses gcbench (a benchmark that
simulates high GC pressure). This commit brings it back.


densh:
Here are some pretty charts from 0.3.3:



Vs this PR in combination with #1085:



Better throughput, faster first iteration and less variance at peak performance.

ekrich:
Excellent 👍

PR #1117, opened 2017-12-12
Improve rebuild workflow and environment variable-based defaults

### LINKED ISSUE #1040: Allow easy switching of GC setting at sbt command line

ekrich:
This approach does not use a root project and addresses most of the issues raised in the first PR review. https://github.com/scala-native/scala-native/pull/1041

I would like to reduce the items in `rebuild` by using `publishLocal` dependencies for `scala` and `test-interface`. This will make developing Scala Native easier if changes span more than one project. `sbtScalaNative/publishLocal` depends on `tools/publishLocal` which in turn depends on `nir/publishLocal` and `util/publishLocal`.

### DIFF
--- a/docs/contrib/build.rst
+++ b/docs/contrib/build.rst
@@ -124,7 +124,7 @@ Setting this as follows will set the value in the plugin when `sbt` is run.
     > show nativeGC
 
 This setting remains unless changed at the `sbt` prompt. If changed, the value
-will be restored the the environment variable value if `sbt` is restarted or
+will be restored to the environment variable value if `sbt` is restarted or
 `reload` is called at the `sbt` prompt. You can also revert to the default
 setting value via `unset SCALANATIVE_GC` and then restarting `sbt`.
 
--- a/docs/contrib/build.rst
+++ b/docs/contrib/build.rst
@@ -1,0 +1,76 @@
+.. _build:
+
+Guide to the sbt build
+======================================
+
+This section gives some basic information and tips about the build system. The
+``sbt`` build system is quite complex and effectively brings together all the
+components of Scala Native. The ``build.sbt`` file is at the root of the project
+along with the sub-projects that make up the system.
+
+Overview
+--------------------------------
+In order to effectively work with Scala Native, a knowledge of the build system
+is very helpful. In general the code is built and published to your local Ivy
+repository so that other components in the system can depend on each other.
+Although the ``build.sbt`` file and other code in the system is the way to learn the system
+thoroughly, the following sections will give information that should be helpful.
+
+The build has roughly four groups of sub-projects as follows:
+
+1.  The Native Scala Compiler plugin and libraries. Each of these depend on the next project
+    in the list.
+
+    - `nscplugin`
+
+    - `nativelib`
+
+    - `javalib`
+
+    - `auxlib`
+
+    - `scalalib`
+
+
+2.  The Scala Native plugin and dependencies (directory names are in parentheses).
+
+    - `sbtScalaNative (sbt-scala-native)`
+
+    - `tools`
+
+    - `nir`
+
+    - `util`
+
+3.  The Scala Native test interface and dependencies.
+
+    - `testInterface (test-interface)`
+
+    - `testInterfaceSerialization (test-interface-serialization)`
+
+    - `testInterfaceSbtDefs (test-interface-sbt-defs)`
+
+4.  Tests and benchmarks (no dependencies on each other).
+
+    - `tests (unit-tests)`
+
+    - `tools` This has tests within the project
+
+    - `(scripted-tests)`
+
+    - `benchmarks`
+
+Each of the groups above also depend on the previous group being compiled and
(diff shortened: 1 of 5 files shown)


ekrich:
This is ready for initial review. I would like to add a section to the developers guide if this PR gets the go ahead. Here are the improvements and advantages to the changes.

- No global settings, they are not needed. How to change settings will be added as proposed above.
- `rebuild` no longer uses the default `sbt` project for `publishLocal` so it does not create a local artifact. This also results in very much fewer calls to `publishLocal` - only once for each project. Before it called `publishLocal` again for many if not all the projects including the ones already published - mostly a no op but extra work/time anyway.
- `publishLocal` dependencies are now explicit for each project like they were for `sbtScalaNative` already. If you are working on `javalib` and `nativelib` at the same time, `javalib/publishLocal` will make sure dependencies such as `nativelib` are compiled and published first.
- The plugin now directly uses the `SCALANATIVE_GC` env var so all the extra code is taken out of the main build. This also means that projects using Scala Native are on equal footing and can use this feature when doing Travis or Jenkins builds.

ekrich:
@densh I'm pretty happy with this now. Scripted now works with nothing built.  I added some documentation for the build which also needs some review.

docs/contrib/build.rst
```diff
@@ -0,0 +1,159 @@
...
+Normal development workflow
+---------------------------
+Let us suppose that you wish to work on the `javalib` project to add some code
+or fix a bug. Once you make a change to the code, run the following command
+at the sbt prompt to compile the code and run the tests:
+
+.. code-block:: text
+
+    > javalib/publishLocal
+    > tests/test
```
pbatko: if there's a change in codegen, is `rebuild` the only option?

ekrich: This was suppose to cover the basics. The next section talks about the codegen but I have never worked on that myself. http://www.scala-native.org/en/latest/contrib/compiler.html

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -395,7 +395,7 @@ object ScalaNativePluginInternal {
       val paths     = apppaths.map(_.abs) ++ opaths
       val compile   = clangpp.abs +: (flags ++ paths)
 
-      logger.time("Linking native code") {
+      logger.time(s"Linking native code - $gc gc") {
```
densh: I'd change the format to `Linking native code ($gc gc)"`.

docs/contrib/build.rst
densh: If you change anything in `tools` (linker, optimizer, codegen), you need to `reload`, not `rebuild`. It's possible because we textually include code of sbt plugin and the toolchain. 

If you change `nscplugin`, `rebuild` is the only option.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -123,7 +123,7 @@ object ScalaNativePluginInternal {
     nativeLinkerReporter in NativeTest := (nativeLinkerReporter in Test).value,
     nativeOptimizerReporter := tools.OptimizerReporter.empty,
     nativeOptimizerReporter in NativeTest := (nativeOptimizerReporter in Test).value,
-    nativeGC := "boehm",
+    nativeGC := Option(System.getenv.get("SCALANATIVE_GC")).getOrElse("boehm"),
```
densh: Lets also add `SCALANATIVE_MODE` similarly to `SCALANATIVE_GC`.

docs/contrib/build.rst
```diff
@@ -0,0 +1,159 @@
+.. _build:
+
+Guide to the sbt build
```
densh: This is extremely valuable! 👍

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
densh: Lets also add similar change to explicitly print current mode in nativeOptimizeNIR:

```
logger.time("Optimizing ($mode mode)") { ... }
```

docs/contrib/build.rst
ekrich: Thanks!

docs/contrib/build.rst
ekrich: I included this in the docs as well.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
ekrich: Done.

densh:
LGTM! 🍰

PR #1067, opened 2017-10-26
Add -Wl and -dead_stripe option to release mode

logistark:
Add -Wl and -dead_stripe option to release mode.

This solves #1063

PR #1126, opened 2017-12-18
Add strip options

### LINKED ISSUE #1063: Strip binaries from debug and symbol information in `Release` mode
As reported in #789 by @DarkDimius, the binaries that Scala Native generates are larger than need be, and we could automatically strip debug and symbols information from the generated binary when compiling in `Release` mode. 

We support Clang 3.7 and later. First, we should figure whether Clang 3.7 supports `-Wl` to pass options to the linker. If not, it would be interesting to know in what version of Clang it was introduced (and consider dropping support for earlier Clang versions).

When in Release mode, we should adapt the options that we pass to Clang to include `-Wl,-s`. This should strip all the unnecessary information from the binary that LLVM generates.

# Notes
- [Definition of the `nativeLinkLL` task](https://github.com/scala-native/scala-native/blob/cc9b8b8a5e9869e25671195565f1d0b8bdd063ac/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L363-L403)
- [How to strip binaries from unnecessary information in LLVM?](https://stackoverflow.com/questions/6085491/gcc-vs-clang-symbol-stripping)

logistark:
Hi,
i screwed PR #1067 that resolves #1063.

So i have created a new PR solving it.

Still, i tried to create a setting named "nativeStripOpts". 
[Here](https://github.com/logistark/scala-native/commit/d4505826b8d937448f110d7f16940e6c53705eb3) is a commit where i try to add this, but seems that have problems with sbt.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -51,6 +51,10 @@ object ScalaNativePlugin extends AutoPlugin {
 
     val nativeGC =
       settingKey[String]("GC choice, either \"none\", \"boehm\" or \"immix\".")
+
+    val nativeEnableStrip =
+      settingKey[Boolean]("Enable stripping of debug symbols in release mode")
+
   }
 
   override def globalSettings: Seq[Setting[_]] =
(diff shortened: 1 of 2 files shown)


logistark:
I have put behind a setting as @densh suggested. Anyway i am not sure, how this should look like.

Names i have used here for settings could be better, what do you think about this?

densh:
Hi @logistark, thanks for the contribution!

As discussed in https://github.com/scala-native/scala-native/pull/1111#issuecomment-402197193, we've changed our policy on what should go in the main sbt plugin vs 3-rd party plugins since the introduction of the build API. 

Unfortunately, this PR falls in the same category, it's a nice-to-have feature, but it should go into a 3-rd party plugin instead of being built-in.

PR #1132, opened 2018-01-03
Add support for CFunctionPtr casts

densh:
Previously they were not supported due to a bug nscplugin.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
@@ -714,72 +714,33 @@ object CodeGen {
 
     def genCall(genBind: () => Unit, call: Op.Call)(
         implicit fresh: Fresh): Unit = call match {
-      case Op.Call(ty, Val.Global(pointee, _), args, Next.None)
-          if lookup(pointee) == ty =>
-        val Type.Function(argtys, _) = ty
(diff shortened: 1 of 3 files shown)


tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -714,7 +714,8 @@ object CodeGen {
 
     def genCall(genBind: () => Unit, call: Op.Call)(
         implicit fresh: Fresh): Unit = call match {
-      case Op.Call(ty, Val.Global(pointee, _), args, Next.None) =>
+      case Op.Call(ty, Val.Global(pointee, _), args, Next.None)
+          if lookup(pointee) == ty =>
```
pbatko: Can we merge these
```
case Op.Call(ty, Val.Global(pointee, _), args, Next.None) =>
case Op.Call(ty, ptr, args, Next.None) =>
```
and these
```
case Op.Call(ty, Val.Global(pointee, _), args, unwind) =>
case Op.Call(ty, ptr, args, unwind) =>
```
?

The only differences seem to be prefixing pointee with either `@` or `%` and optional bitcast

densh: Good call, fixed in 41a732f.

PR #1135, opened 2018-01-06
Add support for extern objects inside other classes

shadaj:
Before, creating an `@extern` object inside a method or another class would result in linking errors, since the constructor for the object would not be eliminated. This changes the constructor for `@extern` objects to return `null`, which works since the constructor is never used, instead of not being generated.


sjrd:
That should be disallowed instead. There should be a compile error when an extern object is inside a class or trait. That's what happens if you have an `@js.native` object inside a class in Scala.js.

At the user level, in such a case, you should declare the extern object as private in the companion object of the class you're writing instead.

shadaj:
@sjrd The reason I had to make this change was to support generating extern definitions in a def macro (in this case for creating bindings to JNI libraries). Is it possible to generate bindings some other way in the macro that doesn't require creating an extern object?

densh:
**Workaround**: use macro annotations instead of def macros (as mentioned by @shadaj on gitter).

PR #1143, opened 2018-01-10
Build API 

Duhemm:
This is still WIP.

I'm opening this pull request to start discussing about the API that should emerge for using the Scala Native toolchain. Expected users of this API are build tools.

At the moment, the following steps have been made:
 - [x] Extract all logic out of the sbt plugin to the tools module.
 - [x] Move to NIO
 - [x] Add more fields to `Config`, so that signatures are not too long.

Next steps:
 - [ ] Have a convincing story for setting up the native library
 - [ ] Hide implementation details, provide more helper functions.
 - [ ] Make sure we don't do too much work. Currently we rely on sbt to simplify the task graph, but that may not be the case for all users of this API.
 - [ ] ...


sjrd:
You may want to have a look at https://github.com/scala-js/scala-js/issues/2998, where we discussed the "ideal" API for the Scala.js linker. It is also a good idea to consider what a CLI needs to do to use the API. Usually CLIs highlight the simplest use cases. For example, for Scala.js: https://github.com/scala-js/scala-js-cli/blob/dd2e0f3c2ea49d7db79cb13ce129cd9a5520df7a/src/main/scala/org/scalajs/cli/Scalajsld.scala#L155-L182

tools/src/main/scala/scala/scalanative/llvm/package.scala
```diff
@@ -0,0 +1,216 @@
...
+import java.nio.file.{Files, Path, Paths}
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+import scala.sys.process._
+
+import tools.{Config, GarbageCollector, IO, LinkerResult, Logger, Mode}
+import IO.RichPath
+
+package object llvm {
```
densh: This one should go within `scalanative.build` namespace.

tools/src/main/scala/scala/scalanative/tools/package.scala
```diff
@@ -38,38 +38,33 @@ package object tools {
...
    *  assumption.
    */
-  def link(config: Config,
-           driver: OptimizerDriver,
-           reporter: LinkerReporter = LinkerReporter.empty): LinkerResult = {
-    val chaDeps  = optimizer.analysis.ClassHierarchy.depends
-    val passDeps = driver.passes.flatMap(_.depends).distinct
-    val deps     = (chaDeps ++ passDeps).distinct
-    val injects  = driver.passes.flatMap(_.injects)
+  def link(config: Config): LinkerResult = {
```
densh: Lets move `tools` API into `scalanative.build` namespace.

tools/src/main/scala/scala/scalanative/build/package.scala
```diff
@@ -0,0 +1,148 @@
...
+      tools.optimize(config, linkerResult.defns, linkerResult.dyns)
+    val generated = {
+      tools.codegen(config, optimized)
+      IO.getAll(config.workdir, "glob:**.ll")
+    }
+    val objectFiles = llvm.compileLL(config, generated, logger)
+    val unpackedLib = unpackNativeLibrary(config.nativeLib, config.workdir)
+
+    val nativeLibConfig =
+      config.withCompileOptions("-O2" +: config.compileOptions)
```
densh: Passing an extra O2 option is duplicated here and in sbt plugin. There is a risk that it's changed in one place and not the other, lets refactor this.

tools/src/main/scala/scala/scalanative/tools/Config.scala
```diff
@@ -1,80 +1,197 @@
...
+  def withClang(value: Path): Config
+
+  /** Create a new config with given path to clang++. */
+  def withClangPP(value: Path): Config
 
   /** Create a new config with given target triple. */
   def withTarget(value: String): Config
 
-  /** Create a new config with given compilation mode. */
-  def withMode(value: Mode): Config
```
densh: Why was mode removed here?

tools/src/main/scala/scala/scalanative/tools/Config.scala
```diff
@@ -1,80 +1,197 @@
...
+  def nativeLib: Path
+
+  /** The driver to use for the optimizer. */
+  def driver: OptimizerDriver
+
+  /** The reporter that receives messages from the linker. */
+  def linkerReporter: LinkerReporter
+
+  /** The reporter that receives messages from the optimizer. */
+  def optimizerReporter: OptimizerReporter
```
densh: Having pluggable `OptimizerReporter` and `LinkerReporter` leaks NIR into the build API. We need to replace with something less powerful. For example: we can add a boolean option that turns on an off current to-file reporters.

tools/src/main/scala/scala/scalanative/tools/Config.scala
```diff
@@ -1,80 +1,197 @@
...
+
 import nir.Global
 
 sealed trait Config {
 
+  /** Path to the nativelib jar. */
+  def nativeLib: Path
+
+  /** The driver to use for the optimizer. */
+  def driver: OptimizerDriver
```
densh: Having customizable optimizer driver implies leaking NIR into the API. If it's not customizable it might as well be removed from the config object (and materialized on demand in the build api implementation).

densh: With `passes` field commented out, it's not leaking NIR any more, but referencing to optimizer-internal class in a public API is still suboptimal.

densh:
There is a merge conflict in `project/build.sbt`. Please rebase on top of latest master.

ekrich:
Do you think there is any way that the compile portion of the following could be added to this PR? It is difficult to add `C` code to user libraries.

**The sbt plugin should offer a way to compile C files** https://github.com/scala-native/scala-native/issues/657

I really like the idea of the `src/main/c` directory. Currently we have the `c` code in `src/main/resources` with some special logic to not include `optional` and unused `gc` code. Maybe the `c` code in Scala Native `nativelib` could be moved to a `src/main/c` directory.

lihaoyi:
Currently Mill's Scala.js' integration is only about 50 lines https://github.com/lihaoyi/mill/blob/master/scalajslib/jsbridges/1.0/src/mill/scalajslib/bridge/ScalaJSBridge.scala

(This doesn't count the plumbing for wiring up a Scala.js "Module", but such wiring will always exist independent of whatever compiler you are using)

Ideally I'd be able to get Mill-ScalaNative support for another 50 lines or less :smile:

Andrei-Pozolotin:
just f.y.i:
* an eager user of this WIP: [scalor-maven-plugin](https://github.com/random-maven/scalor-maven-plugin)
* currently having to use [copy/paste](https://github.com/random-maven/scalor-maven-plugin/blob/develop/src/main/scala/com/carrotgarden/maven/scalor/scalanative/Native.scala)
* please let me know how can I help

tools/src/main/scala/scala/scalanative/build/llvm/package.scala
```diff
@@ -0,0 +1,217 @@
...
+   */
+  def discover(binaryName: String,
+               binaryVersions: Seq[(String, String)]): Path = {
+    val docSetup =
+      "http://www.scala-native.org/en/latest/user/setup.html"
+
+    val envName =
+      if (binaryName == "clang") "CLANG"
+      else if (binaryName == "clang++") "CLANGPP"
+      else binaryName
```
Andrei-Pozolotin: please make configurable `clang` etc. perhaps could use typesafe config.

tools/src/main/scala/scala/scalanative/build/llvm/package.scala
```diff
@@ -0,0 +1,217 @@
...
+            throw new Exception(
+              s"no ${binaryNames.mkString(", ")} found in $$PATH. Install clang ($docSetup)")
+          }
+      }
+    }
+  }
+
+  /** Versions of clang which are known to work with Scala Native. */
+  val clangVersions =
+    Seq(("5", "0"), ("4", "0"), ("3", "9"), ("3", "8"), ("3", "7"))
```
Andrei-Pozolotin: please make configurable `clang` etc. perhaps could use typesafe config.

tools/src/main/scala/scala/scalanative/build/llvm/package.scala
```diff
@@ -0,0 +1,217 @@
...
+          s"at $pathToClangBinary.\nSee http://scala-native.readthedocs.io" +
+          s"/en/latest/user/setup.html for details.")
+    }
+  }
+
+  /** Default compilation options passed to LLVM. */
+  lazy val defaultCompileOptions: Seq[String] = {
+    val includes = {
+      val includedir =
+        Try(Process("llvm-config --includedir").lines_!.toSeq)
```
Andrei-Pozolotin: please make configurable `llvm-config` etc. perhaps could use typesafe config.

tools/src/main/scala/scala/scalanative/build/llvm/package.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  /** Compile the given LL files to object files */
+  def compileLL(config: Config,
+                llPaths: Seq[Path],
+                logger: Logger): Seq[Path] = {
+    val optimizationOpt =
+      config.driver.mode match {
+        case Mode.Debug   => "-O0"
+        case Mode.Release => "-O2"
+      }
```
Andrei-Pozolotin: please make configurable `-O0 -O2` etc. perhaps could use typesafe config.

tools/src/main/scala/scala/scalanative/build/llvm/package.scala
```diff
@@ -0,0 +1,217 @@
...
+    }
+    includes :+ "-Qunused-arguments"
+  }
+
+  /** Default options passed to LLVM's linker. */
+  lazy val defaultLinkingOptions: Seq[String] = {
+    val libs = {
+      val libdir =
+        Try(Process("llvm-config --libdir").lines_!.toSeq)
+          .getOrElse(Seq.empty)
```
Andrei-Pozolotin: please make configurable `llvm-config` etc. perhaps could use typesafe config.

Andrei-Pozolotin:
few notes, from "build integrator" point of view:

1) please externalize all process invocations commands into templates,
say with help of typesafe config https://github.com/lightbend/config

2) it seems there are no provisions in this PR for cache and incremental building, 
i.e. compare to `scala-js` [IRFileCache.scala](https://github.com/scala-js/scala-js/blob/master/linker/shared/src/main/scala/org/scalajs/linker/irio/IRFileCache.scala)

3) again, for cache / incremental reasons, please split `build/Config.scala` into
* static, tools-settings-profile (so that configured pipeline can be cached), 
* and dynamic, project-settings-profile, which can accept incremental delta `*.nir`
i.e. see in `scala-js` [StandardLinker.scala/Config](https://github.com/scala-js/scala-js/blob/master/linker/shared/src/main/scala/org/scalajs/linker/StandardLinker.scala) and `scalor` [scalajs.Cacher.scala](https://github.com/random-maven/scalor-maven-plugin/blob/develop/src/main/scala/com/carrotgarden/maven/scalor/scalajs/Cacher.scala)

.travis.yml
```diff
@@ -53,7 +53,8 @@ script:
       "^^ $SBT_VERSION" \
       rebuild \
       "set scriptedBufferLog in sbtScalaNative := false" \
-      test-all
+      test-all \
+      "tools/mimaReportBinaryIssues"
```
densh: We should probably include mima in `test-all` instead.

bin/release
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 
+set -ex
+
 sbt rebuild
```
densh: `sbt rebuild test-all`

util/src/main/scala/scala/scalanative/io/VirtualDirectory.scala
```diff
@@ -29,28 +28,28 @@ sealed trait VirtualDirectory {
 object VirtualDirectory {
```
densh: Given that VirtualDirectory appears in the `scalanative.build` it becomes part of its public API. Virtual I/O for Scala Native is actually not so virtual after all as LLVM command we invoke require everything to be physically present on disc. We should consider dropping these abstractions altogether, they are not pulling their own weight.

densh:
Also needs a rebase due to conflicting changes on master.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -184,24 +149,39 @@ object ScalaNativePluginInternal {
...
       )
-      val classpath = fullClasspath.value.map(_.data).filter(_.exists)
-      val entry     = nir.Global.Top(mainClass.toString + "$")
-      val cwd       = nativeWorkdir.value
-
-      tools.Config.empty
+      val classpath =
+        fullClasspath.value.map(_.data.toPath).filter(f => Files.exists(f))
+
+      val nativeLibJar =
```
densh: Given that subproject is called `nativelib`, I'd avoid capitalization for `Lib` here and everywhere else in this PR (when referring to `{native, scala, java, aux}lib`).

tools/src/main/scala/scala/scalanative/build/Logger.scala
```diff
@@ -0,0 +1,61 @@
...
+  def warn(msg: String): Unit
+
+  /** Logs `msg` at the errro level. */
+  def error(msg: String): Unit
+
+  /** Logs at the debug level that the command `cmd` will start running. */
+  def running(cmd: Seq[String]): Unit
+
+  /** Executes `f` and logs at the info level how much long it took. */
+  def time[T](msg: String)(f: => T): T
```
densh: `running` and `time` don't have to be abstract here. It would be cleaner to move implementation for those (in `new Logger { ... }`) into the trait itself.

tools/src/main/scala/scala/scalanative/build/Logger.scala
```diff
@@ -0,0 +1,61 @@
...
+}
+
+object Logger {
+
+  def default: Logger =
```
densh: Needs a docstring.

tools/src/main/scala/scala/scalanative/build/Logger.scala
```diff
@@ -0,0 +1,61 @@
...
+    Logger(msg => err.println(s"[debug] $msg"),
+           msg => out.println(s"[info] $msg"),
+           msg => out.println(s"[warn] $msg"),
+           msg => err.println(s"[error] $msg"))
+
+  def apply(debugFn: String => Unit,
```
densh: Needs a docstring.

tools/src/main/scala/scala/scalanative/build/Logger.scala
```diff
@@ -0,0 +1,61 @@
...
+      import java.lang.System.nanoTime
+      val start = nanoTime()
+      val res   = f
+      val end   = nanoTime()
+      info(s"$msg (${(end - start) / 1000000} ms)")
+      res
+    }
+  }
+
+  def toProcessLogger(logger: Logger): ProcessLogger =
```
densh: Needs a docstring.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -0,0 +1,211 @@
+package scala.scalanative
+package build
+
+import java.nio.file.{Path, Paths}
+
+import nir.Global
+
+sealed trait Config {
```
densh: Needs a docstring.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -0,0 +1,211 @@
...
+  /** Create a new config with given behavior for stubs. */
+  def withLinkStubs(value: Boolean): Config
+
+  /** Create a new config with the given logger. */
+  def withLogger(value: Logger): Config
+}
+
+object Config {
+
+  def default(nativeLib: Path,
```
densh: Needs a docstring. Additionally, it's not very clear when one should use `empty` and when `default`.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -0,0 +1,215 @@
...
+          Process(compile, config.workdir.toFile) ! Logger.toProcessLogger(
+            config.logger)
+          Paths.get(outpath)
+        }
+        .seq
+        .toSeq
+    }
+  }
+
+  def linkLL(config: Config,
```
densh: Needs a docstring.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -0,0 +1,215 @@
...
+
+    if (!clangIsRecentEnough) {
+      throw new Exception(
+        s"No recent installation of clang found " +
+          s"at $pathToClangBinary.\nSee http://scala-native.readthedocs.io" +
+          s"/en/latest/user/setup.html for details.")
+    }
+  }
+
+  /** Default compilation options passed to LLVM. */
```
densh: I'd say `clang` instead of LLVM here.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -0,0 +1,215 @@
...
+    val includes = {
+      val includedir =
+        Try(Process("llvm-config --includedir").lines_!.toSeq)
+          .getOrElse(Seq.empty)
+      ("/usr/local/include" +: includedir).map(s => s"-I$s")
+    }
+    includes :+ "-Qunused-arguments"
+  }
+
+  /** Default options passed to LLVM's linker. */
```
densh: `system linker` instead of `LLVM's linker`.

tools/src/main/scala/scala/scalanative/build/package.scala
```diff
@@ -0,0 +1,208 @@
...
+  def codegen(config: Config, assembly: Seq[nir.Defn]): Seq[Path] = {
+    config.logger.time("Generating intermediate code") {
+      scalanative.codegen.CodeGen(config, assembly)
+    }
+    val produced = IO.getAll(config.workdir, "glob:**.ll")
+    config.logger.info(s"Produced ${produced.length} files")
+    produced
+  }
+
+  def build(nativeLib: Path,
```
densh: Needs a big fat docstring. We expect this to be the main entry point so it should be very clear and very prominent that it's the one you want to use if you're implementing Scala Native support in your build tool. It would be nice to put this higher up in the file as well.

tools/src/main/scala/scala/scalanative/build/package.scala
```diff
@@ -0,0 +1,208 @@
...
+    if (!unpacked) {
+      IO.deleteRecursive(lib)
+      IO.unzip(nativeLib, lib)
+      IO.write(jarhashPath, jarhash)
+    }
+
+    lib
+  }
+
+  def compileNativeLib(config: Config,
```
densh: Needs a docstring.

tools/src/main/scala/scala/scalanative/build/GarbageCollector.scala
```diff
@@ -0,0 +1,25 @@
+package scala.scalanative.build
+
+/**
+ * @param dir directory name of the gc
+ * @param links dependencies of the gc
+ */
+sealed abstract class GarbageCollector(val name: String,
```
densh: Lets shorten name of this one to just `GC`.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -178,86 +157,13 @@ object ScalaNativePluginInternal {
...
-      nativeLinkLL.value
+      val logger  = streams.value.log.toLogger
+      val config  = nativeConfig.value.withLogger(logger)
+      val outpath = (artifactPath in nativeLink).value
+
+      build.build(config, outpath.toPath)
+      outpath
     },
     run := {
       val env    = (envVars in run).value.toSeq
```
densh: Wow, just 169 lines in sbt plugin, that's awesome! 👏

tools/src/main/scala/scala/scalanative/tools/Config.scala
Duhemm: The `Reporter`s no longer leak NIR.

tools/src/main/scala/scala/scalanative/tools/Config.scala
Duhemm: Is it okay as it is now?

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -0,0 +1,215 @@
...
+          case (major, minor) =>
+            Seq(s"$binaryName$major$minor", s"$binaryName-$major.$minor")
+        } :+ binaryName
+
+        Process("which" +: binaryNames)
+          .lines_!(SilentLogger)
+          .map(Paths.get(_))
+          .headOption
+          .getOrElse {
+            throw new Exception(
```
densh: We should have a dedicated exception type for exceptions thrown with build api. Something along the lines of `sealed class BuildException(...) extends Exception(...)`.

tools/src/main/scala/scala/scalanative/build/package.scala
```diff
@@ -0,0 +1,210 @@
...
+  type OptimizerDriver = optimizer.Driver
+  val OptimizerDriver = optimizer.Driver
+
+  type OptimizerReporter = optimizer.Reporter
+  val OptimizerReporter = optimizer.Reporter
+
+  def build(nativelib: Path,
+            paths: Seq[Path],
+            entry: String,
+            target: Path,
```
densh: We already have `target` in Config that means something different (LLVM platform target). Lets pick a different name for this parameter to avoid potential confusion.

ekrich:
@Duhemm It looks one of the outputs on linking may be missing from this PR. https://github.com/scala-native/scala-native/pull/1117 

 I see this code. https://github.com/Duhemm/scala-native/blob/topic/build-api/tools/src/main/scala/scala/scalanative/build/package.scala#L106
`Optimizing (debug mode)`

I am not seeing this one for the GC. https://github.com/Duhemm/scala-native/blob/topic/build-api/tools/src/main/scala/scala/scalanative/build/LLVM.scala#L203
`Linking native code (immix gc)`'

Duhemm:
@ekrich Otherwise sbt is going to show the full stack trace on every error.

ekrich:
@Duhemm I guess I'm just not sure how the stack trace gets suppressed but I get the point. In `sbt`
```
final class MessageOnlyException(override val toString: String) extends RuntimeException(toString)
```

Duhemm:
@ekrich See also:
https://github.com/sbt/sbt/blob/05c2c506b2b218d8ed8befcef843951391c07be6/main-command/src/main/scala/sbt/ExceptionCategory.scala#L20-L27
https://github.com/sbt/sbt/blob/34d311f9cef2732d0427060fd7c8e704a46cba52/main/src/main/scala/sbt/EvaluateTask.scala#L306-L314

ekrich:
@Duhemm Very helpful - thanks. They do some pretty tricky stuff in `sbt`.

densh:
Excellent work, @Duhemm ! 👏

I have a few more minor suggestions but I'll open a follow-up PR for those.

ajrnz:
Is there any way `test-interface-serialization` could be published for 2.12? At the moment there is an sbt hack which just includes the source files in the plugin. I'm looking at integrating scala-native into mill and at the moment I just have to copy the serialization code. The files in package `scala.scalanative.sbtplugin.testinterface` (`ComRunner` etc) have the same issue.

PR #1145, opened 2018-01-17
Add support for java.util.Base64

xuwei-k:
copy from scala-js
- https://github.com/scala-js/scala-js/blob/6169a651ad/javalib/src/main/scala/java/util/Base64.scala
- https://github.com/scala-js/scala-js/blob/6169a651ad/test-suite/shared/src/test/require-jdk8/org/scalajs/testsuite/javalib/util/Base64Test.scala

### DIFF
--- a/javalib/src/main/scala/java/util/Base64.scala
+++ b/javalib/src/main/scala/java/util/Base64.scala
@@ -1,0 +1,596 @@
+package java.util
+
+import scala.annotation.tailrec
+
+import java.io._
+import java.nio.charset.StandardCharsets
+import java.nio.ByteBuffer
+
+object Base64 {
+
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/util/Base64.scala
```diff
@@ -0,0 +1,596 @@
+package java.util
```
densh: Please add '// Ported from Scala.js' here.

unit-tests/src/test/scala/java/util/Base64Suite.scala
```diff
@@ -0,0 +1,611 @@
+package java.util
```
densh: Please add '// Ported from Scala.js' here.

PR #1147, opened 2018-01-18
Fix #1087: Use sbt-platform-deps to provide `%%%`, drop sbt-crossproject.

### LINKED ISSUE #1087: Depend on sbt-platform-deps instead of sbt-crossproject
New architecture for cross projects decouples sbt-crossproject from the underlying platform even further. Now, we should only depend on [sbt-platform-deps](https://github.com/portable-scala/sbt-platform-deps), instead of the sbt-crossproject.

sjrd:
This means that sbt-scala-native does not directly provide `crossProject` nor `NativePlatform` anymore. Those will be provided by sbt-scala-native-crossproject, to be added to the portable-scala/sbt-crossproject repository.

See also the corresponding PR that was made in Scala.js: https://github.com/scala-js/scala-js/pull/3197


densh:
Thanks a lot @sjrd, this looks great! 👍

PR #1156, opened 2018-02-02
Fix #1155: Don't add $ to impl class names

### LINKED ISSUE #1155: Cannot link a nested object if it's called CLASS
```scala
trait C {
  def foo = "bar"
}

object C {
  object CLASS extends C
}

object Hello extends App {
  println(C.CLASS)
}
```

```
> run
[info] Compiling 1 Scala source to /Users/eburmako/Projects/tmp/target/scala-2.11/classes...
[info] Linking (1089 ms)
[error] cannot link: @C$CLASS$
[error] unable to link
[error] (compile:nativeLinkNIR) unable to link
[error] Total time: 1 s, completed Feb 1, 2018 2:13:04 PM
```

densh:


sjrd:
You might want to know that this change is not binary compatible.

densh:
Indeed, I'm currently trying to come up with a binary compatibility workaround.

nir/src/main/scala/scala/scalanative/nir/Global.scala
```diff
@@ -67,4 +67,16 @@ object Global {
...
+      name
+    case Global.Top(id) =>
+      if (id.endsWith("$class$")) {
+        Global.Top(id.substring(0, id.length - 1))
+      } else {
+        name
+      }
+    case Global.Member(subname, id) =>
+      Global.Member(stripImplClassTrailingDollar(subname), id)
+  }
```
densh: @sjrd So a potential workaround: if a symbol name ends with "$class$", automatically strip the last dollar. The file name still contains a dollar but the tree that's deserialized from it is not going to have it. I tried it with library from 0.3.6 and a 0.3.7-compiled application and it seems to work.

sjrd: Seems plausible.

PR #1160, opened 2018-02-07
Add autoImport alias to preserve build source compatibility

densh:
In #1147 we renamed `AutoImport` to `autoImport`. This breaks
source compatibility for existing builds. This commit adds
a deprecated alias `autoImport` instead.


densh:
As discovered by @xeno-by while porting sScalameta.

PR #1162, opened 2018-02-08
add java.lang.System.{setIn, setOut, setErr}

xuwei-k:
https://docs.oracle.com/javase/8/docs/api/java/lang/System.html


densh:
LGTM, thanks @xuwei-k !

PR #1082, opened 2017-11-16
Implement #746 Add support for java.lang.{Process, ProcessBuilder}

eatkins:
It took a surprisingly long time for me to get this working reliably due to the lack of thread support in scala-native. There is discussion of this in the commit with message "Implement #746..."

I tested these changes quite extensively with a simple script that created some new processes, read the input streams and waited on the results. I would run the executable in a bash loop and often the script would hang because of deadlock. At this point, I can't get it to lock up, but concurrency is hard and I don't doubt there are some edge cases I've missed.

I view the process monitoring portion of this PR as a partial solution until scala native has full thread support.

I also tested with @lihaoyi's [ammonite](http://ammonite.io/) (which was why I wanted to implement this in the first place) and everything was working fine.

PR #1170, opened 2018-02-21
Time suite

eatkins:


eatkins:
For whatever reason, I didn't see junk with zone.alloc, but I did with manual stdlib.malloc. At any rate, I added a helper to zero out the tm struct.

eatkins:
I think it might make sense to do what is done for stat and create an alternative tm struct (in native c code, i.e. struct scalanative_tm) that only contains the platform independent fields: https://github.com/scala-native/scala-native/blob/master/nativelib/src/main/resources/stat.c

With that approach, we can expose only the standard portions of time.h (i.e. the fields that are in the existing struct defined in time.scala). If users want the extended functionality provided by their os, they can add their own native bindings.

Also, I am not at all committed to this PR. I just wanted an updated version of #1082 to pass against master. I only spent about two hours on it.

eatkins:
Another approach to handle the size of the struct that should work is to define a time.c file and add
`size_t scalanative_sizeof_tm() { return sizeof(tm); }`
We could also add accessors
`int scalanative_tm_sec(tm *t) { return t->tm_sec; }`, etc.
The user api ends up being pretty kludgy though because you have to do things like
`val tmPtr: Ptr[tm] = stackalloc[scala.Byte](tm.getSize()).cast[Ptr[tm]]`
It makes more sense for things like pthread where there is no standard size for the opaque types like pthread_cond_t.

densh:
Hi @eatkins, thanks for the contribution! Please run `bin/scalafmt` to unblock the CI.

densh:
This PR also needs a rebase on top of latest master due to conflicting changes in e03a85c914bcda57f99647d0b3b9d0e671711831

eatkins:
Whoops. Fixed everything up.

eatkins:
Hmm... The ProcessSuite occasionally hangs in testing. My guess is that it's somehow related to signal handling. I'll have to look into that a bit further.

densh:
@eatkins The failure looks spurious, we do get those on travis from time to time.

Andrei-Pozolotin:
+1

nativelib/src/main/scala/scala/scalanative/native/time.scala
```diff
@@ -1,44 +1,76 @@
...
+    def clock(): clock_t                                       = extern
+    def ctime_r(time: Ptr[time_t], buf: Ptr[CChar]): CString   = extern
+    def difftime(time_end: CLong, time_beg: CLong): CDouble    = extern
+    @name("scalanative_gmtime_r")
+    def gmtime_r(time: Ptr[time_t], tm: Ptr[tm]): Ptr[tm] = extern
+    @name("scalanative_localtime_r")
+    def localtime_r(time: Ptr[time_t], tm: Ptr[tm]): Ptr[tm] = extern
+    @name("scalanative_mktime")
+    def mktime(time: Ptr[tm]): time_t = extern
+    def strftime(str: Ptr[CChar],
```
Andrei-Pozolotin: missing `@name("scalanative_strftime")` ?

eatkins: Good catch. Should be fixed now.

nativelib/src/main/scala/scala/scalanative/native/time.scala
```diff
@@ -1,44 +1,77 @@
...
-  def gmtime_s(time: Ptr[time_t], result: Ptr[tm]): Ptr[tm]    = extern
-  def localtime(time: Ptr[time_t]): Ptr[tm]                    = extern
-  def localtime_s(time: Ptr[time_t], result: Ptr[tm]): Ptr[tm] = extern
-  def mktime(time: Ptr[tm]): time_t                            = extern
+  private final val timeBufferSize = 26 // taken from `man asctime`
+  @inline private def zeroAlloc[T](size: CSize)(implicit z: Zone): Ptr[T] = {
+    val buf = z.alloc(size)
+    string.memset(buf, 0, size)
+    buf.cast[Ptr[T]]
+  }
```
densh: In bindings we try to stay 100% close to the original signatures and APIs in C. We do not introduce extra parameters or helper methods. It's not the goal of bindings to increase usability of C code, it's the goal of the javalib or 3-rd party projects to build higher-level abstractions.

nativelib/src/main/scala/scala/scalanative/native/time.scala
```diff
@@ -1,44 +1,77 @@
...
-  def tzname(): Ptr[CStruct2[CString, CString]] = extern
+               time: Ptr[tm]): CSize            = impl.strftime(str, count, format, time)
+  def time(arg: Ptr[time_t]): time_t            = impl.time(arg)
+  def tzset(): Unit                             = impl.tzset()
+  def daylight(): CInt                          = impl.daylight()
+  def timezone(): CLong                         = impl.timezone()
+  def tzname(): Ptr[CStruct2[CString, CString]] = impl.tzname()
+
+  @extern
+  private object impl {
```
densh: `object time` itself has to be extern with all of its methods beings extern. As mentioned above bindings should stay true to the original C API, not provide an additional layer of abstraction.

nativelib/src/main/scala/scala/scalanative/native/time.scala
eatkins: Ok, if that's strictly the aim of the project, that's fine. I was just put off by the functions that returned naked pointers to static buffers.

nativelib/src/main/scala/scala/scalanative/native/time.scala
eatkins: Is there a technical reason for this or is it just the convention that scala native has adopted?

eatkins:
Sort of related, should this time.scala file be moved to nativelib/src/main/scala/scala/scalanative/posix/ instead of nativelib/src/main/scala/scala/scalanative/native/?

densh:
@eatkins Good point, it should be part of posix subpackage.

nativelib/src/main/scala/scala/scalanative/native/time.scala
densh: We need raw bindings and we also need higher-level abstractions. By separating the two we make it clear which is which. Some people may require access to lowest-level APIs without any abstractions, while others might prefer to have ergonomics-oriented higher-level version of it. Both need to exist.

eatkins: That makes sense. Posix is what it is, for better and worse.

eatkins:
I added a second commit that moves the file and cleans up some redundant declarations.

densh:
LGTM, thanks for the contribution @eatkins ! 👍

PR #1179, opened 2018-02-28
Fix dependencies of `testingCompiler`

Duhemm:
It doesn't need to have all the dependencies on all the libraries and
the scalac plugin.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -469,7 +469,7 @@ lazy val benchmarks =
 lazy val testingCompilerInterface =
   project
     .in(file("testing-compiler-interface"))
-    .settings(libSettings)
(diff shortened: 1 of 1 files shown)


ekrich:
The change from `projectSettings` to `libSettings` takes the tail of the settings - https://github.com/scala-native/scala-native/blob/master/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L71-L73

This drops exactly that dependency if I am looking at it correctly.

densh:
I'm closing this PR for now due to the problem mentioned earlier. Feel free to reopen once it's resolved.

Issue #1065, opened 2017-10-26
Support more classes from java.util.function Package


Duhemm:
These classes are being used by more and more classes of the JDK, and are therefore more likely to be used in Scala projects (they already are used in sbt or Dotty for instance). We will need these classes to port any project that uses them.

Please let me know what classes you want to implement and I’ll add your name next to it, so that we don’t duplicate work :)

Classes from the `java.util.function` package:
 - [ ] BiConsumer
 - [x] BiFunction
 - [x] BiPredicate
 - [ ] BinaryOperator
 - [ ] Consumer
 - [x] Function
 - [x] Predicate
 - [x] Supplier @jiminhsieh 
 - [x] UnaryOperator @jiminhsieh 

# Notes
 - Implementation of `java.util.function.Function`, for inspiration
 - [Javadoc for `java.util.function`](https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html)

jiminhsieh:
I want to tackle those 6 classes, but I am really busy. I think what I can do is I can pick two of them every time. If anyone tries to do this too, then they can pick rest of them.  

@Duhemm I want to do `Supplier` and `UnaryOperator` first. Thanks! :)

veera83372:
Hey, I can spend some time on it.Can I pick 2 classes from that list?

pbatko:
@Duhemm Looking at `BiFunction`

jiminhsieh:
@veera83372 @pbatko Do you still want to work on the rest? Otherwise, I will finish the rest.

pbatko:
@jiminhsieh Please take over

PR #1187, opened 2018-03-07
[WIP] Add Supplier.scala and UnaryOperator.scala.

jiminhsieh:
Fix part of #1065.
I am still thinking about unit test, so I labeled it as WIP.

### DIFF
--- a/javalib/src/main/scala/java/util/function/Supplier.scala
+++ b/javalib/src/main/scala/java/util/function/Supplier.scala
@@ -1,0 +1,5 @@
+package java.util.function
+
+trait Supplier[T] {
+  def get(): T
+}
--- a/javalib/src/main/scala/java/util/function/UnaryOperator.scala
+++ b/javalib/src/main/scala/java/util/function/UnaryOperator.scala
@@ -1,0 +1,12 @@
+package java.util.function
+
+trait UnaryOperator[T] extends Function[T, T] { self =>
+
+}
+
+object UnaryOperator {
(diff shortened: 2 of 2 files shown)


jiminhsieh:
I have a question about the unit test. Because of scala-native stays on 2.11, I found I couldn't write like this: 
```scala
  test("get") {
    val string: Supplier[String] = () => "scala"
    assert(string.get() == "scala")
  }
```
I need to write like this: 
```scala
  test("get") {
    val string: Supplier[String] = new Supplier[String] {
      override def get(): String = "scala"
    }
    assert(string.get() == "scala")
  }
```
My question is: Is second snippet fine for you?

Duhemm:
Hey @jiminhsieh,

Thanks for investing time into this! Yes, this is absolutely fine, we don't have a choice at the moment anyway 😄

nightscape:
@jiminhsieh I think you can enable SAM support in Scala 2.11 by adding `-Xexperimental` to the compiler options. See e.g. https://blog.jetbrains.com/scala/2015/07/16/try-experimental-sam-in-scala-plugin-1-7/

jiminhsieh:
@Duhemm Do you have any idea about this?
> I think you can enable SAM support in Scala 2.11 by adding -Xexperimental to the compiler options.

We could only add this compiler option to `javalib`. However, I don't know would this cause other side effects or not.

Adowrath:
Is there any update on this? I'd like to help with #1065, but with this PR being inactive, I don't know what the status of that Issue/"project" is right now.

PR #1190, opened 2018-03-08
Fix ProcessSuite hangs

eatkins:
There are a lot of spurious test failures on travis due to ProcessSuite
hanging or crashing. After some debugging on my mac, I determined that
the tests were likely to crash if the signal handler function was called
during garbage collection. This somewhat hacky fix is to check the
current stack in the signal handler and abort if garbage collection is
ongoing.

This should be fixed in the long run by using a scala monitor thread (as
opposed to the current native pthread) once #1068 is merged.


densh:
#1068 is a PR to multithreading branch, not master. It's probably going to take a while to get this all the way into master. We can not rely on it yet.

javalib/src/main/scala/java/lang/UnixProcess.scala
```diff
@@ -211,17 +218,21 @@ object UnixProcess {
...
-    val procInfo = stackalloc[ProcessMonitor.proc_info]
-    ProcessMonitor.lastProcInfo(procInfo)
-    val pid = !procInfo._1
-    processes get pid match {
-      case Some(process) =>
-        process.setExitValue(!procInfo._2)
-        processes -= pid
-        process.alertWaitingThreads()
-      case _ =>
+    if (ProcessMonitor.inGC != 0) {
```
densh: This doesn't look like a real solution, but just a workaround for the symptom of the problem.

eatkins: I agree that it is a workaround rather than a solution. At the same time, the entire monitoring implementation is a workaround not having threads. Since nothing is thread safe in scala native code, the monitor thread must communicate with the native thread using only data that isn't allocated by the gc. There is cleanup work on the process that must happen when the process exits that cannot be done on the monitor thread because it must modify scala native objects to set the exit value and cleanup the io streams. The existing hack is to interrupt the main thread to perform this work. This works well, except when GC is running.

As ugly as it is, I have found that this change does reliably make the processes work. I wrote a simple scala native project that more or less just called git status (and did some other work to help trigger garbage collection). Prior to this change, the script would crash roughly O(5%) of the time. After this change, the script ran 100000+ times without a crash.

I know that multithreading is still a ways out so I'm not relying on that at all. I'm just saying that whenever that is merged in, the process monitoring code should be re-implemented with scala native threads.

PR #1191, opened 2018-03-08
Resolve #1185.

### LINKED ISSUE #1185: BUG: proper packaging for `scalalib`
1. `scalalib` on central
http://search.maven.org/#artifactdetails%7Corg.scala-native%7Cscalalib_native0.3_2.11%7C0.3.6%7Cjar
* both depends on `scala-library`
```
        <dependency>
            <groupId>org.scala-lang</groupId>
            <artifactId>scala-library</artifactId>
            <version>2.11.12</version>
        </dependency>
```
* and also includes `scala-library` JVM classes `*.class`
```
# cd /tmp
# wget http://repo1.maven.org/maven2/org/scala-native/scalalib_native0.3_2.11/0.3.6/scalalib_native0.3_2.11-0.3.6.jar
# jar tf scalalib_native0.3_2.11-0.3.6.jar | grep "scala/Predef.class"
scala/Predef.class
```

2. this produces various problems in various build systems, for example
* errors in Scala IDE:
```
More than one scala library found in the build path
```
* due to `scalaLibraries` detection:
https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala#L264
```
val pathToPredef = new Path("scala/Predef.class")
```
3. one way to get proper packaging for `scalalib` is to follow `scala-js` approach:
http://search.maven.org/#artifactdetails%7Corg.scala-js%7Cscalajs-library_2.11%7C0.6.22%7Cjar
when only JS-VM `*.sjsir` classes are included in the library jar and JVM `*.class` classes are excluded from the library jar

Andrei-Pozolotin:
Don't include classfiles for scalalib in the packaged jar.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -398,6 +398,14 @@ lazy val scalalib =
       compile in Compile := (compile in Compile)
         .dependsOn(assembleScalaLibrary)
         .value,
+      // Don't include classfiles for scalalib in the packaged jar.
+      mappings in packageBin in Compile := {
+        val previous = (mappings in packageBin in Compile).value
+        previous.filter {
+          case (file, path) =>
+            !path.endsWith(".class")
+        }
+      },
       publishLocal := publishLocal
         .dependsOn(assembleScalaLibrary, publishLocal in auxlib)
         .value


build.sbt
```diff
@@ -398,6 +398,14 @@ lazy val scalalib =
       compile in Compile := (compile in Compile)
         .dependsOn(assembleScalaLibrary)
         .value,
+      // Don't include classfiles for scalalib in the packaged jar.
+      mappings in packageBin := {
```
densh: Builds fails at the moment. It seems that `in Compile` is missing here and on the next line, [similarly to the mapping for javalib](https://github.com/carrot-garden/scala_scala-native/blob/81d76e2c1fdb01698749bf9cd2445a567660bdd5/build.sbt#L321).

Andrei-Pozolotin: changed to `mappings in packageBin in Compile`

Andrei-Pozolotin:
@densh there is still issue with warnings for missing scaladoc links

densh:
@Andrei-Pozolotin Missing scaladoc is unrelated to this PR, we should resolve it separately.

densh:
LGTM, thanks! 👍

Andrei-Pozolotin:
thank you

PR #1192, opened 2018-03-08
Eclipse support: gitignore

Andrei-Pozolotin:
Eclipse support: gitignore: additional excludes entries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,10 @@ sbt-crossproject/
 
 CMakeLists.txt
 cmake-build-debug/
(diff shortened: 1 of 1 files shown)


Andrei-Pozolotin:
bada-boom: 
* sorry I just noticed the problem while looking at #1193
* need to revert this or rename `bin` -> `build` etc folder in source, 
* `bin` is special and is pretty much hard coded all over the place in eclipse

https://github.com/scala-native/scala-native/tree/master/bin

PR #1196, opened 2018-03-20
Add support for POSIX termios

ekrich:
POSIX definitions plus XSI features.

### DIFF
--- a/docs/lib/posixlib.rst
+++ b/docs/lib/posixlib.rst
@@ -80,7 +80,7 @@ C Header          Scala Native Module
 `sys/wait.h`_     N/A
 `syslog.h`_       scala.scalanative.posix.syslog_
 `tar.h`_          N/A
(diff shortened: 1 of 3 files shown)


ekrich:
I implemented the spec including extensions but both Linux and macOS support the iflag `IUTF8` and macOS below is using `OXTABS` and other non-standard stuff.
```
$ stty
speed 9600 baud;
lflags: echoe echoke echoctl pendin
iflags: iutf8
oflags: -oxtabs
cflags: cs8 -parenb
```
I was looking at JLine3 JNA layer too and it looks like it supports at least `IUTF8` but I'm not sure how much if any of this extra stuff would be needed to create something like JLine.
https://github.com/jline/jline3/blob/master/terminal-jna/src/main/java/org/jline/terminal/impl/jna/osx/CLibrary.java#L329

PR #1200, opened 2018-03-21
Resolve #1198: Fix FileChannel.read to have same semantics as JVM

### LINKED ISSUE #1198: FileChannelImpl only reads a single byte into ByteBuffers
Given code similar to the following:

    channel = FileChannel.open(path, StandardOpenOption.READ)
    val idBB = ByteBuffer.allocate(12)
    idBB.order(ByteOrder.LITTLE_ENDIAN)
    channel.read(idBB)
    idBB.flip()
    
    val idBA = idBB.array()
    print("ID BYTES: ")
    for (i <- 0 until idBA.length)
      print(s"${idBA(i)} ")
    println()

The JVM will output `ID BYTES: -85 75 84 88 32 49 49 -69 13 10 26 10`
Scala Native will output `ID BYTES: -85 0 0 0 0 0 0 0 0 0 0 0`

This discrepancy comes from the implementation of `javalib/src/main/scala/java/nio/channels/FileChannelImpl.scala` `def read(buffer: ByteBuffer, pos: Long)`

Particularly, where the `dst` Array is hardcoded as only having a single Byte on line 73.

    val dst = new Array[Byte](1)

dlpatri:
Scala-Native's FileChannel.read operations were only reading a Byte at a time. The JVM's reads from the `ByteBuffer.position() until ByteBuffer.limit()` or less, if there aren't that many bytes to read.

Added/Changed/Fixed tests in FilesSuite and FileChannelSuite to accommodate the changes.

### DIFF
--- a/javalib/src/main/scala/java/nio/channels/FileChannelImpl.scala
+++ b/javalib/src/main/scala/java/nio/channels/FileChannelImpl.scala
@@ -57,20 +57,25 @@ final class FileChannelImpl(path: Path,
                     start: Int,
                     number: Int): Long = {
     ensureOpen()
-    val dst = new Array[Byte](1)
-    val nb  = raf.read(dst)
-    var i   = 0
-    while (i < number) {
-      buffers(start + i).put(dst)
-      i += 1
+
+    var bread: Long = 0
+
+    for (i <- start until number) {
+      val len = buffers(i).limit() - buffers(i).position()
+      val dst = new Array[Byte](len)
+      val nb  = raf.read(dst)
+      if (nb > 0) buffers(i).put(dst)
+      bread += nb
     }
-    nb
+
+    bread
   }
 
   override def read(buffer: ByteBuffer, pos: Long): Int = {
     ensureOpen()
     position(pos)
-    val dst = new Array[Byte](1)
+    val len = buffer.limit() - buffer.position()
+    val dst = new Array[Byte](len)
     val nb  = raf.read(dst)
     if (nb > 0) buffer.put(dst)
     nb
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/nio/channels/FileChannelImpl.scala
```diff
@@ -57,22 +57,40 @@ final class FileChannelImpl(path: Path,
...
-    val dst = new Array[Byte](1)
-    val nb  = raf.read(dst)
-    var i   = 0
-    while (i < number) {
-      buffers(start + i).put(dst)
-      i += 1
+
+    var bread: Long = 0
+
+    for (i <- start until number) {
```
densh: Please use while loops instead of for comprehensions.

javalib/src/main/scala/java/nio/channels/FileChannelImpl.scala
```diff
@@ -57,22 +57,40 @@ final class FileChannelImpl(path: Path,
...
                     number: Int): Long = {
     ensureOpen()
-    val dst = new Array[Byte](1)
-    val nb  = raf.read(dst)
-    var i   = 0
-    while (i < number) {
-      buffers(start + i).put(dst)
-      i += 1
+
+    var bread: Long = 0
```
densh: Ambiguous name wrt 🥖.

dlpatri: Yeah, I thought it'd be a fun way to write bytesRead in way that enforces the correct reading of read (like "red", as opposed to "reed"). But I can change it.

ekrich: I looked at that too and said "what!". Then I figured maybe it was `bytesRead` but given your other short names it seemed out of place. I liked the 🍞 icon though. 😃

densh:
LGTM, thanks @dlpatri! 👍

PR #1201, opened 2018-03-23
Revise Build API

densh:
This PR adds a few final strokes to the Build API introduced in #1143 by @Duhemm:

1. Remove `OptimizerReporter`, `LinkerReporter` and `OptimizerDriver` from public API. Those are internal tools which were not meant to be a part of the API.

2. Remove `nativeMissingDependencies`, `nativeExternalDependencies`, `nativeAvailableDependencies`. Another group of accidental internal helpers leaking to the public API.

3. Refactor `package object build` into `object Build`. Unify two `build` methods into one.

4. Remove `Config.default` and expose and additional `Discover` object instead.

5. Hide concrete instances for `GC` and `Mode` to avoid falsely reassuring exhaustivity checks. 

Overall this boils down to the following invocation of the toolchain using the public parts of the API:

```scala
val classpath: Seq[Path] = ...
val workdir: Path        = ...
val main: String         = ...

val clang     = Discover.clang()
val clangpp   = Discover.clangpp()
val linkopts  = Discover.linkingOptions()
val compopts  = Discover.compileOptions()
val triple    = Discover.targetTriple(clang, workdir)
val nativelib = Discover.nativelib(classpath).get
val outpath   = workdir.resolve("out")

val config =
  Config.empty
    .withGC(GC.default)
    .withMode(Mode.default)
    .withClang(clang)
    .withClangPP(clangpp)
    .withLinkingOptions(linkopts)
    .withCompileOptions(compopts)
    .withTargetTriple(triple)
    .withNativelib(nativelib)
    .withMainClass(main)
    .withClassPath(classpath)
    .withLinkStubs(true)
    .withWorkdir(workdir)

Build.build(config, outpath)
```


Andrei-Pozolotin:
sbt/zinc calls `withClasspath` -> `withClassPath`

Andrei-Pozolotin:
@densh  summary:

1. based on limited experience of integrating scala-native pipeline in maven plugin
https://github.com/random-maven/scalor-maven-plugin/blob/develop/src/main/scala/com/carrotgarden/maven/scalor/scalanative/Native.scala
I must say that #1201 pipeline interface is nice "porcelain" api,  meaning
https://wiki.eclipse.org/JGit/User_Guide#Porcelain_API
and still insufficient for scalor project, which is also fine

2. perhaps alternative tools integration approach could be 
to expose more internal scala-native tools code with `warning: beware`

3. from practical integrator's point of view it is easier and faster 
to fix a breaking change coming from upstream
then dance around and beg to make some `private` feature `public` API, etc. 

4. please release more and faster, and do not look back :-)

densh:
> please release more and faster

Amen. We should release more often indeed.

Andrei-Pozolotin:
after looking through this few times
```
    .withGC(GC.default)
    .withMode(Mode.default)
    .withClang(clang)
```

`withMode` still sticks out as too generic

perhaps:  `withOutputMode` or `withBuildMode` or `withReleaseMode` or `withPipelineMode`

`withPipelineMode` could mean "propagate release-vs-debug distinction" 
to all steps/levels - `*.nir` generation, `LL` compile , etc
while other `modeXXX` could affect more fine-grained features

surely people will start asking for more and better `modes` :-)

densh:
@Andrei-Pozolotin Mode has been part of sbt's public API since the very beginning of the project. We should keep it.

Andrei-Pozolotin:
@densh or duplicate+deprecate, that is what `scala-js` is doing for "rename"

PR #1210, opened 2018-03-28
 Modularize nir parser out of tools

densh:
This removes fastparse dependency from tools jar.


build.sbt
```diff
@@ -231,19 +231,32 @@ lazy val nir =
...
     .settings(toolSettings)
     .settings(mavenPublishSettings)
     .settings(
       libraryDependencies ++= Seq(
-        "com.lihaoyi"    %% "fastparse"  % "0.4.2",
-        "com.lihaoyi"    %% "scalaparse" % "0.4.2",
         "org.scalacheck" %% "scalacheck" % "1.13.4" % "test",
-        compilerPlugin(
-          "org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
-        "org.scalatest" %% "scalatest" % "3.0.0" % "test"
```
Duhemm: I think that you can also remove this from `project/build.sbt`:

https://github.com/scala-native/scala-native/blob/682dab82c7c689d00ebe6d263685694994287ad4/project/build.sbt#L17-L20

build.sbt
```diff
@@ -231,19 +231,32 @@ lazy val nir =
     .settings(mavenPublishSettings)
     .dependsOn(util)
 
+lazy val nirparser =
```
Duhemm: Also, this project should be added to `test-all` and `scala-native/scripts/release`

build.sbt
densh: Indeed. 👍

build.sbt
densh: Good catch on including in `test-all`. We don't have to release it yet, as it's not part of the public API at the moment.

densh: a6e6257

build.sbt
densh: a6e6257

PR #1211, opened 2018-03-28
Fail fast at Travis CI.

jiminhsieh:
I noticed that wrong format of source code didn't stop early, it will continue to do testing. I think it would be better to fail fast.


densh:
I believe it's useful to independently run both formatting failure and test failures. Travis CI latency is quite long and it can save some time to see that tests pass but formatting doesn't.

PR #1212, opened 2018-03-29
Replace to new JVM option.

jiminhsieh:
In Java 8, `MaxPermSize` no longer exists, so I replaced with similar JVM option.


densh:
LGTM, thanks @jiminhsieh and congrats on first contribution! 👍

PR #1217, opened 2018-04-06
Fix regex semantics

eatkins:
When there is a capturing group with optional content and the content is
empty, the Matcher is supposed to return "", not null.


densh:
LGTM, thanks @eatkins ! 👍

Issue #1139, opened 2018-01-09
Proposal: no external dependencies in the core


densh:
We've seen numerous problems with having dependencies on 3-rd party native libraries in the core. This includes difficulties finding and installing latest required versions of all libraries. Additionally, it makes Go-style single-binary deployments harder, as you also need to have the same environment to make the application work.

Our current runtime dependencies are:

- [x] **Boehm GC.** We should make immix a default instead. #1137
- [x] **libunwind.** Include libunwind in the core, build it ourselves, and link statically with it. #1138
- [ ] **RE2.** Finish our port of re2j to Scala and make it the default. #873

densh:
So it looks RE2 is the last man standing on this one.

densh:
Last missing step was fixed in #1498.

PR #1218, opened 2018-04-06
Sys props

eatkins:
This PR makes the os.name, os.version and java.io.tmpdir system variables match the jvm on osx and linux.

### DIFF
--- a/javalib/src/main/scala/java/lang/System.scala
+++ b/javalib/src/main/scala/java/lang/System.scala
@@ -4,6 +4,8 @@ import java.io._
 import java.util.{Collections, HashMap, Map, Properties}
 import scala.scalanative.native._
 import scala.scalanative.posix.unistd
+import scala.scalanative.posix.sys.utsname._
+import scala.scalanative.posix.sys.uname._
 import scala.scalanative.runtime.time
 import scala.scalanative.runtime.Platform
 import scala.scalanative.runtime.GC
@@ -49,6 +51,14 @@ object System {
       sysProps.setProperty("user.country", userCountry)
 
     } else {
+      if (Platform.isMac) {
+        sysProps.setProperty("os.name", "Mac OS X")
+      } else {
+        val u = stackalloc[utsname]
+        uname(u)
+        sysProps.setProperty("os.name", u.sysname)
+        sysProps.setProperty("os.version", u.release)
+      }
       sysProps.setProperty("file.separator", "/")
       sysProps.setProperty("path.separator", ":")
       val userLocale = getenv("LANG")
@@ -72,6 +82,13 @@ object System {
 
     sysProps
   }
+
+  var in: InputStream =
+    new FileInputStream(FileDescriptor.in)
+  var out: PrintStream =
+    new PrintStream(new FileOutputStream(FileDescriptor.out))
+  var err: PrintStream =
+    new PrintStream(new FileOutputStream(FileDescriptor.err))
 
   private var systemProperties = loadProperties()
 
@@ -99,13 +116,6 @@ object System {
 
   def getenv(): Map[String, String] = envVars
   def getenv(key: String): String   = envVars.get(key)
-
-  var in: InputStream =
-    new FileInputStream(FileDescriptor.in)
-  var out: PrintStream =
-    new PrintStream(new FileOutputStream(FileDescriptor.out))
-  var err: PrintStream =
-    new PrintStream(new FileOutputStream(FileDescriptor.err))
 
   def setIn(in: InputStream): Unit =
     this.in = in
--- a/javalib/src/main/scala/java/lang/System.scala
+++ b/javalib/src/main/scala/java/lang/System.scala
@@ -51,14 +51,6 @@ object System {
       sysProps.setProperty("user.country", userCountry)
 
     } else {
-      if (Platform.isMac) {
-        sysProps.setProperty("os.name", "Mac OS X")
-      } else {
-        val u = stackalloc[utsname]
-        uname(u)
-        sysProps.setProperty("os.name", u.sysname)
-        sysProps.setProperty("os.version", u.release)
-      }
       sysProps.setProperty("file.separator", "/")
       sysProps.setProperty("path.separator", ":")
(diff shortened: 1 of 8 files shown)


nativelib/src/main/resources/osx_platform.mm
```diff
@@ -0,0 +1,20 @@
+#ifdef __APPLE__
+#include <Foundation/Foundation.h>
+typedef const char *CString;
+
+extern "C" {
+void scalanative_mac_osx_version(int *major, int *minor, int *patch) {
```
densh: Lets make those to be `scalanative_os_version` and `scalanative_os_tmp_dir` with platform-specific dispatch logic inside. The platform-dependant code is better left on the C side (with ifdefs where appropriate), so that Scala side remains portable.

javalib/src/main/scala/java/lang/System.scala
```diff
@@ -49,6 +51,26 @@ object System {
       sysProps.setProperty("user.country", userCountry)
 
     } else {
+      if (Platform.isMac) {
```
densh: I was not very clear on the previous comment, but I think this branch specifically should rather be in the C code. So that on Scala side we're left with something along the lines of:

```
        val name, tmpdir, version = ...
        Platform.osVersion(major, minor, patch)
        Platform.osName(name)
        Platform.osTmpDir(tmpdir)
        sysProps.setProperty("os.name", ...)
        sysProps.setProperty("os.version", ...)
        sysProps.setProperty("java.io.tmpdir", ...)
```
We should try avoiding platform-dependant branching in Scala code as much as possible (and keep those in C code instead).

ekrich: Yes, we started that platform branching early on and with only a couple of items and platforms it looked reasonable. Now, I agree that we should hide the platform details in the main code line.

nativelib/src/main/resources/platform.c
```diff
@@ -33,3 +46,79 @@ int scalanative_little_endian() {
     int n = 1;
     return (*(char *)&n);
 }
+
+#ifndef __APPLE__
+void scalanative_set_os_props(void (*add_prop)(const char *, const char *)) {
+#ifdef _WIN32
```
eatkins: Note that the windows code compiles, but I could not wire things up on my vm to actually get scala native to run on windows.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -162,7 +165,8 @@ private[scalanative] object LLVM {
       }
       linkerResult.links.map(_.name) ++ librt ++ libunwind ++ config.gc.links
     }
-    val linkopts = config.linkingOptions ++ links.map("-l" + _) ++ Seq(
+    val osxOpts = if (Properties.isMac) Seq("-framework", "Foundation") else Nil
+    val linkopts = config.linkingOptions ++ links.map("-l" + _) ++ osxOpts ++ Seq(
```
densh: It looks like we go from two dynamic dependencies for hello world:
```
# otool -L sandbox/target/scala-2.11/sandbox-out 
sandbox/target/scala-2.11/sandbox-out:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
	/usr/local/opt/llvm/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
```
To four after this change:
```
# otool -L sandbox/target/scala-2.11/sandbox-out 
sandbox/target/scala-2.11/sandbox-out:
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1349.91.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
	/usr/local/opt/llvm/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
```
Given our long-term strive to minimize number of dynamic dependencies (#1139), I don't think supporting macOS version string justifies the added cost. 

I believe that reporting empty string for version on macOS is a better alternative.

nativelib/src/main/resources/platform.c
densh: If we can't test it, I'm afraid it's a dead and unmaintained code that might or might not work in the future. I think we should cross that bridge when we get there, and not prematurely.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
eatkins: I agree that minimizing dynamic dependencies is a great idea. That being said, the two new libraries are not really external dependencies. They are system built-ins that as far as I can tell are guaranteed to be present on any apple platform: https://developer.apple.com/documentation/foundation?changes=_5&language=objc.

Also, I didn't really make this clear in the commit, but I was actually more concerned with the temp directory location than the os version string. I don't think there is any way other than using the framework api to get the correct user temp directory location. Of course, this is probably not a huge deal, `/tmp` still works, but it is not where Files.createTempDirectory puts the new temp directories on the jvm.

nativelib/src/main/resources/platform.c
eatkins: FWIW, I did add the following to platform.c
```c
static void print_prop(const char *name, const char *value) {
  printf("Added prop %s %s\n", name, value);
}
int main(int argc, char **argv) {
  scalanative_set_os_props(&print_prop)
}
```

And it printed:
```
Added os.name Windows 8
Added os.version 6.2
Added java.io.tmpdir C:\Users\ethan\AppData\Local\Temp
```

The os is actually wrong because of changes made to GetVersionEx: https://docs.microsoft.com/en-us/windows/desktop/SysInfo/targeting-your-application-at-windows-8-1. To fix, there would have to be a build step that adds a manifest to the generated binary. But at least the os.name prefix is correct and it gets the right temp directory.

densh:
LGTM, thanks for the contribution @eatkins ! 👏

PR #1219, opened 2018-04-06
Add shell fallback for processes

eatkins:
On the JDK, if the command cannot be exec'd, there is a fallback to
running the command using /bin/sh. It was an oversight that I didn't
originally add this fallback. This came up while I was working on
getting ammonite ops to run under scala native.


densh:
LGTM, thanks for another contribution @eatkins ! 👍

PR #1220, opened 2018-04-06
Add optimized split

eatkins:
It is very slow to have to compile a regex pattern for every call to
split so I provided an optimized version of split that is similar to the
versions provided by other jvm providers. In the case of a happy path
where the separator is a single character that isn't a regex special
character or when the separator is an escaped regex special character,
we use the optimized version.

I made a crude benchmark where I split the string "abcdefghijklnop" * 10
on the letter j 10000 times and the runtime for the fast version was 4us
per iteration while using Pattern.compile.split took 300us.

### DIFF
--- a/javalib/src/main/scala/java/lang/String.scala
+++ b/javalib/src/main/scala/java/lang/String.scala
@@ -738,11 +738,58 @@ final class _String()
   def replaceFirst(expr: _String, substitute: _String): _String =
     Pattern.compile(expr).matcher(this).replaceFirst(substitute)
 
+  def fastSplit(ch: Char, max: Int): Array[String] = {
+    var separatorCount = 0
+    var begin          = 0
+    var end            = 0
+    while (separatorCount + 1 != max && { end = indexOf(ch, begin); end != -1 }) {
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/lang/String.scala
```diff
@@ -733,11 +733,56 @@ final class _String()
...
+  @inline private def isRegexMeta(c: Char) =
+    REGEX_METACHARACTERS.indexOf(c) >= 0
+
   def split(expr: _String): Array[String] =
-    Pattern.compile(expr).split(this)
+    split(expr, 0)
 
   def split(expr: _String, max: Int): Array[String] =
-    Pattern.compile(expr).split(this, max)
+    if (isEmpty) Array("")
```
densh: Please put curly braces around then branch. We tend to either have no curlies at all (small/single-line ifs) or curlies for both then and else cases.

eatkins: Make sense. Done.

densh:
LGTM, thanks @eatkins ! 👍

PR #1222, opened 2018-04-07
Proposal: add shutdown hooks

eatkins:
This is a proposal PR to add shutdown hooks. In order to implement a few javalib apis, like File.deleteOnExit, we need to have a mechanism for running shutdown hooks. I'm not sure if this is the best approach, but it does seem to work.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/DeleteOnExit.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/DeleteOnExit.scala
@@ -1,0 +1,19 @@
+package scala.scalanative.runtime
+
+import scala.collection.mutable
+import scala.scalanative.native.{Zone, stdio, toCString}
+
+object DeleteOnExit {
+  private val toDeleteSet: mutable.Set[String] = mutable.Set.empty
+  private val toDelete: mutable.ArrayBuffer[String] =
+    mutable.ArrayBuffer.empty
+  Shutdown.addHook(() =>
+    toDelete.foreach { f =>
+      Zone { implicit z =>
+        stdio.remove(toCString(f))
+      }
+  })
+  def addFile(name: String) = toDelete.synchronized {
+    if (toDeleteSet.add(name)) toDelete += name
+  }
+}
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Shutdown.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Shutdown.scala
@@ -1,0 +1,24 @@
+package scala.scalanative.runtime
+
+import scala.collection.mutable
+import scala.scalanative.native._
+
+private[runtime] object Shutdown {
+  private val hooks: mutable.ArrayBuffer[() => Unit] = mutable.ArrayBuffer.empty
+  def addHook(task: () => Unit)                      = hooks.synchronized(hooks += task)
+  private def runHooks(): Unit =
+    hooks.foreach { task =>
+      try {
+        task()
+      } catch {
+        case e: Exception => // Maybe add a system propery that adds logging of exceptions?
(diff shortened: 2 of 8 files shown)


nativelib/src/main/scala/scala/scalanative/runtime/DeleteOnExit.scala
```diff
@@ -0,0 +1,22 @@
...
+  Shutdown.addHook(2, new Runnable() {
+    override def run() {
+      toDelete.foreach { f =>
+        Zone { implicit z =>
+          stdio.remove(toCString(f))
+        }
+      }
+    }
+  })
+  def add(s: String) = toDelete.synchronized {
```
densh: It would be more clear with more explicit names here `def deleteOnExit(path: String): Unit = ...`

nativelib/src/main/scala/scala/scalanative/runtime/DeleteOnExit.scala
```diff
@@ -0,0 +1,22 @@
+package scala.scalanative.runtime
+
+import scala.collection.mutable
+import scala.scalanative.native.{Zone, stdio, toCString}
+
+object DeleteOnExit {
+  private[this] val toDeleteSet: mutable.Set[String] = mutable.Set.empty
+  private[this] val toDelete: mutable.ArrayBuffer[String] =
```
densh: Just `private` should be sufficient.

nativelib/src/main/scala/scala/scalanative/runtime/Shutdown.scala
```diff
@@ -0,0 +1,35 @@
...
+  }
+  def addHook(id: Int, runnable: Runnable) =
+    hooks.synchronized(hooks.get(id) match {
+      case Some(runnables) => runnables += runnable
+      case None =>
+        val runnables = mutable.Set(runnable)
+        hooks += id -> runnables
+    })
+  private[this] val hooks: mutable.Map[Int, mutable.Set[Runnable]] =
+    mutable.Map.empty
```
densh: I don't think the `id` is necessary here. Lets just make it a `mutable.Seq[Runnable]`. Every time you add a hook it adds one more hook to run, in the order they were added.

eatkins: The `id` variable name was bad. It should have been priority. On the jvm, shutdown hooks have a priority that determine the order in which they run. For example, on the jvm the delete on exit hooks are always the last to run. The hooks added via `Runtime.addShutdownHook(Thread)` should run first. I didn't add that api because the current Thread implementation has a private constructor so there is no way for a user to add it at the moment. Leaving in the priority field, it would be trivial in the future to add `Runtime.addShutdownHook`. If that really isn't a concern, then we can go ahead and switch to a Seq now.

nativelib/src/main/scala/scala/scalanative/runtime/DeleteOnExit.scala
eatkins: `DeleteOnExit.deleteOnExit` sounds redundant to me. Perhaps `DeleteOnExit.addFile(name: String)`?

densh: That works as well.

nativelib/src/main/scala/scala/scalanative/runtime/DeleteOnExit.scala
eatkins: Yeah, I guess in this particular case there is no reason for private[this]. It's just a habit of mine.

densh:
LGTM, thanks @eatkins ! 👍

PR #1224, opened 2018-04-11
Match Scala JVM hash codes for case classes

shadaj:
Scala Native's `finalizeHash` didn't match Scala JVM, so hash codes for case classes were different between the two platforms. This updates it to use the same implementation.

### DIFF
--- a/auxlib/src/main/scala/scala/runtime/Statics.scala
+++ b/auxlib/src/main/scala/scala/runtime/Statics.scala
@@ -18,7 +18,7 @@ object Statics {
   }
 
   def finalizeHash(hash: Int, length: Int): Int =
-    avalanche(hash & length)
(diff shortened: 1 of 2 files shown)


shadaj:
Hmm, one of the builds failed due to a timeout. Appears unrelated to the changes.

auxlib/src/main/scala/scala/runtime/Statics.scala
```diff
@@ -18,7 +18,7 @@ object Statics {
   }
 
   def finalizeHash(hash: Int, length: Int): Int =
-    avalanche(hash & length)
```
densh: I wonder how did this ever happen. 😲

densh:
Restarted failing job, probably spurious.

PR #1228, opened 2018-04-15
Make optimizer reporter configurable in sbt again

pbatko:


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -25,6 +25,9 @@ import scalanative.sbtplugin.testinterface.ScalaNativeFramework
 
 object ScalaNativePluginInternal {
 
+  val optimizerReporterOutPath =
```
pbatko: Is it ok to have a key for the reporter back if it's in `ScalaNativePluginInternal`? @densh @Duhemm

ekrich: Does it make sense to have a build relative default directory?

ekrich: Well, we could have a switch to turn it on and off and have a project relative destination in the target directory. I think that would work for each project that uses the plugin. @Duhemm What do you think?

densh:
Optimizer reporter is a useful concept but currently it doesn't exist any more due to a complete optimizer rewrite. We need to redesign this feature from scratch as requested again in #1414.

PR #1234, opened 2018-04-27
Split test-runner out of sbt plugin

Duhemm:
This module contains the sources for the test adapter. It can be useful
for other tools such as Mill or Bloop that want to support running tests
with Scala Native projects.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,6 +54,7 @@ addCommandAlias(
   "dirty-rebuild",
   Seq(
     "scalalib/publishLocal",
+    "testRunner/publishLocal",
     "sbtScalaNative/publishLocal",
     "testInterface/publishLocal"
   ).mkString(";", ";", "")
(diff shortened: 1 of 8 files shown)


ajrnz:
Would it be possible to publish test-interface-serialization for 2.12 JVM? (Currently SBT just includes the sources in the pluging to handle the JVM side of the [serialization](https://github.com/scala-native/scala-native/blob/d1b17c78b7e1abca269122bc541c2d28026dc5e4/build.sbt#L315))

I'm looking at scala-native mill integration and currently have to copy this code which is less than ideal going forward.

lihaoyi:
@Duhemm just bumping this since it's a blocker for Scala-Native support in Mill and uPickle =)

Duhemm:
@ajrnz Sorry for taking so long to reply. In this PR, the `test-runner` artefact includes the sources for the serialisation. You should be able to use it it for Mill; it sure worked with Bloop 😄 : https://github.com/scalacenter/bloop/commit/84395b666eafdeae6f2f664a30a39fc86c6bce7d

ekrich:
@Duhemm Is this ready to be merged to SNAPSHOT after review?

Duhemm:
@ekrich Yes, I believe so. I'm not considering this as just WIP if that's what you're asking.

ajrnz:
Thanks @Duhemm. I can confirm that I have everything I need. I now have mill scala-native running against this PR. It's much smaller. Cheers.

Issue #104, opened 2016-05-18
Provide complete bindings for C11's libc


densh:
- [ ] assert.h
- [ ] complex.h
- [ ] ctype.h
- [ ] errno.h
- [ ] fenv.h
- [ ] float.h
- [ ] inttypes.h
- [ ] iso646.h
- [ ] limits.h
- [ ] locale.h
- [ ] math.h
- [ ] setjmp.h
- [ ] signal.h
- [ ] stdalign.h
- [ ] stdarg.h
- [ ] stdatomic.h
- [ ] stdbool.h
- [ ] stddef.h
- [ ] stdint.h
- [ ] stdio.h
- [ ] stdlib.h
- [ ] stdnoreturn.h
- [ ] string.h
- [ ] tgmath.h
- [ ] threads.h
- [ ] time.h
- [ ] uchar.h
- [ ] wchar.h
- [ ] wctype.h

densh:
That's a nice starting task for someone who wants to contribute some work. Just pick one of the header, go through all the functions and write signatures for them in terms of our native types that are defined in https://github.com/scala-native/scala-native/tree/master/nativelib/src/main/scala/scala/scalanative/native

TimothyKlim:
I'll try to finish this task until the end of the week.

mkotsbak:
Isn't this possible to automate using a script?

TimothyKlim:
I'm using own python + clang script to do this: https://gist.github.com/TimothyKlim/a65289901221451e515e408014eb3613

densh:
@TimothyKlim This looks great, would you be interested in expanding this script into full-blown tool for automatic binding generation? We'd love to home something like this in the project.

TimothyKlim:
@densh I can't promise but I'll try to complete this as a tool within this task.

TimothyKlim:
Which convention I need to use to declaring necessary constants in extern object? For example, I have `signal.h` constants as macros which cannot be extracted:
```c
#define	SIGHUP	1	/* hangup */
#define	SIGINT	2	/* interrupt */
#define	SIGQUIT	3	/* quit */
//...
```
and I can't declare these as `final val`s in extern object.

mkotsbak:
@TimothyKlim ah nice. Then you just need to run it on libclang to convert the script to Scala native :)

mkotsbak:
Hmm, regarding constants, maybe use Enumeration instead?

sjrd:
[...]

PR #1235, opened 2018-04-28
Add bindings for float.h

kornilova-l:
See list of all header files that should be supported in #104

float.h contains only macros, all of them are listed in this commit except macros for type `long double` (for example `LDBL_MANT_DIG`) because this type is not supported in Scala.


densh:
LGTM, thanks @kornilova-l ! 👍

PR #1238, opened 2018-05-01
Fix #1137: Make Immix the default garbage collector

### LINKED ISSUE #1137: Make immix a default garbage collector
We've introduced our own garbage collector in `0.3.0`. The overall experience with it has been extremely positive and we should make it the default. This is going to make Boehm GC and optional dependency that's only needed if you explicitly do `nativeGC := "boehm"`. 

/cc @LukasKellenberger

ekrich:
Change the default GC to Immix and fix a typo.

### DIFF
--- a/docs/user/sbt.rst
+++ b/docs/user/sbt.rst
@@ -12,21 +12,21 @@ The easiest way to make a fresh project is to use our official gitter8 template:
 
     sbt new scala-native/scala-native.g8
 
-This includes the following files with the proper version entered (denoted by <version>):
+This generates the following files:
 
 * ``project/plugins.sbt`` to add a plugin dependency::
 
-    addSbtPlugin("org.scala-native" % "sbt-scala-native" % "<scala-native-version>")
+    addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.3.7")
 
 * ``project/build.properties`` to specify the sbt version::
 
-    sbt.version = <sbt-version>
+    sbt.version = 0.13.17
 
 * ``build.sbt`` to enable the plugin and specify Scala version::
 
     enablePlugins(ScalaNativePlugin)
 
-    scalaVersion := "<scala-version>"
+    scalaVersion := "2.11.12"
 
 * ``src/main/scala/Main.scala`` with minimal application::
 
@@ -96,18 +96,18 @@ Scala Native supports two distinct linking modes:
 Garbage collectors
 ------------------
 
-1. **immix.** (default, since 0.3)
+1. **immix.** (default since 0.3.8, introduced in 0.3)
 
    Immix is a mostly-precise mark-region tracing garbage collector.
    More information about the collector is available as part of the original
    `0.3.0 announcement <https://github.com/scala-native/scala-native/releases/tag/v0.3.0>`_.
 
-2. **boehm.**
+2. **boehm.** (default through 0.3.7)
 
    Conservative generational garbage collector. More information is available
    at the `project's page <https://www.hboehm.info/gc/>`_.
 
-3. **none.** (experimental, since 0.2)
+3. **none.** (experimental, introduced in 0.2)
 
    Garbage collector that allocates things without ever freeing them. Useful
    for short-running command-line applications or applications where garbage
--- a/docs/user/sbt.rst
+++ b/docs/user/sbt.rst
@@ -12,21 +12,21 @@ The easiest way to make a fresh project is to use our official gitter8 template:
 
     sbt new scala-native/scala-native.g8
 
-This includes:
+This includes the following files with the <version> filled in:
 
 * ``project/plugins.sbt`` to add a plugin dependency::
 
-    addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.3.7")
+    addSbtPlugin("org.scala-native" % "sbt-scala-native" % "<version>")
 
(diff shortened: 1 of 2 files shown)


ekrich:
@densh Is the `none` GC still experimental?

docs/user/sbt.rst
```diff
@@ -107,7 +107,7 @@ Garbage collectors
    for short-running command-line applications or applications where garbage
    collections pauses are not acceptable.
 
-3. **immix.** (experimental, since 0.3)
+3. **immix.** (default, since 0.3)
```
densh: Lets reorder this list to make immix come first, boehm second and none last. None is still experimental.

docs/user/sbt.rst
```diff
@@ -12,21 +12,21 @@ The easiest way to make a fresh project is to use our official gitter8 template:
...
 
 * ``project/plugins.sbt`` to add a plugin dependency::
 
-    addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.3.7")
+    addSbtPlugin("org.scala-native" % "sbt-scala-native" % "<version>")
 
 * ``project/build.properties`` to specify the sbt version::
 
-    sbt.version = 0.13.17
+    sbt.version = <version>
```
densh: This is seems wrong. Sbt version can not the same as Scala Native version.

ekrich: Yes, agreed. See if the change makes more sense. I wanted to remove the versions that could easily get out of date.

docs/user/sbt.rst
ekrich: Sounds good.

docs/user/sbt.rst
```diff
@@ -12,21 +12,21 @@ The easiest way to make a fresh project is to use our official gitter8 template:
...
 
-    sbt.version = 0.13.17
+    sbt.version = <sbt-version>
 
 * ``build.sbt`` to enable the plugin and specify Scala version::
 
     enablePlugins(ScalaNativePlugin)
 
-    scalaVersion := "2.11.12"
+    scalaVersion := "<scala-version>"
```
densh: After having another look at this, I'm convinced this is guaranteed to confuse people. Previous thing was just copy-paste and run no questions asked and no thinking required. Suggested change leaves no info on what combination of versions is known to work/recommended for use with Scala Native. Concrete versions are much more clear, even if it have to update them manually at every release.

docs/user/sbt.rst
```diff
@@ -96,23 +96,23 @@ Scala Native supports two distinct linking modes:
 Garbage collectors
 ------------------
 
-1. **boehm.** (default)
+1. **immix.** (default, since 0.3)
```
densh: `1. **immix.** (default since 0.3.7, introduced in 0.3)`

docs/user/sbt.rst
```diff
@@ -96,23 +96,23 @@ Scala Native supports two distinct linking modes:
...
+
+   Immix is a mostly-precise mark-region tracing garbage collector.
+   More information about the collector is available as part of the original
+   `0.3.0 announcement <https://github.com/scala-native/scala-native/releases/tag/v0.3.0>`_.
+
+2. **boehm.**
```
densh: `2. **boehm.** (default up until 0.3.6)`

docs/user/sbt.rst
```diff
@@ -96,23 +96,23 @@ Scala Native supports two distinct linking modes:
...
 
    Conservative generational garbage collector. More information is available
    at the `project's page <https://www.hboehm.info/gc/>`_.
 
-2. **none.** (experimental, since 0.2)
+3. **none.** (experimental, since 0.2)
```
densh: `3. **none.** (experimental, introduced in 0.2)`

docs/user/sbt.rst
ekrich: Ok, this makes sense. When the `g8` is updated then the versions in this document need to be updated. You can't really change the `g8` until after the release or right when you make the release.

docs/user/sbt.rst
ekrich: @densh I think this needs to say `0.3.8` (the next release) so the next below should be `through 0.3.7`

docs/user/sbt.rst
densh: We typically change g8 as soon as release syncs to sonatype, but before announcing publicly, so people don't really perceive the gap between the two. I don't think it's an issue we should worry about.

docs/user/sbt.rst
densh: My bad, you're absolutely correct, it should have been `default since 0.3.8` here and `default up until 0.3.7` for boehm.

docs/user/sbt.rst
ekrich: I totally see the logic, I was trying to be fancy to avoid making changes but it could certainly be too confusing especially for new users.

docs/user/sbt.rst
ekrich: I used `through 0.3.7` as I think it is a bit more precise. We used boehm as the default version through `0.3.7` and then switched the default to immix for `0.3.8`. Hopefully you agree.

densh: Sounds good to me.

ekrich: I appreciate your attention to detail in these matters. Not only is documentation important but this also is a record of what happened.

PR #1269, opened 2018-07-10
Decouple s.c.i.Range from formatters

densh:
Here we reimplement Range::description method to use
string builder rather than formatter. This halfs number of
reachable methods on an empty project (from 9k to 4.5k).


densh:
Also decreases binary size for hello world down to 1.9M from 4M in 0.3.7.

PR #1282, opened 2018-07-20
Bundle LLVM libunwind instead of depending on system installation

shadaj:
Part 1 of supporting 32-bit platforms, needed because the installable libunwind versions are mostly broken on x86. In addition, this makes it easier to handle cross-compiling since libunwind compilation is now managed by Scala Native.


densh:
Needs a pass of scalafmt to unblock the CI.

shadaj:
Everything is green 🎉

PR #1286, opened 2018-07-25
Use hidden visibility for runtime code

densh:
This change allows LLVM's LTO to eliminate all
symbols which are not reachable after optimizations
(e.g., methods that are always inlined).

The main goal of this change is to reduce binary size.
Bundling libunwind into the runtime code increased the
hello world binary size to 3.3M in release mode. After
this change it gets down to 3.1M (compared to 2.1M before
bundling).


densh:
**Update:** unfortunately we can not make Scala methods hidden by default as that destroys the symbol name that's necessary for stack traces.

PR #1302, opened 2018-08-08
Run unit tests inside Docker containers

shadaj:
Second part of preparing Scala Native for 32bit platform support.


ci-docker/linux-arm64v8/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+  dpkg -i sbt-$SBT_LAUNCHER_VERSION.deb && \
+  rm sbt-$SBT_LAUNCHER_VERSION.deb && \
+  apt-get update && \
+  apt-get install -y sbt
+
+RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+
+RUN apt-add-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-5.0 main" && apt-get update
+
+RUN apt-get install -y clang-5.0
```
jonas: For scala-native-bindgen we install `llvm-5.0-dev`. From what I remember package will install `llvm-config` which I know is used by Scala Native. The LLVM tools are located in `/usr/lib/llvm-5.0/bin` so would have to be added to `PATH`.

shadaj:
@densh passing (only x86_64 for now) and ready for review!

densh:
@shadaj Is there any way to include platform name in the here:

screen shot 2018-09-02 at 18 26 13

Right now it's not obviously clear which platform failed/succeeded at a glance...

shadaj:
@densh fixed! I ended up going back to just a simple matrix build since splitting up the compile and test steps didn't end up speeding things up that much.

ekrich:
@shadaj Why isn't the `scripts/scalafmt --test` just run as it was in the past?

shadaj:
@ekrich since there are now more (and soon to be a lot more when ARM support is merged) matrix elements, I moved the `scalafmt` test to only run in the first matrix build, which runs all the compiler unit tests. All the other builds are platform specific so they only run the runtime tests.

shadaj:
Ping @densh I think this is ready for another review!

ekrich:
@shadaj I think that is a big improvement to do the source checks as suggested by @olafurpg

ci-docker/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+COPY --from=0 /usr/bin/qemu-* /usr/bin/
+
+RUN dpkg --add-architecture $HOST_DEBIAN_PLATFORM
+
+RUN apt-get update && apt-get install -y openjdk-8-jdk-headless:$HOST_DEBIAN_PLATFORM
+
+ENV PATH "/usr/lib/jvm/java-8-openjdk-$HOST_DEBIAN_PLATFORM/bin:$PATH"
+
+ENV SBT_LAUNCHER_VERSION 1.2.0
```
jonas: Would make sense to keep this in sync with the version in `project/build.properties`, i.e. 0.13.17?

ci-docker/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+ENV LD_LIBRARY_PATH "/usr/local/lib"
+
+ENV LC_ALL "C.UTF-8"
+
+ENV LANG "C.UTF-8"
+
+RUN useradd -ms /bin/bash scala-native
+
+USER scala-native
```
jonas: Will this ensure that the owner of files created inside Docker match the Travis CI user?

.travis.yml
```diff
@@ -8,60 +8,70 @@ dist: trusty
...
-  - scripts/scalafmt --test
-  - >
-    sbt \
-      "-no-colors" \
-      "-J-Xmx2G" \
-      "^^ $SBT_VERSION" \
-      rebuild \
-      "set scriptedBufferLog in sbtScalaNative := false" \
-      test-all
+  - if [ "$MODE" == "source-checks" ]; then
```
densh: Can we factor out the code to the scripts/ folder? It would let us use normal bash syntax without having it embedded in travis configuration.

ekrich: I would consider just copying what @olafurpg does here. https://github.com/scalameta/scalameta/tree/master/bin

If you add an argument to `checkCLA.sh` "Scala Native" and substitute in the script for "Scalameta", this script could be used on any project that needs to check the CLA.

.travis.yml
```diff
@@ -8,60 +8,70 @@ dist: trusty
...
-    - SCALANATIVE_GC=boehm SBT_VERSION=1.2.1
-    - SCALANATIVE_GC=immix SBT_VERSION=0.13.16
-    - SCALANATIVE_GC=immix SBT_VERSION=1.2.1
+    - MODE=source-checks
+    - MODE=compile-tests SBT_VERSION=0.13.16
+    - MODE=compile-tests SBT_VERSION=1.2.1
+    - MODE=runtime-tests TEST_CONTAINER=scala-native-testing:linux-x86_64 TARGET_DOCKER_PLATFORM=library SCALANATIVE_GC=boehm SBT_VERSION=0.13.16 TEST_COMMAND="test-runtime"
+    - MODE=runtime-tests TEST_CONTAINER=scala-native-testing:linux-x86_64 TARGET_DOCKER_PLATFORM=library SCALANATIVE_GC=boehm SBT_VERSION=1.2.1 TEST_COMMAND="sbtScalaNative/scripted"
+    - MODE=runtime-tests TEST_CONTAINER=scala-native-testing:linux-x86_64 TARGET_DOCKER_PLATFORM=library SCALANATIVE_GC=immix SBT_VERSION=0.13.16 TEST_COMMAND="test-runtime"
+    - MODE=runtime-tests TEST_CONTAINER=scala-native-testing:linux-x86_64 TARGET_DOCKER_PLATFORM=library SCALANATIVE_GC=immix SBT_VERSION=1.2.1 TEST_COMMAND="sbtScalaNative/scripted"
```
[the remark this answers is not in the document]
shadaj: @jonas eventually (with the 32-bit support PR), there will be additional matrix elements with `TEST_CONTAINER=scala-native-testing:linux-i386` and `TARGET_DOCKER_PLATFORM=i386`. So this just gets the structure set up for later.

.travis.yml
```diff
@@ -8,60 +8,70 @@ dist: trusty
...
+      fi;
+    fi
+  - if [ "$MODE" == "compile-tests" ]; then
+      sbt "-no-colors" "-J-Xmx2G" "^^ $SBT_VERSION" nscplugin/publishLocal sbtScalaNative/publishLocal test-compile;
+    fi
+  - if [ "$MODE" == "runtime-tests" ]; then
+      ./ci-docker/load-cached-images.sh
+      docker run --rm --privileged multiarch/qemu-user-static:register;
+      docker build --cache-from scala-native-testing:linux-$TARGET_DOCKER_PLATFORM -t scala-native-testing:linux-$TARGET_DOCKER_PLATFORM --build-arg TARGET_DOCKER_PLATFORM=$TARGET_DOCKER_PLATFORM --build-arg HOST_DEBIAN_PLATFORM=amd64 ci-docker;
+      ./ci-docker/run-test-travis.sh;
```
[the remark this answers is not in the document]
shadaj: The main reason I didn't go with that approach was to prevent duplication with definition of environment variables inside the docker-compose config but also referencing the same information when identifying which container to run in Travis. I'll look at setting up a script that automatically builds and runs a specific platform test though.

ci-docker/Dockerfile
shadaj: Some file unit tests check read/write permissions, which fail if the user is not set because the root user can do whatever it wants.

ekrich:
Right, some unit tests will not pass if run as root user. I ran into this problem earlier when trying to support the old docker setup.

shadaj:
@densh / @jonas feedback addressed!

.travis.yml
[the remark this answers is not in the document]
jonas: As I see it you either duplicate it in `.travis.yml` or in the docker-compose.yml file. To me the whole point of using docker is to allow reproducing the build environment everywhere, not just in Travis. Anyway, we will only need docker images for each combination of `x86_64`/`x86_32` and LLVM version. The rest can be managed with environment variables.

jonas:
LGTM 🎉 

I still suggest to use docker-compose instead of shell scripts for building and launching the build environment. I'm up for doing that in a future PR.

densh:
LGTM, thanks @shadaj! 👍

PR #1303, opened 2018-08-08
Improve codegen scalability 

densh:
This change improves work balancing for the
codegen and native compilation steps. Previously
we would emit one file per package. Some packages
contain more definitions than the others, which
causes work skew (a few threads do most of the work)
which in turn leads to poor scalability to 6+ cores.

Instead, the work should be split randomly across a
number of files proportional to the number of available
processors. This minimizes the risk of any single file
taking longer to compile than the others.

This change shaves off 1s of compilation on the tests
project on 6 core machine with hyper threading.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
@@ -6,7 +6,7 @@ import java.nio.ByteBuffer
 import java.nio.file.Paths
 import scala.annotation.tailrec
 import scala.collection.mutable
-import scalanative.util.{Scope, ShowBuilder, unsupported}
+import scalanative.util.{Scope, ShowBuilder, unsupported, partition, procs}
 import scalanative.io.{VirtualDirectory, withScratchBuffer}
 import scalanative.sema.ControlFlow.{Graph => CFG, Block, Edge}
 import scalanative.nir._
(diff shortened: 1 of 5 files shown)


jonas:
One useful thing with per-package files are that it is intuitive to find the generated code for a method. I guess with this change it's a matter of grepping all files?

densh:
@jonas Grep of the signature over the all generate IR files should work for this. It's indeed more annoying than before, but the performance gains from this change justify the inconvience (it decreases both the time to generate llvm ir and the time to compile llvm ir to object files).

ekrich:
Why is the the number of batches equal to the number of processors squared?
https://github.com/densh/scala-native/blob/55e48b6dd61ecabe758bc664cc118b5a287dc243/util/src/main/scala/scala/scalanative/util/package.scala#L52

densh:
@ekrich That's a simple heuristic for a batch size based on empirical tests on 4x, 6x and 10x core machines. The more cores you have, the harder it is to have a balanced distribution. Parallel collections do work stealing under the hood, but it has high overhead if amount of work per item is low (i.e. optimizing very small methods).

densh:
@jonas Another nice way to view the LLVM output:
```
$ llvm-link sandbox/target/scala-2.11/native/*.ll | llvm-dis | less
```
And then you can easily navigate across all symbols as if it was a single LLVM IR file.

jonas:
👍 Cool, will make a mental note to add that to the contributor guide.

PR #1313, opened 2018-08-17
Fix that linker thing (part 1)

densh:
This change completly overhauls the method resolution
handling throughout the toolchain. Previously, we relied on
overrides attribute to figure out when a method overrides
another. This included both class-virtual overrides and
trait-virtual implementations.

Instead, now we do resolution purely based on method signatures.
Unlike our previous scheme, method signatures are not sensitive
towards the exact super class or trait where the overiden method
was defined first. This is going to address a number of issues
we've had wrt to crashes and dummy forwarders needed in our JDK
implementation.

The only part of the pipeline that still relies on overrides
is the linker. Next PR (part 2) is going to rewrite the linker
to perform semantic linking rather than named-based linking
that we use currently.


densh:
As a nice side effect this change decreases the overhead of the virtual dispatch tables on the binary size:
* sandbox binary in debug mode gets 491Ki smaller
* unit-tests binary in debug mode gets 3.57Mi smaller (!)

PR #1315, opened 2018-08-20
Bump to sbt 1.2.1 for travis ci

densh:


densh:
It still looks like sbt 1.x build is 4-6 minutes slower than 0.13.x.

jonas:
Based on how many CI runs?

densh:
@jonas It's not meant to be a rigourous performance evaluation, I did a few restarts, 1.x seems to be consistently slower (usually 4+ more minutes than 0.13.16). I suspect that the difference comes from scripted tests which start a new sbt instance on every test suite which amplifies cold start performance degradation in 1.x.

jonas:
Okay, there was some work planned by Scala Center to make scripted tests faster. I don't know if it got done.

ekrich:
It looks like there may be a workaround - https://gitter.im/sbt/sbt-contrib?at=5b7c5f5258a3797aa398ebc6

densh:
@ekrich We don't publish from scripted, so that shouldn't affect us.

PR #1323, opened 2018-08-27
Fix that linker thing (part 2): semantic reachability

### LINKED ISSUE #375: Linker requires dummy forwarders to implement Java APIs

densh:
This PR is a follow-up to #1313 that completely eliminates `override` and `pin` attributes from NIR. Those were used previously to perform name-based reachability analysis (aka linking step of the `nativeLink`).

Instead, we do a Scala.js-like reachability and semantic analysis in the same pass. This improves precision of the reachability as it now knows about NIR semantics, not just loads all reachable names from the entry point. Smaller reachable set improves compilation time (less code to transform) and the resulting binary size (less code to emit).

Resolves #375 and probably a few more issues related to null-dereferencing segfaults and linking errors when calling across JDK boundary.


densh:
Binary size numbers on linux in debug mode, before #1313:

- sandbox: 2.71Mi full, 2.13Mi stripped
- unit tests: 20.7Mi full, 17.4Mi stripped

Binary size numbers after #1313, #1314, #1323:

- sandbox: 1.58Mi full, 1.16Mi stripped
- unit tests: 15.7Mi full, 12.6Mi stripped (

Issue #202, opened 2016-06-26
Failure to increment extern variable


frgomes:
https://github.com/frgomes/scala-native-testsuite/blob/master/testsuite/0202/src/main/scala/issue0202.scala

``` scala
package scala.scalanative
package native


@extern
object getopt {
  var optarg: CString = extern
  var optind: CInt    = extern
  var opterr: CInt    = extern
  var optopt: CInt    = extern

  def getopt(argc: CInt, argv: Ptr[CString], optstring: CString): CInt = extern

  def getopt_long     (argc: CInt, argv: Ptr[CString], optstring: CString, longopts: Ptr[option], longindex: Ptr[CInt]): CInt = extern
  def getopt_long_only(argc: CInt, argv: Ptr[CString], optstring: CString, longopts: Ptr[option], longindex: Ptr[CInt]): CInt = extern

  @struct
  class option(val name:    CString,
               val has_arg: CInt,
               val flag:    Ptr[CInt],
               val `val`:   CInt)
}
```

``` scala
object Main {
  def main(args: Array[String]): Unit = {
    import scala.scalanative.native.getopt._
    optind = optind + 1
  }
}
```

densh:
Thanks for the report!

Issue #1278, opened 2018-07-17
Interflow


densh:
This issue is meant to track progress on upstreaming Interflow optimizer. More details can be found in [the original announcement](http://www.scala-native.org/en/latest/blog/interflow.html) and [paper preprint](https://github.com/densh/talks/blob/master/2018-06-16-interflow-preprint-v1.pdf).

Prerequisites:

- [x] LLVM integration
   - [x] use `-flto` to optimize across C/Scala boundary (issue #280, pr #1281)
   - [x] annotate hot paths with inline/noinline to guide flto inlining (pr #1284)
- [x] Library 
   - [x] introduce `scalanative.runtime.Array.compare` intrinsic to replace memcmp call in String.equals  (simplifies partial evaluation, pr #1288)
   - [x] don't use unmanaged pointers for String.hashCode (simplifies partial evaluation, pr #1288)
   - [x] annotate performance-critical parts of the Scala runtime with inline/noinline (pr #1290, #1291)
   - [x] minimize use of top-level fields in Scala library and use methods instead for them (simplifies partial evaluation, pr #1292)
- [x] Compiler infrastructure
    - [x] split out lowering passes from optimizer (pr #1295)
    - [x] semantic reachability analysis (pr #1323)
- [x] NIR 
   - [x] include unwind handler in every instruction (pr #1311)
   - [x] dedicated instructions for field load and store (simplifies partial evaluation, pr #1325)
   - [x] dedicated instructions for local vars (simplifies partial evaluation, pr #1326)
   - [x] dedicated instructions for array operations (simplifies partial evaluation, pr #1331)

Interflow:

- [x] initial infrastructure (pr #1339)
- [x] don't bail out on unsafe ops (pr #1401)
- [x] don't bail out on reflective calls (pr #1402)
- [x] introduce nir checker to verify code before and after interflow (pr #1408)
- [x] conservative inlining to test partial evaluation and partial escape analysis (pr #1411)
- [x] aggresive inlining guided by partial escape analysis (pr #1411)
- [x] do DCE in the same pass as interflow (pr #1437)
- [x] do GVN in the same pass as interflow (pr #1440)
- [x] do Canonicalization in the same pass as interflow (#1442)
- [x] do InstCombine in the same pass as interflow (pr #1444)
- [x] optimize in parallel in debug mode (pr #1465)
- [x] do Polymorphic Inlining in the same pass as interflow (pr #1495)
[...]

PR #1325, opened 2018-08-28
Add dedicated instructions for field load/store

densh:
Previously we represented field operations through a mix of `Op.Field` and `Op.Load`/`Op.Store`. The semantics of `Op.Field` was to return an inner pointer that points directly to the field within a class, and then subsequent load and stores would perform an operation through that pointer. This encoding turned out to be tricky to optimize as a high-level operation in Interflow (#1278) because we also use load and store for C's unsafe memory operations. 

This pr introduces two dedicated instructions `Op.Fieldload` and `Op.Fieldstore`. Unlike the previous encoding these are only used for high-level Scala and are never emitted for C-style memory accesses. This makes it easier to optimize the away without doing any additional analysis.


jonas:
Does this fix #202 or merely makes it possible to implement a proper fix?

densh:
@jonas It's unrelated to #202 as far as I can tell but it might make it simpler.

PR #1332, opened 2018-08-29
Fix #1327 - Update clangfmt to version 6.0

### LINKED ISSUE #1327: Update clangfmt script to use clang-format version 6.0
Currently, the latest `clang/llvm` supported by Scala Native is 6.0. This change encourages developers to use the latest support versions of these tools.

References:
https://github.com/scala-native/scala-native/blob/master/scripts/clangfmt#L16
https://github.com/scala-native/scala-native/blob/master/tools/src/main/scala/scala/scalanative/build/Discover.scala#L134-L135

ekrich:


jonas:
CI doesn't currently run clangfmt. Is that something we also want to fix?

ekrich:
That is a good point - I noticed the new `libunwind` stuff needed formatting. Maybe we should require it and format code as part of this PR.

ekrich:
Running `clangfmt` on all the C and C++ files changes 3 already existing files slightly and all the added `libunwind` files that were added recently due to indentation. It would be a good idea to rebase after this PR lands in master.

ekrich:
Well, we are not using LLVM 6.0 in CI. Should we upgrade?
```
$ scripts/clangfmt --test
clang-format version 6.0 not found
Install LLVM version 6.0 and rerun.
```

ekrich:
Formatting breaks the build - nice.

ekrich:
Ok, I have clang and clang-format upgraded to 6.0 but the new`libunwind` seems to be not compiling on the sandbox for Ubuntu trusty with the upgraded clang. I am using clang 6 on macOS and sandbox is working fine.

@shadaj If you could take a look that would be great. You can see the error below.
https://travis-ci.org/scala-native/scala-native/jobs/423099828

ekrich:
Even though it is fun playing with Travis we can wait for improvements until after build is changed to use Docker when we have more control.

PR #1339, opened 2018-09-04
  Interflow: initial infrastructure

densh:
This pr introduces preliminary infrastructure for Interflow (#1278).  

In the first part, we focus on type inference and method duplication. Partial evaluation and partial escape analysis are included as well but they are mostly inactive until we add inlining (in one of the next PRs). 

Apart from the new code that runs before the old optimizer, there are a few more NIR changes included here aswell:

1. minor changes to box and unbox ops.
1. minor changes to case node.
1. all reference types track 'nullability' (i.e., can this reference be a null).
1. class types track 'exactness' (i.e., is it only this class, or class and all subclasses). 

We leave interflow enabled at all times on master to catch potential bugs early. Due to added compile time cost, it's likely we'll make it a release mode only feature in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ script:
   - java -version
   - if [ "$MODE" == "source-checks" ]; then ./scripts/source-checks; fi
   - if [ "$MODE" == "test-tools" ]; then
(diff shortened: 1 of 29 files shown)


ekrich:
I really like the way you rewrote the tests.

PR #1348, opened 2018-09-08
Fix #202: Support assignment to extern variables

### LINKED ISSUE #202: Failure to increment extern variable

jonas:
Implement the unsupported assignment case in
NirGetnExpr.genApplyExternAccessor. As a bonus, add bindings for
getopt.h and use it to test the interoperability.

---

This is only tested with the example from #202, that is assignment of primitive types.

### DIFF
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenName.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenName.scala
@@ -78,7 +78,12 @@ trait NirGenName { self: NirGenPhase =>
     if (sym == String_+) {
       genMethodName(StringConcatMethod)
     } else if (sym.owner.isExternModule) {
-      owner member id tag "extern"
+      if (sym.isSetter) {
+        val id0 = sym.name.dropSetter.decoded.toString
+        owner member id0 tag "extern"
+      } else {
+        owner member id tag "extern"
+      }
     } else if (sym.name == nme.CONSTRUCTOR) {
       owner member ("init" +: mangledParams).mkString("_")
     } else {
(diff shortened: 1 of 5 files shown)


nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenName.scala
```diff
@@ -78,7 +78,12 @@ trait NirGenName { self: NirGenPhase =>
     if (sym == String_+) {
       genMethodName(StringConcatMethod)
     } else if (sym.owner.isExternModule) {
-      owner member id tag "extern"
+      if (sym.isSetter) {
+        val id0 = sym.name.dropSetter.decoded.toString
```
jonas: This drops the `_=` part of the name and then decodes the name as done by `nativeIdOf`.

densh:
LGTM, thanks @jonas ! 👍

PR #1350, opened 2018-09-18
Factor out posixlib into its own subproject

densh:


jonas:
Restarted the build that got stuck.

PR #1357, opened 2018-09-21
Configurable heap size for immix

valdisxp1:
- Change the minimum heap size by setting the `SCALANATIVE_MIN_HEAP_SIZE` environment variable.
- Change the maximum heap size by setting the `SCALANATIVE_MAX_HEAP_SIZE` environment variable.
- Sizes can be specified in bytes or using suffixes `k` or `K` for kilobytes, `m` or `M` for megabytes and `g` or `G` for gigabytes


nativelib/src/main/resources/gc/immix/ImmixGC.c
```diff
@@ -13,8 +14,56 @@
...
-    Heap_Init(&heap, INITIAL_SMALL_HEAP_SIZE, INITIAL_LARGE_HEAP_SIZE);
+    char *minHeapSizeStr = getenv("SCALANATIVE_MIN_HEAP_SIZE");
+    size_t minHeapSize;
+    if (minHeapSizeStr != NULL) {
+        minHeapSize = parseSizeStr(minHeapSizeStr);
+    } else {
+        minHeapSize = DEFAULT_MIN_SMALL_HEAP_SIZE;
+    }
+
+    if (minHeapSize < 2 * BLOCK_TOTAL_SIZE) {
```
densh: Why 2 blocks? This seems to be *really* low, even it works now. Scala Native is not aimed at super lower memory devices with <1M of memory. Lets set a lower limit at 1M.

nativelib/src/main/resources/gc/immix/Heap.h
```diff
@@ -33,8 +33,8 @@ static inline bool heap_isObjectInHeap(Heap *heap, Object *object) {
     return Heap_IsWordInHeap(heap, (word_t *)object);
 }
 
-void Heap_Init(Heap *heap, size_t initialSmallHeapSize,
-               size_t initialLargeHeapSize);
+void Heap_Init(Heap *heap, size_t minSmallHeapSize,
+               size_t initialLargeHeapSize, size_t maxHeapSize);
```
densh: There is a problem with large heap size not being a part of the min/max heap size limits. It's not easy to fix though, but it's something we need to in one of the next PRs, opened an issue here: #1358

nativelib/src/main/resources/gc/immix/ImmixGC.c
```diff
@@ -13,8 +14,56 @@
...
+        case 'K':
+            size *= 1024;
+            break;
+        case 'm':
+        case 'M':
+            size *= 1024 * 1024;
+            break;
+        case 'g':
+        case 'G':
+            size *= 1024 * 1024 * 1024;
```
densh: This can overflow if size_it is 32-bit. This needs to be checked and reported as an error in case it does.

nativelib/src/main/resources/gc/immix/Heap.h
densh: This is actually pretty bad. Imagine someone doing -Xms1G -Xmx1G expecting a constant heap size of 1G, but it's not going to work at all because large object space will default to 1M and won't able to grow if a 2M allocation is requested.

valdisxp1:
This needs more changes before proceeding. No sense in keeping a stale PR.

PR #1364, opened 2018-10-04
Log GC statistics

valdisxp1:
- Records mark and sweep times separately in a csv file.
- Write to file every 10 collections to reduce io overhead.
- Set the output file name via `SCALANATIVE_GC_STATS_FILE` environment variable to enable.
```bash
export SCALANATIVE_GC_STATS_FILE=gc-stats.csv
sandbox/target/scala-2.11/sandbox-out
```
Example `gc-stats.csv`:
```csv
timestamp_us,collection,mark_time_us,sweep_time_us
27862510653,0,764,392
27862538561,1,774,561
27862872738,2,992,794
27863270876,3,999,824
27863428286,4,1147,936
```


nativelib/src/main/resources/gc/immix/Heap.c
```diff
@@ -184,19 +195,68 @@ word_t *Heap_Alloc(Heap *heap, uint32_t objectSize) {
...
+    HeapStats* stats = heap->stats;
+    if (stats != NULL) {
+        uint64_t remainder = heap->stats->collections % GC_STATS_MEASUREMENTS;
+        if (remainder > 0) {
+            // there were some measurements not written in the last full batch.
+            Heap_writeStatsToFile(stats);
+        }
+        fclose(stats->outFile);
+    }
+}
```
densh: Please move all statistics-related logic to a new `Stats` module.

nativelib/src/main/resources/gc/immix/Heap.h
```diff
@@ -5,6 +5,15 @@
...
 #include "datastructures/Stack.h"
+#include <stdio.h>
+
+typedef struct {
+    FILE *outFile;
+    uint64_t collections;
+    uint64_t timestamp_us[GC_STATS_MEASUREMENTS];
+    uint64_t mark_time_us[GC_STATS_MEASUREMENTS];
+    uint64_t sweep_time_us[GC_STATS_MEASUREMENTS];
+} HeapStats;
```
densh: Just `Stats` in its own module.

valdisxp1:
Extracted the stat related code into a separate module. Two things remain
1. I want to keep in the Heap.c is the `!= NULL` check so I do not have to allocate extra 400*8> 3KB if the stats are not collected.
2. I do no see a cleaner way of keeping track of the time than having the `timespec` structs as local variables.

densh:
You should use `scalanative_nano_time` to get the nano time measurement instead.

valdisxp1:
Ok, using nanoseconds

nativelib/src/main/resources/gc/immix/Stats.c
```diff
@@ -0,0 +1,45 @@
...
+    stats->outFile = fopen(statsFile, "a");
+    fprintf(stats->outFile, "timestamp_ns,collection,mark_time_ns,sweep_time_ns\n");
+    stats->collections = 0;
+}
+
+void Stats_RecordCollection(Stats *stats, uint64_t start_ns, uint64_t sweep_start_ns, uint64_t end_ns) {
+    uint64_t index = stats->collections % GC_STATS_MEASUREMENTS;
+    stats->timestamp_ns[index] = start_ns;
+    stats->mark_time_ns[index] = sweep_start_ns - start_ns;
+    stats->sweep_time_ns[index] = end_ns - sweep_start_ns;
```
densh: I don't think we need timestamp_ns at all, mark and sweep times and collection # are more than sufficient on their own.

densh: In fact collection # is redundant as well, it can be trivially recomputed as line number in the output file.

valdisxp1: I would rather keep these two.

Yes, when calculating aggregates for benchmarks, `collection #` and `timestamp` are redundant. However, I wanted to see/show the difference between a few large collections and many smaller ones by using the timestamps.

I also considered how useful it may be for diagnostics. Timestamps are important as they allow correlating other events. `collection #` allows getting a piece of the file and still making sense what's happening. The output file is append mode, so getting the `collection #` from the line number can be not so straight forward.

densh: Timestamp is a poor substitute for this as youd'd want to correlate measurements with the iterations of the benchmark for it to be meaningful. Mutator time can be different as well, making the relative comparisons based on timestamps to be completely off.

densh: It's also safe to assume that append case effectively never happens. If someone cares about more separation of data they can pass a different file on every invocation.

valdisxp1:
@densh  remove the two fields. If I really need them I can always add more and rerun.
Also updated benchmark pr to be able to handle csv format changes.
https://github.com/scala-native/scala-native-benchmarks/pull/3

nativelib/src/main/resources/gc/immix/Stats.c
```diff
@@ -0,0 +1,42 @@
...
+        remainder = GC_STATS_MEASUREMENTS;
+    }
+    FILE *outFile = stats->outFile;
+    for (uint64_t i = 0; i < remainder; i++) {
+        fprintf(outFile, "%lu,%lu\n", stats->mark_time_ns[i], stats->sweep_time_ns[i]);
+    }
+    fflush(outFile);
+}
+
+void Stats_Close(Stats *stats) {
```
densh: `Stats_OnExit`.

valdisxp1:
I made the naming changes.

densh:
LGTM, thanks @valdisxp1 ! 😃

PR #1366, opened 2018-10-16
GC metadata changes (performance improvement)

valdisxp1:
- step towards unified heap and heap size changes
- moves block headers and line headers to separate spaces
- removes ObjectHeader
- simplifies LineHeader (only used for marking now)
- adds two bytemaps to keep per object marking - one for small allocator, one for large allocator
- allocations are now aligned on two words (16 bytes)
- also fixes #1365

Improves performance by ~~3.23%~~ 5.53%  (geomean at 50 percentile) for AMD 1950X.
Full performance report:

https://github.com/valdisxp1/scala-native-benchmark-results/blob/gh-pages/reports/summary_20181018_114237_scala-native-0.3.9-SNAPSHOT-thin-lto_gc-stats_vs_scala-native-0.3.9-SNAPSHOT%40metadata-bytemap-gc-gc/Readme.md


densh:
Performance numbers on Intel 7900X:

screen shot 2018-10-17 at 10 53 08

screen shot 2018-10-17 at 10 53 26

Seems to be a definitive win at 50 percentile. >90 is a bit all over the place, but still a win in terms of geomean across the benchmark suite.

nativelib/src/main/resources/gc/immix/datastructures/Bytemap.h
```diff
@@ -0,0 +1,114 @@
...
+    size_t endIndex = startIndex + WORDS_IN_LINE / ALLOCATION_ALIGNMENT_WORDS;
+    ubyte_t *data = bytemap->data;
+    for (size_t i = startIndex; i < endIndex; i++) {
+        if (data[i] == bm_marked) {
+            data[i] = bm_allocated;
+        } else if (data[i] == bm_allocated) {
+            data[i] = bm_free;
+        }
+    }
+*/
```
densh: This comment needs to go inside the method, and use `//` instead of `/* ... */`.

nativelib/src/main/resources/gc/immix/datastructures/Bytemap.h
```diff
@@ -0,0 +1,114 @@
...
+#include "../GCTypes.h"
+#include "../Constants.h"
+#include "../Log.h"
+
+typedef struct {
+    word_t *firstAddress;
+    size_t size;
+    ubyte_t *end;
+    ubyte_t data[0];
+} Bytemap;
```
densh: I have an impression that this module has two conceptual things merged in one: byte mapping and object metadata. When you access a bytemap you should get a pointer to object metadata back, and getters/setters to modify the header should go into its own module. Object metadata might evolve to store more info, so it's best to keep it separate conceptually.

nativelib/src/main/resources/gc/immix/headers/BlockHeader.h
```diff
@@ -20,85 +20,97 @@ typedef struct {
         int16_t first;
         int32_t nextBlock;
     } header;
-    LineHeader lineHeaders[LINE_COUNT];
 } BlockHeader;
 
-static inline bool Block_IsRecyclable(BlockHeader *blockHeader) {
+static inline bool BlockHeader_IsRecyclable(BlockHeader *blockHeader) {
```
densh: Block metadata is not stored in the header (initial part) of the block. So it's best to rename it to be `BlockMetadata` (or `BlockMeta` to shorten it a bit) instead of `BlockHeader`.

densh: Similarly accessing a bytemap, gives you `ObjectMeta`.

nativelib/src/main/resources/gc/immix/headers/LineHeader.h
```diff
@@ -14,46 +12,18 @@ typedef struct {
 typedef enum {
     line_empty = 0x0,
     line_marked = 0x1,
-    line_contains_object_header = 0x2,
 } LineFlag;
```
densh: `LineMeta`.

tools/src/main/scala/scala/scalanative/codegen/Generate.scala
```diff
@@ -246,19 +247,50 @@ object Generate {
...
+      if (ids != (min to max)) {
+        throw new Exception(
+          s"Ids for all known arrays ($tpes) are not consecutive!")
+      }
+
+      buf += Defn.Var(Attrs.None, arrayIdsMinName, Type.Int, Val.Int(min))
+
+      buf += Defn.Var(Attrs.None, arrayIdsMaxName, Type.Int, Val.Int(max))
+
+    }
```
densh: In current implementation object arrays are indeed all going to get a contiguous sequence of class ids, but please add an assert to verify this here.

nativelib/src/main/resources/gc/immix/Block.c
```diff
@@ -110,20 +114,16 @@ void Block_Recycle(Allocator *allocator, BlockHeader *blockHeader) {
 
 void Block_Print(BlockHeader *block) {
```
densh: I have an impression this method is not used anywhere. If that's the case you should remove it altogether.

tools/src/main/scala/scala/scalanative/codegen/Generate.scala
valdisxp1: It is already done in lines 282-285
```scala
      if (ids != (min to max)) {
        throw new Exception(
          s"Ids for all known arrays ($tpes) are not consecutive!")
      }
```

densh: You're right, missed that one.

densh:
Needs a rebase due to conflicting changes in another PR.

valdisxp1:
@densh  Made the changes you asked. The stress tests look good. Currently re-running the benchmarks to ensure it has not magically regressed.

valdisxp1:
Looks even better:
https://github.com/valdisxp1/scala-native-benchmark-results/blob/gh-pages/reports/summary_20181018_114237_scala-native-0.3.9-SNAPSHOT-thin-lto_gc-stats_vs_scala-native-0.3.9-SNAPSHOT%40metadata-bytemap-gc-gc/Readme.md

densh:
LGTM, well done! 🍰

PR #1368, opened 2018-10-17
Fix regression on json benchmark

densh:
We regressed performance on json compared to 0.3.8
due to changes in #1288. This pr reverts the change
completely.


densh:
Thanks to @valdisxp1 for discovering and reporting the regression.

PR #1371, opened 2018-10-21
Enable testing on arm64v8

shadaj:
work-in-progress, pushing up to see Travis results since my computer is too slow with the overhead of the Docker VM.

Based on #1363


ekrich:
@shadaj Is this one still needed?

shadaj:
@ekrich eventually, we can enable arm64v8 in the GitHub Actions matrix with the existing Docker + QEMU support. But this PR is way out of date so closing it for now.

PR #1373, opened 2018-10-23
Update sbt.version to 1.2.6 in docs

lolgab:
The suggested sbt setup contains, as `sbt.version`, the last `0.13.x`. Scala Native works great on sbt `1.2.x` which is newer and better.
So it is better to change the suggested version.


densh:
Hi @lolgab, please sign Scala CLA here: https://www.lightbend.com/contribute/cla/scala

lolgab:
Signed :+1:

Issue #1358, opened 2018-09-21
Don't reserve max heap size memory twice


densh:
Currently immix has two separate memory spaces for small and large objects. They grow independently and don't share memory in any way. This makes it very tricky to have a clear and understandable min/max heap size configuration. It also over-reserves heap memory which is known to cause problems in some environments (#1312).

Issue #1312, opened 2018-08-16
Make immix heap size configurable


jokade:
The size of the memory chunks `mmap`ed by immix (and GC=`none`) seems to be too large for iOS devices.

`Heap_getMemoryLimit()` returns the full RAM size, and then `mmap` is called twice with this size (each for `smallHeap` and `largeHeap`). The second call to `mmap` returns `0xFFFFFFFF`which results in `Heap_mapAndAlign` returning a `NULL` ptr, which is not checked in the subsequent call to `LargeAllocator_Init()`:
https://github.com/scala-native/scala-native/blob/28a6d64525ef3ff640fffd26057921101975d68d/nativelib/src/main/resources/gc/immix/Heap.c#L66-L68

densh:
@jokade Would adding an equivalent of -Xms/-Xmx setting on JVM fix the issue for you?

Double heap size reservation is an implementation artifact that we should totally address. Currently, large and small subsets of heaps have completely disjoint subheap areas but it's just an implementation artifact. They can (and should) coexist in a single memory area. 

Virtually reserving contiguous area of the max heap size is necessary due to conservative heap scanning. If you know that heap is a monolithic memory area, then checking if the pointer is in heap is trivial bounds check. This is not something we can fix easily.

jokade:
@densh yes, when I hack immix/Heap.c to a reduced heap size (e.g. 64MB), immix heap allocation works fine for iOS/arm64 :) So -Xms/-Xmx should do the job

densh:
Fixed in #1374.

PR #1374, opened 2018-10-25
Unified heap and Heap size settings

valdisxp1:
## Made the heap unified
- `LargeAllocator` now uses superblocks (superblock = multiple consecutive blocks) within the small heap.
- `Heap_Recycle` now delegates to the specific sweepers per block basis.
-  Free blocks are no longer swept.
-  Both `Allocator` and `LargeAllocator` ask a common `BlockAllocator` for free blocks. 

## Added heap size settings
- Change the minimum heap size by setting the `SCALANATIVE_MIN_HEAP_SIZE` environment variable.
- Change the maximum heap size by setting the `SCALANATIVE_MAX_HEAP_SIZE` environment variable.
- They can go as low as 1 MB and as high as your system memory up to about 512 GB.
- The size is specified in bytes. Suffixes can be used to specify  kilobytes(`k`), megabytes(`m`) or gigabytes(`g`). Both upper and lower case is accepted.

Solves #1358 and #1312 

The block allocator does not have a noticeable impact on benchmarks. They get 0.19% faster at 50 percentile.
Full report:
https://github.com/valdisxp1/scala-native-benchmark-results/blob/gh-pages/reports/summary_20181025_144930_scala-native-0.3.9-SNAPSHOT%40log-gc-times.e5ceb7bd.-gc_vs_scala-native-0.3.9-SNAPSHOT%40block-allocator-gc-gc/Readme.md


nativelib/src/main/resources/gc/immix/metadata/BlockMeta.h
```diff
@@ -9,45 +11,90 @@
...
+    union {
+        struct {
+            uint8_t flags;
+            int8_t first;
+        } simple;
+        struct {
+            uint8_t flags;
+            int32_t size : BLOCK_COUNT_BITS;
+        } superblock;
+    } block;
```
densh: BlockMeta struct stays 64-bit sized in the end, which is great. https://godbolt.org/z/9GXIan

densh:
Needs a rebase due to conflicting changes in gc logging PR, otherwise looks good to me!

valdisxp1:
Rebased.
Thanks @ekrich for pointing out the warning on 32 bit compilation,

densh:
LGTM, well done @valdisxp1 ! 👍

PR #1378, opened 2018-11-01
Immix GC performance improvement: Lazy (incremental) sweeping

valdisxp1:
Instead of sweeping all of the heap at once it will only sweep a few blocks at a time until in can allocate again. The minimum number blocks swept is 8 (see `Constants.h`).

GC Stats logging changed to emit an event per lazy sweep.

Improves performance by 1.11% at 50 percentile (on intel i5-3470). Also 2.31% at 90 percentile and 4.03% at 99 percentile.
https://github.com/valdisxp1/scala-native-benchmark-results/blob/intel-i53470/reports/summary_20181108_063048_scala-native-0.3.9-SNAPSHOT%40master_vs_scala-native-0.3.9-SNAPSHOT%40lazy-sweep-u%40origin/Readme.md

Splitting the sweep into smaller batches greatly reduces pause times as seen in this report:
https://github.com/valdisxp1/scala-native-benchmark-results/blob/intel-i53470/reports/summary_20181108_063211_scala-native-0.3.9-SNAPSHOT%40master-gc_vs_scala-native-0.3.9-SNAPSHOT%40lazy-sweep-min8%40origin-gc/Readme.md#gc-pause-time-ms-at-50-percentile


nativelib/src/main/resources/gc/immix/Heap.c
```diff
@@ -290,33 +315,63 @@ bool Heap_shouldGrow(Heap *heap) {
...
+    }
+}
+
+void Heap_Recycle(Heap *heap) {
+    Allocator_Clear(&allocator);
+    LargeAllocator_Clear(&largeAllocator);
+    BlockAllocator_Clear(&blockAllocator);
 
+    heap->sweep.cursor = 0;
+    Heap_sweep(heap, heap->blockCount);
```
valdisxp1: This is actually NOT lazy sweeping.  The line here sweeps the entire heap. The performance improvement probably comes from changes to `BlockList` and `BlockAllocator` that were needed to enable lazy sweeping.

valdisxp1: I will remove this line and run the benchmarks again.

nativelib/src/main/resources/gc/immix/Stats.c
```diff
@@ -10,10 +10,18 @@ void Stats_Init(Stats *stats, const char *statsFile) {
...
+    stats->mark_time_ns[index] = end_ns - start_ns;
+    stats->sweep_time_ns[index] = 0L;
+}
+
+void Stats_RecordLazySweep(Stats *stats, uint64_t start_ns,  uint64_t end_ns) {
+    uint64_t index = stats->collections % STATS_MEASUREMENTS;
+    stats->sweep_time_ns[index] += end_ns - start_ns;
+}
+
+void Stats_RecordCollectionDone(Stats *stats) {
```
densh: I don't think aggregating here is appropriate. With real lazy sweeping we now have two types of pauses: one to mark and one to sweep a little. So pause time distribution should shift to the right with 'happy path' pause time being a cost of sweeping a few blocks. 

To get this info we need to log events like 'mark' and 'sweep' separately as they happen. So instead of:
```scala
mark_time,sweep_time
123,456
```
We should log events (m for mark and s for sweep):
```scala
m, 123
s, 5
s, 6
...
s, 4
```
Where all sweeps ad up to 456 for example.

valdisxp1:
Set a minimum lazy sweep batch size to have a better compromise between extra overhead and reduced pause times.

valdisxp1:
Closing this PR, because I will create another one for concurrent sweep. It will act like a lazy sweeper when the number of GC threads is set to 0. I am doing this to make the rebases easier on me and less error prone.

I will definitely move the sweeper code and to `Sweeper.c` and have an explanation there when I create the concurrent sweep.

densh:
@valdisxp1 If you feel like it takes too much work to make 2 PRs (lazy and concurrent separately) it makes perfect sense to wait for the next one instead! 👍

PR #1375, opened 2018-10-29
Update build and plugin versions

ekrich:
WIP for now - testing in CI

PR #1379, opened 2018-11-06
Downgrade to sbt-pgp 1.0.0

densh:
In #1375 @ekrich bumped sbt-pgp version 1.0.0 but it made `sandbox/run` to run the code on the JVM instead of SN which shouldn't happen.


ekrich:
Hey @densh, sorry about that. That seems very strange.

PR #1398, opened 2018-11-22
Add preliminary support for write barriers for immix

densh:


densh:
@ekrich I don't think so. Each GC implementation is going to specify if it requires a barrier or not, it's not going to be configurable from the end-user perspective.

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -17,10 +17,12 @@ import scalanative.linker.{
...
 object Lower {
 
-  def apply(defns: Seq[Defn])(implicit meta: Metadata): Seq[Defn] =
-    (new Impl).onDefns(defns)
+  def apply(config: build.Config, defns: Seq[Defn])(
+      implicit meta: Metadata): Seq[Defn] =
+    (new Impl(config)).onDefns(defns)
 
-  private final class Impl(implicit meta: Metadata) extends Transform {
+  private final class Impl(config: build.Config)(implicit meta: Metadata)
```
ekrich: Do you think it might be better then to just pass in the `GC` only rather than `build.Config` to the `Impl` class and later to the lower function? I should have been a little more clear why I was asking.

densh: It's likely we might need to rely on more things from config in the future, so might as well pass the whole thing.

muxanick:
@densh I really like this PR, can you please share your thought on when it could be in? Or why it's not in the master yet. Thanks.

PR #1408, opened 2018-12-10
Checker: initial infrastructure

densh:
This PR introduces initial infrastructure for NIR checking. It checks that all instructions and jumps within methods are well-typed with respect to our erased type system before and after the optimizer. We'll probably extend it with more checks over time but this is the minimal subset which is already quite useful.

It discovered a few problems:
1. We seem to use signed an unsigned types interchangeably without explicit casts between them. This is not a problem at runtime as they both map to LLVM's integer types which don't distinguish between signed and unsigned numbers, but it might cause issues in the optimizer.
2. In a few cases Interflow lubs (least-upper-bounds) are less precise than the original declared type coming from Scala.
3. Pointers are not boxed and this is detected by the checker. While this a known issue it's nice to have it checked and reported in an automated fashion.

The checker is only going to be used for the development of the toolchain itself, we'll keep it off for end-user projects (likely under a flag).


densh:
As of latest commit here, pointer boxing is the only thing that doesn't pass the checker. It's a sufficiently major change that deserve it's own PR. We'll keep checker in "warning mode" until then, and once it's fixed we make the failures fatal (i.e. the CI build has to pass with all checks on).

Issue #1412, opened 2018-12-18
Unable to link sconfig due to missing Java classes


ekrich:
I am currently working on getting the `lightbend/config` port, now `ekrich/sconfig` to compile under Scala Native and here is my current output. The goal is to support `scalafmt` on Scala Native.

I will look to see if any of these are internal details. The `java.time.Duration` class is part of the interface.

cc @olafurpg

List of items in the works for a possible `0.3.9` release. I will provide PRs once a branch is added.

- [x] Vector and Stack fixed by using `scala.collection.mutable.ArrayStack` in sconfig
- [x] Forwarders and `trait` additions
- [x] Adding `java.util.IdentityHashMap`
- [x] Adding tests for `IdentityHashMap`
- [x] Adding methods and ctor to `java.io.File`
- [x] Adding tests for `File`
- [x] Adding minimum `java.time` classes and methods
- [x] Adding tests for `java.time` additions
- [x] URL and URLConnection stubs

```
[IJ]sbt:sconfig-test> run
[info] Compiling 2 Scala sources to /Users/eric/workspace/sconfig-test/target/scala-2.11/classes ...
[info] Done compiling.
[info] Linking (2709 ms)
[error] cannot link: @java.io.File::init_java.net.URI
[error] cannot link: @java.io.File::toURI_java.net.URI
[error] cannot link: @java.lang.Class$::forName_java.lang.String_java.lang.Class
[error] cannot link: @java.lang.Class::getConstructor_scala.scalanative.runtime.ObjectArray_java.lang.reflect.Constructor
[error] cannot link: @java.lang.Class::getField_java.lang.String_java.lang.reflect.Field
[error] cannot link: @java.lang.ClassLoader::getResources_java.lang.String_java.util.Enumeration
[error] cannot link: @java.lang.ClassLoader::loadClass_java.lang.String_java.lang.Class
[error] cannot link: @java.lang.Thread::getContextClassLoader_java.lang.ClassLoader
[error] cannot link: @java.lang.reflect.Constructor::newInstance_scala.scalanative.runtime.ObjectArray_java.lang.Object
[error] cannot link: @java.net.URL::toExternalForm_java.lang.String
[error] cannot link: @java.net.URL::toURI_java.net.URI
[error] cannot link: @java.net.URLConnection::connect_unit
[error] cannot link: @java.net.URLConnection::getContentType_java.lang.String
[error] cannot link: @java.net.URLConnection::getInputStream_java.io.InputStream
[...]

PR #1419, opened 2019-01-15
[0.3.x] Add forwarders to support sconfig/scalafmt

ekrich:
This is the first PR in a series to support a Scala Native version of `scalafmt` that needs not be applied to master.

### DIFF
--- a/javalib/src/main/scala/java/util/ArrayList.scala
+++ b/javalib/src/main/scala/java/util/ArrayList.scala
@@ -154,6 +154,7 @@ class ArrayList[E] private (private[this] var inner: Array[Any],
     _size = 0
   }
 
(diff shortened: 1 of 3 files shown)


densh:
I think it's good to prepend 0.3.x-only PR name with `[0.3.x]` to make it clear that it's a backport and/or otherwise a non-master PR.

ekrich:
This is a good idea. I have several more that I would like to target both `0.3.x` and `0.4.0`. They are tested on the former. Once they are accepted, is it ok to rebase and test them against master so they can be applied there as well?

densh:
@ekrich Lets first merge to master and only then backport to 0.3.x.

densh:
LGTM, thanks @ekrich !

ekrich:
Refer to Issue #1412

PR #1420, opened 2019-01-16
[WIP] Linker-level 32bit platform support

shadaj:
Introduces a special `Word` class that represents values that are either 32 bits or 64 bits based on the platform. At link time, this type is lowered to either an `Int` or `Long` based on the target platform.

Right now, just focusing on the infrastructure around `Word` but haven't actually implemented what to do when the target platform is 32 bits so there are still some assumptions around the platform being 64 bits.


shadaj:
Everything's working except runtime tests on Boehm for some reason. Also seeing a couple warnings from the checker around a piece of code where a pointer is being set to the result of an if statement that returns a `Word` on both branches but for some reason the result merging block is expecting an `Object` (don't think that's related to the Boehm failures though). Any ideas on how to debug these issues @densh?

densh:
Hi @shadaj, we have very similar issues with current implementation of `Ptr`. The checker is not happy because there are missing box/unbox conversions between object and `Ptr`. Scalac doesn't insert them because from its point of view, `Ptr` is a reference type and thus requires no boxing.

I've been working on a potential solution in [my topic/boxed-ptr branch](https://github.com/densh/scala-native/tree/topic/boxed-ptr). The core ideas are:

1. Hide the actual pointer type away from the public api and rename it as `RawPtr`. 
2. All operations on pointers become methods in the `Intrinsic` module in `scalanative.runtime`.
3. Actual user-facing `Ptr` type becomes a regular non-magic type that wraps `RawPtr`. All operations on it are implemented as calls to the corresponding intrinsic method.

Currently I'm 90% into solving this, but because pointers are such a core feature (we implement just about everything through them) there are still a few segfaults I'm yet to debug.  I think this approach should work 1-to-1 for your `Word` type here (i.e. introduce `RawWord` similarly to what you have now and make a public-facing `Word` be a wrapper class over it).

densh:
> I'm kinda stuck figuring out how to implement this, since Word is encoded as an object-ish type that's lowered to the appropriate integer type at link time, so I'm not sure how the equality check should be encoded at the NIR level.

Given that Word is an integer type, it should use `Op.Comp(Comp.Ieq,  Type.Word, l, r)`. It's used for all reference and integer types alike.

shadaj:
Closing in favor of #1571

PR #1422, opened 2019-01-17
Add support for raw pointers

densh:
This commit introduces a new internal feature:
a raw pointer. It 1-to-1 maps to the pointer
type in NIR and doesn't attempt to provide any
convenience helpers or abstractions. It's going
to be used as a foundation for boxed pointers
in the coming PRs.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Intrinsics.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Intrinsics.scala
@@ -29,4 +29,97 @@ object Intrinsics {
 
   /** Intrinsified int to unsigned long conversion. */
   def intToULong(v: Int): Long = intrinsic
+
+  /** Intrinsified raw memory load of boolean. */
+  def loadBoolean(rawptr: RawPtr): Boolean = intrinsic
+
+  /** Intrinsified raw memory load of char. */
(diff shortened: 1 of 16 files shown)


nativelib/src/main/scala/scala/scalanative/runtime/Intrinsics.scala
```diff
@@ -29,4 +29,97 @@ object Intrinsics {
...
+  def castRawPtrToInt(rawptr: RawPtr): Int = intrinsic
+
+  /** Intrinsified cast that reinterprets raw pointer as an long. */
+  def castRawPtrToLong(rawptr: RawPtr): Long = intrinsic
+
+  /** Intrinsified cast that reinterprets int as a raw pointer. */
+  def castIntToRawPtr(int: Int): RawPtr = intrinsic
+
+  /** Intrinsified cast that reinterprets long as a raw pointer. */
+  def castLongToRawPtr(int: Long): RawPtr = intrinsic
```
densh: Curious reader might observe that this API maps very closely to memory operations in [`sun.misc.Unsafe`](http://www.docjar.com/docs/api/sun/misc/Unsafe.html).

ekrich:
I just wanted to say that I think this is a brilliant idea.

PR #1404, opened 2018-11-29
Performance improvement: Immix GC: Concurrent sweep

valdisxp1:
Based upon #1378
## Features:
- New `SCALANATIVE_GC` option  `commix` to enable the new GC changes.
- Concurrent sweep. Reduced GC pause times by utilizing additional processor cores during sweeping.
- Configurable number of GC threads with `SCALANATIVE_GC_THREADS` environment variable. Set to number of available processors - 2 up to 8, but no less than 1.
- Lazy sweep when setting `SCALANATIVE_GC_THREADS=0`.
- Stats changed to event-based CSV format.
- Stats now also saves events for collection start and end.
- Stats includes `start_ns` column - time in nanoseconds since application was started.
- Stats is compatible with the new concurrent sweep events.
- Increasing heap size until mark is <=5% of total time. 
## Internals:
- thread-safe `BlockList`s
- thread-safe `BlockRange`s
- thread-safe block counts
- new block flags in `BlockMeta`
## Debugging :
- asserts in Immix can now be enabled by setting `DEBUG_ASSERT` flag
```scala
nativeCompileOptions += "-DDEBUG_ASSERT"
```
This similar to the way you could enable trace level printing and debug symbols.
```scala
// trace level printing in Immix
nativeCompileOptions += "-DDEBUG_PRINT"
// enable debug symbols for C code
nativeCompileOptions += "-g"
```
- More integrity check asserts for sweeping.
## Refactoring:
- Moved `Block_Sweep` to `Allocator_Sweep` in order to match `LargeAllocator_Sweep`.
- Tenamed `block_superblock_middle` to ` block_superblock_tail` in order to avoid confusion. This block flag is used for both middle and last blocks.

On Intel i5-3470:
Reduces GC time on the main thread by 47.65% (geomean) on default heap size and by 97.04%(geomean) on 1GB heap.

https://github.com/valdisxp1/scala-native-benchmark-results/blob/intel-i53470/reports/summary_20190109_110202_scala-native-0.3.9-SNAPSHOT%40master-concurrent-base%40origin-gc_vs_scala-native-0.3.9-SNAPSHOT-commix%40concurrent-sweep-timed-growth%40origin-gc/Readme.md#total-gc-time-on-application-thread-ms
[...]

PR #1423, opened 2019-01-17
Commix: parallel mark and concurrent sweep GC

valdisxp1:
Concurrent sweep changes from #1404.

## Features:
- New `SCALANATIVE_GC` option  `commix` to enable the new GC changes.
- Concurrent, parallel sweep. Reduced GC pause times by utilizing additional processor cores during sweeping.
- Parallel mark. Reduces GC pause times by utilizing additional processor cores during marking.
- Configurable number of GC threads with `SCALANATIVE_GC_THREADS` environment variable. Set to number of available processors - 1 up to 8, but no less than 1.
- Increasing heap size until mark is <=5% of total time. This can be adjusted with `SCALANATIVE_TIME_RATIO`. I.e. `export SCALANATIVE_TIME_RATIO=0.03` for <= 3%.
- Added an option the change the minimum free block ratio `SCALANATIVE_FREE_RATIO`. If there are not enough free blocks after a garbage collection it will increase the heap size.
- The minimum heap size setting renamed from `SCALANATIVE_MIN_HEAP_SIZE` to a shorter `SCALANATIVE_MIN_SIZE`. Updated Immix to match.
- The maxium heap size setting renamed from `SCALANATIVE_MAX_HEAP_SIZE` to a shorter `SCALANATIVE_MAX_SIZE`. Updated Immix to match.
### Changes to GC Statistics
- Stat collection disabled by default to reduce code size. They can be enabled by compiler arguments. See example and table below.
- Stats changed to event-based CSV format.
- Stats now also saves events for collection start and end.
- Stats includes `start_ns` column - time in nanoseconds since application was started.
- Stats is compatible with the new concurrent sweep events.
 Multiple verbosity levels of enabled for stats with compiler arguments. For example:
```scala
nativeCompileOptions += "-DENABLE_GC_STATS_BATCHES"
```

| type| compiler flag| scala-native-benchmarks flag | usage | results size per run|
|-------|-----------------|----------|----------|-----------------|
|basic|ENABLE_GC_STATS | `--gc` | shows GC pauses and total GC time, shows phases in Gantt charts | 1.5MB |
|batches|ENABLE_GC_STATS_BATCHES | `--gcv` | shows batches in Gantt chart - are batch sizes good ?| 180MB |
|sync|ENABLE_GC_STATS_SYNC |`--gcvv` | shows synchronization events in Gantt chart - what is the  synchronization overhead? | 309MB |

> scala-native-benchmarks flags as per suggested changes in https://github.com/scala-native/scala-native-benchmarks/pull/3
> results size for default tests when doing a single run `--runs 1` instead of the default which is 20.

## Internals:
- thread-safe `BlockList`s
- thread-safe `BlockRange`s
- thread-safe block counts
- thread-safe `GreyList`s
- new block flags in `BlockMeta`
- removed marking `Stack`s
## Debugging :
- asserts in Immix can now be enabled by setting `DEBUG_ASSERT` flag
```scala
nativeCompileOptions += "-DDEBUG_ASSERT"
```
This similar to the way you could enable trace level printing and debug symbols.
```scala
// trace level printing in Immix
nativeCompileOptions += "-DDEBUG_PRINT"
// enable debug symbols for C code
nativeCompileOptions += "-g"
```
- More integrity check asserts for sweeping.
## Refactoring:
- Moved `Block_Sweep` to `Allocator_Sweep` in order to match `LargeAllocator_Sweep`.
- Moved `Heap_AllocSmall` and `HeapAllocLarge` to their respective allocators
- Renamed `block_superblock_middle` to ` block_superblock_tail` in order to avoid confusion. This block flag is used for both middle and last blocks


nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,542 @@
...
+        ObjectMeta *objectMeta = Bytemap_Get(allocator.bytemap, (word_t *)object);
+#ifdef DEBUG_ASSERT
+        ObjectMeta_AssertIsValidAllocation(objectMeta, size);
+#endif
+        ObjectMeta_SetAllocated(objectMeta);
+        return (word_t *)object;
+    }
+
+    if (!Sweeper_IsSweepDone(heap)) {
+        object = Sweeper_LazySweep(heap, size);
```
densh: `Sweeper_LazySweep` needs a better name. Based on the code it's more like `Sweeper_LazySweepUntilCanAllocate`.

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,542 @@
...
+
+        if (object != NULL)
+            goto done;
+    }
+
+    // A small object can always fit in a single free block
+    // because it is no larger than 8K while the block is 32K.
+    Heap_Grow(heap, 1);
+    object = (Object *) Allocator_Alloc(&allocator, size);
```
densh: The two lines above look wrong. `Heap_Collect` will grow heap appropriately after collection so that 50% of the heap is free blocks. We have a single mutator, so there is no chance of someone else allocating all of the 50% of the heap before we reach here. 

In the hypothetical case of multiple mutator threads, it's better to do another collection instead of ad-hoc growth here. Because it looks like it might cause unbounded growth to the heap limit if this path is actually taken frequently.

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,542 @@
...
+    Heap_assertIsConsistent(heap);
+#endif
+    heap->mark.lastEnd_ns = heap->mark.currentEnd_ns;
+    heap->mark.currentStart_ns = scalanative_nano_time();
+    Marker_MarkRoots(heap, stats);
+    heap->gcThreads.phase = gc_mark;
+    // make sure the gc phase is propagated
+    atomic_thread_fence(memory_order_release);
+    // start main gc thread (0)
+    sem_post(&heap->gcThreads.start0);
```
densh: This needs to be factored out into a helper function `startMainGCThread` instead of a comment.

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,542 @@
...
+    while (!Marker_IsMarkDone(heap)) {
+        Marker_Mark(heap, stats);
+        if (!Marker_IsMarkDone(heap)) {
+            sched_yield();
+        }
+    }
+    heap->gcThreads.phase = gc_idle;
```
densh: This should also be a helper function. It would be nice to isolate all of the phase/thread management into its own module with clear method names.

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,542 @@
...
+    heap->mark.currentEnd_ns = scalanative_nano_time();
+#ifdef ENABLE_GC_STATS
+    if (stats != NULL) {
+        Stats_RecordEvent(stats, event_mark, heap->mark.currentStart_ns,
+                          heap->mark.currentEnd_ns);
+    }
+#endif
+    Heap_Recycle(heap);
+}
```
densh: Most of the code in `Heap_Collect` belongs to either the `Marker` or phase management module. The heap part of the code should be something along the lines of:
1. Stats boilerplate
2. Call phases to start marking phase.
2. Call marker entry point method to do actual marking on the mutator thread until it's done.
3. Call phases to trigger sweeping phase.
3. Call sweeper entry point.

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,542 @@
...
+    int numberOfBatches = blockCount / SWEEP_BATCH_SIZE;
+    int threadsToStart = numberOfBatches / MIN_SWEEP_BATCHES_PER_THREAD;
+    if (threadsToStart <= 0) {
+        threadsToStart = 1;
+    }
+    if (threadsToStart > gcThreadCount){
+        threadsToStart = gcThreadCount;
+    }
+    GCThread_Wake(heap, threadsToStart);
+}
```
densh: Similarly to `Heap_Collect` most of the code either belongs in sweeper or phase management modules.

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,542 @@
...
+    // use the reserved block so mutator can does not have to lazy sweep
+    // but can allocate imminently
+    BlockAllocator_UseReserve(&blockAllocator);
+
+    // all the marking changes should be visible to all threads by now
+    atomic_thread_fence(memory_order_seq_cst);
+
+    // before changing the cursor and limit values, makes sure no gc threads are
+    // running
+    GCThread_JoinAll(heap);
```
densh: This looks like you're going to wait here *on the mutator thread*, is there any way to avoid that?

nativelib/src/main/resources/gc/commix/Heap.h
```diff
@@ -0,0 +1,99 @@
...
+    word_t *heapStart;
+    word_t *heapEnd;
+    word_t *greyPacketsStart;
+    size_t heapSize;
+    size_t maxHeapSize;
+    uint32_t blockCount;
+    uint32_t maxBlockCount;
+    struct {
+        sem_t start;
+        sem_t start0;
```
densh: What's the difference between `start` and `start0` ?

nativelib/src/main/resources/gc/commix/Sweeper.c
```diff
@@ -0,0 +1,482 @@
...
+Batch 2 is done, thread 2 tries to coalesce. It can coalesce the remaining 2 batches.
+U[FFFF]U|U[F]U
+The free superblock of size 4 gets returned to BlockAllocator.
+U[FFFF]U|U[F]U
+U[FFFF]UU[F]U
+All is coalesced! Sweeping is done.
+
+Note that besides coalescing `Sweeper_LazyCoalesce` also finishes the sweeping of superblocks in some cases.
+See also `block_superblock_start_me` and `LargeAllocator_Sweep`.
+*/
```
densh: A few changes to formatting:
1. Align doc comments on 80 character line limit
2. Use `//` for non-doc comment style comments.
3. Indent examples by 4 spaces and blank line before and after them.
4. Make sure that comment is up-to-date with latest implementation.

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,367 @@
...
+#include <setjmp.h>
+#include "Marker.h"
+#include "Object.h"
+#include "Log.h"
+#include "State.h"
+#include "headers/ObjectHeader.h"
+#include "datastructures/GreyPacket.h"
+#include "GCThread.h"
+#include <sched.h>
+
```
densh: Needs a comment somewhere around here to explain the gray packets and how they are managed.

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,367 @@
...
+    Bytemap *bytemap = heap->bytemap;
+    if (*outHolder == NULL) {
+        GreyPacket *fresh = Marker_takeEmptyPacket(heap, stats);
+        assert(fresh != NULL);
+        *outHolder = fresh;
+    }
+    while (!GreyPacket_IsEmpty(in)) {
+        Object *object = GreyPacket_Pop(in);
+
+        if (Object_IsArray(object)) {
```
densh: I'd move the two branches of this if into two helper methods `Marker_markArrayObject` and `Marker_markRegularObject`.

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,367 @@
...
+                        assert(slice != NULL);
+                        GreyPacket_Push(slice, object);
+                        Marker_giveFullPacket(heap, stats, slice);
+                    }
+                }
+            }
+            // non-object arrays do not contain pointers
+        } else {
+            int64_t *ptr_map = object->rtti->refMapStruct;
+            int i = 0;
```
densh: You can remove the need for this index via pointer arithmetics on a var that points to `ptr_map`, this is idiomatic in C:
```
int64_t *cursor = ptr_map;
for (*cursor != LAST_FIELD_OFFSET) {
   word_t *field = object->fields[*cursor];
   ...
   cursor++;
}
```

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,367 @@
...
+    }
+    while (!GreyPacket_IsEmpty(in)) {
+        Object *object = GreyPacket_Pop(in);
+
+        if (Object_IsArray(object)) {
+            if (object->rtti->rt.id == __object_array_id) {
+                ArrayHeader *arrayHeader = (ArrayHeader *)object;
+                size_t length = arrayHeader->length;
+                word_t **fields = (word_t **)(arrayHeader + 1);
+                if (length <= ARRAY_SPLIT_THRESHOLD) {
```
densh: Array splitting needs helper methods with clear names and comments on how array splitting works.

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,367 @@
...
+                // next == NULL, exits
+                Marker_giveEmptyPacket(heap, stats, in);
+                Marker_giveEmptyPacket(heap, stats, out);
+            }
+        }
+        in = next;
+    }
+}
+
+void Marker_MarkAndScale(Heap *heap, Stats *stats) {
```
densh: 1. Needs a comment with motivation for not using all threads at once. 
2. Some of thread management boilerplate probably belongs to non-marker module (i.e. gcthread module).

nativelib/src/main/resources/gc/commix/Settings.c
```diff
@@ -0,0 +1,88 @@
...
+}
+
+#ifdef ENABLE_GC_STATS
+char *Settings_StatsFileName() { return getenv(STATS_FILE_SETTING); }
+#endif
+
+int Settings_GCThreadCount() {
+    char *str = getenv("SCALANATIVE_GC_THREADS");
+    if (str == NULL) {
+        // default is number of cores - 2, but no less than 1 and no more than 8
```
densh: Number of cores - 2 seems off. We want 1 thread for the mutator, but the rest can be used by the GC.

nativelib/src/main/resources/gc/commix/Stats.c
```diff
@@ -0,0 +1,53 @@
+#include "Stats.h"
+#include "GCTypes.h"
+#include <stdio.h>
+#include <inttypes.h>
+
+#ifdef ENABLE_GC_STATS
+const char *const Stats_eventNames[] = {"mark", "sweep", "concmark",
+                                        "concsweep", "collection",
+                                        "mark_batch", "sweep_batch", "coalesce_batch",
+                                        "mark_waiting", "sync"};
```
densh: Needs a comment explaining what those events mean.

nativelib/src/main/resources/gc/commix/Sweeper.c
```diff
@@ -0,0 +1,482 @@
...
+        if (object == NULL) {
+            sched_yield();
+        }
+    }
+#ifdef ENABLE_GC_STATS
+    if (stats != NULL) {
+        end_ns = scalanative_nano_time();
+        Stats_RecordEvent(stats, event_sweep, start_ns, end_ns);
+    }
+#endif
```
densh: Is there any way to reduce boilerplate of stat recording via a few method-like macros? 
```
val start_ns = Stats_Start(stats);
...
Stats_Record(stats, event_sweep, start_ns);
```
Where definitions of `Stats_Start` and `Stats_Record` do nothing (i.e. just return 0) if stat recording is off?

densh: Currently all the ifdefs really hurt the readability of the code.

nativelib/src/main/resources/gc/commix/Heap.h
```diff
@@ -0,0 +1,99 @@
...
+    assert(lineGlobalIndex >= 0);
+    LineMeta *lineMeta = (LineMeta *)heap->lineMetaStart + lineGlobalIndex;
+    assert(lineMeta < (LineMeta *)heap->lineMetaEnd);
+    return lineMeta;
+}
+
+void Heap_Init(Heap *heap, size_t minHeapSize, size_t maxHeapSize);
+word_t *Heap_Alloc(Heap *heap, uint32_t objectSize);
+word_t *Heap_AllocSmall(Heap *heap, uint32_t objectSize);
+word_t *Heap_AllocLarge(Heap *heap, uint32_t objectSize);
```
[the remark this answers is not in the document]
valdisxp1: Moved

nativelib/src/main/resources/gc/commix/Heap.h
valdisxp1: `start0` is for GC Thread 0 (main GC thread)
`start` is for all other GC threads
I will change all the suffixes 0 to `Main` or `Master` so it is more explicit.

nativelib/src/main/resources/gc/commix/Heap.c
valdisxp1: Sweeping until it is possible to allocate is literally the definition of lazy sweeping. `UntilCanAllocate` is redundant.

nativelib/src/main/resources/gc/commix/GCThread.c
```diff
@@ -0,0 +1,277 @@
...
+#endif
+
+    while (!Marker_IsMarkDone(heap)) {
+        Marker_MarkAndScale(heap, stats);
+        if (!Marker_IsMarkDone(heap)) {
+            sched_yield();
+        }
+    }
+
+#ifdef ENABLE_GC_STATS
```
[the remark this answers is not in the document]
valdisxp1: There are actually 3 types of GC stats.

| type| compiler flag | usage | results size per run|
|-------|-----------------|----------|-----------------|
|basic|ENABLE_GC_STATS | shows GC pauses and total GC time, shows phases in Gantt charts | 1.5MB |
|batches|ENABLE_GC_STATS_BATCHES | shows batches in Gantt chart - are batch sizes good ?| 180MB |
|sync|ENABLE_GC_STATS_SYNC | shows synchronization events in Gantt chart - what is the  synchronization overhead? | 309MB |

The more we log the larger the results and the more noise it adds.  It is important to be able to pick right verbosity of stats for the thing you are measuring and for the space budget you have. For example, I try to save most of my results and the reports in https://github.com/valdisxp1/scala-native-benchmark-results . But github does not allow commits larger than 100MB.

nativelib/src/main/resources/gc/commix/Heap.c
valdisxp1: This still could happen if the reserved space (`SWEEP_RESERVE_BLOCKS`) is exactly as large as the free space after heap growth. When sweep is done, GC master thread takes the reserved blocks out of the free blocks. There is a race condition between this and the mutator allocating. If the GC thread wins and there are no recycled blocks, then the small allocation must grow the heap.

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,380 @@
...
+// contains pointers to objects for marking.
+//
+// Each marker has a grey packet with references to check ("in" packet).
+// When it finds a new unmarked object the marker puts a pointer to
+// it in the "out" packet. When the "in" packet is empty it gets
+// another from the full packet list and returns the empty one to the empty
+// packet list. Similarly, when the "out" packet get full, marker gets another
+// empty packet and pushes the full one on the full packet list.
+//
+// Marking is done when all the packets are empty and in the empty packet list.
```
densh: Needs more details. Especially around:
1. Work balancing: array splitting, packet splitting (including explanation of the constants).
2. Synchronization: only on taking/returning packets from the pool, but need to mention that marking naturally doesn't require fine-grain synchronization.

densh: Also just a brief mention of how roots are handled and how/when are threads spin up/down would be good as well.

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,380 @@
...
+#include "GCThread.h"
+#include <sched.h>
+
+extern word_t *__modules;
+extern int __modules_size;
+extern word_t **__stack_bottom;
+
+#define LAST_FIELD_OFFSET -1
+
+// Marking is done using grey packets. A grey packet is a fixes size list that
```
densh: Are packets actual linked lists?

nativelib/src/main/resources/gc/commix/Marker.c
```diff
@@ -0,0 +1,380 @@
...
+}
+
+int Marker_splitObjectArray(Heap *heap, Stats *stats, GreyPacket **outHolder,
+                            Bytemap *bytemap, word_t **fields, size_t length) {
+    word_t **limit = fields + length;
+    word_t **lastBatch =
+        fields + (length / ARRAY_SPLIT_BATCH) * ARRAY_SPLIT_BATCH;
+
+    assert(lastBatch <= limit);
+    // first
```
densh: This is an extremely uninformative comment. Either remove it or make it more clear.

nativelib/src/main/resources/gc/commix/datastructures/GreyPacket.c
```diff
@@ -0,0 +1,115 @@
...
+        packet->items[packet->size++] = value;
+        return true;
+    }
+}
+Stack_Type GreyPacket_Pop(GreyPacket *packet) {
+    assert(packet->size > 0);
+    return packet->items[--packet->size];
+}
+
+void GreyPacket_Move(GreyPacket *src, GreyPacket *dst, int count) {
```
densh: It looks like GreyPacket_Split would be a better name for this.

nativelib/src/main/resources/gc/commix/Constants.h
```diff
@@ -0,0 +1,94 @@
...
+#define LARGE_OBJECT_MIN_SIZE_BITS 13
+
+#define MIN_BLOCK_SIZE (1UL << LARGE_OBJECT_MIN_SIZE_BITS)
+
+#define LARGE_BLOCK_MASK (~((1 << LARGE_OBJECT_MIN_SIZE_BITS) - 1))
+
+#define EARLY_GROWTH_THRESHOLD (128 * 1024 * 1024UL)
+#define EARLY_GROWTH_RATE 2.0
+#define GROWTH_RATE 1.414213562
+#define GROWTH_MARK_FRACTION 0.05
```
densh: This one should probably be a configurable setting (similarly to heap size).

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,308 @@
...
+    printf("Free: %" PRIu32 "\n", freeBlockCount);
+    printf("Recycled: %" PRIu32 "\n", recycledBlockCount);
+    fflush(stdout);
+#endif
+
+    uint64_t timeInMark = heap->mark.currentEnd_ns - heap->mark.currentStart_ns;
+    uint64_t timeTotal = heap->mark.currentEnd_ns - heap->mark.lastEnd_ns;
+
+    return timeInMark >= GROWTH_MARK_FRACTION * timeTotal ||
+           freeBlockCount * 2 < blockCount ||
```
densh: Free block fraction is 50%, lets make it a publicly configurable setting, similarly to heap size.

nativelib/src/main/resources/gc/commix/Heap.c
```diff
@@ -0,0 +1,308 @@
...
+           4 * unavailableBlockCount > blockCount;
```
densh: This one (a fraction of unavailable block count) should be a constant in Constants.h

nativelib/src/main/resources/gc/commix/Constants.h
densh: i.e., `SCALANATIVE_GC_TIME_RATIO` similarly to `-XX:GCTimeRatio` on the JVM.

nativelib/src/main/resources/gc/commix/Heap.c
densh: i.e., `SCALANATIVE_FREE_RATIO`.

nativelib/src/main/resources/gc/commix/Settings.c
```diff
@@ -0,0 +1,88 @@
...
+            size <<= 30;
+        } else {
+            size = UNLIMITED_HEAP_SIZE;
+        }
+    }
+    return size;
+}
+
+size_t Settings_MinHeapSize() {
+    char *minHeapSizeStr = getenv("SCALANATIVE_MIN_HEAP_SIZE");
```
densh: We can probably shorten this one to just `SCALANATIVE_MIN_SIZE`. It should be clear enough.

nativelib/src/main/resources/gc/commix/Settings.c
```diff
@@ -0,0 +1,88 @@
...
+    if (minHeapSizeStr != NULL) {
+        return Settings_parseSizeStr(minHeapSizeStr);
+    } else {
+        return DEFAULT_MIN_HEAP_SIZE;
+    }
+}
+
+size_t Settings_MaxHeapSize() {
+    char *maxHeapSizeStr = getenv("SCALANATIVE_MAX_HEAP_SIZE");
```
densh: We can probably shorten this one to just `SCALANATIVE_MAX_SIZE`. It should be clear enough.

nativelib/src/main/resources/gc/commix/Allocator.c
```diff
@@ -0,0 +1,427 @@
...
+#endif
+    ObjectMeta_SetAllocated(objectMeta);
+
+    __builtin_prefetch(object + 36, 0, 3);
+
+    assert(Heap_IsWordInHeap(heap, object));
+    return object;
+}
+
+uint32_t Allocator_Sweep(Allocator *allocator, BlockMeta *blockMeta,
```
densh: 1. Why do we have sweeping logic in allocator?
2. Is this function used?

nativelib/src/main/resources/gc/commix/Allocator.c
```diff
@@ -0,0 +1,427 @@
...
+
+INLINE word_t *Allocator_Alloc(Heap *heap, uint32_t size) {
+    assert(size % ALLOCATION_ALIGNMENT == 0);
+    assert(size < MIN_BLOCK_SIZE);
+
+    word_t *start = allocator.cursor;
+    word_t *end = (word_t *)((uint8_t *)start + size);
+
+    // Checks if the end of the block overlaps with the limit
+    if (end >= allocator.limit) {
```
densh: Why is it `>=` here and not `>` ?

nativelib/src/main/resources/gc/commix/Allocator.c
```diff
@@ -0,0 +1,427 @@
...
+            // Otherwise try to get a new line.
+            if (Allocator_getNextLine(allocator)) {
+                return Allocator_tryAlloc(allocator, size);
+            }
+
+            return NULL;
+        }
+    }
+
+    memset(start, 0, size);
```
densh: Why do we need memset here? `tryAlloc` is only called on allocation slow path that returns back to fast path before memory is zeroed.

nativelib/src/main/resources/gc/commix/LargeAllocator.c
```diff
@@ -0,0 +1,375 @@
...
+        if (head == NULL) {
+            return NULL;
+        }
+        newValue = (word_t)head->next;
+    } while (!atomic_compare_exchange_strong(&freeList->head, (word_t *)&head,
+                                             newValue));
+    return head;
+}
+
+Chunk *LargeAllocator_freeListPop_OnlyThread(FreeList *freeList) {
```
densh: Second underscore in the name is not matching our naming convention. Please remove it.

nativelib/src/main/resources/gc/commix/Marker.c
valdisxp1: No, the packets are lists not linked lists. It says that in the comment.

nativelib/src/main/resources/gc/commix/Marker.c
valdisxp1: Removed

nativelib/src/main/resources/gc/commix/datastructures/GreyPacket.c
valdisxp1: How about `GreyPacket_MoveItems` ? I realize `GreyPacket_Move` does not work because it does not move the entire packet. `GreyPacket_Split` also does not fit well because we do not take a packet and return two or more packets.

nativelib/src/main/resources/gc/commix/Allocator.c
valdisxp1: You are right, nice catch! I updated that.

nativelib/src/main/resources/gc/commix/datastructures/GreyPacket.c
densh: `MoveItems` is much better than just `Move`.

nativelib/src/main/resources/gc/commix/Allocator.c
valdisxp1: When the slow path is called the object is returned immediately and the memset in fast path is not called. However I did manage to reduce the call instructions to memset by moving them from `tryAlloc` and `Allocator_overflowAllocation` to `Allocator_allocSlow`.  `tryAlloc` gets inlined in multiple places, so this should reduce the code size somewhat.

nativelib/src/main/resources/gc/commix/LargeAllocator.c
valdisxp1: Removed

nativelib/src/main/resources/gc/commix/Heap.c
valdisxp1: Done

nativelib/src/main/resources/gc/commix/Heap.c
valdisxp1: Done

nativelib/src/main/resources/gc/commix/Constants.h
valdisxp1: Done, only using the `SCALANATIVE_MARK_TIME_RATIO` setting. Because we are not based on the total GC time, just the mark time.

nativelib/src/main/resources/gc/commix/Settings.c
```diff
@@ -62,6 +62,28 @@ size_t Settings_MaxHeapSize() {
     }
 }
 
+double Settings_MaxMarkTimeRatio() {
+    char *str = getenv("SCALANATIVE_MARK_TIME_RATIO");
```
densh: Just `SCALANATIVE_TIME_RATIO`.

nativelib/src/main/resources/gc/commix/Allocator.c
valdisxp1: Both `Allocator_Sweep` and `LargeAllocator_Sweep` are called from `Sweeper_Sweep`.
1. They take their corresponding allocator as the first argument.
2. All the logic that can change the state of small object specific metadata is in `Allocator` and for large objects it is `LargeAllocator`. It is convenient and I see no benefit of  splitting it.
3. This saves LargeAllocator from exposing the chunk and its freelist manipulation logic which is part of its implementation.
4. The obvious alternative is to put them both in `Sweeper` and call them something like `Sweeper_SweepSmall` and `Sweeper_SweepLarge` which I do not like. For the same reason I moved `Allocator_lazySweep` and `LargeAllocator_lazySweep` to these modules.

nativelib/src/main/resources/gc/commix/Constants.h
densh: With concurrent sweep mark time is the direct proxy gc time +/- error.

nativelib/src/main/resources/gc/commix/Allocator.c
densh: Having sweeping logic spread over multiple modules makes it harder to see the whole picture. This needs to be refactored better for clarity.

nativelib/src/main/resources/gc/commix/Constants.h
valdisxp1: Makes sense

nativelib/src/main/resources/gc/commix/Allocator.c
valdisxp1: You are right. After all we do not split the conservative marking logic in `Marker` either. I moved both functions to `Sweeper`. And they are private - not exposed in the `*.h` file.

tools/src/main/scala/scala/scalanative/build/GC.scala
```diff
@@ -32,6 +33,9 @@ object GC {
   /** Mostly-precise mark-region garbage collector. */
   def immix: GC = Immix
 
+  /** Mostly-precise mark-region garbage collector running concurrently. */
+  def commix: GC = Immix
```
AlexandreDubray: Should it be `Commix` instead of `Immix` here ?

valdisxp1: You are right.

densh:
LGTM, thanks a lot for the epic contribution @valdisxp1 ! ✨🍰✨

PR #1425, opened 2019-01-23
Add java.util.IdentityHashMap

ekrich:
This is the second PR needed to support `ekrich/sconfig`.

Needs to be applied to `master` and `0.3.x` but forwarders should not be needed on master and can be removed at a future date.


ekrich:
This PR will be targeted at `0.3.x` first and then `0.4.0`.

Issue #153, opened 2016-06-06
Control-flow constructs inside C-style varargs crash the frontend


densh:
``` scala
import scalanative.native._
import scalanative.libc.stdlib._

object Test {
  def main(args: Array[String]): Unit = {
    fprintf(stdout, c"%d", if ("".isEmpty) 0 else 1)
  }
}
```

Crashes with:

```
scala.MatchError: (Focus(List(),Local(src,2),List(Local(Local(src,0),Module(Top(Test$))), Local(Local(src,1),Class(Top(s
cala.scalanative.runtime.ObjectArray)))),List(Inst(Local(src,3),Module(Top(scala.scalanative.libc.stdlib$))), Inst(Local
(src,4),Load(Ptr,Global(Member(Top(scala.scalanative.libc.stdlib$),extern.__stdoutp),Ptr)))),Const(Chars(%d)),false),if
("".isEmpty())
  native.this.Vararg.apply(scala.Int.box(0), (ClassTag.Int(): scala.reflect.ClassTag))
else
  native.this.Vararg.apply(scala.Int.box(1), (ClassTag.Int(): scala.reflect.ClassTag))) (of class scala.Tuple2)
        at scala.scalanative.nscplugin.NirCodeGen$SaltyCodePhase$$anonfun$31.apply(NirCodeGen.scala:1533)
        at scala.scalanative.nscplugin.NirCodeGen$SaltyCodePhase$$anonfun$31.apply(NirCodeGen.scala:1533)
        at scala.collection.LinearSeqOptimized$class.foldLeft(LinearSeqOptimized.scala:124)
        at scala.collection.immutable.List.foldLeft(List.scala:84)
        at scala.scalanative.nscplugin.NirCodeGen$SaltyCodePhase.genExpandRepeatedArg(NirCodeGen.scala:1533)
        at scala.scalanative.nscplugin.NirCodeGen$SaltyCodePhase$$anonfun$genMethodArgs$2.apply(NirCodeGen.scala:1513)
        at scala.scalanative.nscplugin.NirCodeGen$SaltyCodePhase$$anonfun$genMethodArgs$2.apply(NirCodeGen.scala:1510)
        at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:733)
        at scala.collection.immutable.List.foreach(List.scala:381)
        at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:732)
[...]

Issue #396, opened 2016-11-18
Make type tagged intrinsic work in generic context


tindzk:
`sizeof` correctly resolves the implicit `ClassTag` parameter:

```scala
println(42 * sizeof[Int])
```

This should be equivalent to: 

```scala
def size[T](length: Int)(implicit ct: ClassTag[T]): CSize = length * sizeof[T]
println(size[Int](42))
```

However, the compiler crashes with the following exception:

```
scala.scalanative.util.package$UnsupportedException: ct (class scala.reflect.internal.Trees$Ident)
```

densh:
Thanks for the bug report!

densh:
Current API that rely on tags that are affected by this:

1. Indirect function calls through function pointers.
2. C-style varargs.
3. Stack and zone allocation. 
4. Sizeof.
5. Reinterpret casts. 
6. Pointer operations.

jokade:
I've encountered a similar problem. I get 
```
java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:347)
	at scala.None$.get(Option.scala:345)
	at scala.scalanative.nscplugin.NirGenUtil$class.unwrapTagOption(NirGenUtil.scala:74)
```
during NIR generation for
```scala
class NSArray[T <: NSObject] protected (implicit evidence$1: ClassTag[T]) extends NSObject with NSCopying with NSMutableCopying with NSSecureCoding {
  // ...
  @inline def arrayByAddingObject(anObject: T): NSArray[T] = _root_.objc.helper.objc_msgSend1[T](this, NSArray.__sel_arrayByAddingObject_, anObject).cast[Object].asInstanceOf[NSArray[T]];
}
```
where `_root_.objc.helper.objc_msgSend1` is a def macro:
```scala
 @inline def objc_msgSend1[T1](self: id, op: SEL, arg1: T1): id = macro Macros.objc_msgSend1Impl[T1]

class Macros(val c: Context) {
  import c.universe._

  val tPtrByte = weakTypeOf[scalanative.native.Ptr[Byte]]

  def objc_msgSend1Impl[T1: c.WeakTypeTag](self: c.Tree, op: c.Tree, arg1: c.Tree): c.Expr[id] =
[...]

Issue #1142, opened 2018-01-09
Cannot pass repeated CVararg params on to external function


jokade:
Is this somehow related to #396? I couldn't find an open issue for this (although I'd thought that's a quite common use case)...

```scala
def foo(args: CVararg*): Unit = Foo.bar(args:_*)

@extern
object Foo {
  def bar(args: CVararg*): Unit = extern
}
```

```
java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:347)
	at scala.None$.get(Option.scala:345)
	at scala.scalanative.nscplugin.NirGenExpr$ExprBuffer$$anonfun$genMethodArgs$2.apply(NirGenExpr.scala:1665)
```

densh:
Finally found a design that works with this in #1426. Instead of compiling to C's varargs, we'll compile to `va_list`. Unlike the previous design it can be abstracted over and passed as normal value as requested in this issue.

densh:
Fixed in #1426

Issue #155, opened 2016-06-07
Add support for linking Scala code as a native library


densh:
We already have an interop layer across C ABI: `@extern object`. 

- [ ] Add support for linking without injecting `main` entry point in the compiler
- [ ] Don't disallow having non-extern methods in extern objects (https://github.com/scala-native/scala-native/issues/79)

This will let our toolchain produce "*.so"/"*.dylib"/"*.dll" files.

RimashMohomed:
@densh interesting to contribute for this feature can you please guide me through

densh:
We need to cook this one a bit longer before we are confident to include it in the official feature set.

drdozer:
I'm potentially interested in this. I can imagine aggressively compiling kernels that don't need to allocate objects down to native code that's basically just maths and pointers, and then loading this back in as a black-box native routine. I'm thinking things like lookups on static tables and indexes.

densh:
Related: #450

velvia:
This would be an awesome feature.  How cooked is the SBT setting?

densh:
@velvia Roughly as raw as meat in steak tartare.

nornagon:
It'd be super neat to be able to dynamically load SN-generated `*.so` files from SN, and call functions in them. I'm imagining a live-reload scenario for game dev :D

jstoiko:
How big would the resulting dynamic library be? Is there any way to get an idea?

densh:
@jstoiko Our current output for the applications is in low single-digit megabytes. We haven't spent much time optimizing for binary size yet. At the moment our goal is to optimize for runtime performance first and foremost.

rwhaling:
.so files could also be useful for integration with Nginx Unit, which dynamically loads language-specific modules with dlopen().

hepin1989:
Highly needed, my friend @WayneWang12 can't use scala-native for the competition of alicloud...

hepin1989:
Any update on this? Would be very interesting to implement out tql with this.

Zizico2:
Any progress?

sheiksadique:
I see that there is an open PR. Are there plans to continue this work and merge the PR?

WojciechMazur:
> I see that there is an open PR. Are there plans to continue this work and merge the PR?

@sheiksadique 
Yes, we are going to continue work in this area. It definitely still needs proper review and a good set of additional tests (especially in terms of performance and ABI compatibility)

sheiksadique:
[...]

PR #1426, opened 2019-01-24
Add support for boxed pointers

### LINKED ISSUE #367: Ptr[T] must box when it escapes to generic context

densh:
This PR introduces two major changes:

### Makes `Ptr[T]` a managed object

So now pointers are:

* Can be used with `asInstanceOf`, so there is no more need for our own special `a.cast[T]` syntax.
* Be passed to generic context (i.e., generic methods that expect boxed representation).
* Have meaningful `toString`, `hashCode` and `equals` based on the value of underlying pointer.
* Are stored in arrays and Scala collections in boxed form. 

Moreover:

* All pointer operations are now well-behaved in generic context as long as there is a provided instance of `native.Tag`. This makes it possible to write type-polymorphic code over the contents of the pointer itself.

### Makes `CArray` a managed object

* Well behaved with respect to boxing.
* Has meaningful `toString`, `hashCode` and `equals` based on the value of underlying pointer.
* Can access elements directly without going back to pointers.

Additionally, we also change encoding for `Nat`. Intead of `Nat.Digit[A, B]` we have multiple overloaded `Nat.DigitN` that correspond to number of digits in the number. So for example `128` is represented as `Nat.Digit3[Nat._1, Nat._2, Nat._8]`.

### Makes `CStruct` a managed object

* Well behaved with respect to boxing.
* Has meaningful `toString`, `hashCode` and `equals` based on the value of underlying pointer.
* Can access elements directly without going back to pointers.

### Changes encoding of function pointers to new SAM-based `CFuncPtrN`.

Similarly to normal pointers:

* Function pointers are now regular objects without any special semantics.
* Declaring a function pointer auto-generates an extern-accessible bridge method.
* Passing a function pointer to extern methods resolves a bridge method and passes that as a pointer value.

Note this still doesn't allow function pointers to have any state. 

Fixes #367 #396


densh:
This pr depends on #1411 for performance reasons. Current implementation is semantically correct but introduces a major performance regression until we get inlining merged.

nativelib/src/main/scala/scala/scalanative/native/Ptr.scala
```diff
@@ -1,142 +1,187 @@
...
-import runtime._
+import scala.runtime.BoxesRunTime._
+import scala.reflect.ClassTag
+import scalanative.runtime._
+import scalanative.runtime.Intrinsics._
+import scalanative.runtime.Boxes._
 
-/** The C `const T *` pointer. */
-final abstract class Ptr[T] {
+final class Ptr[T](private[scalanative] val rawptr: RawPtr) {
```
densh: @shadaj Here is a concrete example of what I meant in the discussion of #1420. `Ptr` type is a wrapper over `RawPtr`. User-facing API is safe from the boxing point of view and encapsulates raw unmanaged `RawPtr`. 

Similarly for `Word` boxing to work, `Word` itself could be a wrapper over `RawWord` which maps to a numeric type under the hood:
```
// .../native/Word.scala
final class Word(private[scalanative] val rawword: RawWord) {
  ...
  @inline def +(other: Word): Word =
     new Word(Intrinsics.addWord(this.rawword, other.rawword))
  ...
}

// .../runtime/RawWord.scala
final abstract class RawWord

// .../runtime/Intrinsics.scala
object Intrinsics {
  ...
  def addWord(l: RawWord, r: RawWord): RawWord = intrinsic
  ...
}
```

shadaj: Awesome, thanks for the example! The semester just started so still adjusting to the workload in my new schedule, but I'll try to update my PR to use this style sometime this week.

unit-tests/src/test/scala/scala/scalanative/native/SizeofSuite.scala
```diff
@@ -75,9 +75,9 @@ object SizeofSuite extends tests.Suite {
                  CStruct3[Int, Int, CStruct4[Byte, Int, Short, Byte]]]] == 40)
   }
 
-  type _32   = Nat.Digit[Nat._3, Nat._2]
-  type _128  = Nat.Digit[Nat._1, Nat.Digit[Nat._2, Nat._8]]
-  type _1024 = Nat.Digit[Nat._1, Nat.Digit[Nat._0, Nat.Digit[Nat._2, Nat._4]]]
+  type _32   = Nat.Digit2[Nat._3, Nat._2]
+  type _128  = Nat.Digit3[Nat._1, Nat._2, Nat._8]
+  type _1024 = Nat.Digit4[Nat._1, Nat._0, Nat._2, Nat._4]
```
densh: /cc @rwhaling This is the new Nat encoding that I mentioned at some point. IMO it looks nicer actually. 😄

densh: With this change, sizeof is always constant folded away and is zero cost despite being completely implemented in the library code.

densh:
Recent changes also make `CArray` and `CStructN` behave like regular objects, similarly to `Ptr`.

ekrich:
This is going to make a huge improvement to making coding easier. 👍

densh:
There are no major performance regressions after this change on our benchmarks. Typical uses of pointers are optimized away and boxing doesn't cause any major overhead. This is quite amazing given how much indirection is there in the current implementation (we removed special-case code from the nscplugin and moved it to the library so optimizer has more work to do).

densh:
@ekrich I'm currently experimenting with improving usability of structs. In current prototype `Ptr -> CStructN` is a no-op conversion that can be done implicitly. I'm not sure if I'm going to keep yet though.

ekrich:
@densh Not explicitly de-referencing `Ptr` looks nice in code but I can understand if you didn't want to keep it. Does the pass by value for `struct` look attainable?

densh:
Delayed structs seem to be really tricky to get right so I'll do it in a separate PR.

densh:
@LeeTibbert It looks like unit testing overflow bug has hit me this time. Time to get this over with, it's caused enough pain already.

densh:
Current if you mix boxing, c-style varargs and new revised structs in this PR:
```
import scalanative.native._
import scalanative.libc._

object Main {
  type Point = CStruct2[CInt,CInt]
  def main(args:Array[String]):Unit = {
    val p:Ptr[Byte] = stdlib.malloc(sizeof[Point])
    val point:Ptr[Point] = p.asInstanceOf[Ptr[Point]]
    point._1 = 5
    point._2 = 8
    stdio.printf(c"(printf) point at %p: (%d,%d)\n",point,point._1,point._2)
    println(s"(println) point: $point, x: ${point._1}, y: ${point._2}")
  }
}
```
It doesn't do what you expect and prints wrong value for the printf call. As discovered by @rwhaling.

densh:
So thanks to the nudge from @rwhaling we now have 100% generic and abstraction-friendly `CVarArg` and `CVarArgList`. This fixes #153, and #1142.

densh:
@ekrich Currently we only have one-way interop: Scala can call C. If/when we make it work the other way around (i.e. #155) we might add support for reading values from `CVarArgList` as well.

nativelib/src/main/scala/scala/scalanative/native/CVarArgList.scala
```diff
@@ -0,0 +1,110 @@
...
+import scalanative.runtime.{intrinsic, RawPtr, toRawPtr, libc, LongArray}
+
+/** Type of a C-style vararg list (va_list in C). */
+final class CVarArgList private[scalanative] (
+    private[scalanative] val rawptr: RawPtr)
+
+object CVarArgList {
+  // Implementation below is based on VarArgs.swift from apple/swift repo.
+  // Currently we only support X86_64, we'll need to revisit the code below
+  // if and when we add any more officially supported architectures.
```
densh: /cc @shadaj

PR #1427, opened 2019-02-05
Don't publish docs and sources if SNAPSHOT

jvican:
Publishing sources and docs is not required for local versions and slows
down the workflow cycle when doing `rebuild` or depending on `package`
in the compilation of certain Scala Native modules.

To speed up the `publishLocal`/`publish` workflow, we don't package
source and doc classifiers, meaning there will be no -src and -doc jars
published in the local repository if the version is a SNAPSHOT.


densh:
Hi @jvican, please run `scripts/scalafmt` to unblock the CI.

jvican:
Done.

PR #1429, opened 2019-02-10
[0.3.x] Move UnicodeBlock and Subset under Character object

### LINKED ISSUE #1388: java.lang.Character.UnicodeBlock.of segfaults
I'm using version 0.3.8

The following code causes a segfault:

```
object Main {
    java.lang.Character.UnicodeBlock.of('a')
  }
}
```

When I copied the minimal code from [Character.scala](https://github.com/scala-native/scala-native/blob/v0.3.8/javalib/src/main/scala/java/lang/Character.scala) to a local file it did not crash. However, there's something strange about this file, `object UnicodeBlock` is defined inside `class Character` and not inside `object Character`, so I am not sure why even the call works, on my local copy I removed UnicodeBlock to be a top-level object.

thesamet:
This is a port of 494b7499 to 0.3.x.

Currently those classes are defined under the Character class. As a
result they do not work properly because member fields (such as BLOCKS)
is not initialized.

Fixes #1388 for 0.3.x so we can fix lihaoyi/fastparse#215, scalapb/scalapb#510 scalapb/scalapb#540

### DIFF
--- a/javalib/src/main/scala/java/lang/Character.scala
+++ b/javalib/src/main/scala/java/lang/Character.scala
@@ -176,758 +176,6 @@ class Character(val _value: scala.Char)
   protected def %(x: scala.Long): scala.Long     = _value % x
   protected def %(x: scala.Float): scala.Float   = _value % x
   protected def %(x: scala.Double): scala.Double = _value % x
-
-  // Ported from Harmony
-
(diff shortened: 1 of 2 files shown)


densh:
LGTM, thanks @thesamet !

PR #1437, opened 2019-02-26
Interflow: code motion

densh:
This PR introduces one more optimization as a part of the Interflow's main pass: code motion. Previously we never reordered operations in the original code, they could either be partially evaluated or where emitted into the optimized output in the same order as they were defined originally. We used to run DCE in a separate pass to remove pure operations whose result was never used. Instead, we add support for "delayed" ops into Interflow's optimizer state. 

For example:
```
%entry:
   %x = add[Int] %y, 3
   %cond = gt[Int] %y, 0 
   if %cond then %ifTrue else %ifFalse
%ifTrue:
   %z = mul[Int] %x, 42
   jump %merge(%z)
%ifFalse:
   jump %merge(%y)
%merge(result: Int):
   ...
```

Here `%x` is defined in `%entry` but only used in `%ifTrue`. Interflow now can sink the definition of `%x` into the `%ifTrue` branch, and completely eliminate it from other branches:
```
%entry:
   %cond = gt[Int] %y, 0 
   if %cond then %ifTrue else %ifFalse
%ifTrue:
   %x = add[Int] %y, 3
   %z = mul[Int] %x, 42
   jump %merge(%z)
%ifFalse:
   jump %merge(%y)
%merge(result: Int):
   ...
```
The optimization extends infrastructure for Partial Escape Analysis: pure operations are removed from the code when they are visited, and later materialized when the virtual value escapes (i.e. used by other emitted op). This turns materialization into a graph scheduling problem, so state materialization and block invalidation had to be improved a bit to support this use case.

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Ops.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Ops.scala
@@ -54,10 +54,14 @@ sealed abstract class Op {
     case _: Op.Elem | _: Op.Extract | _: Op.Insert | _: Op.Comp | _: Op.Conv |
         _: Op.Is | _: Op.Copy | _: Op.Sizeof =>
       true
-    // Division and modulo on integers are not pure as
-    // they may throw if the divisor is zero.
-    case Op.Bin(Bin.Sdiv | Bin.Udiv | Bin.Srem | Bin.Urem, _: Type.I, _, _) =>
-      false
+    // Division and modulo on integers is only pure if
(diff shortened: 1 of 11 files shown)


densh:
This improves throughput on some of our benchmarks by 3-5%, while most are unaffected.

densh:
Hello world binary on macOS (release mode, thin lto) dropped to 944Ki, down from 1.06Mi (-12%).

PR #1438, opened 2019-02-26
Reduce size overhead of runtime type information

densh:
This PR reduces binary size overhead of dynamic hash maps (underlying data structure used for the implementation of the Scala's reflective method calls). Additionally, the desugaring of reflective calls uses similar pattern introduced in #1432 by delaying the slow path until the end of the method and sharing it as long as unwind handler is the same.

### DIFF
--- a/nativelib/src/main/resources/gc/immix/headers/ObjectHeader.h
+++ b/nativelib/src/main/resources/gc/immix/headers/ObjectHeader.h
@@ -19,19 +19,9 @@ typedef struct {
         int32_t id;
         int32_t tid;
         word_t *name;
-        int8_t kind;
     } rt;
-    int64_t size;
-    struct {
-        int32_t from;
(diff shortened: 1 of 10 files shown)


nativelib/src/main/resources/gc/immix/headers/ObjectHeader.h
```diff
@@ -19,19 +19,9 @@ typedef struct {
...
-        int32_t to;
-    } range;
-    struct {
-        int32_t dyn_method_count;
-        word_t *dyn_method_salt;
-        word_t *dyn_method_keys;
-        word_t *dyn_methods;
-    } dynDispatchTable;
+    int32_t size;
+    int32_t idRangeUntil;
```
densh: /cc @valdisxp1 You'd need to apply the same change in commix PR.

PR #1441, opened 2019-02-27
Interflow: infer module constructor purity

densh:
This change relaxes the definition of "pure" module constructor to allow the following inside the module constructor:

1. Any non-throwing control-flow instructions
1. Any pure `let n = op` instructions
1. Loads and stores to currently accessed module
1. Loads and stores to any values allocated inside the module constructor
1. Access to other pure modules

This also allows for modules which store other modules values as "optimization" (i.e. many modules in standard library do this).


densh:
Thanks to @sjrd for the idea.

PR #1442, opened 2019-02-28
Interflow: canonicalize constants to be on the right

densh:
Cannonicalize constants to always go on the right of the
commutative binary and comparisons ops.

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Ops.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Ops.scala
@@ -84,6 +84,26 @@ sealed abstract class Op {
     case _ =>
       false
   }
+
+  final def isCommutative: Boolean = this match {
+    case Op.Bin(bin, _, _, _) =>
+      import Bin._
+      bin match {
+        case Iadd | Imul | And | Or | Xor =>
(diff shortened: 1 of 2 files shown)


nir/src/main/scala/scala/scalanative/nir/Ops.scala
```diff
@@ -84,6 +84,26 @@ sealed abstract class Op {
...
       false
   }
+
+  final def isCommutative: Boolean = this match {
+    case Op.Bin(bin, _, _, _) =>
+      import Bin._
+      bin match {
+        case Iadd | Imul | And | Or | Xor =>
+          true
+        case Fadd | Isub | Fsub | Fmul | Sdiv | Udiv | Fdiv | Srem | Urem | Frem |
```
sjrd: If `Fmul` represents floating-point multiplication, it's commutative. (It's not associative, though.)

densh: 👍

densh: `Fadd` is commutative as well it seems.

PR #1444, opened 2019-03-01
Interflow: combine instructions 

densh:
This PR revives InstCombine optimization, only as part of the Interflow's main optimization pass rather than a stand-alone pass. The goal of InstCombine is to combine pairs or triples of connected instructions and reduce them down to a simpler alternative based on algebraic equivalences. The implementation is based on the delayed instructions introduced in #1437.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/interflow/Combine.scala
+++ b/tools/src/main/scala/scala/scalanative/interflow/Combine.scala
@@ -88,6 +88,10 @@ trait Combine { self: Interflow =>
           case (BinRef(Iadd, x, y1), y2) if y1 == y2 =>
             x
 
+          // (x + y) - x ==> y
+          case (BinRef(Iadd, x1, y), x2) if x1 == x2 =>
+            y
+
           case _ =>
             fallback
         }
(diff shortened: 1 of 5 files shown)


tools/src/main/scala/scala/scalanative/interflow/Combine.scala
```diff
@@ -0,0 +1,615 @@
...
+package interflow
+
+import scalanative.nir._
+import scalanative.linker._
+import scalanative.util.{unreachable, And}
+import nir.Bin.{And => Iand, _}
+import nir.Comp._
+import nir.Conv._
+
+trait Combine { self: Interflow =>
```
densh: @sjrd It would be great if you could give it a brief look in case anything looks obviously wrong.

densh:
Dropped binary op overloads on `Byte`, `Char`, `Short` as they are never used on Scala code. Fixed a bug in redistribution on shift ops.

tools/src/main/scala/scala/scalanative/interflow/Combine.scala
```diff
@@ -0,0 +1,554 @@
...
+          Some(Val.Int(jl.Integer.numberOfTrailingZeros(v)))
+        case Val.Long(v) if isPowerOfTwo(v) =>
+          Some(Val.Long(jl.Long.numberOfTrailingZeros(v)))
+        case _ =>
+          None
+      }
+    }
+
+    def isPowerOfTwo(x: Long): Boolean =
+      (x & (x - 1)) == 0
```
tarsa: That will return `true` for `Long.MinValue`, but `false` for any other negative Long value.

densh: Yep, it matches but the optimization that uses it still works even on min value. I renamed the helper method to be more explicit that it also matches min value.

tarsa: java.lang.Math contains methods that end with `Unsigned` but take normal longs and ints. With such convention the name of above function would be `isPowerOfTwoUnsigned`. Would that fit?

densh: Treating is an unsigned power of 2 makes perfect sense. Given that some of rules here work on signed integers as well, I'd probably keep the current name.

PR #1451, opened 2019-03-08
Reduce binary size overhead of the new floating point to string conversion

densh:
This PR adds a few minor changes that shrink down the binary size overhead of the #1436 to around 25k (split roughly in half between code and data for the arrays).

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/ieee754tostring/ryu/RyuRoundingMode.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/ieee754tostring/ryu/RyuRoundingMode.scala
@@ -35,34 +35,19 @@ package scala.scalanative
 package runtime
 package ieee754tostring.ryu
 
-object RyuRoundingMode extends Enumeration {
+abstract class RyuRoundingMode {
+  def acceptUpperBound(even: Boolean): Boolean
+  def acceptLowerBound(even: Boolean): Boolean
+}
 
-  val CONSERVATIVE: RyuRoundingMode = new conservativeRM()
-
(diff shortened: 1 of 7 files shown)


nativelib/src/main/scala/scala/scalanative/runtime/ieee754tostring/ryu/RyuRoundingMode.scala
```diff
@@ -35,34 +35,19 @@ package scala.scalanative
 package runtime
 package ieee754tostring.ryu
 
-object RyuRoundingMode extends Enumeration {
```
densh: Friends don't let friends use `scala.Enumeration`. This used to add a few hundred KB in transitive dependencies.

nativelib/src/main/scala/scala/scalanative/runtime/ieee754tostring/ryu/RyuRoundingMode.scala
```diff
@@ -35,34 +35,19 @@ package scala.scalanative
...
-  }
-
-  private[this] class conservativeRM extends RyuRoundingMode {
+abstract class RyuRoundingMode {
+  def acceptUpperBound(even: Boolean): Boolean
+  def acceptLowerBound(even: Boolean): Boolean
+}
 
+object RyuRoundingMode {
+  object Conservative extends RyuRoundingMode {
```
densh: By making the instances of `RoundingMode` be `object`-s, Interflow now specializes the floatToString/doubleToString to the rounding mode. This means two (up to) copies of the method where rounding mode is statically known and moth of the `acceptUpperBound` and `acceptLowerBound` are constant and folded away. In practice we only use conservative mode now, so it's exactly one copy.

PR #1475, opened 2019-03-20
Test release mode on travis

densh:
Thanks to socket fixes by @LeeTibbert we can do this now without having to blacklist any tests.


ci-docker/Dockerfile
```diff
@@ -47,4 +47,4 @@ USER scala-native
 
 WORKDIR /home/scala-native/scala-native
 
-CMD sbt -no-colors -J-Xmx3G "^^ $SBT_VERSION" rebuild "set scriptedBufferLog in sbtScalaNative := false" "$TEST_COMMAND"
+CMD sbt -no-colors -J-Xmx${SBT_HEAP} "^^ $SBT_VERSION" rebuild "set scriptedBufferLog in sbtScalaNative := false" "$TEST_COMMAND"
```
densh: @shadaj Do we need to manually republish the docker image every time we change the dockerfile?

shadaj: Nope, the images are built from source during each Travis run!

PR #1478, opened 2019-03-20
Add unimplemented method to java.util.Properties

ekrich:
This is an extra method needed to link `ekrich/sconfig` that was found when working on the `java.net.URL` code change but it seems better as a separate PR. We have had some PRs working on `Properties` but nothing solid yet and this should be easy to rebase for WIP.

EDIT: This is targeted for both `0.3.x` and `master`.

### DIFF
--- a/javalib/src/main/scala/java/util/Properties.scala
+++ b/javalib/src/main/scala/java/util/Properties.scala
@@ -3,6 +3,7 @@ package java.util
 import java.{util => ju}
 
 import scala.collection.JavaConverters._
+import scala.scalanative.native.stub
 
 class Properties(protected val defaults: Properties)
     extends ju.Hashtable[AnyRef, AnyRef] {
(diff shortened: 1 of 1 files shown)


javalib/src/main/scala/java/util/Properties.scala
```diff
@@ -14,6 +14,8 @@ class Properties(protected val defaults: Properties)
 
   def load(inStream: java.io.InputStream): Unit = ???
 
+  def load(reader: java.io.Reader): Unit = ???
```
densh: Please add `@stub` annotation.

ekrich: In general, are `@stub` suppose to be everywhere there is an empty implementation?

densh: Responded on gitter: :point_up: [March 21, 2019 12:20 PM](https://gitter.im/scala-native/scala-native?at=5c9373675547f774483ec225)

densh:
LGTM, thanks @ekrich ! 👍

ekrich:
Refer to Issue #1412

Issue #745, opened 2017-05-26
Add support for polymorphic inline caching


densh:
Inline caching is a well-known optimization that uses per-call-site statistics to optimize virtual dispatch. It's typically employed in JIT-ed VM environments where such statistics is collected in early stages of the application run (typically in interpreter on in compiled code generated by baseline compiler) and is then used by the best-tier compiler to produce fully optimized code.

Given our design constraints, we can not do this at runtime (i.e. Scala Native is an AOT compiler) but we can still optimize this by providing means for user to collect statistics in a separate run and use it for subsequent fully-optimized compilation.

densh:
Fixed in #1495 based on static analysis rather than profile data.

PR #1495, opened 2019-03-22
 Interflow: polymorphic inlining

densh:
This PR adds support for polymorphic inlining based on static target prediction. This gives a major performance boost on DeltaBlue, Tracer and Permute benchmarks.


densh:
Fixes #745

PR #1499, opened 2019-03-24
Stdbool bindings

sorenbug:
I followed what the `float.h` binding PR did (i.e. I had no real clue what I was doing).

### DIFF
--- a/clib/src/main/scala/scala/scalanative/libc/stdbool.scala
+++ b/clib/src/main/scala/scala/scalanative/libc/stdbool.scala
@@ -1,0 +1,15 @@
+package scala.scalanative
+package libc
+
+import scalanative.native._
+
+/**
+ * Bindings for stdbool.h
+ */
+@extern
+object stdbool {
+  @name("scalanative_bool_true")
+  def `true`: CBool = extern
+  @name("scalanative_bool_false")
+  def `false`: CBool = extern
+}
--- a/nativelib/src/main/resources/wrap.c
+++ b/nativelib/src/main/resources/wrap.c
@@ -4,6 +4,7 @@
 #include <math.h>
 #include <errno.h>
 #include <float.h>
+#include <stdbool.h>
 
 // This file contains functions that wrap libc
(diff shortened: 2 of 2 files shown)


avdv:
Hi.

According to the ISO C standard, the `true` and `false` identifiers are simply macros, expanding to the integers `1` or `0`, respectively.

Since `CBool` is just an alias for `Boolean` in Scala Native, there's already perfect interop and I see no reason one would use <tt>\`true\`</tt> instead of just `true`...

densh:
@avdv Agreed, I think there is very little reason to use binding to C's stdbool over just `true` and `false` in Scala.

sorenbug:
So then, what about `stdbool` in #104? Does it even make sense in that context?

ekrich:
No I think in that issue and in the documents needs to be removed including any other we know that we don't need. In general, we do the `extern` interfaces to C lib using types that are already included in other interfaces or the ones in https://github.com/scala-native/scala-native/blob/master/nativelib/src/main/scala/scala/scalanative/unsafe/package.scala . This way we can avoid duplication and nicely tie back to the `Word` and other standard types for our eventual 32 bit support.

sorenbug:
Ok. Thanks for the clarification! :)

ekrich:
There are quite a few moving parts and the beast is evolving as well. Your PR was well meaning so thank-you.

densh:
@sorenbug Sorry for the confusion, that ticket just lists all headers in C library without much thought put in which of those are useful in Scala.

sorenbug:
@densh So then, what will be done about bindgen on libraries that use stdbool?

densh:
stdbool defines:

* `bool` type, it maps to `Boolean` in Scala with 1-to-1 the same semantics and ABI.
* `true` is a macro which doesn't exist on ABI level.
* `false` is a macro which doesn't exist on ABI level.

I believe bindgen already uses `Boolean` for `bool`, so there is nothing to do really.

PR #1512, opened 2019-03-27
Interflow: make sure all reachable methods are visited

### LINKED ISSUE #1278: Interflow

densh:
Previously in a number of conditions Interflow could not reach some of the methods:

1. Methods reachable from methods on which optimization bailed out. We fix this by introducing a "conservative" reachability-like visiting mode which had to be done in case of bail out. Additionally, we also add new `Opt` definition attributes, which describe if the optimizer was able to optimize this method, and if not than what was the reason of bail out. 

2. Recursive methods used to short-circuit to the original signature but in some cases it would not be visited. This would cause recursive methods to jump into (potentially large) unoptimized subset of the program. We need to short-circuit recursion due to our whole-program type type inference. But this doesn't mean short-circuited version can not be optimized.

3. Interflow did not consider extern forwarders which are passed by value as reachable, but they are as they could be called by C code. 

This fixes our last regression with respect to the original interflow announcement performance numbers vs current master.

Fixes #1278


densh:
An unfortunate side-effect of this change is that it makes linking tests in release mode take forever. Previously, major fraction of the test suite wasn't optimized due to the issues mentioned above.

PR #1536, opened 2019-04-12
Split mangled and unmangled signatures

densh:
This change splits global signatures into two categories:

1. Mangled (default) just a string containing a mangled version of the signature.
2. Unmangled (recovered on-demand) an ADT that lets you pattern-match on signature structure.

This improves performance of signatures and globals when they are used as keys in hash maps.


jvican:
:+1: Will follow up with my changes to the linker

PR #1561, opened 2019-04-24
Update scalafmt version

ekrich:
The primary purpose of this PR is to update `scalafmt` to the latest production version `1.5.1`.

The secondary changes are are as follows:
- Change the `scalafmt` CLI runner to use the Scala 2.12 version
- Fix the `sbt` startup hook so Metals can index all the projects.
- Update minor versions of Scala and sbt.
- Add `version = 1.5.1` to the `.scalafmt.conf` file so you can format from inside VSCode.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,3 +1,4 @@
+version = 1.5.1
 style = defaultWithAlign
 docstrings = JavaDoc
 assumeStandardLibraryStripMargin = true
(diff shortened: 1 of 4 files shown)


ekrich:
@jvican This should interest you. If it passes CI then it should be ready for review.

densh:
LGTM, thanks @ekrich ! 👍

PR #1566, opened 2019-04-25
Partial Fix #1560: Add thread stubs for scalafmt

### LINKED ISSUE #1560: Linking errors for scalafmt cli
```
```

```
```

ekrich:


ekrich:
@densh @olafurpg I can do a backport once this is merged. API for reference/review. https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html

densh:
LGTM! 👍

PR #1602, opened 2019-05-22
Fix #1197: rename scalanative.native into scalanative.unsafe

### LINKED ISSUE #1197: Rename scalanative.native to scalanative.unsafe
1. motivation:
scala-native uses packages names, such as `scala.scalanative.native`:
https://github.com/scala-native/scala-native/tree/master/nativelib/src/main/scala/scala/scalanative/native
which contains `java` reserved keywords, such as `native`,
which results in various problems in Scala IDE, 
including inability to view class files from such packages

2. a solution could be just to rename such packages, 
i.e. `native` -> `nativelib`, instead of fighting various IDEs

3. the word `native` itself is heavily overused and overloaded
perhaps `scala.scalanative.clang` or `scala.scalanative.interop` are more descriptive terms

densh:


ekrich:
This might require a change to the `bindgen` project when it outputs the code.

densh:
@ekrich both this and boxed ptr will affect bindgen.

ekrich:
Thanks, just didn't know if you were thinking about that or not.

PR #1603, opened 2019-05-23
Release 0.4.0-M1

densh:


densh:
@LeeTibbert I had to replace all scaladoc comments in `scalanative.regex` with `//` comments instead. Scaladoc is crashing on one of the files and I don't have time to debug this right now. Given that API is not officially public this is not a blocker.

PR #1607, opened 2019-05-24
 Release 0.4.0-M2

densh:


densh:
We build M1 on JDK11 which has different binary signatures for some of the methods in ByteBuffer APIs, causing it not to work on JDK8. M2 is literally the same release as M1 only build on JDK8 instead.

PR #1632, opened 2019-06-22
[WIP] Support embedding of external source strings in NIR

jokade:
**Description**:
This PR provides a new annotation `ExternalSource(language: String, source: String)` which allows to embed source snippets into the NIR generated for a class or object, e.g.
```scala
@ExternalSource("C","int foo(int i) { /* ... */ }")
@extern
object Foo {
  def foo(i: Int): Int = extern
}
```
These source snippets can then be processed with sbt plugins by registering an `ExternalSourceHook` to the task key `nativeExternalSourcesHooks`.

Use Cases:
- embed small external helper functions
- on-the-fly generation of wrapper code (e.g. for C++ libraries) with macros (i.e. the macro generates the `ExternalSource` snippet that contains the C wrapper around some C++ type).

**Changes**:
- add new annotation `ExternalSource(source,language)`
- add new NIR tag `ExternalSrcAttr`
- handle tag in BinarySerializer/BinaryDeserializer
- collect all ExternalSource attrs during linking
- add Config.externalSourceHooks; registered hooks are called with the list of ExternalSource attrs after optimization is completed
- provide sbt settingKey nativeExternalSourcesHooks to register  hooks that should be called after optimization is completed


ekrich:
Related: https://github.com/scala-native/scala-native/issues/657

coreyoconnor:
This reminds me of "inline-c" for GHC haskell. Which has definitely been proved to be useful.

The name "ExternalSource" sounds odd to me. The `@extern` makes sense as short for "has an external implementation". However `ExternalSource` doesn't really follow the same conceptual expansion IMO

How about mirroring haskell and using "InlineSource"?

coreyoconnor:
for reference: https://hackage.haskell.org/package/inline-c

jokade:
@coreyoconnor good point, I didn't put too much thought into it. I'll go with your proposal.

lolgab:
Do we need this at this point?
It has not been touched for 5 years and Scala Native is moving away from C and C++ code more and more.

keynmol:
Honestly I feel like this would be a good idea if the annotations were on functions – for cases where there is naked struct passing the forwarders are very mechanical and I wish they just lived in the Scala source as well, instead of having to ship a C file: https://gist.github.com/keynmol/642af3a015d3aa05cefdadfb8635f5d3#file-clay-glue-c-L9-L12

lolgab:
sn-bindgen would still need to generate c source code, just in a different place, without autocomplete, source highlighting, etc.
I don't see much value in this feature but I'm happy to be proved wrong.

PR #1636, opened 2019-06-26
Fix #1606: Add printf vararg helpers

### LINKED ISSUE #1606: Provide printf and scanf family of functions in clib
Since we document how to provide an idiomatic printf in the docs now, we might as well provide the essential printf/scanf functions as part of clib.  

If we provide them as part of the SN distribution, that should cover 99% of uses of CVarArgs that will occur - they are otherwise quite uncommon in library and application code.

The functions we'd need are:

printf()
fprintf()
sprintf()
snprintf()
scanf()
fscanf()
sscanf()

rwhaling:
Adds a libc package object, which provides CVarArg* wrappers around stdio for the printf/scanf family of functions, which are incompatible with new interop in their previous form.


ekrich:
@rwhaling Looks like it needs some `scalafmt` love.

densh:
LGTM!

Issue #1352, opened 2018-09-18
Proposal: modularize core libraries


densh:
Currently Scala Native is mostly a monorepo-style project. The monorepo includes a few coarse-grained library projects within it:

1. nativelib -- core native interop, core runtime classes, libc bindins, posix bindings
2. javalib -- java library compatibility layer
3. scalalib -- scala library cross-built for native

This has caused problems in terms of portability to new platforms/architectures as all-or-nothing approach means that we can only support new platform if 100% of the existing code work on it (e.g. windows doesn't have posix). 

Instead I suggest to split the libraries into finer grained modules by introducing extra modules for:
- [x] clib -- libc bindings (pr #1353)
- [x] posixlib -- posix bindings (pr #1350)
- [ ] javacorelib -- absolute basics of java namespace such as `java.lang`, `java.util` etc. 
- [ ] javafilelib -- file i/o (currently based on posix)
- [ ] javanetlib -- network i/o (currently based on posix)
- [ ] javaziplib -- zip and jar support (currently based on zlib)

Having multiple smaller projects will allow us to introduce early support for new platforms as soon as basics such as nativelib, clib and javacorelib are working on it. Additional libraries may adopt new platforms with some delay (if at all). Some libraries might remain platform-dependent forever such as posixlib. 

/cc @muxanick @jonas

jonas:
How will this affect developers? For example, will a developer have to opt-out of `javaziplib` if it is not available for the specific platform?

densh:
@jonas If some API (lets say zlib) is not available on the target system, the project that uses it will fail to link at LLVM link time. This is just as true now if for example if you try to use boehm gc without having it installed.

muxanick:
Quick question, should I split unit tests? Like ProcessSuit.scala import posix

muxanick:
@densh I need a help to begin this task, please, I don't understand some basics like how to split `java.io` into two libs if there are `private[io]` methods doesn't work after split.
I have tried to move out `java.nio` but `String` used it a lot and other files as well.
Should I rewrite them all or I misunderstood something?
There are many small things like: `Filesystem` used `java.net.URI` class which is in another lib, which uses `Filesystem` and depends on it. I can try to rewrite `URI` class if needed but I want to know if that is desirable result.

sjrd:
* There are 3 `private[io]` defs in the javalib currently, and none of them are used from other logical components, so the split won't affect them.
[...]

PR #1637, opened 2019-06-26
Fix #657: Give libraries a way to include native C code to be compiled.

### LINKED ISSUE #657: The sbt plugin should offer a way to compile C files

ekrich:
Currently, C or C++ code added into the `resources` directory is used by Scala Native `nativelib`. The basic need for adding C code to a library is to support C macros and any adapter functions for C libraries that pass `struct`s by value. Adding the raw C is portable and a good solution. I have two use cases I am targeting at this time.

- [x] A published library that contains C or C++ code - currently this code goes in `resources` which gets copied to the root of the jar archive. The `nativelib` and any added `nativeLibraryDependencies`, a new plugin setting,  are unpacked and compiled as part of the application. ~~Currently, you also must add to `nativeLinkingOptions` for each library to link.~~ Edit: This was incorrect as Scala Native keeps track and adds these to the link phase.

The `Config` class contains a deprecation and also a runtime deprecation warning as 3rd party tools such as `coursier` may need some minor changes to allow this feature to work. We may need to deprecate and refactor some of the other public `tools` APIs as well depending on need.

- [x] You have an application that needs a little C code. This use case adds an additional plugin setting `nativeCodeInclude` which is false by default. The intent is to combine any native code to the compile and link sequence when set to `true`. This way the current `publishLocal` methodology will continue to function without change.

There are quite a few other ideas in the issue but I think these 2 items are a good start and enable libraries and local coding without local changes to the build system.

Edit: 2019-07-19
This conceivable allows Scala Native to distribute `clib` or `posixlib` as separate artifacts if the code is not used in `javalib` or other core parts of the system.


ekrich:
I have removed the WIP as this could probably use some review or discussion at this point.

Here are some highlights of the implementation.

1. The code handles the Scala Native `nativelib` exactly as it does for third party jars with native code with the following exceptions:
  - Always unpacks `nativelib`
  - Only unpacks `otherlib` jars if listed on the `nativeLibraryDependencies` classpath. The `otherlib` will be unpacked in a directory named the same as your artifact name in Maven.
2. If you specify `nativeCodeInclude := true` then you can add C code to your project and the Scala Native plugin will copy the code to the `native` build area and create a sub-directory named the same as you project and will compile and link that code.
3. The code is designed to be used with other build tools with nothing specific to `sbt` except in the plugin for ease of use with `sbt`.

The changes deprecate `nativelib` in `Config` with a runtime deprecation Warning message. The preferred method is now to provide a `Seq[NativeLib]` as `nativelibs` rather than just `Path` as `nativelib`.

ekrich:
Now, directory C/C++/S files are hashed to avoid a copy when they have not changed. This is similar to when a jar doesn't change, it is not unpacked. Computing the hash of the `nativelib` jar on my machine takes about 30 ms and I am guessing that any sort of unpacking of a 5284045 byte `nativelib` jar and writing to disk is slower so I think that is why the original code hashed the jar. So now if the project is large, this optimization will apply for directory copying.

ekrich:
Ready for Review.

ekrich:
Given an interest in https://github.com/scala-native/scala-native/issues/155, this does allow a new class of libraries that were not possible before without hacking.

This PR allows you to distribute a pure Scala Native library that can include C/C++/S code in a standard jar format exactly how the Scala Native `nativelib` is distributed. Your library can be developed locally with tests or example applications so you can link the library.

My test case library is [stensorflow](https://github.com/ekrich/stensorflow) which is currently just a Scala Native C binding with no high level Scala code. It is not 100% working yet. It currently uses an application to test itself.

The PR above still uses the standard static link setup but does allow the full Scala Native optimization system to work on any application or library (with tests or examples to force a link) for the library of interest. It is very easy to use in a natural way with the current Scala Native plugin or raw tool chain.

muxanick:
@densh What the current plan about it? It is very useful change

ekrich:
I have been thinking about this PR and I think it might be helpful to put the native code in a directory specified by the full org and project name to have a unique namespace. So instead of `nativelib` it would be `org_scala-native_nativelib` or for `stensorflow` it would be `org_ekrich_stensorflow`.

ekrich:
@lolgab I added namespaces based on the organization and project name in the last commit.

ekrich:
### Recap

Allow internal Scala Native components and external libraries to include C code which will be compiled on demand by the Scala Native plugin.

### Vision of the future

Rather than Scala Native be a platform where everything is included, this opens up the potential where Scala Native can consist of components that are included from the repository based on the platform or potentially some future configuration. This could help us make Scala Native scale from smaller devices and different platforms, other than Linux/UNIX, to more enterprise usages. Some examples follow:

- Windows support:  sharing what it can via C conditional compilation and separate projects where this strategy does not make sense.
- Robotics and 32bit support as demonstrated by Shadaj Laddad.
- Potentially WASM
- Potentially different concurrency approaches

We still want to support Java libraries like we do now to promote a robust set of portable cross projects. This is a change that can allow an evolution of the platform guided by the leadership of this project and the needs and contributions from the community. 

I know there has been some frustration especially related to Windows support as the requested changes would have been too disruptive and potentially destabilizing. This should allow some of the development to run in parallel in separate projects that will not affect the stability of the current system.

ekrich:
Looking at `jsDependencies` and the plugin for that leads me to think that this is really not the same type of problem we are trying to solve. 

This really is not about compiling C per se. If you are building an application and you can't really create an Scala Native interface to C with `extern` directly to call a C library because Scala Native does not support a perfect C interopt then this allows you to toss a little C code into your project to fill the gap. Many people have just given up at this point since you can't do this and there are numerous hacked solutions out there to get this to work. Also, if you are building a library that has the same problem, you can include a little C code that can be compiled when your library is used. This is the `tensorflow` example. Remarkable BLAS as in `sblas` doesn't need any C code so a full BLAS interface is available for Scala Native.

Compiling C is already baked into the `tools` project but only for the `nativelib` artifact. I am not sure exactly how this PR would be done without the `nativeLibraryDependencies` because you would have to search the entire classpath to find the native code looking inside each jar and also we use the `org` and `name` of the artifact to create a unique directory name to place the code to be compiled. The sbt plugin basically just calls into the `tools` API knowing which jars to expand so that the C code can be included for the C compiler as it is being used in the normal tool chain.

I'm not sure if this helps explain or answers you questions exactly. I do think something other than this solution is going to be a much bigger development project.

I tried to minimize the complexity/changes and put only what needs to be put in the plugin and putting the rest into the `tools` project. I think the changes are quite modest and should be maintainable. Currently, the `nativelib` is one big blob that doesn't allow the code to be segregated to the internal sub-project/artifact where it belongs like `posixlib` or `clib` either so somehow we need to solve that problem because `posixlib` needs to be excluded for Windows. I am thinking this may help with that as well but have not tried to reorganize Scala Native using this PR. I added a `scripted` test to test the `nativeCodeInclude` feature and tested `stensorflow` with this PR for a library test.

I am really not sure about how `coursier` or other tools that may call the `tools` API and will deal with the additional case classes that are in the `Discover` object when creating a `Config`. Not sure who has experience in that area.

@rwhaling  @lolgab Could you take a look and see what you think about this change and approach?

shadaj:
> Compiling C is already baked into the `tools` project but only for the `nativelib` artifact. I am not sure exactly how this PR would be done without the `nativeLibraryDependencies` because you would have to search the entire classpath to find the native code looking inside each jar and also we use the `org` and `name` of the artifact to create a unique directory name to place the code to be compiled.

A bit late to the discussion, but the strategy for searching the entire classpath is actually implemented in my fork for 32-bit support + JNI bridging (see https://github.com/shadaj/scala-native/commit/68887063af05964e86f0e65a0e5e030ad8cf3892). To ensure unique folder names, I just used the name of the JAR file the sources were extracted from. Then, any library just needs to add C sources they need with a extra module dependency (https://github.com/Team846/scala-native-wpilib/blob/master/scalaNativeJNINativeLib/src/main/resources/mockjni.cpp).

While not a large-scale usage of the feature, since we only had one externally provided C source, this strategy seemed to work pretty well since there were no changes needed to projects using these extra C sources as they were automatically brought into the classpath by regular dependency resolution.

ekrich:
This looks like you used `nativeUnpackLib` as a `Seq` maybe and was also before the code was moved out of the plugin and into `tools` so it looks different now. I see the `jar` naming too thanks. I originally was going to use just the artifact name (similar to jar name) but worried that 2 people could name their library the same and added the `org` name to make it more unique.

sjrd:
> If you are building an application and you can't really create an Scala Native interface to C with `extern` directly to call a C library because Scala Native does not support a perfect C interopt then this allows you to toss a little C code into your project to fill the gap. Many people have just given up at this point since you can't do this and there are numerous hacked solutions out there to get this to work.

I think this is the root problem that needs to be solved. We need to make Scala Native able to directly interface with *any* C library. It needs perfect interop. I managed to do that for Scala.js with a perfect interop with JS, and interop with JS is fundamentally much more complex than interop with C, so I'm pretty confident we can achieve perfect interop for C.

ekrich:
I agree that "perfect interopt" is the ultimate solution but I don't think that is the only issue this is solving.

1. The current plugin unpacks the `nativelib` and compiles the C/C++ along with the other code.
2. Each project should hold its own C code including perhaps separate projects for each GC so this can be worked on in an independent manner.
3. Having all the C/C++ code in the same project means that each project itself lacks encapsulation and is going to be a maintenance issue.
4. In order to properly support Windows we could segment the Java library code for example such as Threads and IO where the windows libraries differ greatly from Standard C or POSIX. Then we could build the proper code on different platforms. This could also allow the current release to support Windows without affecting the more stable Linux/macOS platforms.
5. Sometimes writing some C glue code is just a better thing to do in certain situations even on top of the underlying library.
6. Would allow my `tensorflow` interface to be published in the interim. https://github.com/ekrich/stensorflow

The issues are probably not low hanging fruit otherwise @densh would have probably had a solution.

1. Structs on the stack.
2. Preprocessor macros. https://www.geeksforgeeks.org/interesting-facts-preprocessors-c/

I know we should carefully look at these changes and make sure we get the best solution we can based on current and future needs.

Perhaps others can help provide some guidance and feedback on this PR as well.

ekrich:
Splitting the projects with their own C code.
https://github.com/scala-native/scala-native/issues/1352

ekrich:
@sjrd I now just use the `classpath` to discover the `nativelib` and other libs that contain native code. It is not complete or polished but was wondering if this approach would be more acceptable?

ekrich:
### Highlights for this version
The plugin code now looks for jars on the classpath and if it finds `.c`, `.cpp` or `.S` code it looks for the organization and name in the manifest and uses the concatenation of those to provide a unique place to put the native code to be included in the build.

There is no need to pass the `nativelib` location to the build. The `nativelib` is discovered and provided to the build. Specifying the `nativelib` is deprecated and completely ignored. We know the `nativelib` coordinates so the discovery is rock solid. The `nativelib` build currently is special because it selects the `GC` and `optional` code.

If the project build has a main with native code, this requires `nativeCodeInclude := true`. This required a small amount of code in the plugin that should be easy to provide in tools other than `sbt`.

### Benefits

1. Plugin code is simplified
2. No need for third parties to discover `nativelib`
3. Allow libraries to use C macros and defines as well as code for pass by value and structs
4. More code is internal to the build tools so changing the internals is easier
5. Simple demos can include C code - after all Scala Native is easy to work with C
6. All third party libraries with native code are also discovered and handled by the build
7. Number 6 will allow Scala Native to support Tensorflow which motivated this PR.


### Other Info
This PR still requires that code is delivered in jar form to provide a unique directory for code for each project. This only applies to Scala Native internal libraries. When someone uses the plugin we expect all the libraries to come in jar form via the dependency manager.

If we decide to build locally without jars, we would need to figure out how manage the code and discover the `scalalib`. It is unclear also whether putting the relevant native code into the `javalib` jar will just work or we would have to enhance the build.

C code uses a flat namespace so it needs to be prefixed to not conflict with other code. This is not new but just mentioned as a reminder.

sjrd:
Using the organization and artifact name just to figure out a unique directory name to unpack things into is not the best solution. An obvious argument against it is that, as you observed, it requires jars. Requiring jars for arbitrary entries on the classpath goes against the spirit of classpaths, and makes life unnecessary difficult for users. We should be able to deal with native files anywhere on the classpath, including in directories.

A much easier scheme that supports all of that is simply to use *fresh names*, something that compilers do all the time. A dumb version is simply to take numbers corresponding to the position in the classpath. Since classpath are ordered collections, this is correct:
```scala
val extractPaths = for ((entry, index) <- config.classpath.zipWithIndex) yield {
  entry -> workdir.resolve(s"native-code-$index")
}
```
We can enhance this a little bit to include at least part of the name, which can be useful for debugging purposes:
```scala
val extractPaths = for ((entry, index) <- config.classpath.zipWithIndex) yield {
  val simpleName = entry.getName(entry.getNameCount() - 1).toString().stripSuffix(".jar")
  entry -> workdir.resolve(s"native-code-$name-$index")
}
```
This produces descriptive and unique names for all entries on the classpath, without having to look inside the manifests, and supports non-jar entries on the classpath.

docs/user/sbt.rst
```diff
@@ -55,25 +55,25 @@ Scala Native Version       Scala Versions
...
+0.1   ``nativeClangPP``             ``File``           Path to ``clang++`` command
+0.1   ``nativeCompileOptions``      ``Seq[String]``    Extra options passed to clang verbatim during compilation
+0.1   ``nativeLinkingOptions``      ``Seq[String]``    Extra options passed to clang verbatim during linking
+0.1   ``nativeMode``                ``String``         One of ``"debug"``, ``"release-fast"`` or ``"release-full"`` (2)
+0.2   ``nativeGC``                  ``String``         One of ``"none"``, ``"boehm"`` or ``"immix"`` (3)
+0.3.3 ``nativeLinkStubs``           ``Boolean``        Whether to link ``@stub`` definitions, or to ignore them
+0.4.0 ``nativeLTO``                 ``String``         One of ``"none"``, ``"full"`` or ``"thin"`` (4)
+0.4.0 ``nativeCheck``               ``Boolean``        Shall the linker check intermediate results for correctness?
+0.4.0 ``nativeDump``                ``Boolean``        Shall the linker dump intermediate results to disk?
+===== ============================= ================== =================================================================
```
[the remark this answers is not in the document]
ekrich: When `nativeLibraryDependencies` was in the list it was longer so everything got shifted. Now that it is gone, it looks like a change with no purpose. I can go ahead and revert that change.

ekrich:
@sjrd Thanks for the review. I wanted to do directories as well as jars but all I could come up with was a hash for the directory name. These are great ideas that will make this a much more general solution.

sjrd:
> Can the last entry in the path have spaces or anything that should be replaced?

If it's in the input path, then it's already a valid name in a path ;) So nothing needs to be replaced.

ekrich:
Thanks, just want to ask just to make sure. Still haven't had any coffee 😄

docs/user/sbt.rst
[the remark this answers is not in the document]
ekrich: Reverted.

build.sbt
```diff
@@ -20,8 +20,8 @@ def projectName(project: sbt.ResolvedProject): String = {
 
 // Provide consistent project name pattern.
 lazy val nameSettings: Seq[Setting[_]] = Seq(
-  normalizedName := projectName(thisProject.value),         // Maven <artifactId>
-  name := s"Scala Native ${projectName(thisProject.value)}" // Maven <name>
+  normalizedName := projectName(thisProject.value), // Maven <artifactId>
+  name := projectName(thisProject.value)            // Maven <name>
 )
```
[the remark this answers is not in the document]
ekrich: They were needed for the current manifest reading implementation for consistency for lookup and for the resulting directory name. They can be moved to a new PR given the new approach.

docs/contrib/build.rst
```diff
@@ -44,6 +44,17 @@ You can run only the test of interest by using one of the following commands:
     > tests/testOnly java.lang.StringSuite
     > tests/testOnly *StringSuite
 
+Scripted tests are used when you need to interact with the file system,
```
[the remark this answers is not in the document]
ekrich: Agreed.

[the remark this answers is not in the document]
ekrich: Resolved: https://github.com/scala-native/scala-native/pull/1821

ekrich:
How would you feel about adding a text file like `org_scala-native_nativelib.txt` to the native lib so we can identify it 100% whether it is in directory or jar form?

scripted-tests/run/native-code-include/src/main/resources/stdlib2.c
```diff
@@ -0,0 +1,3 @@
+#include <stdlib.h>
+
+long long scalanative_llabs(long long *in) { return llabs(*in); }
```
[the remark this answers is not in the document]
ekrich: This was a great idea, I was thinking way to complex.

build.sbt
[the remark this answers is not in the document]
ekrich: This change has been rolled back.

ekrich:
@sjrd Thanks for the great review. I think I have a pretty good solution. I'm sure there will be some small changes and such, especially in the `Discover` object.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -12,14 +12,94 @@ import scalanative.build.IO.RichPath
...
+    IO.existsInDir(path, srcPatterns) match {
+      case true  => Some(path)
+      case false => None
+    }
+
+  private def readJar(path: Path): Option[Path] =
+    IO.existsInJar(path, isNativeFile) match {
+      case true  => Some(path)
+      case false => None
+    }
```
sjrd, line 101: Thinking more about this, I think we should restrict the search for native code in a specific directory within the Dir/Jar. For example inside `META-INF/scala-native-native-libs/`.

Otherwise, codebases will never be able to put `.c`, `.cpp` or `.S` files in their `resources` without them being considered by the Scala Native linker as native libs that it should necessarily try to compile and link. That's not a great user experience.

tools/src/main/scala/scala/scalanative/build/IO.scala
```diff
@@ -89,40 +134,79 @@ private[scalanative] object IO {
     val stream       = Files.newInputStream(path)
     val digestStream = new DigestInputStream(stream, digest)
     val buf          = new Array[Byte](bufSize)
```
sjrd, line 136: The `val buf =` should also be inside the `try`.

tools/src/main/scala/scala/scalanative/build/IO.scala
```diff
@@ -89,40 +134,79 @@ private[scalanative] object IO {
...
+      digest.digest()
+    } finally {
+      digestStream.close()
+    }
+  }
+
+  /** Compute a SHA-1 hash of `files`. */
+  def sha1files(files: Seq[Path], bufSize: Int = 1024): Array[Byte] = {
+    val digest = MessageDigest.getInstance("SHA-1")
+    val buf    = new Array[Byte](bufSize)
```
sjrd, line 148: The `val buf =` should also be inside the `try`.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -12,14 +12,94 @@ import scalanative.build.IO.RichPath
...
+  /** GC variant used from SCALANATIVE_GC env var or default. */
+  def GC(): String =
+    getenv("SCALANATIVE_GC").getOrElse(build.GC.default.name)
+
+  private[build] def findNativeLibs(classpath: Seq[Path],
+                                    workdir: Path): Seq[NativeLib] = {
+    val nativeLibPaths = classpath.flatMap { path =>
+      if (isJar(path)) readJar(path)
+      else readDir(path)
+    }
```
sjrd, line 51: Is this early filter even necessary? Later, `unpackNativeCode` will anyway copy only relevant files, ending up being a no-op if there is no relevant file. The early filter has a cost, since we must walk through all the files.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -1,51 +1,89 @@
...
+
+  /**
+   * Copy project code from project `myproj` to `workdir/myproj`
+   * where is can be compiled and linked.
+   * Not used. Saved here in case non jar based is desired.
+   *
+   * @param nativelib    The native lib directory with native code.
+   * @param workdir   The working directory as described above.
+   * @return An Option with a Path to the `nativeCodeProject` work dir
+   *         or None if there is `NativeCodeProject is not configured.
```
sjrd, line 59: The `def` returns a `Path`, not an option. This comment seems obsolete.

sjrd: Moreover, this is *always* `nativelib.dest`, and it's the same in `unpackNativeJar`. It would be better to return `Unit` from both, and from `unpackNativeCode` as well, then directly use `nativelib.dest` at call site.

tools/src/main/scala/scala/scalanative/build/Logger.scala
```diff
@@ -21,7 +21,7 @@ trait Logger {
 
   /** Logs at the debug level that the command `cmd` will start running. */
   def running(cmd: Seq[String]): Unit = {
-    val msg = "running" + nl + cmd.mkString(nl + "\t")
+    val msg = "Running" + nl + cmd.mkString(nl + "\t")
```
sjrd, line 24: Spurious change?

tools/src/main/scala/scala/scalanative/build/ScalaNative.scala
```diff
@@ -11,7 +11,7 @@ import scalanative.codegen.CodeGen
 import scalanative.interflow.Interflow
 import scalanative.checker.Check
 
-/** Internal utilities to instrument Scala Native linker, otimizer and codegen. */
+/** Internal utilities to instrument Scala Native linker, optimizer and codegen. */
```
sjrd, line 14: Spurious change?

docs/contrib/build.rst
```diff
@@ -44,18 +44,6 @@ You can run only the test of interest by using one of the following commands:
     > tests/testOnly java.lang.StringSuite
     > tests/testOnly *StringSuite
 
-Scripted tests are used when you need to interact with the file system,
```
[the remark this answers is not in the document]
ekrich: I removed this as requested and created a new PR. https://github.com/scala-native/scala-native/commit/4d639629d7e1947975adc7374a2cf931d10e404f
It must have gotten removed on rebase. For future reference, should I leave the change in the old PR too?
What is the best way to resolve this now with GIT?

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -17,16 +17,15 @@ object Build {
...
-   *  val linkopts = Discover.linkingOptions()
-   *  val compopts = Discover.compileOptions()
-   *  val triple   = Discover.targetTriple(clang, workdir)
-   *  val outpath  = workdir.resolve("out")
+   *  val clang     = Discover.clang()
+   *  val clangpp   = Discover.clangpp()
+   *  val linkopts  = Discover.linkingOptions()
+   *  val compopts  = Discover.compileOptions()
+   *  val triple    = Discover.targetTriple(clang, workdir)
+   *  val outpath   = workdir.resolve("out")
```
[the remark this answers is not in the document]
ekrich: Yes, this was not intended.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -51,23 +49,28 @@ object Build {
...
+    val nativelibs   = Discover.findNativeLibs(config.classPath, workdir)
+    val nativelib    = Discover.findNativeLib(nativelibs)
+    val unpackedLibs = nativelibs.map(LLVM.unpackNativeCode(_))
 
-    val unpackedLib = LLVM.unpackNativelib(config.nativelib, config.workdir)
     val objectFiles = config.logger.time("Compiling to native code") {
       val nativelibConfig =
         config.withCompileOptions("-O2" +: config.compileOptions)
-      LLVM.compileNativelib(nativelibConfig, linked, unpackedLib)
+      // head needs to be the nativelib which is required
```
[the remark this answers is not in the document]
ekrich: Yes, this is an artifact of an old version.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -12,14 +12,94 @@ import scalanative.build.IO.RichPath
...
+    nativeLib match {
+      case Some(nl) => nl.dest
+      case None =>
+        throw new BuildException(s"Native Library not found: $nativeLibs")
+    }
+  }
+
+  private def isNativeFile(name: String): Boolean =
+    srcExtensions.map(name.endsWith(_)).exists(identity)
```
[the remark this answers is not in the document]
ekrich: Good suggestion. Not sure what I was thinking 😄

docs/contrib/build.rst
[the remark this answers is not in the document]
sjrd: The best way to resolve this is to rebase on top of the latest master.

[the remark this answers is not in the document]
ekrich: I was rebased on top of that commit. That is why I am a bit confused on what I did wrong.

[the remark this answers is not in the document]
ekrich: Fixed.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -48,11 +50,9 @@ object ScalaNativePluginInternal {
     nativeClangPP := interceptBuildException(Discover.clangpp().toFile),
     nativeCompileOptions := Discover.compileOptions(),
     nativeLinkingOptions := Discover.linkingOptions(),
-    nativeMode := Option(System.getenv.get("SCALANATIVE_MODE"))
-      .getOrElse(build.Mode.default.name),
+    nativeMode := Discover.mode(),
     nativeLinkStubs := false,
-    nativeGC := Option(System.getenv.get("SCALANATIVE_GC"))
-      .getOrElse(build.GC.default.name),
+    nativeGC := Discover.GC(),
```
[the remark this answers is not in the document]
ekrich: Yes, but I was trying to use the Boy Scout Rule. Also, I think having them all consistent and in the build tools rather than in the `sbt` plugin is a good change.

tools/src/main/scala/scala/scalanative/build/Logger.scala
ekrich: No, this was intended as all the other logging messages start with a capital.

tools/src/main/scala/scala/scalanative/build/ScalaNative.scala
ekrich: Spelling fix.

tools/src/main/scala/scala/scalanative/build/Discover.scala
ekrich: This filter reduces the search space down to only the classpath entries that need native processing. This native group then gets named/numbered according to your scheme. This list has to have at least one entry for the `nativelib` otherwise there is a problem. This much smaller list is then searched to find the `nativelib`.

We could also cache this if desired. The only downside would be if you decided to add a C file to a project that was not already in the list you would have to reload `sbt`. This is not too much different that if you add or change a dependency you must restart `sbt`. In this case the dependency would be changing from one that has no native code to one that does. I did not time or benchmark but I can't detect any slow down for searching.

tools/src/main/scala/scala/scalanative/build/IO.scala
ekrich: I see why, thank-you.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -12,14 +12,94 @@ import scalanative.build.IO.RichPath
...
+          path
+            .getName(path.getNameCount() - 1)
+            .toString()
+            .stripSuffix(jarExtension)
+        NativeLib(src = path,
+                  dest = workdir.resolve(s"native-code-$name-$index"))
+      }
+
+    if (extractPaths.isEmpty)
+      throw new BuildException(s"No Native Libraries found: $classpath")
```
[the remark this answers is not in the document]
ekrich: I think it makes sense as the message will print the classpath. There does have to be a minimum of one - the `nativelib` itself. The next call `findNativeLib` will just be searching an empty list and their will be no context in the `BuildException` as the list is empty.
https://github.com/scala-native/scala-native/pull/1637/files#diff-cec470ec94d2401cc4e1360e1d09c1ecR62-R63

tools/src/main/scala/scala/scalanative/build/IO.scala
ekrich: This would create a new buffer for each file rather than just reusing the buffer.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
ekrich: This documentation has been fixed. The returned value from the function is used downstream. I need to look at this closer.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -12,14 +12,94 @@ import scalanative.build.IO.RichPath
  */
 object Discover {
 
-  /** Compilation mode name that takes SCALANATIVE_MODE into account or default otherwise. */
+  /** List of source patterns used */
+  val jarExtension  = ".jar"
+  val srcExtensions = Seq(".c", ".cpp", ".S")
+  val srcPatterns   = srcExtensions.mkString("glob:**{", ",", "}")
```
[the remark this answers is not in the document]
ekrich: I moved the NativeLib code to a new file and made it package private since Discovery was getting messy.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -12,14 +12,94 @@ import scalanative.build.IO.RichPath
...
+      case None =>
+        throw new BuildException(s"Native Library not found: $nativeLibs")
+    }
+  }
+
+  private def isNativeFile(name: String): Boolean =
+    srcExtensions.map(name.endsWith(_)).exists(identity)
+
+  private def hasMarkerFileInJar(name: String): Boolean =
+    name.endsWith(nativeLibMarkerFile)
```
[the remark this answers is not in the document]
ekrich: Yes, you are correct. Fixed.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -12,14 +12,94 @@ import scalanative.build.IO.RichPath
...
+  val srcExtensions = Seq(".c", ".cpp", ".S")
+  val srcPatterns   = srcExtensions.mkString("glob:**{", ",", "}")
+
+  /** Original jar or dir path and generated dir path for native code */
+  case class NativeLib(src: Path, dest: Path)
+
+  /** To positively identify nativelib */
+  val nativeLibMarkerFile = "org_scala-native_nativelib.txt"
+
+  private lazy val dirMarkerFilePattern = "glob:**" + nativeLibMarkerFile
```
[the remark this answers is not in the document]
ekrich: It is a path matcher: https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-

I tried removing it but I guess it needs a glob or a regex to match the end of the path. I could write special code but I was reusing the `IO.existInDir` function.
```
[error] java.lang.IllegalArgumentException
[error] 	at sun.nio.fs.UnixFileSystem.getPathMatcher(UnixFileSystem.java:288)
[error] 	at scala.scalanative.build.IO$.existsInDir(IO.scala:67)
```

tools/src/main/scala/scala/scalanative/build/Discover.scala
ekrich: I am not super excited about that idea after reading online about how `META-INF` is used.

A couple of thoughts.

1. Our only customers are Scala library authors wishing to make their library work with Scala Native.
2. What are the chances that they are putting these files in the resource directory and for what purpose?

We could decide on a special directory inside resources but then we would have change the current code. Nothing we could come up with is 100% foolproof and would restrict library authors so I am not sure it is worth the effort for a hypothetical. 

That said, we could internally migrate code to `resources/native` and recommend that to library authors so that if we ever have a conflict we could change the code to use that directory only. It looks to me like it would take another PR to make the Scala Native `nativelib` compile if the code was moved into `resources/native`. Any other library with native code shouldn't matter. I have tested this with the scripted test and have documentation prepared if you agree.

tools/src/main/scala/scala/scalanative/build/Discover.scala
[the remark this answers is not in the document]
ekrich: We also have Mill and coursier and other tools that could use the Build API.

tools/src/main/scala/scala/scalanative/build/Discover.scala
ekrich: Just as a data point searching all of Github. `"src/resources/native" in:path`
You can also add the that `langauge:java` or Scala and see even less.

It is used but very little and no source code matches I could find. If we wanted to be super safe we could go with `scala-native` which is more specific but not too ugly.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
ekrich: I looked this closer and in `Build.build` the unpacking process returns the `NativeLib.dest` which is the generated destination path which is passed on later in the processing. From a functional programming point of view it is doing a side-effect and returning a value.

ekrich:
@sjrd I have addressed or at least commented for each items on your review. Please re-review at your convenience.

tools/src/main/scala/scala/scalanative/build/Discover.scala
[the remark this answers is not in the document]
ekrich: I looked at this again. It looks for the following for example which is across directory boundaries.
```
/Users/eric/workspace/scala-native/nativelib/target/scala-2.11/classes/org_scala-native_nativelib.txt
```

tools/src/main/scala/scala/scalanative/build/Discover.scala
ekrich: If we went with `resources/scala-native` we could do a followup PR to enforce it.

LeeTibbert:
I did not do a line by line review, but I just created a standalone project by following the
setup.rst instructions.  Wow!  This PR makes things one, and probably closer to two 
orders of magnitude easier.   I makes some of the issues I reported in issue #468 back
in December 2016 easier irrelevant or simple enough to live with.

Nice work, Eric!  

So far LGTM to me. I think there will be many happy end users, myself
included, once this is merged in the fullness of time.

ekrich:
@LeeTibbert I improved the documentation. Let me know if that helps, especially for users with little or no experience.

@sjrd I made a couple of other changes - one requested and some other improvements. We can followup with another PR to enforce the `resources/scala-native` directory if you like the idea.

docs/user/sbt.rst
```diff
@@ -168,28 +168,80 @@ package resolution system.
...
 In order to create standalone native projects with native code use the
-following procedure.
+following procedure. You can start with the basic Scala Native template.
+
+Add C/C++ code into `src/main/resources/scala-native`. The code can be put in
+subdirectories as desired inside the `scala-native` directory. As an example,
+create a file named `myapi.c` and put it into your `scala-native` directory
+as described above.
+
+.. code-block:: scala
```
LeeTibbert, line 198: Why is the `code-block` language `scala`.  I would have expected `c`.  I think 
Sphinx supports that out-of-the-box.   

Good that you provided the name "myapi.c" just above to help me orient on
language.  Sorry, I get confused easily.

docs/user/sbt.rst
```diff
@@ -168,28 +168,80 @@ package resolution system.
...
+code on the respective platforms.
+
+If the dependency contains native code, Scala Native will identify the
+library as a dependency that has native code and will unpack the library.
+Next, it will compile, link, and optimize any native code along with the
+Scala Native runtime and your application code. No additional information
+is needed in the build file other than the normal dependency so it is
+transparent to the library user.
+
+This feature can be used in combination with the feature to put native
```
LeeTibbert, line 243: The revised section seems much clearer, thank you.

The short & to the point working example will make it easier for people
with a 'see one, do one' learning style to follow along at home.

Slight glitch at the phrase "with the feature to". Thank you for leaving that
for me to find ;-)

ekrich: Yes, that is not great, do you have a suggested rewording?

docs/user/sbt.rst
ekrich: Yes, I changed that to C and the `long long` changed to be highlighted in red, like the Scala types. Nice catch.

docs/user/sbt.rst
ekrich: How about this? *This feature can be used in combination with the feature above that
allows native code in your application.*

ekrich: Ok, I pushed that as a change.

docs/user/sbt.rst
```diff
@@ -240,8 +240,8 @@ Scala Native runtime and your application code. No additional information
 is needed in the build file other than the normal dependency so it is
 transparent to the library user.
 
-This feature can be used in combination with the feature to put native
-code in you application as described above.
+This feature can be used in combination with the feature above that
+allows native code in your application.
```
LeeTibbert, line 244: Reads much better to my eyes, thank you.

ekrich:
@sjrd Rebased, again. Hopefully it is close.

tools/src/main/scala/scala/scalanative/build/Discover.scala
ekrich: We only want `unpackNativeCode` to work on libraries with native code. This blindly unpacks jars and copies directories and hashes the jars and directory contents. Later on in the process the code is found to be compiled. On a previous commit I combined the search into one operation using one glob for each file extension.

tools/src/main/scala/scala/scalanative/build/Discover.scala
sjrd: > What are the chances that they are putting these files in the resource directory and for what purpose?

It's not a question of "chances". If there is a chance, we need to take it into account. In language design, you have to account for *all* possible cases, otherwise you will block *some*one, somewhere.

> If we wanted to be super safe we could go with scala-native which is more specific but not too ugly.

`scala-native` sounds good.

Issue #1654, opened 2019-07-18
Commix doesn't link on MacOs with clang 10.0.1


lolgab:
Setting `nativeGC := "commix"` on a MacOs Mojave machine with provided clang (10.0.1) gives an error about the use of deprecated members. Probably -Werror is on by default.
```
[error] /Users/lorenzo/scala/scala-native-loop/examples/server/target/scala-2.11/native/lib/gc/commix/Settings.c:7:10: fatal error: 'sys/sysinfo.h' file not found
[error] #include <sys/sysinfo.h>
[error]          ^~~~~~~~~~~~~~~
[error] 1 error generated.
[error] /Users/lorenzo/scala/scala-native-loop/examples/server/target/scala-2.11/native/lib/gc/commix/Phase.c:8:5: warning: 'sem_init' is deprecated [-Wdeprecated-declarations]
[error]     sem_init(&heap->gcThreads.startWorkers, 0, 0);
[error]     ^
[error] /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/sys/semaphore.h:55:42: note: 'sem_init' has been explicitly marked deprecated here
[error] int sem_init(sem_t *, int, unsigned int) __deprecated;
[error]                                          ^
[error] /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/sys/cdefs.h:176:40: note: expanded from macro '__deprecated'
[error] #define __deprecated    __attribute__((deprecated))
[error]                                        ^
[error] /Users/lorenzo/scala/scala-native-loop/examples/server/target/scala-2.11/native/lib/gc/commix/Phase.c:9:5: warning: 'sem_init' is deprecated [-Wdeprecated-declarations]
[error]     sem_init(&heap->gcThreads.startMaster, 0, 0);
[error]     ^
[error] /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/sys/semaphore.h:55:42: note: 'sem_init' has been explicitly marked deprecated here
[error] int sem_init(sem_t *, int, unsigned int) __deprecated;
[error]                                          ^
[error] /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/sys/cdefs.h:176:40: note: expanded from macro '__deprecated'
[error] #define __deprecated    __attribute__((deprecated))
```

[...]

PR #1658, opened 2019-07-20
Commix: named semaphores

valdisxp1:
Using named semaphores only for comparability with MacOs see #1654 .
Needs changes from #1657 to compile on Mac.

- [x] tested on Linux
- [ ] tested on MacOs

### DIFF
--- a/nativelib/src/main/resources/gc/commix/Phase.c
+++ b/nativelib/src/main/resources/gc/commix/Phase.c
@@ -23,8 +23,8 @@ void Phase_Init(Heap *heap, uint32_t initialBlockCount) {
         sem_open(startMasterName, O_CREAT | O_EXCL, 0644, 0);
     // clean up when process closes
     // also prevents any other process from `sem_open`ing it
-    sem_unlink("startWorkers");
-    sem_unlink("startMaster");
+    sem_unlink(startWorkersName);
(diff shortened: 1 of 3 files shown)


lolgab:
Tried on my Mac machine and it links nicely now.

nativelib/src/main/resources/gc/commix/Phase.c
```diff
@@ -3,10 +3,28 @@
...
+    // only reason for using named semaphores here is for compatibility with
+    // MacOs we do not share them across processes
+    heap->gcThreads.startWorkers =
+        sem_open(startWorkersName, O_CREAT | O_EXCL, 0644, 0);
+    heap->gcThreads.startMaster =
+        sem_open(startMasterName, O_CREAT | O_EXCL, 0644, 0);
+    // clean up when process closes
+    // also prevents any other process from `sem_open`ing it
+    sem_unlink("startWorkers");
+    sem_unlink("startMaster");
```
densh: Shouldn't it use the same semaphore name here as the one given to `sem_open` ?

valdisxp1: Nice catch! You are absolutely right.

densh:
LGTM, thanks @valdisxp1 ! 👍

PR #1660, opened 2019-07-21
Fix #1631: Add func ptr conversions

### LINKED ISSUE #1631: Cast CFuncPtr to Ptr[Byte] (and reverse)
For certain unsafe, void-pointer slinging C API's, it can be useful to convert a CFunctPtr to a Ptr[Byte] (or vice versa).  For example, libcurl in particular needs this idiom to set callbacks for various events.

Currently, I'm implementing this in user-space code like so:

```
  def func_to_ptr(f:Object):Ptr[Byte] = {
    Boxes.boxToPtr[Byte](Boxes.unboxToCFuncRawPtr(f))
  }
```

Which, if cleaned up a bit, would be great to include in the `unsafe` package.  And for the sake of completeness/generality, we should be able to implement the reverse cast as well.

rwhaling:
Small PR to add conversions between Ptr and CFuncPtr, useful for stuffing functions into random struct data fields.

Adds a method `toCFuncPtr` to the `Ptr` class, as well as a `fromCFuncPtr` method to the `Ptr` companion object; not 100% sure either is quite right.

For toCFuncPtr, I'm a little sketchy on using the asInstanceOf to convert a CFuncRawPtr to any CFuncPtr type; the relationship between CFuncRawPtr and the CFuncPtr trait itself is a little weird, and I don't really understand how the runtime creates these.  But it seems like it's safe in practice?

for `fromCFuncPtr`, I originally wanted to add this conversion as a method on the CFuncPtr trait instead, but then CFuncPtr would no longer be a SAM.  Also not sure how performant this is, or if there's a better way to do it.


nativelib/src/main/scala/scala/scalanative/unsafe/Ptr.scala
```diff
@@ -61,4 +65,7 @@ object Ptr {
 
   @alwaysinline implicit def ptrToCStruct[T <: CStruct](ptr: Ptr[T])(
       implicit tag: Tag[T]): T = !ptr
+
+  @alwaysinline def fromCFuncPtr[T](f: CFuncPtr) =
```
densh: Shouldn't this one return `Ptr[T]` rather than `Ptr[Byte]` ?

rwhaling: good catch!

rwhaling: @densh - does this look good to you now?

densh: Would be good to explicitly declare the return type as well.

PR #1712, opened 2019-10-23
Upgrade to sbt 1.3.7 and drop support for sbt 0.13.x.

ekrich:
The goal here is to simplify things for `0.4.x`. So much time has elapsed we are getting behind and missing improvements. In my opinion, `0.13.x` is obsolete and dropping it will not affect users moving forward. This will also reduce our CI load and making working on the project much easier.


lolgab:
Just out of curiosity, why in travis are we testing on `2.11.8` and not on `2.11.12` ?

ekrich:
That is what is in Travis. I'll go ahead and change that. In `build,sbt` we do have this as well because Scala Native should work in slightly older versions as well if I understand correctly.

```Scala
val libScalaVersion       = "2.11.12"
val libCrossScalaVersions = Seq("2.11.8", "2.11.11", libScalaVersion)
```

ekrich:
Build has been tested with both `jdk8` and `jdk11`. The plugin code has been left in the `sbt 1.0` directory because it is unclear why there is more code for `1.0.x` as there was for `0.13.x` and whether that code should be merged back in with the plugin code. Some other cross plugin code has been left in the `build.sbt` file in case it is needed for future `2.13` support. All the `0.13.x` specific code and comments should be removed.

Ready for review.

build.sbt
```diff
@@ -232,11 +209,10 @@ lazy val noPublishSettings = Seq(
 lazy val toolSettings =
   baseSettings ++
     Seq(
-      crossSbtVersions := List(sbt13Version, sbt10Version),
+      crossSbtVersions := List(sbt10Version),
```
eatkins: You might want to set this to an older version of sbt unless you are using sbt 1.3.x features. Otherwise you might publish a version of the plugin that works on sbt 1.3.x but not earlier versions. See https://github.com/sbt/sbt/issues/5154.

ekrich: Thanks for the review. What version of sbt `1.x` do you recommend?

sideeffffect: `1.2.8` according to https://github.com/sbt/sbt/issues/5154#issuecomment-538666595

eatkins:
I would actually probably target 1.0.4 or so if that compiles. Because backward compatibility of sbt is guaranteed in the 1.x series, the lowest version that compiles is ideal.

ekrich:
@eatkins I went ahead and used version `1.0.4` and tested a Scala Native program using `1.0.4`, `1.1.6`, `1.2.8` and `1.3.3` and they all worked. These are the latest point releases in each series. The build still uses `1.3.3` but the plugin should be published using the `1.0.4` version if I understand correctly.

eatkins:
@ekrich that sounds right.

ekrich:
The sbt `Process` related code has been moved back into the plugin area. I also upgraded to the latest sbt release although the releases are pretty frequent.

@errikos @sjrd Please review at your convenience.

errikos:
Since we are migrating SBT version to 1.x, maybe it would also be nice to migrate to the / syntax. I have a commit doing that in my own fork: https://github.com/errikos/scala-native/commit/c05d5c7e977bc2f2f29c9c2e1bbf21525fed2bf0

Maybe you could add that to the PR as well? Thanks!

ekrich:
Sure, I'll add that later this evening. Thanks for the commit for reference.

build.sbt
```diff
@@ -325,9 +301,9 @@ lazy val nscplugin =
       scalaVersion := libScalaVersion,
       crossScalaVersions := libCrossScalaVersions,
       crossVersion := CrossVersion.full,
-      unmanagedSourceDirectories in Compile ++= Seq(
-        (scalaSource in (nir, Compile)).value,
-        (scalaSource in (util, Compile)).value
+      Compile / unmanagedSourceDirectories ++= Seq(
+        (nir / Compile / scalaSource).value,
+        (util / Compile / scalaSource).value,
```
errikos: I accidentally introduced a trailing comma here.
```suggestion
        (util / Compile / scalaSource).value
```

ekrich:
Ok, should be good now.

.travis.yml
```diff
@@ -14,20 +14,13 @@ services:
...
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_GC=commix SBT_VERSION=1.2.6 TEST_COMMAND="sbtScalaNative/scripted"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=boehm SBT_VERSION=0.13.17 TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=boehm SBT_VERSION=1.2.6 TEST_COMMAND="sbtScalaNative/scripted"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=immix SBT_VERSION=0.13.17 TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=immix SBT_VERSION=1.2.6 TEST_COMMAND="sbtScalaNative/scripted"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=commix SBT_VERSION=0.13.17 TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=commix SBT_VERSION=1.2.6 TEST_COMMAND="sbtScalaNative/scripted"
+    - MODE=test-tools SBT_VERSION=1.3.7
```
sjrd: `SBT_VERSION` should match the value of `sbt10Version` in `build.sbt`, not the `sbt.version` of `project/build.properties`. So this should be `1.0.4`.

Same in all the lines below.

build.sbt
```diff
@@ -226,17 +203,16 @@ lazy val noPublishSettings = Seq(
...
-      crossSbtVersions := List(sbt13Version, sbt10Version),
+      crossSbtVersions := List(sbt10Version),
       scalaVersion := {
-        (sbtBinaryVersion in pluginCrossBuild).value match {
-          case "0.13" => sbt13ScalaVersion
-          case _      => sbt10ScalaVersion
+        (pluginCrossBuild / sbtBinaryVersion).value match {
+          case _ => sbt10ScalaVersion
         }
       },
```
sjrd: You can replace all that with
```scala
scalaVersion := sbt10ScalaVersion
```

ci-docker/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN sed -i -e "s/deb http/deb [arch=$HOST_ARCHITECTURE] http/g" /etc/apt/host-so
 RUN apt-get update && apt-get install -y openjdk-8-jdk-headless:$HOST_ARCHITECTURE
 ENV PATH "/usr/lib/jvm/java-8-openjdk-$HOST_ARCHITECTURE/bin:$PATH"
 
-ENV SBT_LAUNCHER_VERSION 0.13.7
+ENV SBT_LAUNCHER_VERSION 1.3.3
```
sjrd:
```suggestion
ENV SBT_LAUNCHER_VERSION 1.3.7
```

build.sbt
ekrich: Yes, we don't need that extra code.

ci-docker/Dockerfile
ekrich: Good catch - forgot this one.

.travis.yml
ekrich: Thanks for the review.

So we run the following directly in Travis outside of Docker:
`sbt "-no-colors" "-J-Xmx3G" "^^ $SBT_VERSION" nscplugin/publishLocal sbtScalaNative/publishLocal test-tools`
So I guess the idea is we test the tools with the minimum `sbt` and launcher version.

Then we test the runtimes using the latest `sbt` so hopefully it is better and faster?

ekrich: I looked in the CI logs and it is using the `sbt` launcher version `1.3.7` but then using `1.0.4` later in the build `[info] Setting` `` `sbtVersion in pluginCrossBuild` `` `to 1.0.4`. Also, it seems that some deprecations may not be able to be removed until the `sbt` minimum version is updated based on the warnings. Hopefully, if someone if working on the tools then some of this code can be updated to remove deprecations.

I guess too if someone is using an older launcher and it fails then it is ok for them too update the launcher. They can update `sbt` in their build on a different schedule.

ekrich:
@sjrd This should be ready for another review now.

.travis.yml
sjrd: It should be set to 1.0.4 for *all* builds. This enjoyment variable only affects the version of sbt that the sbt plugin is built for. The build itself will always use the version of sbt that is defined in `build.properties`.

In fact, since we now only support sbt 1.x for the sbt plugin, this environment variable is redundant. It would be better to get rid of it entirely, and get rid of the `^^ ...` commands. All of that is redundant now.

ekrich: @sjrd The sbt slash syntax was not offered until 1.1.0 and now the scripted tests are running in 1.0.4 so this doesn't work. The idea was to support the latest in each series, 1.0.4, 1.1.6, 1.2.8 and 1.3.x. Maybe we should bump the minimum version to 1.1.6? Otherwise we need change the 3 scripted tests to use the non-slash syntax. I tested the 1.1.6 idea and that works.

ekrich: I pushed the `1.1.6` change but can go the other way if desired.

sjrd:
Thanks @ekrich for your patience :)

PR #1714, opened 2019-10-31
Fix #1709: Upgrade to scalafmt 2.4.2

### LINKED ISSUE #1709: Upgrade to newer scalafmt
When upgrading to version `2.1.0` for example, a default changed. If we wish to retain the old format for some reason we should use the following:
```
align.openParenCallSite = true
```
I personally like the newer format. We mostly use defaults so I think that is in line with our current formatting style.

Reference:
https://scalameta.org/scalafmt/docs/configuration.html#alignopenparencallsite

ekrich:
I have done an upgrade and tried to minimize changes via lots of trial and error.

Here is a list of changes that I couldn't avoid but it is upgrading from `1.5.1` to`2.2.1`.

1. Added `// format: off` in niocharset to avoid some changes.
2. Trailing comments are now aligned.
3. Wrapped return types are now are 4 spaces instead of 2. This is using the default but trying to alter these settings makes more of a diff.
4. Curly brace expansion when they are inlined, `if`, `try` and inside `if({})`. The first 2 are ok but I really dislike the last.
Note: `scalanative.unsafe.Tag` and generated files have the most changes.

If anyone wants to see if they can improve on this, just grab the `.scalafmt.conf` file and give it a try with a current branch.


ekrich:
Note: The three `literal` settings should probably be removed as they standardize hex and numbers to the most readable format.  If people think the defaults are not the most readable then there are options that can at least make them all standard.

Other settings could be removed or adjusted to accept a more `scalafmt` standard format but that would cause a bigger diff and we have allot of outstanding PRs which would need to be rebased.

ekrich:
I went ahead and pushed the change to accept the default number constants. They are designed for readability and this way they will be standard going forward. I was waiting for the `2.4.x` version to be released but it errors on our code base. See https://github.com/scalameta/scalafmt/issues/1720

There is one other error which is really like a warning that we have had as long as I can remember.
```
[warn] Error in /Users/eric/workspace/scala-native/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala: 
org.scalafmt.Error$SearchStateExploded: Search state exploded around line 1271
```
@sjrd Maybe this is a good place to stop and then revisit the `2.4.x` version as another PR. WDYT?

unit-tests/src/test/scala/java/util/regex/PatternSuite.scala
```diff
@@ -586,13 +584,11 @@ object PatternSuite extends tests.Suite {
...
-    assertThrowsAnd[PatternSyntaxException](Pattern.compile(pattern))(
-      e => {
-        e.getDescription == description &&
-        e.getPattern == pattern
-        e.getIndex == index
-      }
-    )
+    assertThrowsAnd[PatternSyntaxException](Pattern.compile(pattern))(e => {
+      e.getDescription == description &&
+      e.getPattern == pattern
```
sjrd: There's a missing `&&` here.

unit-tests/src/test/scala/scala/scalanative/regex/PatternSuite.scala
```diff
@@ -403,13 +401,11 @@ object PatternSuite extends tests.Suite {
...
-    assertThrowsAnd[PatternSyntaxException](Pattern.compile(pattern))(
-      e => {
-        e.getDescription == description &&
-        e.getPattern == pattern
-        e.getIndex == index
-      }
-    )
+    assertThrowsAnd[PatternSyntaxException](Pattern.compile(pattern))(e => {
+      e.getDescription == description &&
+      e.getPattern == pattern
```
sjrd: Missing `&&` here too.

javalib/src/main/scala/java/lang/Double.scala
```diff
@@ -211,7 +211,7 @@ object Double {
     }
 
   @inline def doubleToLongBits(value: scala.Double): scala.Long =
-    if (value != value) 0x7ff8000000000000L
+    if (value != value) 0x7FF8000000000000L
```
sjrd: Is this configurable? Can we ask scalafmt to preserve lowercase for hexadecimal digits?

unit-tests/src/test/scala/scala/scalanative/regex/PatternSuite.scala
ekrich: I suppose that could be missing but it wasn't in the original either. @LeeTibbert Perhaps this is some code you are familiar with?

javalib/src/main/scala/java/lang/Double.scala
ekrich: These are the defaults for just the character at the end as far as I know. It looks like they just uppercase or lowercase the whole number rather than just the last character.
```
literals.long = Upper
literals.float = Lower
literals.double = Lower
```
I also looked checked and if the hexidecimal number is not a `long` it leaves those alone so in `java.lang.Character` some hexadecimals are lowercase and some are uppercase.

I think I should just roll back that change and report it to `scalafmt` until a change can be made. 

The last character casing above makes sense but not changing the hex portion unless they are changed everywhere. Also, if you have E Notation like `10E-3f` it changes to `10e-3f` and if you have `10e-3L` it would change to `10E-3L`. This is clearly weird.

ekrich: The whole idea is you have common decent standards and then forget about it as the tool does it for you.

ekrich:
I agree we should wait to see about the `0.3.x` branch but honestly I am the only one who has asked about this version at all in over 6 months or more so I doubt there is any interest. We could also backport this PR if needed.

ekrich:
@sjrd What is the proper way to reverse the last change. Should I reintroduce the previous change that was removed or remove the last commit and force push?

sjrd:
In the Scala.js repo, it would be to remove the last commit and force push. Here it doesn't really matter because the policy has always been to squash and merge PRs at the end of the day.

ekrich:
I would do the former but then sometimes things go missing in GitHub so I wanted to ask.

ekrich:
@sjrd I removed the last commit. I will add an issue to `scalafmt` about the literal numbers. I will ping again when I get something to look at but I want to at least get to the latest version, `2.4.2` as of now.

ekrich:
2.4.2 fixed the `if ({ })` wrapping problem so I'm pushing that with the previous normal formatting.
This change, https://github.com/scalameta/scalafmt/pull/1660, makes quite a few changes to the current code base. Any strong opinions of this?

unit-tests/src/test/scala/java/util/regex/PatternSuite.scala
ekrich: Thanks @LeeTibbert. See https://github.com/scala-native/scala-native/issues/1740

unit-tests/src/test/scala/scala/scalanative/regex/PatternSuite.scala
ekrich: Thanks @LeeTibbert. See https://github.com/scala-native/scala-native/issues/1740

ekrich:
Added the following issues https://github.com/scalameta/scalafmt/issues/1845 https://github.com/scalameta/scalafmt/issues/1844

@sjrd There are some more changes but if they are ok this could get us to the finish line if we have decided not to backport to `0.3.9`.

sjrd:
This looks mostly good, now. There's only changes to `//` with vertical alignment across different levels, which are very confusing. One example of what I'm talking about is
https://github.com/scala-native/scala-native/pull/1714/files#diff-f10107669801b1d4ad2e437090bc7a1bR1549-R1553
Can we disable vertical alignment of `//` comments? If yes, we should do it, otherwise we'll merge as is.

ekrich:
Let me look into that.

Also there is a `2.5.0-RC1` that has the hexadecimal formatting that I tested last night. Would you rather we wait for those changes or just do one big one? 

It also formats like this by default, does that look ok?
```Scala
-  @inline def anyHash(x: Object): Int = x match {
-    case null                => 0
-    case x: java.lang.Long   => longHash(x.longValue)
-    case x: java.lang.Double => doubleHash(x.doubleValue)
-    case x: java.lang.Float  => floatHash(x.floatValue)
-    case _                   => x.hashCode
-  }
+  @inline def anyHash(x: Object): Int =
+    x match {
+      case null                => 0
+      case x: java.lang.Long   => longHash(x.longValue)
+      case x: java.lang.Double => doubleHash(x.doubleValue)
+      case x: java.lang.Float  => floatHash(x.floatValue)
+      case _                   => x.hashCode
+    }
```

sjrd:
Hum, the existing format (the `-` in your diff) is pretty standard formatting in Scala. If we can keep it, that would be better.

Let's do a separate one for the hexadecimal formatting. It probably won't touch the lines touched by this PR, and will likely interfere less with in-flight PRs.

ekrich:
@sjrd Unfortunately, it doesn't seem like there is direct control over the trailing comment alignment. 

- Where infixes like `|`, `&&` wrapped funny I hand edited so the code didn't wrap inside a function call.
- Trailing comments that are too long wrap the code they comment so in some cases it makes sense to put them above the code they comment - in some cases they look good.
- The majority of the changes are a wrap after `=` with a `match` on the next line.
- `implicit`s are wrapped differently now. I don't know the preferred style but I'm hoping the `scalafmt` team has the data.

One really good thing is that `scalafmt` is getting much better. If we want the hexadecimal formatting to be default, we can just remove those lines in another commit. That was fixed practically as fast as I could add the enhancement requests. The more default our style, the less maintenance that needs to be done. Tweaking styles IMHO defeats the purpose of a common readable style. I did tweak some to avoid a bunch more diffs early on but even those could be not desired.

@kitbellew and others have been doing a great job. They even have a Scala.js default style.

sjrd:
Ouch, I really really don't like the changes brought by 2.5.0-RC2. I mean, none of it looks remotely good and some seems utterly incorrect.

Please revert the last commit, stay on 2.4.2 which I had reviewed and was basically happy with. Then make sure to rebase on the latest master so that the new formatting can be applied to changes that have happened on master since, and we can merge.

Let's leave discussion for 2.5.0 to a different PR, please, or it will delay a lot more the already accepted changes.

ekrich:
@sjrd Agreed, this is very difficult given the rapid change occurring in `scalafmt`.

> Hum, the existing format (the - in your diff) is pretty standard formatting in Scala. If we can keep it, that would be better.

This change alone, I liked it too, was really massive. I'm hoping we can work with the `scalafmt` team on any undesirable changes.

PR #1720, opened 2020-01-10
Fix #1684: Fix def commix to point to Commix instead of Immix

### LINKED ISSUE #1684: Commix function points to Immix
I don't know if it is used at all, but this should be a typo. 
https://github.com/scala-native/scala-native/blob/b650b901f2efc765315b16e618176f96ce52385f/tools/src/main/scala/scala/scalanative/build/GC.scala#L37

lolgab:


sjrd:
I disagree with the proposed changes here. Those methods are public entry points for an application that would directly build a `build.Config` object, using a typo-safe reference to the GC they want.

I do see that one of them is erroneous (`def commix = Immix`), which should be fixed. To make this more robust for the future, we should use those public entry points also in `def apply()`.

lolgab:
Thank you for the review @sjrd.
I'm ready for another review round.

PR #1724, opened 2020-01-31
Fix #1723: Generate lowered.hnir file outside of project root directory

### LINKED ISSUE #1723: Generate lowered.hnir file outside of project root directory
At the moment Scala Native is generating a file called `lowered.hnir` in project's root directory.
It is annoying since you need to gitignore this file, that could be generated in the `target` directory (which is usually already gitignored).

ekrich:


ekrich:
Ready for review.

PR #1727, opened 2020-02-18
Fix makeTestMain with classes in the empty package

lolgab:


unit-tests/src/test/scala/RootPackageTest.scala
```diff
@@ -0,0 +1,3 @@
+object RootPackageTest extends tests.Suite {
```
sjrd:
```suggestion
object EmptyPackageTest extends tests.Suite {
```
This whole issue is actually cause by the JVM *empty* package. It's not the root package. If it were the root package, the previous implementation would be correct. Let's use the correct terminology.

sjrd:
Thanks for the contribution. I'm happy to take this in after the above comment has been addressed, even though I expect all of that will become a non-issue once we're through with #1728 and then basing the testing interface on top of `Reflect`, like Scala.js does.

unit-tests/src/test/scala/RootPackageTest.scala
lolgab: Done ✅

lolgab:
Thank you very much for the review :)
Didn't know there was already work being done in that direction, otherwise I would have not taken this :D

PR #1728, opened 2020-02-19
Fix #1279: Enable reflective instantiation via static initializers

### LINKED ISSUE #1279: Add support for portable-scala/reflect
portable-scala/reflect offers a minimalistic set of reflection primitives:

- reflectively instantiate classes that were explicitly annotated to be reachable through reflection
- reflectively instantiate modules that were explicitly annotated to be reachable through reflection

The API is meant to have the same semantics portable across all three platforms (JVM/JS/Native).

errikos:
This PR enables users to reflectively instantiate objects, by making use of static class initialisers (not to be confused with JVM class initialisers). Related issue is https://github.com/scala-native/scala-native/issues/1279.

To achieve the above:
- The NIR member signatures are enhanced with an extra attribute, called `clinit`, which is mangled with `I`. Static initialiser functions take no arguments and return `Unit`.
- For the static initialiser functions to be called, they must be made reachable. For this to work, we need to iterate over all program classes (each of them in a separate `.nir` file) and check whether they contain a `clinit` function. To avoid deserialising the whole file for each class, I added a single byte at the beginning of each file, to indicate whether the type it contains has a `clinit` function. If that byte is set, then the `clinit` function of the type in that file is added to the reachable set.
- When generating the code for the `main` method (native), we have to call the discovered `clinit` fuctions. We do this just before calling the main Scala method.

On top of that, we implement the same `Reflect` API and `@EnableReflectiveInstantiation`-based mechanism that Scala.js has.

Objectives list:
- [x] Add support for static initialisers to the NIR
- [x] Add the `@EnableReflectiveInstantiation` annotation
- [x] Implement the desired annotation behaviour for modules
- [x] Implement the desired annotation behaviour for normal classes

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
@@ -12,7 +12,7 @@ final class BinaryDeserializer(buffer: ByteBuffer) {
   private val header: Map[Global, Int] = {
     buffer.position(0)
 
-    val hasClInit = getBool
+    val hasEntryPoint = getBool
 
     val magic    = getInt
     val compat   = getInt
--- a/nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
@@ -12,6 +12,8 @@ final class BinaryDeserializer(buffer: ByteBuffer) {
   private val header: Map[Global, Int] = {
     buffer.position(0)
 
+    val hasClInit = getBool
+
     val magic    = getInt
     val compat   = getInt
     val revision = getInt
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
@@ -336,6 +336,14 @@ trait NirGenStat { self: NirGenPhase =>
 
     def genRegisterReflectiveInstantiationForNormalClass(
         cd: ClassDef): Seq[Inst] = {
+      val ctors =
+        if (curClassSym.isAbstractClass) Nil
+        else
+          curClassSym.info
+            .member(nme.CONSTRUCTOR)
+            .alternatives
+            .filter(_.isPublic)
+
       unsupported(cd)
     }
 
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
@@ -261,7 +261,7 @@ trait NirGenStat { self: NirGenPhase =>
           nir.Type.Function(Seq(Type.Ref(predefGlobal), Type.Ref(objectGlobal)),
                             Type.Unit),
           methodVal,
-          Seq(moduleVal, Val.String("Static initializer called")),
+          Seq(moduleVal, Val.String(s"${cd.name}: static initializer called")),
           unwind(curFresh)
         )
         exprBuf.ret(Val.Unit)
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
@@ -402,7 +402,7 @@ trait NirGenStat { self: NirGenPhase =>
       def genLazyClassInstantiationMethod(exprBuf: ExprBuffer,
                                           ctors: Seq[global.Symbol]): Val = {
         val applyMethodSig =
-          Sig.Method("apply", Seq(Type.Array(jlObjectType), jlObjectType))
+          Sig.Method("apply", Seq(jlObjectType, jlObjectType))
 
         val tuple2Name = Global.Top("scala.Tuple2")
         val tuple2Type = Type.Ref(tuple2Name)
@@ -452,7 +452,8 @@ trait NirGenStat { self: NirGenPhase =>
               nir.Type.Function(Seq(Type.Ref(reflInstBuffer.name),
                                     Type.Array(jlObjectType)),
                                 jlObjectType),
-              body)
+              body
+            )
           }
 
           // Generate the class instantiator constructor.
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
@@ -440,10 +440,19 @@ trait NirGenStat { self: NirGenPhase =>
               // Extract and cast arguments to proper types.
               val argsVals =
                 (for ((arg, argIdx) <- ctorSig.args.tail.zipWithIndex) yield {
-                  exprBuf.arrayload(Type.box.getOrElse(arg, arg),
-                                    argsArg,
-                                    Val.Int(argIdx),
-                                    unwind(curFresh))
+                  val elem =
+                    exprBuf.arrayload(Type.box.getOrElse(arg, arg),
+                                      argsArg,
+                                      Val.Int(argIdx),
+                                      unwind(curFresh))
+                  // If the actual argument type can be boxed (i.e. is a primitive
+                  // type), then we need to unbox it before passing it to C.
+                  Type.box.get(arg) match {
+                    case Some(bt) =>
+                      exprBuf.unbox(bt, elem, unwind(curFresh))
+                    case None =>
+                      elem
+                  }
                 })
 
               // Allocate a new instance and call C.
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
@@ -282,7 +282,7 @@ trait NirGenStat { self: NirGenPhase =>
 
     // Generate the constructor for the class instantiator class,
     // which is expected to extend one of scala.runtime.AbstractFunctionX.
-    def genReflectiveInstantiationConstructor(
+    private def genReflectiveInstantiationConstructor(
         reflInstBuffer: ReflectiveInstantiationBuffer,
         superClass: Global): Unit = {
       withFreshExprBuffer { exprBuf =>
@@ -310,6 +310,21 @@ trait NirGenStat { self: NirGenPhase =>
           body
         )
       }
+    }
+
+    // Allocate and construct an object, using the provided ExprBuffer.
+    private def allocAndConstruct(exprBuf: ExprBuffer,
+                                  name: Global,
+                                  argTypes: Seq[nir.Type],
+                                  args: Seq[Val]): Val = {
+      val alloc = exprBuf.classalloc(name, unwind(curFresh))
+      exprBuf.call(
+        Type.Function(Type.Ref(name) +: argTypes, Type.Unit),
+        Val.Global(name.member(Sig.Ctor(argTypes)), Type.Ptr),
+        alloc +: args,
+        unwind(curFresh)
+      )
+      alloc
     }
 
     def genRegisterReflectiveInstantiationForModuleClass(
@@ -361,14 +376,7 @@ trait NirGenStat { self: NirGenPhase =>
           Seq(Global.Top("scala.Serializable")))
 
         // Allocate and return an instance of the generated class.
-        val alloc = exprBuf.classalloc(reflInstBuffer.name, unwind(curFresh))
-        exprBuf.call(
-          Type.Function(Seq(Type.Ref(reflInstBuffer.name)), Type.Unit),
-          Val.Global(reflInstBuffer.name.member(Sig.Ctor(Seq())), Type.Ptr),
-          Seq(alloc),
-          unwind(curFresh)
-        )
-        alloc
+        allocAndConstruct(exprBuf, reflInstBuffer.name, Seq(), Seq())
       }
 
       withFreshExprBuffer { exprBuf =>
@@ -399,13 +407,21 @@ trait NirGenStat { self: NirGenPhase =>
       val srAbstractFunction1Name =
         Global.Top("scala.runtime.AbstractFunction1")
 
+      val tuple2Name = Global.Top("scala.Tuple2")
+      val tuple2Type = Type.Ref(tuple2Name)
+
+      // Create a new Tuple2 and initialise it with the provided values.
+      def createTuple2(exprBuf: ExprBuffer, _1: Val, _2: Val): Val = {
+        allocAndConstruct(exprBuf,
+                          tuple2Name,
+                          Seq(jlObjectType, jlObjectType),
+                          Seq(_1, _2))
+      }
+
       def genLazyClassInstantiationMethod(exprBuf: ExprBuffer,
                                           ctors: Seq[global.Symbol]): Val = {
         val applyMethodSig =
           Sig.Method("apply", Seq(jlObjectType, jlObjectType))
-
-        val tuple2Name = Global.Top("scala.Tuple2")
-        val tuple2Type = Type.Ref(tuple2Name)
 
         // Constructors info is an array of Tuple2 (tpes, inst), where:
         // - tpes is an array with the runtime classes of the constructor arguments.
@@ -456,14 +472,10 @@ trait NirGenStat { self: NirGenPhase =>
                 })
 
               // Allocate a new instance and call C.
-              val alloc = exprBuf.classalloc(fqSymName, unwind(curFresh))
-              exprBuf.call(
-                Type.Function(ctorSig.args, Type.Unit),
-                Val.Global(fqSymName.member(Sig.Ctor(ctorSig.args.tail)),
-                           Type.Ptr),
-                alloc +: argsVals,
-                unwind(curFresh)
-              )
+              val alloc = allocAndConstruct(exprBuf,
+                                            fqSymName,
+                                            ctorSig.args.tail,
+                                            argsVals)
 
               exprBuf.ret(alloc)
               exprBuf.toSeq
@@ -491,13 +503,7 @@ trait NirGenStat { self: NirGenPhase =>
 
           // Allocate an instance of the generated class.
           val instantiator =
-            exprBuf.classalloc(reflInstBuffer.name, unwind(curFresh))
-          exprBuf.call(
-            Type.Function(Seq(Type.Ref(reflInstBuffer.name)), Type.Unit),
-            Val.Global(reflInstBuffer.name.member(Sig.Ctor(Seq())), Type.Ptr),
-            Seq(instantiator),
-            unwind(curFresh)
-          )
+            allocAndConstruct(exprBuf, reflInstBuffer.name, Seq(), Seq())
 
           // Create the current constructor's info. We need:
           // - an array with the runtime classes of the ctor parameters.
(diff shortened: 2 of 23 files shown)


ekrich:
Is this the issue in the tracker? https://github.com/scala-native/scala-native/issues/1279

sjrd:
Yes indeed, this is for #1279.
I'll assess #1727 tomorrow.

nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
```diff
@@ -12,6 +12,8 @@ final class BinaryDeserializer(buffer: ByteBuffer) {
   private val header: Map[Global, Int] = {
     buffer.position(0)
 
+    val hasClInit = getBool
```
sjrd: I believe the meaning of this byte should be generalized from

> there is a static initializer in this file

to

> there is at least one reason to fully load this file even if it's not referenced elsewhere

*At the moment*, this would be equivalent to "there is a clinit", because nothing else could justify having to load the file. But if we add another feature in the IR in the future that would also require the file to be loaded, we would be able to do so without breaking backward binary compatibility. If we stick to "this file has a clinit", then we would have to add *another* byte for the new feature, breaking everything.

tools/src/main/scala/scala/scalanative/linker/ClassPath.scala
```diff
@@ -7,6 +7,8 @@ import nir.serialization.deserializeBinary
 import java.nio.file.{FileSystems, Path}
 import scalanative.io.VirtualDirectory
 import scalanative.util.Scope
+import java.io.FileInputStream
+import java.nio.ByteBuffer
```
sjrd: These imports do not seem to be used (at least in this commit).

nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
sjrd: From the following commit, I believe you already have ta name for this concept: `classesWithEntryPoint`. So this could simply be renamed to `hasEntryPoint`.

tools/src/main/scala/scala/scalanative/linker/ClassPath.scala
```diff
@@ -50,5 +54,13 @@ object ClassPath {
           deserializeBinary(directory.read(file))
         }
       })
+
+    def classesWithEntryPoints(): Iterable[Global.Top] = {
+      files.filter {
+        case (top, file) =>
+          val buffer = directory.read(file, len = 1)
+          buffer.get != 0
```
sjrd:
```suggestion
          buffer.get() != 0
```
`get()` is side-effecting (calling it twice will not return the same thing), so it should always be called with explicit `()`. This is a commonly used convention in Scala.

sjrd: This information should really be cached. Otherwise all files will have to reread on every linking request.

tools/src/main/scala/scala/scalanative/linker/ClassLoader.scala
```diff
@@ -41,6 +47,18 @@ object ClassLoader {
...
+    def classesWithEntryPoints(): Iterable[Global] = {
+      scopes.filter {
+        case (top, defns) =>
+          defns.exists {
+            case defn: Defn.Define =>
+              val Global.Member(_, sig) = defn.name
+              sig.isClinit
+            case _ => false
+          }
+      }.keySet
```
sjrd: Even in memory, I believe we should cache this result, because it is expensive to compute, and used on every linking request.

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -138,6 +139,10 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
             }
           }
         }
+        val clinit = cls.name.member(Sig.Clinit())
+        if (loaded(cls.name).contains(clinit)) {
+          reachGlobal(clinit)
+        }
```
sjrd: There's something fishy here. If you unroll the code above this snippet to see all of `def reachEntry`, you can see that when `reachEntry` is called, as soon as the class is not abstract, `reachAllocation(cls)` will be called, which will make instances of the class reachable. Calling `reachEntry` for classes with a static initializer will therefore cause too much methods to be reachable, because a non-abstract class with a static initializer will have its instances reachable, although the presence of a static initializer would not necessarily cause the class to be instantiated.

I believe you cannot reuse `reachEntry`. You need a different `def reachClinit` that focuses on reaching the static initializers, and not reach the allocation.

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -219,6 +224,8 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
                 update(Rt.JavaHashCodeSig)
               case sig if sig.isMethod || sig.isCtor || sig.isGenerated =>
                 update(sig)
+              case sig if sig.isClinit =>
+                update(sig)
```
sjrd: It looks like this could be merged with the previous `case`, just adding `|| sig.isClinit`.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
```diff
@@ -82,6 +82,7 @@ trait NirGenStat { self: NirGenPhase =>
       val body   = cd.impl.body
 
       buf += Defn.Class(attrs, name, None, Seq.empty)
+      genReflectiveInstantiation(cd)
```
sjrd: I don't think we should generate reflective instantiation info for structs. Or at least we should very carefully consider the consequences of doing so.

FTR, no reflective instantiation info are ever generated for JS classes in Scala.js, so there is precedent for ignoring that for interop-specific "stuff".

.gitignore
```diff
@@ -19,3 +19,8 @@ bin/
 .settings/
 .externalToolBuilders/
 .cache*
+
+# metals
+/.bloop/
+/.metals/
+/project/metals.sbt
```
sjrd: This should be in a separate PR ;)

tools/src/main/scala/scala/scalanative/codegen/Generate.scala
```diff
@@ -138,6 +138,8 @@ object Generate {
         Next.Unwind(exc, Next.Label(handler, Seq(exc)))
       }
 
+      lazy val clinitCalls = genClinitCalls(unwind)
```
sjrd: A `lazy val` that calls a thing with by-name argument that has side effects is dangerous. Since `clinitCalls` is always called exactly once anyway, it would be easier to reason about by simply inlining `genClinitCalls(unwin)` at the only place where `clinitCalls` is used.

tools/src/main/scala/scala/scalanative/codegen/Generate.scala
```diff
@@ -155,19 +157,33 @@ object Generate {
...
-          Inst.Let(Op.Call(entryMainTy, entryMain, Seq(module, arr)), unwind),
-          Inst.Let(Op.Call(RuntimeLoopSig, RuntimeLoop, Seq(module)), unwind),
-          Inst.Ret(Val.Int(0)),
-          Inst.Label(handler, Seq(exc)),
-          Inst.Let(Op.Call(PrintStackTraceSig, PrintStackTrace, Seq(exc)),
-                   Next.None),
-          Inst.Ret(Val.Int(1))
+                   unwind)
         )
+          ++ clinitCalls // call the class initializers before calling main
```
sjrd: Wouldn't it be better to insert this between the `Let`s for `rt` and `arr`? That would avoid the latter two's lifetimes to extend over the calls to the static initializers.

sjrd:
For tests, you should be able to just copy https://github.com/scala-js/scala-js/blob/3534fda873127e5da1badbefa376ce0c7d951d03/test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala and adapt the package names and imports. Everything should work the same as in Scala.js.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
```diff
@@ -218,6 +229,354 @@ trait NirGenStat { self: NirGenPhase =>
...
+                  // If the expected argument type can be boxed (i.e. is a primitive
+                  // type), then we need to unbox it before passing it to C.
+                  Type.box.get(arg) match {
+                    case Some(_) =>
+                      // first, cast the primitive type to java.lang.Number
+                      val num = exprBuf.as(jlNumberRef, elem, unwind(curFresh))
+                      // then, convert to the desired type
+                      val conv = exprBuf.method(num,
+                                                Type.primConvSig(arg),
+                                                unwind(curFresh))
```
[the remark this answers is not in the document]
errikos: I think that we also need to take care of shrinking conversions. If a constructor expects a `Short` and someone provides e.g. `42`, which is an `Int`, we still have a mismatch.

I think that we need something more generic, like:

```scala
  private def primConversionIfRequired(arg: Any,
                                       paramType: Class[_]): Any = {
    paramType match {
      case java.lang.Byte.TYPE =>
        arg.asInstanceOf[Number].byteValue()
      case java.lang.Short.TYPE =>
        arg.asInstanceOf[Number].shortValue()
      case java.lang.Integer.TYPE =>
        arg.asInstanceOf[Number].intValue()
      case java.lang.Long.TYPE =>
        arg.asInstanceOf[Number].longValue()
      case java.lang.Float.TYPE =>
        arg.asInstanceOf[Number].floatValue()
      case java.lang.Double.TYPE =>
        arg.asInstanceOf[Number].doubleValue()
      case _ => arg
    }
  }
```
It is basically what I was doing in the codegen, moved to the library side.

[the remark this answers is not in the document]
errikos: > This won't work for booleans and characters.

Why do we need to convert `Boolean`s and `Character`s? They are not numeric types.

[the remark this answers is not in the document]
sjrd: If you follow the link to the spec I gave, you'll see that narrowing conversions shouldn't be accepted, and that characters can be converted to `int` and bigger.

As validation, here is a transcript of a REPL on the JVM to show that this is indeed what happens:
```scala
scala> case class Foo(x: Int, y: java.lang.Long)
defined class Foo

scala> val c = classOf[Foo].getDeclaredConstructors()(0)
c: java.lang.reflect.Constructor[_] = public Foo(int,java.lang.Long)

scala> c.newInstance(1, 2L)
res4: Object = Foo(1,2)

scala> c.newInstance(1, 2)
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided

scala> c.newInstance(1, 2.toShort)
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided

scala> c.newInstance(1L, 2L)
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided

scala> c.newInstance(1.toShort, 2L)
res8: Object = Foo(1,2)

scala> c.newInstance(1.4, 2L)
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided

scala> c.newInstance('A', 2L)
res10: Object = Foo(65,2)

scala> case class Foo(x: Int, y: java.lang.Long, z: Short, c: Char)
defined class Foo

scala> val c = classOf[Foo].getDeclaredConstructors()(0)
c: java.lang.reflect.Constructor[_] = public Foo(int,java.lang.Long,short,char)

scala> c.newInstance('A', 2L, 'A', 'B')
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided

scala> c.newInstance('A', 2L, 5, 'B')
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided

scala> c.newInstance('A', 2L, 5.toByte, 'B')
res13: Object = Foo(65,2,5,B)

scala> c.newInstance('A', 2L, 5.toShort, 'B')
res14: Object = Foo(65,2,5,B)

scala> c.newInstance('A', 2L, 5.toShort, 5.toShort)
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided

scala> c.newInstance('A', 2L, 5.toShort, 5.toByte)
java.lang.IllegalArgumentException: argument type mismatch
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  ... 34 elided
```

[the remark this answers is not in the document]
sjrd: > 
> 
> > This won't work for booleans and characters.
> 
> Why do we need to convert `Boolean`s and `Character`s? They are not numeric types.

`Boolean`s need not be converted, but the code that is currently in the PR would try to cast `Boolean`s to `Number`s anyway, even if the expected target type was `Boolean`. Or maybe I misunderstood the code. For `Char` we in fact need to convert them to numeric types, as I showed above.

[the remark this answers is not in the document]
errikos: Okay, thanks for the clarification.

I got confused from a test case from Scala.js. In the `testClassCtorWithArgs` test, the `Short` constructor is requested and I though it was "converted" from the `Int` one. However I now see the presence of the `VC` class. The current implementation does not support such a scenario.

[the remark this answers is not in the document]
sjrd: The current implementation in Scala.js seems incomplete about the conversions. We'll need to fix that when I get the chance.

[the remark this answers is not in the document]
errikos: So, if I understand this correctly, instantiating a `Short` constructor with `42`, which is [supported](https://github.com/scala-js/scala-js/blob/3534fda873127e5da1badbefa376ce0c7d951d03/test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala#L179) in Scala.js, should not be supported in Native.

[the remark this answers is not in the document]
errikos: Therefore, if I understand this correctly, calling a `Short` constructor with a value of `42` (which is an `Int`), which is [supported by Scala.js](https://github.com/scala-js/scala-js/blob/3534fda873127e5da1badbefa376ce0c7d951d03/test-suite/js/src/test/scala/org/scalajs/testsuite/library/ReflectTest.scala#L179), should not be supported by Native?

I suppose that Scala.js treats numeric values differently, in a JS-specific way. I am a little confused as to which semantics we would like to "follow". In C for example, it is OK (to some extent) to do all kinds of conversions. Why not do the same in Native and why is the behaviour of Scala.js different in that aspect?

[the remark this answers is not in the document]
sjrd: > Therefore, if I understand this correctly, calling a Short constructor with a value of 42 (which is an Int), which is supported by Scala.js, should not be supported by Native?

That's right.

> I suppose that Scala.js treats numeric values differently, in a JS-specific way. I am a little confused as to which semantics we would like to "follow". In C for example, it is OK (to some extent) to do all kinds of conversions. Why not do the same in Native and why is the behaviour of Scala.js different in that aspect?

We have to follow the JVM semantics first. Copying C or Scala.js is secondary.

Scala.js treats boxed primitive values differently than the JVM across the board (see https://www.scala-js.org/doc/semantics.html) for JS-specific reasons, and is therefore not a gold standard to follow if the JVM does not agree. Basically in Scala.js we *cannot* tell apart a `j.l.Integer` with value 42 from a `j.l.Integer` with value 42, so we're forced to accept it.

> On the other hand, Scala seems to accept the above scenario:

That's with *compile-time* adaptation. Java does the same in that situation. But here we are concerned with *run-time* adaptation as performed by Java reflection, since that is the capability we are trying to emulate.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenSymbols.scala
```diff
@@ -0,0 +1,25 @@
...
+  val serializable = Global.Top("scala.Serializable")
+
+  val jlClass    = Global.Top("java.lang.Class")
+  val jlClassRef = Type.Ref(jlClass)
+
+  val jlObject    = Global.Top("java.lang.Object")
+  val jlObjectRef = Type.Ref(jlObject)
+
+  val jlNumber    = Global.Top("java.lang.Number")
+  val jlNumberRef = Type.Ref(jlNumber)
```
sjrd, line 16: This is now unused, I think.

errikos:
From what I understand, this is going to be squashed and merged with the first comment being the commit message. Will edit today with a summary of the changes.

sjrd:
Oops, cross-action. I took the PR description to make the squashed commit message. It contained the most important info.

errikos:
I would add a few more details, but it was good enough. Thanks!

PR #1731, opened 2020-02-26
Ignore bloop/metals directories/files

errikos:
Add metals/bloop files/directories to `.gitignore`.


lolgab:
In my machine bloop generates a `/project/.bloop` directory too. Don't know if it's something in my setup

PR #1738, opened 2020-03-06
Fix crash when HOME env variable is not set

lolgab:
Steps to reproduce:
- Link any Scala Native binary using the `java.lang.System` object
- unset the `HOME` environment variable
- run the binary:
```bash
$ unset HOME
$ target/scala-2.11/hello-out
```
And it crashes with:
```
java.lang.NullPointerException
        at java.util.Hashtable.put(Unknown Source)
        at java.util.Properties.setProperty(Unknown Source)
        at java.lang.System$.loadProperties(Unknown Source)
        at java.lang.System$.<init>(Unknown Source)
        at <none>._SM17java.lang.System$G4load(Unknown Source)
        at scala.sys.package$.env(Unknown Source)
        at JsonAction$class.main(Unknown Source)
        at Main$.main(Unknown Source)
```

### DIFF
--- a/javalib/src/main/scala/java/lang/System.scala
+++ b/javalib/src/main/scala/java/lang/System.scala
@@ -63,13 +63,14 @@ object System {
         sysProps.setProperty("user.language", userLang)
         sysProps.setProperty("user.country", userCountry)
       }
-      val home = {
+      {
         val buf = stackalloc[pwd.passwd]
         val uid = unistd.getuid()
-        pwd.getpwuid(uid, buf)
-        fromCString(buf.pw_dir)
+        val res = pwd.getpwuid(uid, buf)
+        if (res == 0 && buf.pw_dir != null) {
+          sysProps.setProperty("user.home", fromCString(buf.pw_dir))
+        }
       }
-      sysProps.setProperty("user.home", home)
       val buf = stackalloc[scala.Byte](1024)
       unistd.getcwd(buf, 1024) match {
         case null =>
--- a/javalib/src/main/scala/java/lang/System.scala
+++ b/javalib/src/main/scala/java/lang/System.scala
@@ -61,7 +61,8 @@ object System {
         sysProps.setProperty("user.language", userLang)
         sysProps.setProperty("user.country", userCountry)
       }
-      sysProps.setProperty("user.home", getenv("HOME"))
+      val home = getenv("HOME")
+      if (home != null) sysProps.setProperty("user.home", home)
       val buf = stackalloc[scala.Byte](1024)
       unistd.getcwd(buf, 1024) match {
         case null =>
--- a/javalib/src/main/scala/java/lang/System.scala
+++ b/javalib/src/main/scala/java/lang/System.scala
@@ -63,7 +63,7 @@ object System {
         sysProps.setProperty("user.language", userLang)
         sysProps.setProperty("user.country", userCountry)
       }
-      {
+      locally {
         val buf = stackalloc[pwd.passwd]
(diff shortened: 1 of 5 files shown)


lolgab:
On Mac Os X and Linux (tried with alpine on docker) it is set correctly regardless of the `HOME` environment variable. So this PR doesn't solve the problem. I faced the crash on a docker environment.. I don't know if we want to have a proper solution with platform dependent code, for sure I don't want my applications to suddenly crash in some environments.
As a workaround you can call:
```scala
import scala.scalanative.posix.stdlib.setenv
setenv(c"HOME", c"", 0)
```
as first instruction in the main.

ekrich:
Here is a discussion I found on SOF - https://stackoverflow.com/questions/2910377/get-home-directory-in-linux
Not sure without playing with this if it could be an option.

ekrich:
@lolgab Maybe some code like this? Not sure if you need to look at HOME at all.
Tested in Xcode on macOS. The only question here is POSIX ok for the core system if we are trying to get more modular?
```c
#include <unistd.h>
#include <pwd.h> // POSIX only

char* home = getenv("HOME"); // STD C
printf("HOME: %s\n", home);    
unsetenv("HOME");
const char *homedir;

if ((homedir = getenv("HOME")) == NULL) {
    printf("HOME NULL\n");
    uid_t uid = getuid();
    struct passwd *pw = getpwuid(uid);
    homedir = pw->pw_dir;
}
printf("HOME2: %s\n", homedir);
```
Output:
```
HOME: /Users/eric
HOME NULL
HOME2: /Users/eric
```

lolgab:
Nice! I'll try it on Linux and update the PR if it works!

lolgab:
@ekrich I applied your suggestion! Thank you 🙏

ekrich:
Based on your JVM test, I was thinking that is what they might be doing.

ekrich:
The function you are calling (https://github.com/scala-native/scala-native/pull/1738/commits/5fc118979ef02e8ef56b674005b0892149d83402#diff-9f67b7df156c5096453ed90792be0ad2R69) is a Scala Native wrapper function that you are proposing to remove in https://github.com/scala-native/scala-native/pull/1743

This is the reason I was confused. We are not using error number at all after calling `struct passwd *getpwuid(uid_t);` , just some logic from the wrapper. https://github.com/scala-native/scala-native/blob/master/nativelib/src/main/resources/pwd.c#L25-L30

I am unclear what should be done as the POSIX `pwd` needs updating too. These wrappers and struct copying perhaps were created earlier because we didn't have `UInt` and `ULong` support.

If you add the real `getpwuid` function in this PR and just check whether the `Ptr` returned is `NULL` or not that might be good for this PR and then the other PR can also stand as well.

javalib/src/main/scala/java/lang/System.scala
```diff
@@ -61,7 +63,14 @@ object System {
         sysProps.setProperty("user.language", userLang)
         sysProps.setProperty("user.country", userCountry)
       }
-      sysProps.setProperty("user.home", getenv("HOME"))
+      {
```
sjrd, line 66: Consider using
```suggestion
      locally {
```
to make this more readable.

Also, consider wrapping the following code for `user.dir` in a similar `locally {}` block, since it also has a `val buf`.

lolgab: Did not know about `locally` existence! Thank you :-) I'll implement the suggested changes!

lolgab:
@sjrd Ready for another review round.

PR #1745, opened 2020-04-02
Fix warning with musl libc due to wrong include

lolgab:
Source of the warning can be found [here](https://git.musl-libc.org/cgit/musl/tree/include/sys/errno.h?h=v1.2.0#n1)


lolgab:
Thank you for merging 🙏

PR #1746, opened 2020-04-04
Fix #1652: Allow declaration of external functions with varying signatures in separate objects

### LINKED ISSUE #1652: Allow declaration of external functions with varying signatures in separate objects
Declaration of external functions with multiple signatures works when all signatures are declared in the same object, for example:

```C
void* foo(void* p) { return p; }
```
```Scala
@extern
object Foo {
  def foo(i: Int): Int = extern

  @name("foo")
  def bar(cstr: CString): CString = extern
}
```

However, the above example fails during compilation of LLIR, if `foo` and `bar` are declared in separate objects.

This is an issue if multiple libraries declare their own binding to the same external function, or if bindings to external function are generated on the fly and hence cannot be declared in a single extern object.

lolgab:


sjrd:
I don't understand why this works, or why it did not work before.

The code added in this PR appears to deduplicate extern `Global`s with the same `sig`, so that they don't end up in `deps` by way of `touch`. However, there is already code in `genDeps()` to deduplicate globals with the same `mangled(n)`. Now, `mangled(n)` only depends on `sig` for extern globals, so if they have the same `sig`, they must have the same `mangled(n)`.

What does the early deduplication of the extern globals achieve that is not already achieved by the deduplication of `deps`?

lolgab:
First, thank you very much for the review, as usual :)
Second, it is hard to remember all the reasons behind the changes here. 
If I remember correctly, the problem was that calls are generated before globals (genDefns comes before genDeps [here](https://github.com/scala-native/scala-native/blob/7f193e2a2d1e36f9db6e078e28cd9feee5014a22/tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala#L100-L105))
So if two `Call`s point to two `Global`s with different `sig`, later the globals are deduplicated and only one remains, but still one of the two calls point the wrong one.
Hope this makes it clearer!

sjrd:
But even then, the generated *calls* also use `mangled(pointee)` (through `genGlobal(pointee)`) as the only generated name that depends on the global being accessed. And as explained above, `mangled(g)` only depends on the `sig` for externs, so they would also have the same generated names in the *calls*. Therefore, there cannot be one that points to the wrong one: they necessarily point to the same thing.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -730,7 +730,20 @@ object CodeGen {
...
+                case gg @ Global.Member(_, `sig`) => gg
+              } match {
+                case Some(gg) if gg != g =>
+                  Op.Call(ty, Val.Global(gg, valty), args)
+                case _ => call
+              }
+            case _ =>
+              call
+          }
+          genCall(genBind, withGeneratedExtern, unwind)
```
WojciechMazur, line 746: I don't really like idea of traversing `deps` each time. You may consider adding `val externSigMembers = mutable.Map.empty[Sig, Global.Member]` and use it instead of `deps` traversal. Also I would consider it as hacking so we should provide some comment describing it. 

```suggestion
         /*In case if it's a call to extern method with same mangled Sig as some already defined call
            in other module we need to manually enforce getting into second case of `genCall`
            (when lookup(pointee) != call.ty). By replacing `call.ptr` with ptr of that already defined call we would
            enforce creating call bitcasts to correct type.

            Due to deduplication in `genDeps` and since mangling Sig.Extern is not based on function types
            each extern method in deps is generated only once in IR file. In this case LLVM linking
            would result in call arguments type mismatch.
          */
          val callDef = call.ptr match {
            case Val.Global(m @ Global.Member(_, sig), valty) if sig.isExtern =>
              val glob = externSigMembers.getOrElseUpdate(sig, m)
              if (glob == m) call
              else call.copy(ptr = Val.Global(glob, valty))
            case _ => call
          }
          genCall(genBind, callDef, unwind)
```

lolgab: Thank you for the suggestion @WojciechMazur. I'm going to apply it as soon as I can.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -88,10 +88,11 @@ object CodeGen {
...
 
-    val copies    = mutable.Map.empty[Local, Val]
-    val deps      = mutable.Set.empty[Global]
-    val generated = mutable.Set.empty[String]
-    val builder   = new ShowBuilder
+    val copies           = mutable.Map.empty[Local, Val]
+    val deps             = mutable.Set.empty[Global]
+    val generated        = mutable.Set.empty[String]
+    val externSigMembers = mutable.Map.empty[Sig, Global.Member]
+    val builder          = new ShowBuilder
```
lolgab, line 95: Question: Does it make sense to make these mutable things `private`? In case it does, should this be part of this pull request?

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -681,7 +682,22 @@ object CodeGen {
...
         case call: Op.Call =>
-          genCall(genBind, call, unwind)
+          /* When a call points to an extern method with same mangled Sig as some already defined call
+           * in another extern object we need to manually enforce getting into second case of `genCall`
+           * (when lookup(pointee) != call.ty). By replacing `call.ptr` with the ptr of that already
+           * defined call so we can enforce creating call bitcasts to the correct types.
+           * Because of the deduplication in `genDeps` and since mangling Sig.Extern is not based
+           * on function types, each extern method in deps is generated only once in IR file.
+           * In this case LLVM linking would otherwise result in call arguments type mismatch.
+           */
```
lolgab, line 692: I rephrased a bit your comment @WojciechMazur, tell me if you like the new phrasing :)

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
sjrd: The class itself is private is private, so it doesn't matter much, but feel free to make them explicitly private if they are not used by other classes in the same file. It won't hurt :)

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
WojciechMazur: It's fine I guess

sjrd (approved):

lolgab:
Thank you very much for merging this 🥳🥳🥳

jokade:
Great - that brings us a big step closer to having SN on iOS with ObjC interop 👏 👏 👏

PR #1751, opened 2020-04-07
Fix user.home lookup

ekrich:
@lolgab This is what I was suggesting. I don't know if you want to just work on your branch or just clean this up and go with it. We probably should change `gwtpwnam` too.


ekrich:
This is totally not going to work without `struct`s with named field access. This is the reason the `C` helper functions copy from the `C` struct to the Scala Native `struct` to ensure the number of members and that the correctly named ones get transferred in the C code.

PR #1756, opened 2020-04-09
Fix regression in Travis file on commit db3bea25

ekrich:
@LeeTibbert Thanks for finding this problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,12 @@ env:
   matrix:
     - MODE=source-checks
     - MODE=test-tools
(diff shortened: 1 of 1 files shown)


ekrich:
@sjrd If this could be merged once it clears CI that would be great. We weren't running the correct tests in CI.

sjrd:
What's the difference between `MODE` and `TEST_COMMAND`?

ekrich:
The `MODE` is used before in the `script` portion of the Travis file to control what gets run but then the `TEST_COMMAND` gets past to the Docker - that is a bit confusing to have both the `sbt` alias `test-runtime` and the switch string for the Travis `script` to be the same.

ekrich:
Thanks for the quick turn around.

PR #1759, opened 2020-04-18
Make the 'nir' and 'util' packages cross-compilable with dotty

marcesquerra:
As a learning exercise, I thought on giving it a try to migrate `scala-native` to `dotty`. I know it's probably a lot of work, but I want to see how far I can go.

While doing that, I've seen that probably, the first logical step would be to try and cross-compile the `util` and `nir` packages with `dotty`.

Since this can be done cross-compiling, while keeping the whole build working, I thought on creating this PR back into the main repo, if you are interested.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,7 @@ val sbt10Version          = "1.1.6"
 val sbt10ScalaVersion     = "2.12.10"
 val libScalaVersion       = "2.11.12"
 val libCrossScalaVersions = Seq("2.11.8", "2.11.11", libScalaVersion)
+val dottyVersion          = "0.23.0-RC1"
 
 // Convert "SomeName" to "some-name".
 def convertCamelKebab(name: String): String = {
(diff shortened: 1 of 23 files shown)


sjrd:
Hi @marcesquerra. Thanks for your interest and for your PR.

Sorry for the delay. PRs haven't been reviewed for a while. I am currently walking down the PR queue trying to review as many as possible.

It seems like most of the changes here are actually related to upgrade to the Scala 2.13 stdlib, rather than fundamentally about dotty.

So I believe we should hold off a bit with this, and first have our tools, including util and nir, cross-compile with 2.13. The changes for dotty are less stable from a build perspective, as dotty is still a moving target. Moreover, cross-compiling with 2.13 would already allow dotty users to use the relevant artifacts, since dotty can consume Scala 2.13 libraries. The cost/benefit ratio of including dotty in this codebase seems too high at the moment.

PR #1767, opened 2020-04-28
WIP - BufferedInputStream - mark and async close

coreyoconnor:
This is a re-implementation of `BufferedInputStream` to correct the behavior WRT `mark` and asynchronous `close`.

Personally, I think the `BufferedInputStream` (and referenced `InputStream`) specifications are problematic. Still, I aimed for what's in the spec plus what I could infer from the Harmony implementation and bug reports.

TODO:

- [ ] formatting
- [ ] do a few performance tweaks
- [ ] review
- [ ] test async close. I'm out of the loop WRT thread support on scala-native

An alternative is to copy the harmony implementation. Which is probably fine... but less fun. ;)


coreyoconnor:
I'm unconvinced of the value of optimizing `skip`. Is there a use case where the current implementation (iterating `read`) is going to be problematic? Is there some application that uses read to skip significant number of bytes while using a large buffer, for instance?

As for optimizing `read` into an array. I didn't do anything fancy, but instead just applied `@tailrec` to avoid the stack recursion. Could invest more there but that should be sufficient IMO.

Given the above, I think it's ready for review :)

javalib/src/main/scala/java/io/BufferedInputStream.scala
```diff
@@ -1,129 +1,247 @@
...
-
-  /** The position of the last element in the buffer excluded */
-  private[this] var end = 0
-
-  override def available(): Int = {
-    if (closed) throw new IOException()
-    end - pos + in.available()
+  protected[this] var pos: Int = 0
+
+  override def available(): Int = Option(in) match {
```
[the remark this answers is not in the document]
sjrd: Well, in fact the above `ensureOpen()` is not good enough, because `read()` can reallocate a bigger `buf`, which means that it can concurrently set another non-null `buf` after the `buf = null`.

Therefore, I believe a dedicated
```scala
private[this] var closed: Boolean = false
```
and then do:
```scala
def ensureOpen(): (InputStream, Array[Byte]) = {
  /* First read `in` and `buf`, then `closed`. Since `closed` is the first thing
   * that is set to `true` in `close()`, we know that if `closed` is false, the
   * `in` and `buf` are non-null.
   */
  val in = this.in
  val buf = this.buf
  if (closed)
    throw new IOException("Operation on closed stream")
  (in, buf)
}
```
while `close()` starts by setting `closed` to `true`:
```scala
def close(): Unit = {
  if (!closed) {
    closed = true
    val in = this.in
    if (in != null)
      in.close()
    in = null
    buff = null
  }
}
```

javalib/src/main/scala/java/io/BufferedInputStream.scala
```diff
@@ -1,129 +1,247 @@
...
+  private def fillMarkedBuffer(sourceBuffer: Array[Byte],
+                               source: InputStream): Option[Array[Byte]] =
+    if (count < sourceBuffer.length) {
+      fillMarkedBufferUnsafe(sourceBuffer, source)
+    } else {
+      val newBuffer = new Array[Byte](sourceBuffer.length * 2)
+      sourceBuffer.copyToArray(newBuffer)
+      buf = newBuffer
+      fillMarkedBufferUnsafe(newBuffer, source)
+    }
```
sjrd, line 128: Consider the simpler:
```scala
val bigEnoughBuf = if (count < sourceBuffer.length) {
  sourceBuffer
} else {
  val newBuffer = new Array[Byte](sourceBuffer.length * 2)
  sourceBuffer.copyToArray(newBuffer)
  buf = newBuffer
  newBuffer
}
// then the current content of fillMarkBufferUnsafe, replacing sourceBuffer by bigEnoughBuf
```

sjrd: Also after setting `buf = newBuffer`, we should check again whether `closed` is true, and if yes, reset it back to `null`:
```scala
buf = newBuffer
if (closed)
  buf = null
```
Doing the naively simpler
```scala
if (!closed)
  buf = newBuffer
```
is not correct, since we could first read `closed` as false, then `close()` would execute and set `closed = true` and `buf = null`, then we'd come back here and do `buf = newBuffer`.

javalib/src/main/scala/java/io/BufferedInputStream.scala
```diff
@@ -1,129 +1,247 @@
...
-    } else -1
-  }
+  private def unsafeRead(b: Array[Byte],
+                         off: Int,
+                         len: Int,
+                         sourceBuffer: Array[Byte],
+                         source: InputStream): Int =
+    unsafeRead(b, off, len, sourceBuffer, source, 0)
+
+  @annotation.tailrec
```
sjrd, line 190: I'm really not a fan of supposedly-tailrec functions that actually mutate outside state (`pos`, the content of `b`, the call to `fillBuffer`, etc.) in addition to changing the value of some accumulators (`bytesRead`). It means that when reading this, I have to reason about the explicit state *plus* the implicit state. That's too much for my brain to handle.

Please use a good old loop with a `var bytesRead` instead.

unit-tests/src/test/scala/java/io/BufferedInputStreamSuite.scala
```diff
@@ -27,9 +25,17 @@ object BufferedInputStreamSuite extends tests.Suite {
...
+
+  test("simple array reads") {
+
+    val arrayIn = new ByteArrayInputStream(exampleBytes0, 0, 10)
+
+    val in = new BufferedInputStream(arrayIn)
 
     val a = new Array[Byte](7)
 
+    in.skip(3)
```
sjrd, line 38: You could verify the result returned by `skip`.

coreyoconnor:
> Thanks for the PR. Here is a first round of review.

Thanks for the comments! Haven't had a chance to work on this lately. Still in my queue. Should be easy enough to resolve the requests. :)

PR #1775, opened 2020-05-07
Upgrade for tools to Scala 2.12.11, ScalaTest 3.1.1 and ScalaCheck 1.14.3.

ekrich:
Added a comment in the build as well.

### DIFF
--- a/tools/src/test/scala/scala/scalanative/nir/SigManglingSuite.scala
+++ b/tools/src/test/scala/scala/scalanative/nir/SigManglingSuite.scala
@@ -2,8 +2,9 @@ package scala.scalanative
 package nir
 
 import org.scalatest._
+import org.scalatest.funsuite.AnyFunSuite
 
-class SigManglingSuite extends funsuite.AnyFunSuite {
+class SigManglingSuite extends AnyFunSuite {
   Seq(
     Sig.Field("f"),
     Sig.Field("len"),
--- a/tools/src/test/scala/scala/scalanative/nir/SigManglingSuite.scala
+++ b/tools/src/test/scala/scala/scalanative/nir/SigManglingSuite.scala
@@ -3,7 +3,7 @@ package nir
 
 import org.scalatest._
 
-class SigManglingSuite extends FunSuite {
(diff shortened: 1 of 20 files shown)


errikos:
I believe we should upgrade `org.scalamacros:paradise` to `2.1.1`. It seems that `2.1.0` is not (yet?) [published](https://mvnrepository.com/artifact/org.scalamacros/paradise) for 2.12.11.

ekrich:
@errikos Oh gosh, I didn't run the tools test. I will go ahead and update that plus maybe some other dependencies too as part of our overall refresh.

tools/src/test/scala/scala/scalanative/nir/SigManglingSuite.scala
```diff
@@ -3,7 +3,7 @@ package nir
 
 import org.scalatest._
 
-class SigManglingSuite extends FunSuite {
+class SigManglingSuite extends funsuite.AnyFunSuite {
```
sjrd, line 6: In all other places you added an import for that. Could you do the same here, please?

sjrd:
Anyone knows what we even use `paradise` for?

tools/src/test/scala/scala/scalanative/nir/SigManglingSuite.scala
ekrich: Sure, I originally did that and then went back so I must have missed that one.

ekrich:
It looks like it may have only been needed for Scala 2.10 - https://github.com/lihaoyi/fastparse/issues/72

sjrd:
> 
> 
> It looks like it may have only been needed for Scala 2.10 - [lihaoyi/fastparse#72](https://github.com/lihaoyi/fastparse/issues/72)

Do you want to try and get rid of it, while you're there, instead of upgrading it?

ekrich:
Yes, please. If we don't need it I think that is best. I also would like to update fastparse but in another PR.

ekrich:
@sjrd @errikos Thanks!

PR #1779, opened 2020-05-07
Fix IDEA warnings + optimise imports in build.sbt

errikos:
IntelliJ IDEA shows a lot of syntax highlighting errors in `build.sbt`.

Fix them by:
- Adding type annotations to settings declarations.
- Removes the two unused imports.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,5 @@
 import java.io.File.pathSeparator
+
 import scala.util.Try
 
 val sbt10Version          = "1.1.6" // minimum version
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -202,15 +201,14 @@ lazy val publishSettings = Seq(
       browseUrl = url("https://github.com/scala-native/scala-native"),
       connection = "scm:git:git@github.com:scala-native/scala-native.git"
     )),
-  pomExtra := (
+  pomExtra :=
```
sjrd, line 204: It is good practice to surround XML literals with `()`, due to quirks in how several nodes may or may not be bundled in a `NodeSeq`. I'm surprise IntelliJ flags this. In any case, it would be better to keep the `()`.

errikos: Well, it was a suggestion. Good to know, will revert.

PR #1789, opened 2020-05-10
WIP: update fastparse to 2.3.0

ekrich:
The following documents were use to help convert the code from `fastparse` `1.0` to `2.3`. Things change quite a bit on the `2.0` release. This upgrrade could not be performed until we dropped Scala `2.10` support used for `sbt` `0.13.x`. It should provide better performance.
http://www.lihaoyi.com/fastparse/#GettingStarted
https://javadoc.io/doc/com.lihaoyi/fastparse_2.12/latest/fastparse/index.html

The following areas need attention as the port is not straightforward.

1. This is used as a base for all the NIR parsers.
https://github.com/ekrich/scala-native/blob/topic/fastparse/nirparser/src/main/scala/scala/scalanative/nir/parser/Base.scala#L8-L13

2. This is used for whitespace and cannot be directly ported.
https://github.com/ekrich/scala-native/blob/topic/fastparse/nirparser/src/main/scala/scala/scalanative/nir/parser/Base.scala#L16-L31

3. To get this section to compile `p` was changed to `P(p)`. I doubt this is correct because the argument `p` is `P[String]` so this doesn't make sense.
https://github.com/ekrich/scala-native/blob/topic/fastparse/nirparser/src/main/scala/scala/scalanative/nir/parser/Base.scala#L78

4. The tests do not compile but this is probably due to **1.** above.

This is all the information I have at this time.


ekrich:
Fastparse `2.3.0` only supports Scala `2.12` and `2.13`. Early versions in the `2.x.x` series either do not support `2.13` or `2.11`. If the code would work with different versions it is possibly to maybe support `2.11` with `2.1.2`, and later Scala versions with `2.1.3+`. 

See https://mvnrepository.com/artifact/com.lihaoyi/fastparse

PR #1797, opened 2020-05-12
Small doc changes due to recent sbt updates

ekrich:


hepin1989:
IIRC, only sbt 1.3.x dosen't need the first semicolon, better keep it.

ekrich:
Yeah, I guess if `1.1.6` is min, we should probably keep them. Any other opinions?

ekrich:
Thanks!

PR #1802, opened 2020-05-15
Use reflection instead of source generation in tests

errikos:
Instead of generating sources that collect, register and run the test classes, leverage the newly introduced `@EnableReflectiveInstantiation` annotation for the testing infrastructure.

In particular:
- Load the tests and frameworks via reflection. To that end, annotate `tests.Suite` and `sbt.testing.Framework` with `@EnableReflectiveInstantiation`.
- Move `org.scalajs.testinterface.TestUtils` to `scala.scalanative.testinterface.TestUtils`, make it uses reflection and deprecate. This will be removed in future versions (probably after 0.4.0). Frameworks should use the reflection API to load test classes.
- Get rid of `PreloadedClassLoader`.
- Get rid of the `NativeTest` SBT configuration.
- Do not generate a test main anymore. Rename `scala.scalanative.testinterface.TestMainBase` to `scala.scalanative.testinterface.TestMain` and use it as the test main class.

A subtle change is in `scala.scalanative.testinterface.ScalaNativeFramework`. The `name` and `fingerprints` are no longer obtained via the `Runner`, but via the enclosed `Framework` object (which is a `NativeFramework`). This is required, since we removed the `NativeTest` SBT configuration (it is now basically just the `Test` configuration) and the `Runner` is not available at the moment the above attributes are requested. So, we use the `NativeFramework` from the JVM to obtain them. It is not very elegant, but does the job for the time being. We should refactor this in a less ugly solution, similar to what Scala.js does with its testing adapter. But that deserves its own PR.


test-runner/src/main/scala/scala/scalanative/testinterface/ScalaNativeFramework.scala
```diff
@@ -35,11 +33,4 @@ class ScalaNativeFramework(val framework: Framework,
   }
 
   private[testinterface] def runDone(): Unit = _runner = null
-
-  private def fetchFrameworkInfo(): FrameworkInfo = {
-    _runner.send(Command.SendInfo(id, None))
-    val Command.SendInfo(_, Some(infos)) = _runner.receive()
-    infos
-  }
```
sjrd, line 43: Normally I wouldn't say that, but let's keep this, commented out, with a comment explaining why we're not using it at the moment, and that we should reenable it in the future.

unit-tests/src/main/scala/tests/NativeFramework.scala
```diff
@@ -1,8 +1,10 @@
 package tests
 
 import sbt.testing.{Fingerprint, Framework, Runner}
-import scala.scalanative.testinterface.PreloadedClassLoader
 
+import scala.scalanative.reflect.annotation.EnableReflectiveInstantiation
+
+@EnableReflectiveInstantiation
```
sjrd, line 7: This should not be necessary. `NativeFramework` extends `sbt.testing.Framework`, and the latter has `@EnableReflectiveInstantiation`. `NativeFramework` therefore receives this by inheritance.

errikos: I thought so too, but for some reason this is not the case. Removing this annotation results in a InstantiationError at `FrameworkLoader`.

Following the `sbt.testing.Interface` import in this file gives two options in IntelliJ: the one from `test-interface-sbt-defs` of Scala Native and one from `org.scala-sbt:test-interface`. It seems that `unit-tests` project depends on both. Maybe it resolves to the second one, which does not have the annotation?

sjrd: Could you check `show unit-tests/test:dependencyClasspath`? Does it have an entry for the JVM version of sbt-testing-interface? If yes, the build needs to be fixed so that that doesn't happen.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -169,51 +151,31 @@ object ScalaNativePluginInternal {
...
-                                      testBinary,
-                                      envVars))
+    scalaNativeConfigSettings ++
+      Seq(
+        mainClass := Some("scala.scalanative.testinterface.TestMain"),
+        loadedTestFrameworks := {
+          val frameworks = loadedTestFrameworks.value
+          val logger     = streams.value.log
+          val testBinary = nativeLink.value
+          val envVars    = (test / Keys.`envVars`).value
```
sjrd, line 161:
```suggestion
          val envVars    = (test / Keys.envVars).value
```
The backquotes were not there before, and seem unnecessary.

test-interface/src/main/scala/scala/scalajs/testinterface/TestUtils.scala
```diff
@@ -0,0 +1,31 @@
+package scala.scalajs
+package testinterface
```
sjrd, line 2: This was originally in the package `org.scalajs.testinterface`, not `scala.scalajs.testinterface`.

test-interface/src/main/scala/scala/scalajs/reflect/annotation/package.scala
```diff
@@ -0,0 +1,8 @@
+package scala.scalajs.reflect
+
+package object annotation {
+  @deprecated(message = "Use scala.scalanative.reflect.Reflect instead.",
```
sjrd, line 4:
```suggestion
  @deprecated(message = "Use scala.scalanative.reflect.annotation.EnableReflectiveInstantiation instead.",
```

test-runner/src/main/scala/scala/scalanative/testinterface/ComRunner.scala
```diff
@@ -69,7 +68,7 @@ class ComRunner(bin: File,
   def receive(timeout: Duration = Duration.Inf): Message =
     synchronized {
       in.mark(Int.MaxValue)
-      val savedSoTimeout = socket.getSoTimeout()
+      val savedSoTimeout = socket.getSoTimeout
```
sjrd, line 71: The changes in this file look like they could/should be in a different PR as well.

errikos:
Resolved comments + rebased to latest master.

test-interface/src/main/scala/org/scalajs/testinterface/TestUtils.scala
```diff
@@ -1,32 +1,31 @@
...
-// We substitute it with our implementation, expecting that
-// the `ClassLoader` that we'll receive is a `PreloadedClassLoader`,
-// from which we'll directly get the instance we're interested in.
+@deprecated(message = "Use scala.scalanative.reflect.Reflect instead.",
+            since = "0.4.0")
 object TestUtils {
 
-  def newInstance(name: String, loader: ClassLoader)(
-      args: Seq[AnyRef]): AnyRef =
+  def newInstance(name: String, loader: ClassLoader)(args: Seq[AnyRef]): Any =
```
sjrd, line 10: Oh now that git figures out that this is the same file, it's painfully obviously that the result type should be `AnyRef` to keep compatibility with previous code.

test-interface/src/main/scala/org/scalajs/testinterface/TestUtils.scala
```diff
@@ -1,32 +1,31 @@
...
     newInstance(name, loader, Seq.fill(args.length)(null))(args)
 
-  def newInstance(name: String, loader: ClassLoader, paramTypes: Seq[Class[_]])(
-      args: Seq[Any]): AnyRef = {
+  def newInstance(fqcn: String, loader: ClassLoader, paramTypes: Seq[Class[_]])(
+      args: Seq[Any]): Any = {
```
sjrd, line 14: Same here: it should return `AnyRef`.

You'll need to cast the result of `.newInstance(args: _*)` to `AnyRef` for the code to typecheck. (such a cast always works, so it's fine)

test-interface/src/main/scala/org/scalajs/testinterface/TestUtils.scala
```diff
@@ -1,32 +1,31 @@
...
   }
 
-  def loadModule(name: String, loader: ClassLoader): AnyRef =
-    loader match {
-      case l: PreloadedClassLoader => l.loadPreloaded(name)
-      case other =>
-        val clazz = other.loadClass(name + "$")
-        clazz.getField("MODULE$").get(null)
-    }
+  def loadModule(fqcn: String, loader: ClassLoader): Any = {
```
sjrd, line 25: And same here.

sjrd:
It was `AnyRef`, not `AnyVal` 😅

errikos:
> It was `AnyRef`, not `AnyVal` 😅

Oh sweet Moses.

PR #1838, opened 2020-07-08
Fix #1758: Accept a non-directory file in Files.walk()

### LINKED ISSUE #1758: java.nio.file.Files.walk should not fail when called on a file.
Current implementation is based on `java.nio.file.Files.list` that correctly throws `java.nio.file.NotDirectoryException` when launched on a file instead of a directory.
`walk` instead should not fail e list that single file.

```scala
@ Files.walk(Paths.get("/Users/lorenzo/afile")).iterator.next
res6: Path = /Users/lorenzo/afile

@ Files.list(Paths.get("/Users/lorenzo/afile")).iterator.next
java.nio.file.NotDirectoryException: /Users/lorenzo/afile
  sun.nio.fs.UnixFileSystemProvider.newDirectoryStream(UnixFileSystemProvider.java:406)
  java.nio.file.Files.newDirectoryStream(Files.java:457)
  java.nio.file.Files.list(Files.java:3451)
  ammonite.$sess.cmd7$.<clinit>(cmd7.sc:1)

```

WojciechMazur:
Resolves #1758 bug. 
In cases when `Files.walk` was used with path directing to file instead of directory second element of stream resulted in throwing exceptions. 
Also extracted conversion between `FileVisitOption[]` -> `LinkOption[]` used in several places.

### DIFF
--- a/javalib/src/main/scala/java/nio/file/Files.scala
+++ b/javalib/src/main/scala/java/nio/file/Files.scala
@@ -251,14 +251,7 @@ object Files {
            matcher: BiPredicate[Path, BasicFileAttributes],
            options: Array[FileVisitOption]): Stream[Path] = {
     val stream = walk(start, maxDepth, 0, options, SSet.empty).filter { p =>
-      //
-      // See comments for FileVisitOption & linkOpts in _walkFileTree
-      val linkOpts =
-        if (options.contains(FileVisitOption.FOLLOW_LINKS))
-          Array.empty[LinkOption]
-        else
-          Array(LinkOption.NOFOLLOW_LINKS)
-
+      val linkOpts = linkOptsFromFileVisitOpts(options)
       val attributes =
         getFileAttributeView(p, classOf[BasicFileAttributeView], linkOpts)
           .readAttributes()
@@ -586,7 +579,7 @@ object Files {
                    options: Array[FileVisitOption],
                    visited: SSet[Path]): SStream[Path] = {
     start #:: {
-      if (!isDirectory(start, Array.empty)) SStream.empty
+      if (!isDirectory(start, linkOptsFromFileVisitOpts(options))) SStream.empty
       else
         FileHelpers
           .list(start.toString, (n, t) => (n, t))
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/nio/file/Files.scala
```diff
@@ -585,28 +578,34 @@ object Files {
...
-          val newVisited =
-            if (options.contains(FileVisitOption.FOLLOW_LINKS)) visited + path
-            else visited
-          walk(path, maxDepth, currentDepth + 1, options, newVisited)
-        case (name, _) =>
-          start.resolve(name) #:: SStream.Empty
-      }
+    start #:: {
+      if (!isDirectory(start, linkOptsFromFileVisitOpts(options))) SStream.empty
+      else
```
sjrd, line 583: Consider adding braces around that `else`, since it isn't a one-liner:
```suggestion
      else {
```

javalib/src/main/scala/java/nio/file/Files.scala
```diff
@@ -624,15 +623,22 @@ object Files {
...
   private def _walkFileTree(start: Path,
                             options: Set[FileVisitOption],
                             maxDepth: Int,
                             visitor: FileVisitor[_ >: Path]): Path = {
-    val stream = walk(start,
-                      maxDepth,
-                      0,
-                      options.toArray.asInstanceOf[Array[FileVisitOption]],
-                      SSet.empty)
+    val optsArray  = options.toArray.asInstanceOf[Array[FileVisitOption]]
```
sjrd, line 640: While you're here: is the `asInstanceOf` actually necessary?

WojciechMazur: Unfortunately it is necessary. `toArray` on Java collections results in `Array[Object]` (or `Array[AnyRef` in scala compiler). 
Altough another syntax may be used which is ```options.toArray(new Array[FileVisitOption](options.size))``` 
I can change it, but I think it looks cleaner as it is right now.

sjrd: Ah, it's the Java collections! Then it's worse than I thought. This cast is actually *invalid*, since `toArray` returns an `Array[Object]`, which is *not* an `Array[FileVisitOption]`. It seems that Scala Native lets that go through, but that's clearly undefined behavior and we should not rely on it.

Please switch to the other alternative that you mentioned (with the `new Array`).

PR #1839, opened 2020-07-08
Fix "too many open files" after consecutive runs

errikos:
This bug was introduced by PR #1728. It basically leaves
the files open in the VirtualDirectory.read(path, len) method.
After 3-4 consecutive runs in sbt, the build would complain with
"too many open files", which makes sense since the above method
is used to look for classes with entry points, so it is essentially
used to open every single .nir file in the build.


ekrich:
Thanks, I was seeing that problem pretty often.

PR #1841, opened 2020-07-09
Fix #1256: Add JUnit support

### LINKED ISSUE #1256: Consider porting JUnit support from Scala.js
It seems to be generally platform-independent, and it would allow us to share tests as-is wrt Scala.js.

errikos:
Add JUnit support for Scala Native and its projects. See `docs/user/testing.rst` for usage instructions.

The feature was mostly ported from Scala.js, with some adaptations for Scala Native.

### Notes
- The existing Scala Native test runner (`TestMain`) does not support multiple testing frameworks. For that reason, we now spawn one `TestMain` process per testing framework. This can be addressed in a future PR.
- There is a runtime bug that is triggered by the `(String, Throwable)` constructor of `org.junit.AssumptionViolatedException`. The bug is only present in JUnit self-tests. At runtime, the native test runner exits unexpectedly when the AssumptionViolation is tested. This is why the aforementioned constructor is not available at the moment. Some details:
  - The constructor is always made reachable, no matter if it is used or not. This suggests maybe some bug in the reachability analysis and/or code generation.
  - After some digging around, I found out that the one to blame is the `Throwable` argument. When the argument is changed to anything else (or removed entirely), then everything works as expected. That is, the constructor is only made reachable when used and the runtime error goes away.

### Testing
- The `test-all` command alias now also runs the JUnit self-tests.
- There is an example JUnit test added in `unit-tests/src/test/scala/junit/JUnitExampleTest.scala`.

### Usage
To enable JUnit support for a Scala Native project, add the following lines to your `build.sbt` file:
```scala
libraryDependencies += "org.scala-native" %%% "junit-runtime" % "0.4.0-SNAPSHOT"
addCompilerPlugin("org.scala-native" % "junit-plugin" % "0.4.0-SNAPSHOT" cross CrossVersion.full)
```
If you want to get more detailed output from the JUnit runtime, also include the following line:
```scala
testOptions += Tests.Argument(TestFrameworks.JUnit, "-a", "-s", "-v")
```
You may then use `test` or `testOnly` in the sbt shell to run all or specific tests, respectively.


junit-runtime/src/main/scala/org/hamcrest/CoreMatchers.scala
```diff
@@ -0,0 +1,117 @@
+/*
+ * Ported from https://github.com/hamcrest/JavaHamcrest/
+ */
+package org.hamcrest
+
+import org.hamcrest.core._
+
+object CoreMatchers {
+  // Commented matchers where implemented using reflexion. It is possible that
+  // some of them could be reimplemented from scratch without using reflexion.
```
lolgab, line 10: Is the word spelled reflexion on purpose?
```suggestion
  // Commented matchers where implemented using reflection. It is possible that
  // some of them could be reimplemented from scratch without using reflection.
```

errikos: This came straight from Scala.js.

Probably confused with the French version of reflection, which is "réflexion". 😅

junit-runtime/src/main/scala/scala/scalanative/junit/Bootstrapper.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+@EnableReflectiveInstantiation
+trait Bootstrapper {
+  def beforeClass(): Unit
+  def afterClass(): Unit
+  def before(instance: AnyRef): Unit
+  def after(instance: AnyRef): Unit
+
+  def tests(): Array[TestMetadata]
+  def invokeTest(instance: AnyRef, name: String): Try[Unit]
```
sjrd, line 24: Even if you dropped support for async tests to make things work for now, we must prepare the interface of `Bootstrapper` for that support later on. Therefore, `invokeTest` should still return `Future[Try[Unit]]` and the JUnit plugin should still wrap results in `Future.successful`.

junit-runtime/src/main/scala/scala/scalanative/junit/JUnitTask.scala
```diff
@@ -0,0 +1,199 @@
...
+                                      test: TestMetadata,
+                                      reporter: Reporter): Int = {
+    reporter.reportTestStarted(test.name)
+
+    val result = runTestLifecycle {
+      catchAll(bootstrapper.newInstance())
+    } { instance => catchAll(bootstrapper.before(instance)) } { instance =>
+      handleExpected(test.annotation.expected) {
+        catchAll(bootstrapper.invokeTest(instance, test.name)) match {
+          case Success(f) => f.recover { case t => Failure(t) }
```
sjrd, line 86: We can escape the unsupported `Future`s here with:
```suggestion
          case Success(f) => f.value.get.flatten
```
so that the rest of `JUnitTask` can stay synchronous.

Since the JUnit plugin will only create `Future.successful` for now, we know that `.value` will always return a non-empty value.

unit-tests/src/test/scala/junit/JUnitExampleTest.scala
```diff
@@ -0,0 +1,10 @@
+package junit
+
+import org.junit.Assert._
+import org.junit.Test
+
+class JUnitExampleTest {
+  @Test def exampleTest(): Unit = {
+    assertTrue("This assertion should pass", condition = true)
```
sjrd, line 8:
```suggestion
    assertTrue("This assertion should pass", true)
```
(the parameter name is not part of the API, and should not be referred to)

project/build.sbt
```diff
@@ -11,6 +11,7 @@ Compile / unmanagedSourceDirectories ++= {
   ).map(dir => root / s"$dir/src/main/scala")
 }
 
+addSbtPlugin("de.heikoseeberger"  % "sbt-header"        % "5.0.0")
```
sjrd, line 14: It seems that you're not using `sbt-header` anywhere. At least I haven't seen its configuration, nor invocations of it in the CI scripts, etc.

I recommend removing it if it's not going to be used.

build.sbt
```diff
@@ -630,6 +651,103 @@ lazy val testRunner =
...
+
+// JUnit modules and settings ------------------------------------------------
+
+lazy val junitRuntime =
+  project
+    .in(file("junit-runtime"))
+    .enablePlugins(MyScalaNativePlugin)
+    .settings(mavenPublishSettings)
+    .settings(
+      nameSettings,
```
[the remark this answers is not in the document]
errikos: This was addressed by renaming the project `val`s from `jUnit...` to `junit...`.

build.sbt
```diff
@@ -630,6 +651,103 @@ lazy val testRunner =
...
+  Test / parallelExecution := false,
+  Test / unmanagedSourceDirectories +=
+    baseDirectory.value.getParentFile / "shared/src/test/scala",
+  Test / testOptions ++= Seq(
+    Tests.Argument(TestFrameworks.JUnit, "-a", "-s", "-v"),
+    Tests.Filter(_.endsWith("Assertions"))
+  ),
+  Test / scalacOptions --= Seq("-deprecation", "-Xfatal-warnings"),
+  Test / scalacOptions += "-deprecation:false",
+  Test / logBuffered := false
```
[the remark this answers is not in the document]
errikos: Maybe the `logBuffered` can go, but not the other two, as the JUnit self-tests also test deprecated APIs.

junit-plugin/src/main/scala/scala/scalanative/junit/plugin/ScalaNativeJUnitPlugin.scala
```diff
@@ -0,0 +1,283 @@
+package scala.scalanative
+package junit.plugin
```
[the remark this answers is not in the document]
errikos: Hmm, do we have a standard header in Scala Native?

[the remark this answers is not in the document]
sjrd: No, apparently not.

build.sbt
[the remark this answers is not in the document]
sjrd: Yes, I was only referring to `logBuffered`.

build.sbt
```diff
@@ -1,5 +1,7 @@
 import java.io.File.pathSeparator
 
+import sbt.Keys.libraryDependencies
+
```
sjrd, line 4: This import is not necessary. `sbt.Keys._` is already imported implicitly in `build.sbt`.

junit-test/shared/src/test/scala/scala/scalanative/junit/AsyncTest.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Test
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.scalanative.junit.async._
+import scala.scalanative.junit.utils._
+
+class AsyncTest {
+  @Test
+  def success(): Unit = await {
```
sjrd, line 14:
```suggestion
  def success(): AsyncResult = await {
```
That's the whole point of `junit.async._`.

Also for the two methods below.

junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
```diff
@@ -0,0 +1,12 @@
+package scala.scalanative.junit
+
+import scala.concurrent.Future
+import scala.util.Try
+
+package object async {
+  type AsyncResult = Try[_]
```
sjrd, line 7: This should be `Unit`, since that is the only result type supported by the JUnit plugin.
```suggestion
  type AsyncResult = Unit
```
Currently `await` returns a `Try`, but `@Test` methods calling `await` are declared to return `Unit`. That means that the `Try` is discarded, hence any exception caught in the `Try` is discarded!

junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
```diff
@@ -0,0 +1,12 @@
...
+  def await(future: Future[_]): AsyncResult = {
+    scala.scalanative.runtime.loop()
+    future.value.get
```
sjrd, line 10: Since `AsyncResult` will be `Unit`, this should have an extra `.get` to unwrap the `Try` and potentially throw its exception:
```suggestion
    future.value.get.get
```

ekrich:
I saw a comment about normalized names but I think this takes care of it unless this setting is not used. https://github.com/scala-native/scala-native/blob/master/build.sbt#L23

errikos:
> I saw a comment about normalized names but I think this takes care of it unless this setting is not used. https://github.com/scala-native/scala-native/blob/master/build.sbt#L23

Yep, basically that was the part that introduced the problem. It converts the camelCase name to kebab-case.

So, `jUnitPlugin` became `j-unit-plugin`, which is ugly. I renamed to `junitPlugin` so that it becomes `junit-plugin`.

ekrich:
@errikos Cool, I saw the code after it changed to `junit...` so I was behind the happenings.

PR #1845, opened 2020-07-10
Fix #1631: CFuncPtr <-> Ptr[Byte] conversion

### LINKED ISSUE #1631: Cast CFuncPtr to Ptr[Byte] (and reverse)

### LINKED ISSUE #1850: Unable to apply CFuncPtrN assigned to CStruct field
Reproducible test case: 
```
object Test{
  type FnCtx = CStruct1[CFuncPtr0[CInt]]
  val fn1: CFuncPtr0[CInt] = new CFuncPtr0[CInt] {
    override def apply(): CInt = 1
  }
  val x = stackalloc[FnCtx]
  x._1 = fn1
  
 fn1.apply() // This works 
 x._1.apply() //This results in scala.scalanative.runtime.UndefinedBehaviorError
 val fn = x._1 
 fn() //This also would result in UndefinedBehaviorError
```

Stacktrace for my test case
```
[info]          scala.scalanative.runtime.UndefinedBehaviorError
[info]                  at java.lang.Throwable.fillInStackTrace(Unknown Source)
[info]                  at scala.scalanative.runtime.package$.throwUndefined(Unknown Source)
[info]                  at scala.scalanative.runtime.CFuncRawPtr.apply(Unknown Source)
[info]                  at scala.scalanative.unsafe.PtrOpsSuite$$anonfun$2.apply$mcV$sp(Unknown Source)
```

Mentioned error (and stacktrace) where also observed when trying to cast `Ptr[Byte]` to `CFuncPtrN`.

### LINKED ISSUE #1889: dynamically compose dynamically loaded functions
I am trying to define a helper method to dynamically compose functions which are dynamically loaded.

```scala
import scala.scalanative.native._

@extern
@link("dl")
object dl {
  def dlopen(path: CString, mode: CInt): Ptr[Byte] = extern
  def dlsym(handle: Ptr[Byte], sym: CString): Ptr[Unit] = extern
}

object Sample extends App {
  type Lambda = CFunctionPtr1[Ptr[Unit], Ptr[Unit]]

  def compose[A,B](fs: Lambda*)(in: Ptr[A]): Ptr[B] =
    fs.foldLeft(in.asInstanceOf[Ptr[Unit]]){case (acc, f) =>
      f.apply(acc)
    }.asInstanceOf[Ptr[B]]

  def compose[A, B](f: Lambda, g: Lambda)(in: Ptr[A]): Ptr[B] =
    g(f(in.asInstanceOf[Ptr[Unit]])).asInstanceOf[Ptr[B]]

  Zone { implicit alloc =>
    val handle = dl.dlopen(toCString("functions.so"), 1)
    val fOne = dl.dlsym(handle, toCString("one")).asInstanceOf[Lambda]
    val fTwo = dl.dlsym(handle, toCString("two")).asInstanceOf[Lambda]
    val ptr = stdlib.malloc(sizeof[CInt]).asInstanceOf[Ptr[CInt]]

    val result = compose[CInt, CBool](fOne, fTwo)(ptr)
    println(!result)
  }
}
```

The above code works with the fixed args `compose` but not with the varargs `compose`.

In the latter case I am getting:
```
[error] (Compile / nativeLink) scala.scalanative.util.UnsupportedException: can't cast from Class(Top(java.lang.Object)) to Ptr
```

I am using scala native `0.3.9` with scala `2.11.12`.

WojciechMazur:
This PR changes the implementation of `CFuncPtr` allowing to convert it to or create from `Ptr[Byte]`

- `CFuncPtr` contains helper methods `fromPtr` to `toPtr` allowing conversion between `Ptr[Byte]` and `CFuncPtrN`.
- `CFuncPtrN` is now sealed abstract class. It can now be only created using the public implicit conversion method `fromScalaFunction` or scalanative private `fromRawPtr`.  
- New NIR primitives were introduced `CFUNCPTR_FROM_FUNCTION` and `CFUNCPTR_APPLY`
- `CFuncPtrN.apply` method takes additional implicit type evidence arguments `T: scalanative.unsafe.Tag` for each of function arguments and its return type. They're used to correctly box/unbox call arguments and returned value.
- generating `CFuncPtr.$extern$forwarder` was moved from `NirGenStat` to `NirGenExpr`. 
- `CFuncRawPtr` was removed. 
- abstract class `CFuncPtrTag` was introduced. It contains basic common logic for all `CFuncPtrNTag` classes. It contains single abstract method `fromRawPtr` used to convert `Ptr[_]` to `CFuncPtrN` 
- All usages of `CFuncPtr` were replaced to use a new interop schema in either explicit on the implicit way.

NIR primitive `CFuncPtrApply` apply is now used to call `CFuncPtr` underlying function pointer. It was introduced in order to support `CFuncPtr` created from `Ptr[_]` in an unsafe manner. Based on passed implicit evidence arguments method arguments are unboxed before calling the raw function pointer and the returned value is being boxed. 

Resolves #1631 
Resolves #1850 
Partially resolves #1889


WojciechMazur:
I was thinking about renaming method `Ptr.ptrToCFuncPtr` to `Ptr.toCFuncPtr`, but I've opted after staying with already defined naming convention which includes `Ptr.ptrToCStruct` and `Ptr.ptrToCArray`

unit-tests/src/test/scala/scala/scalanative/unsafe/CFuncPtrOpsSuite.scala
```diff
@@ -14,4 +14,10 @@ object CFuncPtrOpsSuite extends tests.Suite {
       randFunc.asInstanceOf[CFuncPtr1[Int, Int]] // wrong arity
     }
   }
+
+  test("CFuncPtr.toPtr casts to Ptr[Byte]") {
+    assertTrue {
+      randFunc.toPtr.isInstanceOf[Ptr[Byte]]
```
sjrd, line 20: Testing that the result of a method that returns a `Ptr[Byte]` is an instance of a `Ptr[Byte]` is redundant, and does not actually anything beside the fact that the method does not crash.

End-to-end tests for conversions should always test that we can actually *do* something meaningful with the result.

This test does not prevent me from implementing `toPtr` as always returning `unsafe.alloc[Byte]`. A meaningful test would invalidate that implementation.

unit-tests/src/test/scala/scala/scalanative/unsafe/PtrOpsSuite.scala
```diff
@@ -24,4 +24,23 @@ object PtrOpsSuite extends tests.Suite {
...
+    type ExpectedFnType = CFuncPtr0[CInt]
+    type OtherFnType    = CFuncPtr1[CInt, CString]
+    val funcPtr: Ptr[Byte] = new ExpectedFnType {
+      override def apply(): CInt = stdlib.rand()
+    }.toPtr
+
+    assertTrue {
+      Ptr
+        .ptrToCFuncPtr[ExpectedFnType](funcPtr)
+        .isInstanceOf[ExpectedFnType]
```
sjrd, line 38: Same comment about the meaning of the test. We should prevent the implementation from returning a dummy function or something like that.

nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
```diff
@@ -10,7 +10,11 @@
 package scala.scalanative
 package unsafe
 
-sealed trait CFuncPtr
+import scala.scalanative.runtime.Boxes
+
+sealed trait CFuncPtr {
+ def toPtr: Ptr[Byte] = Boxes.boxToPtr[Byte](Boxes.unboxToCFuncRawPtr(this))
```
sjrd, line 16: I am a bit wary of adding custom members to this trait which is a bit special. Consider defining `Ptr.cFuncPtrToPtr` instead.

ekrich:
I like your naming suggestion about. We have toCString so toCFuncPtr seems to be more regular.

unit-tests/src/test/scala/scala/scalanative/unsafe/PtrBoxingSuite.scala
```diff
@@ -182,4 +182,31 @@ object PtrBoxingSuite extends tests.Suite {
...
+      x._2 = stringLength
+
+      val loadedGetInt       = x._1
+      val loadedStringLength = x._2
+
+      val testStr = toCString("hello_native")
+      assertEquals(getInt(), x._1.apply())
+      assertEquals(getInt(), loadedGetInt())
+      assertEquals(stringLength(testStr), x._2.apply(testStr))
+      assertEquals(stringLength(testStr), loadedStringLength(testStr))
```
sjrd, line 209: Always use known results on the left. If you compare the results of two different things you're testing, they could both be wrong in the same way, and the test would pass:
```suggestion
      assertEquals(42, getInt())
      assertEquals(42, x._1.apply())
      assertEquals(42, loadedGetInt())
      assertEquals(12, stringLength(testStr))
      assertEquals(12, x._2.apply(testStr))
      assertEquals(12, loadedStringLength(testStr))
```

nativelib/src/main/scala/scala/scalanative/unsafe/Tag.scala.gyb
```diff
@@ -244,7 +250,45 @@ object Tag {
...
+
+	% for N in range(0, 23):
+	%   targs 		= ", ".join(["T" + str(i) 											for i in range(1, N+1)] + ["R"])
+	%   tArgsDecl = ", ".join(["T" + str(i) + " : c.WeakTypeTag" 	for i in range(1, N+1)] + ["R: c.WeakTypeTag"])
+	%   tps	 			= ", ".join(["$T" + str(i) 											for i in range(1, N+1)] + ["$R"])
+	%   tpsDef		= "\n".join(["val T" + str(i) + " = weakTypeOf[T" + str(i) + "].dealias" for i in range(1, N+1)])
+	%   args		  = ", ".join(["$fnPtr"] + ["arg" + str(i) 									for i in range(1, N+1)])
+	%   argsDecl  = ", ".join("arg" + str(i) + ": $T" + str(i) 							for i in range(1, N+1))
+	%   CFuncPtr  = "CFuncPtr{}[{}]".format(N, tps)
+    def materializeCFuncPtr${N}[${tArgsDecl}]: c.Tree = {
```
sjrd, line 278: My main comment at this point is here: why do we need macros for this? Why can't we write user-level code to define the `Tag`s of `CFuncPtr`s?

nativelib/src/main/scala/scala/scalanative/unsafe/Ptr.scala
```diff
@@ -62,4 +60,12 @@ object Ptr {
...
   @alwaysinline implicit def ptrToCStruct[T <: CStruct](ptr: Ptr[T])(
       implicit tag: Tag[T]): T = !ptr
+
+  @alwaysinline implicit def ptrToCFuncPtr[F <: CFuncPtr](ptr: Ptr[Byte])(
+      implicit tag: Tag.CFuncPtrTag[F]): F =
+    Boxes.boxToCFuncRawPtr(ptr.rawptr).as[F]
+
+  @alwaysinline implicit def cFuncPtrToPtr[T](ptr: CFuncPtr): Ptr[Byte] = {
+    Boxes.boxToPtr[Byte](Boxes.unboxToCFuncRawPtr(ptr))
+  }
```
sjrd, line 70: I don't think those two methods should be implicit. Not all `Ptr[Byte]`s are function pointers, especially not of an arbitrary `F`, and so that conversion should be explicit in the users' code.

This is fundamentally different from `ptrToCArray` and `ptrToCStruct` above: for them, the conversion is always safe when it typechecks, and hence it can be implicit.

sjrd: Additionally, I'm not sure this is the best place for these functions. `Ptr` seems too general a place for them. Consider declaring them in `CFuncPtr`, so that they would be called as `CFuncPtr.fromPtr(ptr)` and `CFuncPtr.toPtr(cFuncPtr)`, for example.

nativelib/src/main/scala/scala/scalanative/unsafe/Ptr.scala
```diff
@@ -62,4 +60,12 @@ object Ptr {
 
   @alwaysinline implicit def ptrToCStruct[T <: CStruct](ptr: Ptr[T])(
       implicit tag: Tag[T]): T = !ptr
+
+  @alwaysinline implicit def ptrToCFuncPtr[F <: CFuncPtr](ptr: Ptr[Byte])(
+      implicit tag: Tag.CFuncPtrTag[F]): F =
+    Boxes.boxToCFuncRawPtr(ptr.rawptr).as[F]
```
sjrd, line 66: `.as` is going to immediately unwrap the `CFuncRawPtr` that was created by `boxToCFuncRawPtr`. Consider inlining the body of `as` directly here, so that we don't need this transient wrapper.

This also seems to be the only use of `as`, so perhaps it should be removed altogether.

sjrd: Heck, this might be the only remaining use of `CFuncRawPtr` itself! If yes, then we could remove that one altogether, which would be a huge win!

sjrd (changes requested):

nativelib/src/main/scala/scala/scalanative/unsafe/Tag.scala.gyb
```diff
@@ -243,8 +247,21 @@ object Tag {
...
+	%   args		  = ", ".join(["$fnPtr"] + ["arg" + str(i) 			for i in range(1, N+1)])
+	%   argsDecl  = ", ".join("arg" + str(i) + ": $T" + str(i) 	for i in range(1, N+1))
+	%   CFuncPtrN 	  = "CFuncPtr{}[{}]".format(N, tps)
+	%   CFuncPtrNDecl = "CFuncPtr{}[{}]".format(N, tpsDecl)
+
+  @alwaysinline implicit def materialize${CFuncPtrNDecl}: CFuncPtrTag[unsafe.${CFuncPtrN}] = {
+		new CFuncPtrTag[unsafe.${CFuncPtrN}] {
+			override def load(ptr: unsafe.Ptr[unsafe.${CFuncPtrN}],
+												loadPtr: Boolean): unsafe.${CFuncPtrN} = {
+				val rawPtr = if (loadPtr) loadRawPtr(ptr.rawptr) else ptr.rawptr
```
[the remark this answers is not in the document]
WojciechMazur: I didn't like it either. Depending on context we can get here `Ptr[CFuncPtr]` or Ptr[Ptr[CFuncPtr]]`

In first case would be reached in following:
```
  val fn0: CFuncPtr0[CInt] = () => 1
  val fnPtr: Ptr[Byte] = CFuncPtr.toPtr(fn0)
  val fnFromPtr        = CFuncPtr.fromPtr[CFuncPtr0[CInt]](fnPtr)
```

and second when storing and loading from array/struct : 
```
  type GetInt = CFuncPtr0[CInt]
  type Functions = CStruct1[GetInt]
  val x: Ptr[Functions] = stackalloc[Functions]
  x._1 = CFuncPtr0.fromScalaFunction0(getInt) // x.unary_!(x)(<rhs>)
  val loadedGetInt: GetInt = x._1             // x.unary_!(x)
```

I'm not really sure which all them is not behaving corectlly

nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala
```diff
@@ -11,78 +11,213 @@
...
-  def apply(arg1: T1, arg2: T2, arg3: T3): R
+final class CFuncPtr2[T1, T2, R](val rawptr: RawPtr) extends CFuncPtr {
+  def apply(arg1: T1,
+            arg2: T2)(implicit ev1: Tag[T1], ev2: Tag[T2], evRet: Tag[R]): R =
+    intrinsic
 }
 
-// ###sourceLocation(file: "nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb", line: 17)
+object CFuncPtr2 {
+  implicit def fromScalaFunction2[T1, T2, R](
```
WojciechMazur, line 63: Can we change the implicit convertions methods naming to `fromScalaFunction`? It should be cleaner and compiler still should distinguish different methods since they have different argument lists

nativelib/src/main/scala/scala/scalanative/unsafe/Tag.scala.gyb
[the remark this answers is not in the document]
WojciechMazur: Actually, both of them are behaving as they should. In first case we're casting `Ptr[Byte]` containg pointer to function to `CFuncPtr`. In the seconds case we're loading a pointer containing a pointer to function. 
I think we can either add comment why we're not loading underlying pointer or add an additional method in CFuncPtrTag, eq. `def cast(ptr: Ptr[Byte]]): CFuncPtrN  = CFuncPtrN.fromRawPtr(ptr.rawptr)`

nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
```diff
@@ -10,13 +10,41 @@
...
+%   FunctionN    = "Function{}[{}]".format(N, allTps)
 
-trait CFuncPtr${N}[${", ".join(["T" + str(i) for i in range(1, N+1)] + ["R"])}] extends CFuncPtr {
-  def apply(${args}): R
+final class ${CFuncPtrN}(val rawptr: RawPtr) extends CFuncPtr {
+  def apply(${args})(implicit ${evidences}): R = intrinsic
 }
 
+object CFuncPtr${N}{
+  implicit def fromScalaFunction[${allTps}](fn: ${FunctionN}): ${CFuncPtrN} = intrinsic
```
sjrd, line 44: I wonder whether we should also take `evidence`s of `Tag`s here as well. We might not need them now in the codegen, but I'm not sure that will never be the case. Also, even if we don't use them, it would act as a nice compile-time check that the function's parameter types are static enough (if they're generic type params, the codegen would not do the right thing, I believe, and that would be caught by no Tag being found).

sjrd (changes requested):
This is very nice now.

I only have some cleanup comments, at this point.

nativelib/src/main/scala/scala/scalanative/runtime/Boxes.scala.gyb
```diff
@@ -0,0 +1,73 @@
...
+
+% for N in range(0, 23):
+%   args 				 = ", ".join("arg" + str(i) + ": T" + str(i) for i in range(1, N+1))
+%   allTps    	 = ", ".join(["T" + str(i) for i in range(1, N+1)] + ["R"])
+%   evidences    = ", ".join(["ev{}: Tag[T{}]".format(i, i) for i in range(1, N+1)] + ["evRet: Tag[R]"])
+%   CFuncPtrN    = "CFuncPtr{}[{}]".format(N, allTps)
+  @inline def boxTo${CFuncPtrN}(v: RawPtr): ${CFuncPtrN} =
+    if (v == null) null else CFuncPtr${N}.fromRawPtr[${allTps}](v)
+  @inline def unboxTo${CFuncPtrN}(o: java.lang.Object): RawPtr =
+    if(o == null) null else o.asInstanceOf[CFuncPtr].rawptr
```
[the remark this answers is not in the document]
WojciechMazur: There are no usages in code, but they're actually used when generating `Op.Box` / `Op.Unbox` All methods in this file are used there. If we want to revert to the previous approach we need to restore `CFuncRawPtr` to which all `CFuncPtr` was boxed/unboxed.

nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
WojciechMazur: Should we also move it to generic `object CFuncPtr`? It would allow doing single import to allow implicit conversion of all method types

sjrd: No, I don't think so. You typically won't need any import anyway, because if the expected type is `CFunctPtr2`, then the `implicit`s in the `object CFunctPtr2` are in scope without import.

nativelib/src/main/scala/scala/scalanative/runtime/Boxes.scala.gyb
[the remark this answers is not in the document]
sjrd: Ah OK, that makes sense. Let's keep this as is, then.

javalib/src/main/scala/java/lang/System.scala
```diff
@@ -92,10 +92,11 @@ object System {
...
   private val systemProperties = loadProperties()
-  Platform.setOSProps(new CFuncPtr2[CString, CString, Unit] {
-    def apply(key: CString, value: CString): Unit =
-      systemProperties.setProperty(fromCString(key), fromCString(value))
-  })
+  Platform.setOSProps { (key: CString, value: CString) =>
+    {
+      val _ = systemProperties.setProperty(fromCString(key), fromCString(value))
+    }
+  }
```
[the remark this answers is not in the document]
WojciechMazur: I've removed te braces, but we need to explicitly return `Unit`, either via `val _ = ???` or `???; ()`

docs/user/interop.rst
```diff
@@ -197,15 +197,33 @@ One can declare it as follows in Scala Native:
 
     def test(f: unsafe.CFuncPtr1[CString, Unit]): Unit = unsafe.extern
 
-`CFuncPtrN` types are SAM (single abstract method) traits. You
-can define them by creating a class that inherits from the corresponding
-trait:
+`CFuncPtrN` types are final classes containing pointer to underlying
+C function pointer. They automatically handle boxing call arguments
+and unboxing result. You can create them from C pointer using `Ptr` helper methods:
```
sjrd, line 202: Is this doc still accurate? Isn't it `CFunctPtrN` now?

docs/user/interop.rst
```diff
@@ -197,15 +197,33 @@ One can declare it as follows in Scala Native:
...
 .. code-block:: scala
 
-   val myfuncptr = new unsafe.CFuncPtr0[Unit] {
-     def apply(): Unit = println("hi there!")
-   }
+    def fnDef(str: CString): CInt = ???
+
+    val anyPtr: Ptr[Byte] = Ptr.cFuncPtrToPtr {
+      CFuncPtr1.fromScalaFunction(fnDef)
+    }
```
sjrd, line 210: Is this still accurate?
```suggestion
    val anyPtr: Ptr[Byte] = CFuncPtr1.toPtr {
      CFuncPtr1.fromScalaFunction(fnDef)
    }
```
?

Perhaps the implicit conversion even kicks in now, so we could write:
```suggestion
    val anyPtr: Ptr[Byte] = CFuncPtr1.toPtr(fnDef(_))
```
?

docs/user/interop.rst
```diff
@@ -197,15 +197,33 @@ One can declare it as follows in Scala Native:
...
+It's also possible to create `CFuncPtrN` from Scala `FunctionN`.
+You can do this by using implicit method conversion method
+from the corresponding companion object.
+
+.. code-block:: scala
+   import scalanative.unsafe.CFuncPtr0
+   def myFunc(): Unit = println("hi there!")
+
+   val myFuncPtr: CFuncPtr0[Unit] = CFuncPtr0.fromScalaFunction(myFunc)
+   val myImplFn: CFuncPtr[Unit] = myFunc _
```
sjrd, line 225:
```suggestion
   val myImplFn: CFuncPtr0[Unit] = myFunc _
```

nativelib/src/main/scala/scala/scalanative/runtime/Boxes.scala.gyb
```diff
@@ -0,0 +1,73 @@
...
+  @inline def boxToCVarArgList(v: RawPtr): CVarArgList =
+    if (v == null) null else new CVarArgList(v)
+  @inline def unboxToCVarArgList(o: java.lang.Object): RawPtr =
+    if (o == null) null else o.asInstanceOf[CVarArgList].rawptr
+
+% for N in range(0, 23):
+%   args 				 = ", ".join("arg" + str(i) + ": T" + str(i) for i in range(1, N+1))
+%   allTps    	 = ", ".join(["T" + str(i) for i in range(1, N+1)] + ["R"])
+%   evidences    = ", ".join(["ev{}: Tag[T{}]".format(i, i) for i in range(1, N+1)] + ["evRet: Tag[R]"])
+%   CFuncPtrN    = "CFuncPtr{}[{}]".format(N, allTps)
```
sjrd, line 66:
```suggestion
%   args         = ", ".join("arg" + str(i) + ": T" + str(i) for i in range(1, N+1))
%   allTps       = ", ".join(["T" + str(i) for i in range(1, N+1)] + ["R"])
%   evidences    = ", ".join(["ev{}: Tag[T{}]".format(i, i) for i in range(1, N+1)] + ["evRet: Tag[R]"])
%   CFuncPtrN    = "CFuncPtr{}[{}]".format(N, allTps)
```

nativelib/src/main/scala/scala/scalanative/runtime/Boxes.scala.gyb
sjrd, line 70:
```suggestion
    if (o == null) null else o.asInstanceOf[CFuncPtr].rawptr
```

sjrd (approved):

docs/user/interop.rst
WojciechMazur: The second one would not work right now, because it should use `CFuncPtr.toPtr(_: CFuncPtr)`. For some reason, the compiler does not try to make implicit conversion here.  
Or maybe there is a cleaner way to write following, which would work 
```
CFuncPtr.toPtr {
      val x: CFuncPtr[CString, CInt] = fnDef _  //would be converted using implicit converison
      x
}
```

sjrd (approved):

sjrd:
@WojciechMazur Could you write up a good commit message that reflects the final changes of this PR? The PR description is too loaded with all the changes that went into the design.

PR #1866, opened 2020-07-29
Fix runtime/tests with disabled Interflow

errikos:
Fix a `NullPointerException` that is thrown at runtime if compilation is done with the Interflow optimiser switched off.

The `NullPointerException` is thrown by the `snapshot` methods that can be found in the various implementations (`ByteArray`, `IntArray`, etc.) of the `Arrays` module in `nativelib`.

In particular, during the lowering phase, the `arrayalloc` operation generation would pass a `Val.Null` as the `this` pointer for the `snapshot` method. During Interflow, the `snapshot` method is always inline, so the problem would not reveal itself.

This PR introduces the following changes:
- Load and pass the correct object reference as `this` pointer in `arrayalloc` operation generation during lowering.
- Introduce extra runs in the CI, with Interflow disabled, to make sure we don't regress in the future.
- To accomplish the above, introduce a new boolean environment variable called `SCALANATIVE_OPTIMIZE` with valid values `true` and `false`, with the obvious meaning.
- Fix some errors in tests that surfaced with Interflow disabled:
  - In `IdentityHashMapSuite`, we used to test for primitive types as keys of the map. These tests are not correct for an `IdentityHashMap`, as each primitive is boxed to a different value and thus lookup would fail even if the underlying primitive is a valid key.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,20 @@ env:
   matrix:
     - MODE=source-checks
     - MODE=test-tools
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_GC=boehm TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_GC=immix TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_GC=commix TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=boehm TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=immix TEST_COMMAND="test-runtime"
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=commix TEST_COMMAND="test-runtime"
+    # without the Interflow optimiser
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=boehm TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=immix TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=commix TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=boehm TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=immix TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=commix TEST_COMMAND="test-runtime"
+    # with the Interflow optimiser
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_OPTIMISE=true SCALANATIVE_GC=boehm TEST_COMMAND="test-runtime"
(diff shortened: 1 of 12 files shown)


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -51,6 +51,9 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeDump =
       settingKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
+
+    val optimise =
```
lolgab, line 55: Have you considered the American spelling `optimize`?

ekrich:
I worked with @sjrd on IdentityHashMap in Scala.js so hopefully we can get those implementations and tests back into Native at some point. My tests and implementation in Native were not as good so thanks for the fixes.

.travis.yml
```diff
@@ -15,12 +15,20 @@ env:
...
-    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_GC=commix TEST_COMMAND="test-runtime"
+    # without the Interflow optimiser
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=boehm TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=immix TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=debug SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=commix TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=boehm TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=immix TEST_COMMAND="test-runtime"
+    - MODE=test-runtime TARGET_DOCKER_PLATFORM=library SCALANATIVE_MODE=release-fast SCALANATIVE_OPTIMISE=false SCALANATIVE_GC=commix TEST_COMMAND="test-runtime"
```
sjrd, line 24: Do you think it's important to test every GC in this configuration? If not I would suggest to only test e.g. immix in the "without the Interflow optimizer" category, for the sake of not exploding the CI time.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
sjrd: I agree. Usually we use the American spelling of things in CS.

sjrd: Also, now is perhaps not a good time to add yet another settingKey, when we're in the process of merging them all in one `nativeConfig`.

Perhaps for now we can just do the `System.getenv` directly when building the `build.Config`.

unit-tests/src/main/scala/tests/Suite.scala
```diff
@@ -88,15 +88,16 @@ abstract class Suite {
...
-    tests += Test(name, { () =>
-      try {
-        body
-        TestResult(true, None)
-      } catch {
-        case thrown: Throwable => TestResult(false, Option(thrown))
-      }
-    })
+  def test(name: String, cond: Boolean = true)(body: => Unit): Unit =
+    if (cond)
```
sjrd, line 92: Consider explicitly writing the braces for this long if branch:
```suggestion
    if (cond) {
```

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
errikos: Given the choice, I usually prefer the British spelling.

(We also have other references to "optimiser" in the docs.)

cc @sjrd

.travis.yml
ekrich: @LeeTibbert and I had a discussion about doing a sparse matrix of tests hoping that we could add `macOS` back into the mix. It was dropped because there were so many failures related to running of Unit tests on `macOS`.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
ekrich: Maybe `Discover.optimize()` to central these env vars to be used by other build tools as well. Hate for everyone to duplicate logic

.travis.yml
ekrich: @LeeTibbert and I had a discussion about using a sparse matrix. We previously dropped `macOS` because the `unit-tests` failed to much but it would be great to get that back. We could also add additional architectures like `arm`, `32bit`, Windows or whatever is most important.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
errikos: @ekrich I like that as a solution.

docs/contrib/build.rst
```diff
@@ -103,6 +103,10 @@ whereas the `release` mode performs additional optimizations and takes longer
 to compile. The `release-fast` mode builds faster, performs less optimizations,
 but may perform better than `release`.
 
+The `optimize` setting is controlled via the `SCALANATIVE_OPTIMISE` environment
```
sjrd, line 106: One that was forgotten:
```suggestion
The `optimize` setting is controlled via the `SCALANATIVE_OPTIMIZE` environment
```

errikos: Ah, I did case sensitive search and replace.

PR #1875, opened 2020-08-12
Rework Properties tests to remove ordering problems

ekrich:
@errikos Could you try this to see if it works for `2.12`?

### DIFF
--- a/unit-tests/src/test/scala/java/util/PropertiesTest.scala
+++ b/unit-tests/src/test/scala/java/util/PropertiesTest.scala
@@ -5,6 +5,7 @@ package java.util
 
 import java.io._
 import java.{util => ju}
+import java.nio.charset.StandardCharsets
 
 import org.junit.Test
 import org.junit.Ignore
@@ -329,48 +330,6 @@ class PropertiesTest {
                  prop.store(null: OutputStream, ""))
   }
 
-  def storeStream(props: Properties,
-                  header: String = ""): ByteArrayOutputStream = {
-    val out = new ByteArrayOutputStream()
-    props.store(out, header)
-    out
-  }
-
-  def storeWriter(props: Properties,
-                  header: String = ""): ByteArrayOutputStream = {
-    val out = new ByteArrayOutputStream()
-    props.store(new OutputStreamWriter(out), header)
-    out.close()
-    out
-  }
-
-  def loadByteArrayOutputStream(out: ByteArrayOutputStream): Properties = {
-    val props = new Properties()
-    props.load(new ByteArrayInputStream(out.toByteArray))
-    props
-  }
-
-  def assertAll(expected: Properties, actual: Properties): Unit = {
-    val e = expected.propertyNames()
-    while (e.hasMoreElements) {
-      val nextKey = e.nextElement().asInstanceOf[String]
-      assertEquals(actual.getProperty(nextKey), expected.getProperty(nextKey))
-    }
-  }
-
-  def loadStream(in: String): Properties = {
-    val prop = new java.util.Properties()
-    prop.load(new ByteArrayInputStream(in.getBytes("ISO8859_1")))
-    prop
-  }
-
-  def loadReader(in: String): Properties = {
-    val prop = new java.util.Properties()
-    prop.load(
-      new InputStreamReader(new ByteArrayInputStream(in.getBytes("UTF-8"))))
-    prop
-  }
-
   @Test def storeOutputStreamCommentsLoadInputStreamRoundtrip(): Unit = {
     val prop1 = new Properties()
     prop1.put("Property A", " aye\\\f\t\n\r\b")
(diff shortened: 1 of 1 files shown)


unit-tests/src/test/scala/java/util/PropertiesTest.scala
```diff
@@ -180,40 +180,40 @@ class PropertiesTest {
   @Test def non_string_values(): Unit = {
     val properties = new Properties
 
-    properties.put("age", Int.box(18))
+    properties.put("age", new Integer(18))
```
sjrd, line 183: The correct way to do this is in fact
```suggestion
    properties.put("age", Integer.valueOf(18))
```

ekrich:
The output formats are the only ones that are tested individually, that they do the proper escapes. The tests did work fine previously if the order from the hashmap was stable but I thought it was weird to sort the results. The same properties are checked for every entry on roundtrips.

The file load tests look at all sorts of entries with continuing values etc. I also used MUnit to test against the JVM (where I found trailing whitespace is significant after `\` ) so I am pretty confident the implementation and tests from Harmony are solid.  I didn't do the original development or ports of the tests either.

There could be tests that are missing, and of course there is no substitute for battle testing.

sjrd:
I know that the tests worked before and that therefore the implementation is correct today. The point of tests is to prevent the implementation from becoming doing in the future. If we decrease test coverage today, we open the door to introducing bugs in the implementation in the future. We should not do that.

ekrich:
Right, but I didn't remove any test I just broke the `list()` test up in steps to test each element rather than all together. Since the implementation uses an unordered list, then you could argue that the old test was invalid as it failed on Scala.js and 2.12 for ordering differences.

Anyway, we are testing the list method which has little on the specification side and as it says _This method is useful for debugging._
https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#list-java.io.PrintStream-https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#list-java.io.PrintStream-
The tests look for the header, a normal property, and one that is too long. These must have been added by checking the JVM output and just writing the rules for those three cases.

Except for the `Integer` change, all the other changes where just refactoring out the common bits to make things shorter and easier to read.

unit-tests/src/test/scala/java/util/PropertiesTest.scala
```diff
@@ -358,32 +349,51 @@ class PropertiesTest {
...
+
+  def storeWriter(props: Properties,
+                  header: String = ""): ByteArrayOutputStream = {
+    val out = new ByteArrayOutputStream()
+    props.store(new OutputStreamWriter(out), header)
+    out.close()
+    out
+  }
+
+  def load(out: ByteArrayOutputStream): Properties =
```
sjrd, line 367: This doesn't seem to be used.

unit-tests/src/test/scala/java/util/PropertiesTest.scala
```diff
@@ -358,32 +349,51 @@ class PropertiesTest {
...
+    load(new ByteArrayInputStream(out.toByteArray))
+
+  def load(in: ByteArrayInputStream): Properties = {
```
sjrd, line 370: It seems you're always calling this method with a `new ByteArrayInputStream(someBytes)`. You could take the `bytes: Array[Byte]`s directly as an argument to avoid that repetition.

Further, you could even take a `String`, since in virtually all cases you pass in `someString.getBytes()`. Speaking of which, you should use `.getBytes(StandardCharsets.ISO_8859_1)` not to depend on the system charset.

unit-tests/src/test/scala/java/util/PropertiesTest.scala
```diff
@@ -358,32 +349,51 @@ class PropertiesTest {
...
+    val props = new Properties()
+    props.load(in)
+    in.close()
+    props
+  }
+
+  def assertAll(props1: Properties, props2: Properties): Unit = {
```
sjrd, line 377: Consider using more descriptive names:
```suggestion
  def assertAll(expected: Properties, actual: Properties): Unit = {
```

unit-tests/src/test/scala/java/util/PropertiesTest.scala
```diff
@@ -358,32 +349,51 @@ class PropertiesTest {
...
+    val e = props1.propertyNames()
+    while (e.hasMoreElements) {
+      val nextKey = e.nextElement().asInstanceOf[String]
+      assertEquals(props2.getProperty(nextKey), props1.getProperty(nextKey))
+    }
+  }
+
   // used for next two tests, \b prints as \u0008
   val prop1 = new Properties()
```
sjrd, line 386: This now only used in `store_OutputStream_comments_load_InputStream_roundtrip()`. Consider moving it inside that method, along with `out1`. And then redeclare another `out1` based on some trivial `Properties` in the test `check_comment_formatted_correctly()` (or merge that test with the previous one to avoid even that).

ekrich:
Thanks for the suggestions. At least everything can be cleaned up a bit and your idea to rewrite I think turned out well and potentially is much easier to read and maintain. Compared to when I took over the code and you first reviewed, it is much much better.

errikos:
This seems to work now with Scala 2.12, thanks @ekrich !

ekrich:
@errikos I'll get it fixed up ASAP.

unit-tests/src/test/scala/java/util/PropertiesTest.scala
ekrich: OK

unit-tests/src/test/scala/java/util/PropertiesTest.scala
ekrich: I used this suggestion which really helped.

unit-tests/src/test/scala/java/util/PropertiesTest.scala
ekrich: Done, thank you.

unit-tests/src/test/scala/java/util/PropertiesTest.scala
ekrich: I fixed most everything I could see up. I also made the test names consistent with toggle case.

ekrich:
@sjrd I think this looks much better now. I was probably rushing a bit to help out the 2.12 effort but in the end fixing this up is much better for the long term maintainability.

PR #1876, opened 2020-08-15
Fix #1970: Restrict native code to a specified subdirectory

ekrich:
The following features are included:

- Restricts detection of native libraries with code only in `classes/scala-native` if in a directory.
- Restricts detection of native libraries with code only in `scala-native` if in a jar.
- Once the jar is unpacked or the directory is copied into the generated directory it looks for the following pattern to include the native files:
-- `glob:**/native/*/scala-native.**{.c, .cpp, .S}` 
-- Any Path to `native/generated-dir-0/scala-native/` with any trailing path.
-- Copies all files with one directory pass vs. 3 for previous code.
- Once code is compiled it again finds object files by restricting path.
-- The `scalalib` compiles everything and then deletes the `.o` files of extra GCs and optional.
-- `glob:**/native/*/scala-native.**.o`
-- Any Path to `native/generated-dir-0/scala-native/` with any trailing path.
-- Again only one pass.

This works for many libraries all at once.

Edit: since `val workdir = crossTarget.value / "native"` is determined in the plugin but is a param `workdir` to `Config`, the code has been made so the directory can vary.

Note: source detection in `crossTarget.value / "classes"` is not controlled by the build so `"classes"` is a constant.


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -80,6 +80,8 @@ object ScalaNativePluginInternal {
       crossTarget.value / (moduleName.value + "-out")
     },
     nativeWorkdir := {
+      // all build tools should have the workdir
+      // ending in `native` for users' consistency
```
sjrd, line 84: I don't understand this.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -69,6 +69,6 @@ object Build {
       LLVM.compile(config, generated)
     }
 
-    LLVM.link(config, linked, objectFiles, unpackedLibs, outpath)
+    LLVM.link(config, linked, objectFiles, outpath)
```
sjrd, line 72: `unpackedLibs` is not used anymore. That seems very fishy to me.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -1,27 +1,95 @@
...
+import java.io.File
 import java.nio.file.Path
+import java.util.regex._
 
 /** Original jar or dir path and generated dir path for native code */
 private[scalanative] case class NativeLib(src: Path, dest: Path)
 
+/** Utilities for dealing with native library code */
 private[scalanative] object NativeLib {
+  private val fs           = File.separator
```
sjrd, line 13: I read this as "file system", not as "file separator". Consider renaming it as `fileSep`.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -1,27 +1,95 @@
...
+  val srcExtensions = Seq(".c", cppExt, ".S")
+
+  /**
+   * Name of directory that contains native code: "scala-native"
+   */
+  val codeDir = "scala-native"
+
+  /** Used to find native source files in directories */
+  val srcPatterns = srcExtensions.mkString(
+    s"glob:**${fs}classes${fs}${codeDir}${fs}**{",
```
sjrd, line 32: The fact that `classes` is here is worrying. The tools shouldn't know about `classes`, which is sbt-specific. We should receive the classpath from the build tool, which would give us entries that end in `classes` if it's sbt, and look into those directories (for each one, in the subdirectory `scala-native`), but nowhere else.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -1,27 +1,95 @@
...
+   * into the `native` directory and also in the `scala-native`
+   * sub directory gets picked up for compilation.
+   *
+   * @param workdir the base working directory
+   * @return the source pattern
+   */
+  def destSrcPatterns(workdir: Path): String = {
+    val workdirName = workdir.getFileName()
+    srcExtensions.mkString(
+      s"glob:**${fs}${workdirName}${fs}*${fs}${codeDir}${fs}**{",
```
sjrd, line 53: This pattern looks weird. Why does it only take the file name part of `workdir` into account? It should use the full path. Or it shouldn't even be part of the patterns, but the search should be restricted to be inside `workdir`.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -1,27 +1,95 @@
...
+  }
 
   /** To positively identify nativelib */
-  val nativeLibMarkerFile = "org_scala-native_nativelib.txt"
+  private val nativeLibMarkerFile = "org_scala-native_nativelib.txt"
 
-  val dirMarkerFilePattern = "glob:**" + nativeLibMarkerFile
+  /** Note: assumes that code is compiled into the `classes` directory */
+  private val dirMarkerFilePattern =
+    s"glob:**${fs}classes${fs}" + nativeLibMarkerFile
```
sjrd, line 75: Another `classes` that should not be there.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
ekrich: I put that in there as a suggestion. If you use `sbt` the code goes in a `native` directory. If `mill` for example also put it in a `native` then it would be easier on the developers to find the code.

tools/src/main/scala/scala/scalanative/build/Build.scala
ekrich: All the unpacked libs are in the same directory so the `glob` looks in this case for `native/*/scala-native` so the individual paths are not needed. I could possibly build the pattern for the `unpackedLibs` such as the following:
`{"dir-0", "dir-1", ...}` - perhaps that is better. I would like to avoid traversing the the directory more than once.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: Ok sure thing.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: Right, I didn't like that either. Somehow I need to get the path to the `classes` dir no matter what it is called. Could we add that to `Config` so it is not essentially hard coded?

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: Yes, I think I could do that and remove the `**` from the front of the `glob`. I will try that.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: RIght, but that is more specific. I don't expect anyone to have a text file with that name. This is related to the discussion above about `classes`.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: This is fixed too to use full path and only the unpacked dirs which is basically the correct solution.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
ekrich: I can just remove that or reword it somehow.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: You are correct, the classpath solves all the problems. Fixed.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: I use the full path here as well to be explicit. It could be `s"glob:**${nativeLibMarkerFile}"` but then we can drop the `**`.

ekrich:
I am still working one issue I noticed revolving around hashing.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: Just using the classpath worked. Thanks for the suggestion.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
ekrich: Fixed as well.

ekrich:
@sjrd Would you please look at this again when you get a chance?

sjrd (approved):
All right. I believe this is correct now. There's one tiny detail below, and it will need a rebase. After that it's good to go.

ekrich:
@sjrd I added another return type but then realized that the method should be private so I made that change too.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -1,27 +1,119 @@
...
+  private def srcPathPattern(path: Path): String =
+    s"${path.toString()}${fileSep}${codeDir}${fileSep}"
+
+  /**
+   * Used to create hash of the directory to copy
+   *
+   * @param path The classpath entry
+   * @return the file pattern
+   */
+  def allFilesPattern(path: Path) =
```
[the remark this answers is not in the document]
ekrich: Rebased and this change made as well.

sjrd (approved):

PR #1877, opened 2020-08-17
Add Scala 2.12 support

errikos:
### Summary
Add Scala 2.12 support for Scala Native. You can consult the Scala 2.12 changelog [here](https://www.scala-lang.org/news/2.12.0/). This PR takes into account all the changes and adapts them for Scala Native.

### Lambdas
In order to keep the NIR changes to a minimum, we continue generating a `FunctionN` class with an `apply` method for each lambda. All the lambda captures are stored inside the class instance and initialised via its constructor. The `apply` method calls the generated `$anonfun` method, also extracting and forwarding the captures.

### NIR compatibility
Scala 2.12 now compiles trait methods as default methods in Java interfaces. This means that the NIR compiled with 2.12 is not compatible with that compiled with 2.11.

### Scala 2.11 cross-compilation
In order to keep the project compilable with Scala 2.11, a `scala.scalanative.nscplugin.NirCompat` component is introduced in `nscplugin`. This allows the compiler plugin to function with both Scala 2.11 and 2.12.

In addition, some tests need to know under which Scala version they are running. We use `sbt-buildinfo` to provide Scala version information to the projects that need them (mainly for `unit-tests` and `tools` tests).

### DIFF
--- a/javalib/src/main/scala/java/lang/Boolean.scala
+++ b/javalib/src/main/scala/java/lang/Boolean.scala
@@ -43,7 +43,8 @@ final class Boolean(val _value: scala.Boolean) extends Comparable[Boolean] {
 }
 
 object Boolean {
-  final val TYPE           = scala.Predef.classOf[scala.scalanative.runtime.PrimitiveBoolean]
+  final val TYPE =
+    scala.Predef.classOf[scala.scalanative.runtime.PrimitiveBoolean]
   final val TRUE: Boolean  = new Boolean(true)
   final val FALSE: Boolean = new Boolean(false)
 
--- a/nir/src/main/scala/scala/scalanative/nir/Types.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Types.scala
@@ -125,11 +125,6 @@ object Type {
       false
   }
 
-  def isPtr(ty: Type): Boolean = ty match {
-    case Type.Ptr => true
-    case _        => false
-  }
-
   val typeToArray = Map[Type, Global](
     Type.Bool    -> Global.Top("scala.scalanative.runtime.BooleanArray"),
     Type.Char    -> Global.Top("scala.scalanative.runtime.CharArray"),
--- a/nir/src/main/scala/scala/scalanative/nir/Types.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Types.scala
@@ -125,6 +125,11 @@ object Type {
       false
   }
 
+  def isPtr(ty: Type): Boolean = ty match {
+    case Type.Ptr => true
+    case _        => false
+  }
+
   val typeToArray = Map[Type, Global](
     Type.Bool    -> Global.Top("scala.scalanative.runtime.BooleanArray"),
     Type.Char    -> Global.Top("scala.scalanative.runtime.CharArray"),
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
@@ -633,9 +633,12 @@ trait NirGenExpr { self: NirGenPhase =>
 
       // Generate fields to store the captures.
 
-      val captureTypes = captureSyms.map(sym => genType(sym.tpe))
+      // enclosing class `this` reference + capture symbols
+      val captureSymsWithEnclThis = curClassSym.get +: captureSyms
+
+      val captureTypes = captureSymsWithEnclThis.map(sym => genType(sym.tpe))
       val captureNames =
-        captureSyms.zipWithIndex.map {
+        captureSymsWithEnclThis.zipWithIndex.map {
           case (sym, idx) =>
             val name = anonName.member(nir.Sig.Field("capture" + idx))
             val ty   = genType(sym.tpe)
@@ -709,7 +712,7 @@ trait NirGenExpr { self: NirGenPhase =>
               curMethodEnv.enter(sym, unboxedOrCast)
           }
 
-          captureSyms.zip(captureNames).foreach {
+          captureSymsWithEnclThis.zip(captureNames).foreach {
             case (sym, name) =>
               val value = buf.fieldload(genType(sym.tpe), self, name, Next.None)
               curMethodEnv.enter(sym, value)
@@ -717,9 +720,9 @@ trait NirGenExpr { self: NirGenPhase =>
 
           val sym      = targetTree.symbol
           val method   = Val.Global(genMethodName(sym), Type.Ptr)
-          val values   = buf.genMethodArgs(sym, functionArgs)
+          val values   = buf.genMethodArgs(sym, Ident(curClassSym.get) +: functionArgs)
           val sig      = genMethodSig(sym)
-          val res      = buf.call(sig, method, Val.Null +: values, Next.None)
+          val res      = buf.call(sig, method, values, Next.None)
           val boxedRes = {
             val expectedResTy = genType(funSym.tpe.resultType)
             if (res.ty != expectedResTy) {
@@ -740,7 +743,7 @@ trait NirGenExpr { self: NirGenPhase =>
       // passing all of the captures as arguments.
 
       val alloc       = buf.classalloc(anonName, unwind)
-      val captureVals = captureSyms.map { sym => genExpr(Ident(sym)) }
+      val captureVals = curMethodThis.get.get +: captureSyms.map { sym => genExpr(Ident(sym)) }
       buf.call(ctorTy, Val.Global(ctorName, Type.Ptr), alloc +: captureVals, unwind)
       alloc
     }
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
@@ -723,19 +723,15 @@ trait NirGenExpr { self: NirGenPhase =>
           val values   = buf.genMethodArgs(sym, Ident(curClassSym.get) +: functionArgs)
           val sig      = genMethodSig(sym)
           val res      = buf.call(sig, method, values, Next.None)
-          val boxedRes = {
-            val expectedResTy = genType(funSym.tpe.resultType)
-            if (res.ty != expectedResTy) {
-              res.ty match {
-                case _: Type.Array =>
-                  wrapAsArrayOps(res, buf)
-                case _ =>
-                  buf.boxValue(callTree.tpe, res)
-              }
-            } else {
-              res
-            }
+
+          // Get the result type of the lambda after erasure, when entering posterasure.
+          // This allows to recover the correct type in case value classes are involved.
+          // In that case, the type will be an ErasedValueType.
+          val resTyEnteringPosterasure = enteringPhase(currentRun.posterasurePhase) {
+            targetTree.symbol.tpe.resultType
           }
+
+          val boxedRes = ensureBoxed(res, resTyEnteringPosterasure, callTree.tpe)(buf)
           buf.ret(boxedRes)
           buf.toSeq
         }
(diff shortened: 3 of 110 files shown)


sjrd:
For the issue with `Array`s, I think the underlying issue is that you're not handling value classes correctly in `genFunction`. The problem is that you need to recover value classes in parameters and result types, and box/unbox them appropriately. You'll have to replicate the uses of `enteringPhase(currentRun.posterasurePhase) { ... }` and handling of `ErasedValueType`s that we do in Scala.js. Some pointers:

* [`ensureBoxed` and `fromAny`](https://github.com/scala-js/scala-js/blob/de973d6dbae77720565c2ac46cc54df462077924/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala#L4786-L4833), which deal with the `ErasedValueType`s
* [one example of where we use `enteringPhase`](https://github.com/scala-js/scala-js/blob/de973d6dbae77720565c2ac46cc54df462077924/compiler/src/main/scala/org/scalajs/nscplugin/GenJSCode.scala#L6069-L6080) to compute the correct `tpe`s to give to `ensureBoxed` and `fromAny`

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -76,14 +83,15 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
...
+      val name = todo.last
+      todo = todo.init
       if (!done.contains(name)) {
         reachDefn(name)
       }
     }
 
   def reachDefn(name: Global): Unit = {
-    stack.push(name)
+    stack :+= name
```
lolgab, line 94: Have you considered to use a `mutable.ListBuffer` instead of a `var` of a `List`?
Appending to a `List` like this has a linear cost and creates garbage.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -1618,14 +1902,14 @@ trait NirGenExpr { self: NirGenPhase =>
         }
       val args = genMethodArgs(sym, argsp)
       val method =
-        if (isImplClass(owner) || statically || owner.isStruct || owner.isExternModule) {
+        if (isImplClass(owner) || statically || owner.isStruct || owner.isExternModule || self.ty == nir.Type.Ptr) {
```
ekrich, line 1905: `self.ty == nir.Type.Ptr` Should this have a name like `isPointer` or something?

sandbox/Test.scala
```diff
@@ -1,5 +1,5 @@
 object Test {
   def main(args: Array[String]): Unit = {
-    println("Hello, World!")
+    println("Hello, world!")
```
ekrich, line 3: This might be changing the originally hello world from Scala Days and the first release.

scalalib/overrides-2.12/scala/App.scala
```diff
@@ -0,0 +1,77 @@
+/*
```
ekrich, line 1: Are the overrides just standard for `2.12` even in the Scala repo?

tools/src/main/scala/scala/scalanative/linker/Reach.scala
errikos: I did not like that either, but the deprecation note says:

> Stack is an inelegant and potentially poorly-performing wrapper around List. **Use a List assigned to a var instead.**

scalalib/overrides-2.12/scala/App.scala
errikos: Hmm, what do you mean?

sandbox/Test.scala
errikos: Yeah, will revert this.

tools/src/main/scala/scala/scalanative/linker/Reach.scala
lolgab: In that case you need to push on the left (using `stack ::= m`) instead of pushing on the right, so you don't recreate an entire new List from scratch on every push.

lolgab: Oh, by _that_ you meant using a `var` of a `List`? Now I understand your answer..
Well, I think that If you push on the head (as I said in the last comment) you get a pretty performant stack, but I wouldn't bet anything on it being better/worse than `mutable.ListBuffer` 😄

scalalib/overrides-2.12/scala/App.scala
ekrich: I am just asking a general question about why these are here. They are not changed for Scala Native are they?

errikos: Oh yeah, we leave them as is. Basically, where applicable, I applied the changelog from the `overrides-2.11` to the 2.12 files.

unit-tests/src/test/scala/tests/SuiteSuite.scala
```diff
@@ -66,7 +66,9 @@ object SuiteSuite extends Suite {
         }
 
       case exc: Throwable =>
-        if (exc == AssertionFailed) // got case object, expected case class
+        if (exc.isInstanceOf[
+              AssertionFailed
+            ]) // got case object, expected case class
```
sjrd, line 71: scalafmt does something horrible here. Consider moving the comment above the if or inside the branch so that this doesn't happen.

util/src/main/scala/scala/scalanative/util/Platform.scala
```diff
@@ -0,0 +1,5 @@
+package scala.scalanative.util
+
+object Platform {
+  def scalaVersion: String = scala.util.Properties.versionNumberString
```
sjrd, line 4: This is only used in the `tools`. Consider moving this `Platform` object to the tools, and make it package-private.

javalib/src/main/scala/java/lang/Boolean.scala
```diff
@@ -43,7 +43,8 @@ final class Boolean(val _value: scala.Boolean) extends Comparable[Boolean] {
 }
 
 object Boolean {
-  final val TYPE           = classOf[scala.Boolean]
+  final val TYPE =
+    scala.Predef.classOf[scala.scalanative.runtime.PrimitiveBoolean]
```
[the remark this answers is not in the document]
errikos: This was actually changed in the prototype branch by @densh.

Keeping `scala.Boolean` results in:
```
[error] Error while emitting Boolean.scala
[error] assertion failed:
[error]   expected class type in classOf[T], found primitive type Z
[error]      while compiling: /Users/ergys/dev/scala/native/javalib/src/main/scala/java/lang/Boolean.scala
[error]         during phase: jvm
[error]      library version: version 2.12.12
[error]     compiler version: version 2.12.12
[error]   reconstructed args: ...
[error]
[error]   last tree to typer: TypeTree(class Array)
[error]        tree position: line 78 of /Users/ergys/dev/scala/native/javalib/src/main/scala/java/io/DataOutputStream.scala
[error]             tree tpe: Array[Byte]
[error]               symbol: final class Array in package scala
[error]    symbol definition: final class Array[T] extends Object (a ClassSymbol)
[error]       symbol package: scala
[error]        symbol owners: class Array
[error]            call site: constructor UnixPath$$anon$1 in package fs

```

util/src/main/scala/scala/scalanative/util/Platform.scala
errikos: Yes, I have to fix this.

Basically, this should change with the Scala version of the linked code that is tested, so it can't go to `tools`. This means that it should probably be refactored to be obtained via `sbt-buildinfo`. However, I can't really find an appropriate place to move it to. The only cross-compiled project `tools` depends on is `testingCompilerInterface`, but I am not sure we can enable `sbt-buildinfo` there.

unit-tests/src/test/scala/java/util/ArrayDequeSuite.scala
```diff
@@ -109,7 +111,7 @@ object ArrayDequeSuite extends tests.Suite {
 
   test(s"addFirst(e)") {
     locally {
-      type E = Float
+      type E = AnyRef
```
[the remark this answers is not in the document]
errikos: The assertions below `assertThrows[NullPointerException]` fail when a primitive type is provided in 2.12. In Scala 2.11 they seem to succeed, but to be honest, I don't understand why, as e.g. `null.asInstanceOf[Float]` is `0.0` and does not throw a `NullPointerException`, both in 2.11 and 2.12.

We could also use `AnyVal` here or `Number`.

util/src/main/scala/scala/scalanative/util/Platform.scala
sjrd: We don't have *any* project that cross compiles. There's no such thing. A project is either for the JVM or for Native. Cross compilation is an illusion provided by having two projects with the same sources. But even that we don't have in this build.

But you really don't need that. What you need is, in tools, the value of `nativelib / scalaVersion`. That's easy to obtain with sbt-buildinfo enabled on the tools project.

errikos: But, if we enable `sbt-buildinfo` in `tools`, which is compiled with 2.12, then the `scalaVersion` in the generated sources will be 2.12. Am I missing something here?

nir/src/main/scala/scala/scalanative/nir/Types.scala
```diff
@@ -125,6 +125,11 @@ object Type {
       false
   }
 
+  def isPtr(ty: Type): Boolean = ty match {
```
WojciechMazur, line 128: Couldn't it be simply `ty.isInstanceOf[Type.Ptr]`?

ikamthania:
Hi. I was wondering if this pr is been under active review. Is there anything pending that community can take a stab at ?

WojciechMazur:
@ikamthania Hey, it's still under review. I hope we can merge it (as well as other remaining 0.4.0 blockers) and make a release

ikamthania:
> @ikamthania Hey, it's still under review. I hope we can merge it (as well as other remaining 0.4.0 blockers) and make a release by end of this month

Thanks for update. No issues. These things take time understandably.

nir/src/main/scala/scala/scalanative/nir/Types.scala
sideeffffect: `ty.isInstanceOf[Type.Ptr.type]`?

unit-tests/src/test/scala/tests/SuiteSuite.scala
```diff
@@ -66,10 +66,12 @@ object SuiteSuite extends Suite {
         }
 
       case exc: Throwable =>
-        if (exc == AssertionFailed) // got case object, expected case class
-          throw AssertionFailed("expected message yet none found")
-        else
-          throw exc
+        exc match {
+          case _: AssertionFailed =>
+            // got case object, expected case class
```
sjrd, line 71: The comment is in contradiction with the code. Testing for `case _: AssertionFailed` means that we got an instance of the class, not of the case object.

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -41,18 +43,23 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
...
+    // Optimizer and codegen may never increase reachability
+    // past what's known now, so it's safe to do this.
     infos.values.foreach {
       case cls: Class =>
-        val entries = cls.responds.toArray
-        entries.foreach {
+        cls.responds.foreach {
           case (sig, name) =>
             if (!done.contains(name)) {
               cls.responds -= sig
```
sjrd, line 54: It is not safe to mutate a collection while you're iterating on it. That was the purpose of the `.toArray` in the previous code. It should be kept.

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -41,18 +43,23 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
...
             }
         }
+        cls.defaultResponds.foreach {
+          case (sig, name) =>
+            if (!done.contains(name)) {
+              cls.defaultResponds -= sig
```
sjrd, line 60: Same problem here.

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -183,11 +191,33 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
...
+          case _ =>
+            util.unreachable
+        }
+        loaded(info.name).foreach {
+          case (_, defn: Defn.Define) =>
+            val Global.Member(_, sig) = defn.name
+
+            info.responds(sig) = defn.name
+
+          case _ =>
```
sjrd, line 218: I suggest removing the blank lines here:
```suggestion
            info.responds(sig) = defn.name
          case _ =>
```

sjrd (changes requested):
I finally dedicated enough continuous time to properly review this.

This looks very good. I only minor comments here and there. I'm looking forward to merging it :)

sjrd:
This will also need to be rebased on master to sit on top of #1912.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirPrimitives.scala
```diff
@@ -128,7 +128,6 @@ abstract class NirPrimitives {
     addPrimitive(ULongToFloatMethod, ULONG_TO_FLOAT)
     addPrimitive(UIntToDoubleMethod, UINT_TO_DOUBLE)
     addPrimitive(ULongToDoubleMethod, ULONG_TO_DOUBLE)
-    HashMethods.foreach(addPrimitive(_, HASH))
```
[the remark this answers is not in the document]
errikos: I am really not sure. `HashMethods` was removed in the first commit of the PR, authored by @densh.

scalalib/overrides-2.11/scala/runtime/ScalaRunTime.scala
```diff
@@ -201,7 +200,7 @@ object ScalaRunTime {
     val high = (lv >>> 32).toInt
     low ^ (high + lowSign)
   }
-  def hash(x: Number): Int  = runtime.BoxesRunTime.hashFromNumber(x)
+  def hash(x: Number): Int = x.hashCode
```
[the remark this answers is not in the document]
errikos: This is needed because of the removed `HashMethods` value mentioned below. I think that, as a result, the `hashFromNumber` method is not intrinsified anymore and the one in `BoxesRunTime` is not implemented.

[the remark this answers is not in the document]
errikos: I don't know why they were removed in the first place. Maybe we could only remove them for 2.12.x onwards.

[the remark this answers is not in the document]
sjrd: > I don't know why they were removed in the first place. Maybe we could only remove them for 2.12.x onwards.

That would probably make more sense.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -1617,14 +1840,14 @@ trait NirGenExpr { self: NirGenPhase =>
         }
       val args = genMethodArgs(sym, argsp)
       val method =
-        if (isImplClass(owner) || statically || owner.isStruct || owner.isExternModule) {
+        if (isImplClass(owner) || statically || owner.isStruct || owner.isExternModule || isPtr(self.ty)) {
```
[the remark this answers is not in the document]
errikos: I really don't remember! It seems that it's not needed, though! 😅

errikos:
There seems to be a test failing in `tools` due to recent changes, I will look into it.

WojciechMazur:
> There seems to be a test failing in `tools` due to recent changes, I will look into it.

From what I can see there is problem with order of mangled arguments for closures:
expected: `D6loop$1iL16java.lang.StringzL16java.lang.StringEPT6xyz.B$`
actual:      `D6loop$1zL16java.lang.StringiL16java.lang.StringEPT6xyz.B$`

And here's actual method: 
```
  final def encode(arg1: Int, arg2: String, arg3: Boolean): String = {
    def loop(): String = {
      if (arg3) "hello"
      else encodeLoop(arg1, arg2)
    }
    loop()
  }
```

I think that order in which compiler creates mangled names for closures have changed in 2.12. Since our signatures/mangled names are based on Scala compiler output we won't be able to change it.  

I would suggest to change test definition to expect method signature in both orders on arguments. I can provide patch with this tomorrow.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirPrimitives.scala
```diff
@@ -128,7 +128,16 @@ abstract class NirPrimitives {
...
+
+    {
```
errikos, line 132: I was not able to use the sbt-buildinfo plugin here to condition for the Scala version, because it resulted to an infinite recursion in sbt when resolving the projects that depend on `nscplugin` (tried to enable sbt-buildinfo for `nscplugin`).

I went with `nsc.settings.ScalaVersion` instead. I don't know if the latter has any version guarantees.

errikos:
This should be ready for another pass @sjrd. I think that all the comments are resolved and the CI seems to be green!

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirPrimitives.scala
```diff
@@ -128,7 +128,16 @@ abstract class NirPrimitives {
...
+      import scala.tools.nsc.settings._
+      ScalaVersion.current match {
+        case SpecificScalaVersion(2, 11, _, _) =>
+          HashMethods.foreach(addPrimitive(_, HASH))
```
errikos, line 136: I was not able to use the sbt-buildinfo plugin here to condition for the Scala version, because it resulted to an infinite recursion in sbt when resolving the projects that depend on `nscplugin` (tried to enable sbt-buildinfo for `nscplugin`).

I went with `nsc.settings.ScalaVersion` instead. I don't know if the latter has any version guarantees.

sjrd (approved):
Ship it!

Thank you @errikos for this huge work!

PR #1878, opened 2020-08-18
Store source code positions in NIR

WojciechMazur:
This PR adds basic support for embedding source code positions inside NIR format, which is first step in enabling debugging in Scala Native.

Positions metadata was added to `Defn` and `Inst` NIR objects via implicit parameters. Within generated code/transformations positions may be preserved if it's possible. In multiple cases code generated via Scala Native plugin contains mocked up Position. 
Serialization of Positions was ported from Scala.js

### DIFF
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
@@ -330,7 +330,6 @@ trait NirGenExpr { self: NirGenPhase =>
                     implicit exprPos: nir.Position): Val = {
       val cases = catches.map {
         case cd @ CaseDef(pat, _, body) =>
-          implicit val pos: nir.Position = cd.pos
           val (excty, symopt) = pat match {
             case Typed(Ident(nme.WILDCARD), tpt) =>
               (genType(tpt.tpe), None)
@@ -341,14 +340,14 @@ trait NirGenExpr { self: NirGenPhase =>
           }
           val f = { () =>
             symopt.foreach { sym =>
-              val cast = buf.as(excty, exc, unwind)
+              val cast = buf.as(excty, exc, unwind)(cd.pos)
               curMethodEnv.enter(sym, cast)
             }
             val res = genExpr(body)
             buf.jump(mergen, Seq(res))
             Val.Unit
           }
-          (excty, f, pos)
+          (excty, f, exprPos)
       }
 
       def wrap(cases: Seq[(nir.Type, () => Val, nir.Position)]): Val =
@@ -618,6 +617,7 @@ trait NirGenExpr { self: NirGenPhase =>
     def genFunction(tree: Function): Val = {
       val Function(paramTrees,
                    callTree @ Apply(targetTree @ Select(_, _), functionArgs)) = tree
+      implicit val pos: nir.Position = tree.pos
 
       val funSym    = tree.tpe.typeSymbolDirect
       val paramSyms = paramTrees.map(_.symbol)
@@ -706,9 +706,9 @@ trait NirGenExpr { self: NirGenPhase =>
           paramSyms.zip(functionArgs.takeRight(sigTypes.length)).zip(params).foreach {
             case ((sym, arg), value) =>
               val unboxedOrCast = {
-                val unboxed = buf.unboxValue(sym.tpe, partial = true, value)
+                val unboxed = buf.unboxValue(sym.tpe, partial = true, value)(arg.pos)
                 if (unboxed == value) // no need to or cannot unbox, we should cast
-                  buf.genCastOp(genType(sym.tpe), genType(arg.tpe), value)
+                  buf.genCastOp(genType(sym.tpe), genType(arg.tpe), value)(arg.pos)
                 else
                   unboxed
               }
@@ -734,7 +734,7 @@ trait NirGenExpr { self: NirGenPhase =>
             targetTree.symbol.tpe.resultType
           }
 
-          val boxedRes = ensureBoxed(res, resTyEnteringPosterasure, callTree.tpe)(buf)
+          val boxedRes = ensureBoxed(res, resTyEnteringPosterasure, callTree.tpe)(buf, callTree.pos)
           buf.ret(boxedRes)
           buf.toSeq
         }
@@ -753,7 +753,7 @@ trait NirGenExpr { self: NirGenPhase =>
     }
 
     def ensureBoxed(value: Val, tpeEnteringPosterasure: Type, targetTpe: Type)(
-        implicit buf: ExprBuffer): Val = {
+        implicit buf: ExprBuffer, pos: nir.Position): Val = {
       tpeEnteringPosterasure match {
         case tpe if isPrimitiveValueType(tpe) =>
           buf.boxValue(targetTpe, value)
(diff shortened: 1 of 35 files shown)


nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -191,18 +192,18 @@ trait NirGenExpr { self: NirGenPhase =>
...
 
       val cond = genExpr(condp)
-      buf.branch(cond, Next(thenn), Next(elsen))
+      buf.branch(cond, Next(thenn), Next(elsen))(condp.pos)
       locally {
-        buf.label(thenn)
+        buf.label(thenn)(thenp.pos)
         val thenv = genExpr(thenp)
-        buf.jump(mergen, Seq(thenv))
+        buf.jump(mergen, Seq(thenv))(NoPosition)
```
sjrd, line 199: IMO there should be virtually no reason to generate `NoPosition` from the compiler back-end, ever. Even for code that seems "generated", you should always take the position of the node that caused the generated instruction to appear. In this case for example, it should be the `pos` of the `If` node.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -1504,10 +1520,11 @@ trait NirGenExpr { self: NirGenPhase =>
       val ty       = genType(st)
       val args     = genSimpleArgs(argsp)
       var res: Val = Val.Zero(ty)
+      val argspVec = argsp.toVector
```
sjrd, line 1523: Consider `zip`ing over `argsp` in addition to the `zipWithIndex` so that you don't have to create this copy of `argsp`.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -1517,11 +1534,11 @@ trait NirGenExpr { self: NirGenPhase =>
       val Seq(lengthp) = argsp
       val length = genExpr(lengthp)
 
-      buf.arrayalloc(genType(targ), length, unwind)
+      buf.arrayalloc(genType(targ), length, unwind)(targ.sym.pos)
```
sjrd, line 1537: That would be the position of the declaration of the element type. Nothing to do with the position of the `new Array`. You need to use the `pos` of the `Tree` that caused the call to `genApplyNewArray`.

In general, you can never use the position of a `Symbol`. You must always use the position of a `Tree`.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -1517,11 +1534,11 @@ trait NirGenExpr { self: NirGenPhase =>
...
     }
 
     def genApplyNew(clssym: Symbol, ctorsym: Symbol, args: List[Tree]): Val = {
-      val alloc = buf.classalloc(genTypeName(clssym), unwind)
+      val alloc = buf.classalloc(genTypeName(clssym), unwind)(clssym.pos)
```
sjrd, line 1541: This is another instance of the above problem. I guess there are others scattered here and there.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
```diff
@@ -94,20 +94,21 @@ trait NirGenStat { self: NirGenPhase =>
...
     def genStructAttrs(sym: Symbol): Attrs = Attrs.None
 
     def genFuncRawPtrExternForwarder(cd: ClassDef): Defn = {
-      val attrs = Attrs(isExtern = true)
-      val name  = genFuncPtrExternForwarderName(cd.symbol)
-      val sig   = Type.Function(Seq.empty, Type.Unit)
+      val attrs                      = Attrs(isExtern = true)
+      val name                       = genFuncPtrExternForwarderName(cd.symbol)
+      val sig                        = Type.Function(Seq.empty, Type.Unit)
+      implicit val pos: nir.Position = Position.generated
```
sjrd, line 107: Use `cd.pos` instead.

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
```diff
@@ -687,6 +698,8 @@ trait NirGenStat { self: NirGenPhase =>
       val fresh = curFresh.get
       val buf   = new ExprBuffer()(fresh)
 
+//      implicit val pos: nir.Position = dd.pos
```
sjrd, line 701: To be removed.

sandbox/Test.scala
```diff
@@ -1,5 +1,11 @@
 object Test {
+  def foo(x: Int): Unit = {
```
sjrd, line 2: These changes should be reverted before merging.

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -269,92 +270,97 @@ object Lower {
...
-        genUnreachable(buf)
-        buf.label(fresh(), Seq(Val.Local(n, op.resty)))
-      case _ =>
-        genOp(buf, n, op)
-    }
+    def genLet(buf: Buffer, n: Local, op: Op)(implicit pos: Position): Unit =
+      op.resty match {
+        case Type.Unit =>
+          genOp(buf, fresh(), op)
+          buf.let(n, Op.Copy(unit), unwind)(Position.generated)
```
sjrd, line 277: In this case here as well, never use a generated position for code that is the product of transforming a tree. You should carry the position that you receive from above.

tools/src/main/scala/scala/scalanative/interflow/Inline.scala
```diff
@@ -181,7 +182,7 @@ trait Inline { self: Interflow =>
               case Inst.Throw(value, unwind) =>
                 val excv = block.end.materialize(value)
                 emit ++= block.toInsts.init
-                emit.raise(excv, unwind)
+                emit.raise(excv, unwind)(block.cf.pos)
```
sjrd, line 185: This is probably redundant with the `implicit val` above.

sjrd (changes requested):

tools/src/main/scala/scala/scalanative/interflow/PolyInline.scala
```diff
@@ -6,6 +6,8 @@ import scalanative.nir._
 import scalanative.linker._
 
 trait PolyInline { self: Interflow =>
+//  private implicit val pos = Position.generated
+
```
sjrd, line 10: Should be removed.

tools/src/main/scala/scala/scalanative/linker/Infos.scala
```diff
@@ -7,6 +7,7 @@ import scalanative.nir._
 sealed abstract class Info {
   def attrs: Attrs
   def name: Global
+  implicit def position: Position
```
sjrd, line 10: That this is implicit is very confusing. Consider:
```suggestion
  def position: Position
```

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -358,6 +358,7 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
 
   def reachVar(defn: Defn.Var): Unit = {
     val Defn.Var(attrs, name, ty, rhs) = defn
+    import defn.pos
```
sjrd, line 361: The fact that `defn.pos` is itself an `implicit` here is an accident and we should not rely on it. It is too confusing. Consider instead:
```suggestion
    implicit val pos = defn.pos
```
Same later in this file.

sjrd (changes requested):
I finally dedicated the time to come back to this PR.

This looks very good. I only have a few minor, localized comments.

It will need a rebase on top of the 2.12 PR, unfortunately.

sjrd (approved):

PR #1891, opened 2020-09-09
Update gitignore for new Metals

ekrich:
The newest Metals supports `sbt` build files and generates a `metals.sbt` in `project` and in `project/project`. This now ignores both cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,7 @@ bin/
 # metals
 /.bloop/
 /.metals/
-metals.sbt
+/project/**/metals.sbt
 
 # vscode
 /.vscode/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -24,7 +24,7 @@ bin/
 # metals
 /.bloop/
 /.metals/
-/project/metals.sbt
+metals.sbt
```
sjrd, line 27: I would prefer that we be more strict. I've been using
```
/project/**/metals.sbt
```
in my projects.

sjrd (changes requested):

.gitignore
ekrich: I didn't even know that `**` worked. Thanks.

sjrd (approved):

PR #1892, opened 2020-09-09
Update java.util.Properties to match Scala.js changes

ekrich:
Take advantage of changes made while creating a PR for Scala.js.
1. Make sure to use ScalaOps not Scala collection converters.
2. Add parens as auto-application is deprecated in future version of Scala when not calling Java.


ekrich:
@sjrd I would like to change the two `File` based checks to exactly match the `String` based file checks. The stream from `String` approach and the extra added test has the same code coverage.

sjrd:
Yes, that would be better, I think.

sjrd (approved):

PR #1894, opened 2020-09-14
Fix #1435: Cache instances of j.l.Class

### LINKED ISSUE #1435: Cache instances of java.lang.Class
Currently we don't cache instances of `java.lang.Class`, which means two instances which correspond to the same type might not compare equals in term of reference identity and `java.lang.System.identityHashCode`. This is inconsistent with the behavior on the JVM.

WojciechMazur:
Resolves #1435 

This PR moves allocation of java.lang.Class instances to linker phase and introduces caching of them.
In previous implementation Class instances were created by compiler backend, which lead to multiple instances of this type in runtime and made impossible to cache them.
Current implementation introduces new Val literal type `Val.ClassOf(n: Global)` which replaces allocations from previous implementation. At lowering phase this literals are transformed into `Object.getClass` calls with argument of RTTI object (as previous).  RTTI type was also extended to store reference of associated with it java.lang.Class. Inside `Object.getClass` it's checked whether RTTI class pointer is not empty and  instantiated in otherwise. Newly created instance of j.l.Class is added to registry in order to prevent GC from collecting it. 

Immix/Commix implementation was fixed to handle new RTTI layout.

### DIFF
--- a/nativelib/src/main/scala/java/lang/Class.scala
+++ b/nativelib/src/main/scala/java/lang/Class.scala
@@ -19,6 +19,7 @@ object rtti {
 }
 import rtti._
 
+/** @param rawty - Pointer with underlying Rt.Type info */
 final class _Class[A](val rawty: RawPtr) {
   @alwaysinline private def ty: Ptr[Type] =
     fromRawPtr[Type](rawty)
--- a/nativelib/src/main/scala/java/lang/Class.scala
+++ b/nativelib/src/main/scala/java/lang/Class.scala
@@ -62,9 +62,8 @@ final class _Class[A](val rawty: RawPtr) {
   def isInstance(obj: Object): scala.Boolean =
     is(obj.getClass.asInstanceOf[_Class[_]].ty, ty)
 
-  @alwaysinline private def is(cls: Class[_]): Boolean = {
-    rawty == toRawType(cls)
-  }
+  @alwaysinline private def is(cls: Class[_]): Boolean =
+    this eq cls.asInstanceOf[_Class[A]]
 
   private def is(left: Ptr[Type], right: Ptr[Type]): Boolean =
     // This replicates the logic of the compiler-generated instance check
--- a/nativelib/src/main/scala/java/lang/Class.scala
+++ b/nativelib/src/main/scala/java/lang/Class.scala
@@ -28,14 +28,14 @@ final class _Class[A](val rawty: RawPtr) {
     obj.asInstanceOf[A]
 
   def getComponentType(): _Class[_] = {
-    if (rawty == toRawType(classOf[BooleanArray])) classOf[scala.Boolean]
-    else if (rawty == toRawType(classOf[CharArray])) classOf[scala.Char]
-    else if (rawty == toRawType(classOf[ByteArray])) classOf[scala.Byte]
-    else if (rawty == toRawType(classOf[ShortArray])) classOf[scala.Short]
-    else if (rawty == toRawType(classOf[IntArray])) classOf[scala.Int]
-    else if (rawty == toRawType(classOf[LongArray])) classOf[scala.Long]
-    else if (rawty == toRawType(classOf[FloatArray])) classOf[scala.Float]
-    else if (rawty == toRawType(classOf[DoubleArray])) classOf[scala.Double]
+    if (is(classOf[BooleanArray])) classOf[scala.Boolean]
+    else if (is(classOf[CharArray])) classOf[scala.Char]
+    else if (is(classOf[ByteArray])) classOf[scala.Byte]
+    else if (is(classOf[ShortArray])) classOf[scala.Short]
+    else if (is(classOf[IntArray])) classOf[scala.Int]
+    else if (is(classOf[LongArray])) classOf[scala.Long]
+    else if (is(classOf[FloatArray])) classOf[scala.Float]
+    else if (is(classOf[DoubleArray])) classOf[scala.Double]
     else classOf[java.lang.Object]
   }
 
@@ -46,20 +46,25 @@ final class _Class[A](val rawty: RawPtr) {
     getName().split('.').last.split('$').last
 
   def isArray(): scala.Boolean =
-    (rawty == toRawType(classOf[BooleanArray]) ||
-      rawty == toRawType(classOf[CharArray]) ||
-      rawty == toRawType(classOf[ByteArray]) ||
-      rawty == toRawType(classOf[ShortArray]) ||
-      rawty == toRawType(classOf[IntArray]) ||
-      rawty == toRawType(classOf[LongArray]) ||
-      rawty == toRawType(classOf[FloatArray]) ||
-      rawty == toRawType(classOf[DoubleArray]) ||
-      rawty == toRawType(classOf[ObjectArray]))
+    is(classOf[BooleanArray]) ||
+      is(classOf[CharArray]) ||
+      is(classOf[ByteArray]) ||
+      is(classOf[ShortArray]) ||
+      is(classOf[IntArray]) ||
+      is(classOf[LongArray]) ||
+      is(classOf[FloatArray]) ||
+      is(classOf[DoubleArray]) ||
+      is(classOf[ObjectArray])
+
   def isAssignableFrom(that: Class[_]): scala.Boolean =
     is(that.asInstanceOf[_Class[_]].ty, ty)
 
   def isInstance(obj: Object): scala.Boolean =
     is(obj.getClass.asInstanceOf[_Class[_]].ty, ty)
+
+  @alwaysinline private def is(cls: Class[_]): Boolean = {
+    Intrinsics.castObjectToRawPtr(this) == Intrinsics.castObjectToRawPtr(cls)
+  }
 
   private def is(left: Ptr[Type], right: Ptr[Type]): Boolean =
     // This replicates the logic of the compiler-generated instance check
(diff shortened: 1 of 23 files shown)


WojciechMazur:
Looks like this change introduced significant regression in build time or performance. CI jobs takes in debug mode around 15-25% longer and in release-fast it take ~50% more time. Due to this 3 last CI jobs has timed out (there's a limit of 50 minutes). I'll take a look at this

ekrich:
Do you think it might be helpful to have some diagrams showing the runtime layout? Isn't one of the big advantages over the JVM is that the JVM carries a bunch of overhead in each object? I don't want to slow down the release maybe a follow on issue if you agree.

nativelib/src/main/scala/java/lang/ClassInstancesRegistry.scala
```diff
@@ -0,0 +1,36 @@
+package java.lang
+
+/** Registry for created instances of java.lang.Class
+ * It's only purpose is to prevent GC from collecting instances of java.lang.Class
```
sjrd, line 4:
```suggestion
 *  Its only purpose is to prevent the GC from collecting instances of java.lang.Class
```

nativelib/src/main/scala/java/lang/ClassInstancesRegistry.scala
```diff
@@ -0,0 +1,36 @@
...
+ **/
+object ClassInstancesRegistry {
+  type Arr[T] = scala.Array[T]
```
sjrd, line 7: This type alias seems superfluous. You can directly use `Array[T]` instead. (AFAICT, the `scala.` prefix is not necessary.)

nativelib/src/main/scala/java/lang/ClassInstancesRegistry.scala
```diff
@@ -0,0 +1,36 @@
+package java.lang
+
+/** Registry for created instances of java.lang.Class
+ * It's only purpose is to prevent GC from collecting instances of java.lang.Class
+ **/
+object ClassInstancesRegistry {
```
sjrd, line 6: You can make this `private[lang]` to make it clear that this is not a public API:
```suggestion
private[lang] object ClassInstancesRegistry {
```

nativelib/src/main/scala/java/lang/ClassInstancesRegistry.scala
```diff
@@ -0,0 +1,36 @@
...
+  type Arr[T] = scala.Array[T]
+  type Bucket = Arr[_Class[_]]
+
+  final val BucketSize = 256
```
sjrd, line 10:
```suggestion
  private final val BucketSize = 256
```

nativelib/src/main/scala/java/lang/ClassInstancesRegistry.scala
```diff
@@ -0,0 +1,36 @@
...
+    bucketId = lastId / BucketSize
+    itemId = lastId % BucketSize
+  }
+
+  def add(cls: _Class[_]): _Class[_] = {
+    nextIds()
+    if (bucketId >= buckets.length) {
+      val newSize: Int = buckets.length * 2
+      val newArr       = new Arr[Bucket](newSize)
+      Array.copy(buckets, 0, newArr, 0, buckets.length)
```
sjrd, line 27: I suggest trying very hard not to use any Scala-specific API here. Consider using `System.arraycopy` instead.

nativelib/src/main/scala/java/lang/Object.scala
```diff
@@ -18,8 +19,18 @@ class _Object {
...
+    val rtti   = getRawType(this)
+    val clsPtr = elemRawPtr(rtti, 16)
+
+    if (loadRawPtr(clsPtr) == null) {
+      val newClass = new _Class[Any](rtti)
+      storeObject(clsPtr, newClass)
+      ClassInstancesRegistry.add(newClass)
+    } else {
+      loadObject(clsPtr).asInstanceOf[_Class[_]]
+    }
```
sjrd, line 32: I suggest moving all these low-level manipulations to a new method `runtime.toClass(rawType: RawPtr): Class[_]`, so that this method becomes simply
```scala
runtime.toClass(runtime.getRawType(this))
```
that way we keep the runtime-related concepts inside `runtime.*`.

tools/src/main/scala/scala/scalanative/codegen/Generate.scala
```diff
@@ -115,6 +115,31 @@ object Generate {
       )
     }
 
+    def genClassInitializers(): Unit = {
```
sjrd, line 118: Why was this extracted here? It doesn't seem to have anything to do with the encoding of `j.l.Class`es.

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -154,18 +170,33 @@ object Lower {
...
+        util.unsupported("Lowering ClassOf needs nir.Buffer")
+      case Val.Global(ScopeRef(node), _) => rtti(node).const
+      case Val.String(v)                 => genStringVal(v)
+      case Val.Unit                      => unit
+      case _                             => super.onVal(value)
+    }
+
+    def genClassOf(buf: Buffer, node: ScopeInfo): Val = {
+      val tpePtr = Val.Const(rtti(node).const)
+      buf.call(JavaGetClassSig, JavaGetClass, Seq(tpePtr), Next.None)
```
sjrd, line 189: This does not seem right. It seems you're calling `java.lang.Object.getClass` to get a `j.l.Class[_]`. But you don't have any object to use as the `this` object. IIUC this line passes `tpePtr` itself as the receiver, but that would not be type-correct, since `tpePtr` is a raw pointer to a C struct, not an instance of `java.lang.Object` or its subclasses.

This should be calling `runtime.toClass` (which I suggested to create above) with the singleton instance of `runtime.package$` as receiver, and `typePtr` as argument.

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -154,18 +170,33 @@ object Lower {
 
       buf ++= handlers
 
-      eliminateDeadCode(buf.toSeq.map(super.onInst))
+      eliminateDeadCode {
+        buf.toSeq.map(super.onInst)
+      }
     }
 
     override def onVal(value: Val): Val = value match {
```
sjrd, line 178: Is method still called from anywhere in practice? If not, it should unconditionally throw to detect problems early. If yes, why is not an issue that it doesn't support `Val.ClassOf` for the places that still call it?

unit-tests/src/test/scala/java/lang/ClassSuite.scala
```diff
@@ -135,4 +135,22 @@ object ClassSuite extends tests.Suite {
     assert(!classOf[java.lang.Class[_]].isInterface)
     assert(classOf[java.lang.Runnable].isInterface)
   }
+
+  private def assertEqualClass(l: java.lang.Class[_],
+                               r: java.lang.Class[_]): Unit = {
+    assert(l == r, s"$l != $r")
+    assert(System.identityHashCode(l) == System.identityHashCode(r),
+           s"$l diff ident hash $r")
```
sjrd, line 143: What you really want here is reference equality (`eq`). The `identityHashCode` is not necessary:
```suggestion
    assert(l eq r, s"$l ne $r")
```

unit-tests/src/test/scala/java/lang/ClassSuite.scala
```diff
@@ -135,4 +135,22 @@ object ClassSuite extends tests.Suite {
...
+
+  test("class instances are cached") {
+    val cls1 = "asd".getClass
+    val cls2 = "xyz".getClass
+
+    assertEqualClass(classOf[String], cls1)
+    assertEqualClass(cls1, cls2)
+    assertEqualClass(classOf[String], classOf[String])
+
+    assert(123L.getClass != 42.getClass)
```
sjrd, line 154: There should be tests with array classes as well, since they are quite special.

sjrd (changes requested):

tools/src/main/scala/scala/scalanative/codegen/Generate.scala
WojciechMazur: In one of previous versions I moved here both Ergys class initializers for Reflect support as well as my j.l.Class initializers, which were later removed, however I've left the method. Right now it does nothing to do with j.l.Classes although it makes code a bit more clear so I thought it may stay here.

sjrd: OK, that makes sense, but then this change must be on a separate PR. Otherwise it's going to get lumped together in the squashed commit for this PR, and the next person looking at that commit during a git blame will get super confused.

WojciechMazur: I've reverted this change

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
WojciechMazur: All instances of `Val.ClassOf` are being handled before this point with `genVal`.
I've tried to test it against `Inst.If` which is not being handled directly, for expr `if(???) classOf[Int] else classOf[String]` it compiles without problem which means it was handled before `super.onInst`.
Same result when using `match`

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -1038,6 +1113,10 @@ object Lower {
   val throwSig  = Type.Function(Seq(Type.Ptr), Type.Nothing)
   val throw_    = Val.Global(throwName, Type.Ptr)
 
+  val toClassName = Rt.Runtime.name.member(Rt.ToClassSig)
+  val toClassSig  = Type.Function(Seq(Rt.Runtime, Type.Ptr), Rt.Class)
+  val toClass     = Val.Global(toClassName, Type.Ptr)
```
sjrd, line 1118: I'm not sure, but I think this should be similar to how we deal with other methods of `runtime`, for example `throwOutOfBounds` down below. (both the definitions of these constants, and how we use then in `buf.call` in `genClassOf` -- although we don't need the `unwind` handler like in the methods that throw)

sjrd (changes requested):

sjrd:
This probably needs a rebase on top of master, now that #1946 has been merged.

nativelib/src/main/scala/java/lang/Class.scala
```diff
@@ -45,21 +46,26 @@ final class _Class[A](val rawty: RawPtr) {
...
+      is(classOf[ObjectArray])
+
   def isAssignableFrom(that: Class[_]): scala.Boolean =
     is(that.asInstanceOf[_Class[_]].ty, ty)
 
   def isInstance(obj: Object): scala.Boolean =
     is(obj.getClass.asInstanceOf[_Class[_]].ty, ty)
 
+  @alwaysinline private def is(cls: Class[_]): Boolean = {
+    rawty == toRawType(cls)
```
sjrd, line 66: I believe this should now be `this eq cls`.

nativelib/src/main/scala/scala/scalanative/runtime/ClassInstancesRegistry.scala
```diff
@@ -0,0 +1,35 @@
+package scala.scalanative.runtime
+
+/** Registry for created instances of java.lang.Class
+ * Its only purpose is to prevent the GC from collecting instances of java.lang.Class
+ **/
```
sjrd, line 5:
```suggestion
/** Registry for created instances of java.lang.Class
 *  Its only purpose is to prevent the GC from collecting instances of java.lang.Class
 */
```

sjrd (changes requested):
Looks good. I only have two very small comments.

sjrd (approved):
@WojciechMazur could you rebase this on top of latest master, just to make sure it still works with the latest CI setup?

PR #1901, opened 2020-09-21
Fix #1627: Allow passing generic functions ptr as method args

### LINKED ISSUE #1627: Calling extern functions with generic CFuncPtr fails at runtime
With **SN-0.4.0-M2**, calling an extern function that has a `CFuncPtr` parameter (instead of the more specific `CFuncPtrN[]`) fails at runtime. Example:

```C
typedef int (*Callback0) (void);

int exec(Callback0 f) {
  return f();
}
```

```scala
@extern
object ext {
  @name("exec")
  def exec0(f: CFuncPtr0[Int]): Int = extern
  def exec(f: CFuncPtr): Int = extern
}

val cb = new CFuncPtr0[Int] { override def apply(): Int = 42 }

// works as expected
ext.exec0(cb)

// fails at runtime
ext.exec(cb)
```

WojciechMazur:
Resolves #1627 

This PR fixes bug leading to segmentation fault when trying to pass `CFuncPtrN` to method with argument of type `CFuncPtr`. 
Bug originated from invalid check inside `NirGenExpr.genMethodArgs` - `CFuncPtr` base trait was not being checked which lead to bad type inheritance.


unit-tests/src/test/scala/scala/scalanative/unsafe/ExternSuite.scala
```diff
@@ -50,4 +50,13 @@ object ExternSuite extends tests.Suite {
...
     assert(string.strcmp(buf2, c"1") == 0)
   }
+
+  val cb: CFuncPtr0[CInt] = new CFuncPtr0[Int] {
+    override def apply(): Int = 42
+  }
+  test("allows to use generic function as argument") {
+    val res0 = testlib.exec0(cb) //expected CFuncPtr0[Int]
+    val res1 = testlib.exec(cb)  //expected CFuncPtr
+    assert(res0 == res1)
```
sjrd, line 60: Consider making this test more precise by explicitly testing the expected result of each call, rather than testing that the two are equal. Currently, this test would pass even if `exec0` and `exec` both return the same wrong result (e.g., `0`).

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenType.scala
```diff
@@ -25,10 +25,20 @@ trait NirGenType { self: NirGenPhase =>
...
+    /** Tests if this type inherits from CFuncPtr with exclusion of CFuncRawPtr
+     * (wrapper for all CFuncPtr types) and should be handled differently.
+     */
+    def isCFuncPtrClass: Boolean = sym != CFuncRawPtrClass && {
+      sym == CFuncPtrClass ||
+      sym.info.parents.exists(_.typeSymbol == CFuncPtrClass)
+    }
+
+    /** Tests if this type inherits from CFuncPtr and contains function definition (apply method)
+     * CFuncPtr itself does not contain apply method */
```
sjrd, line 37: This comment does not seem to have anything to do with the implementation.

sjrd (changes requested):
The fix looks good. I have minor comments.

sjrd (approved):

PR #1902, opened 2020-09-22
Update gitignore for new Metals

ekrich:
Align with Scala.js gitignore


ekrich:
Oops, already merged. Getting confused by doing things here and in Scala.js

PR #1905, opened 2020-09-26
Allow setting target triple and show when compiling

ekrich:
This is nice to have confirmation to have while compiling.
```
[info] Linking (1346 ms)
[info] Checking intermediate code (278 ms)
[info] Dumping intermediate code (linked) (306 ms)
[info] Discovered 592 classes and 3545 methods
[info] Optimizing (debug mode) (1426 ms)
[info] Checking intermediate code (56 ms)
[info] Dumping intermediate code (optimized) (139 ms)
[info] Dumping intermediate code (lowered) (213 ms)
[info] Generating intermediate code (1135 ms)
[info] Produced 8 files
[info] Compiling to native code (x86_64-apple-macosx10.14.0) (3453 ms)
[info] Linking native code (immix gc, none lto) (147 ms)
```
This would have been nice to have for my recent Raspberry Pi demo to show the target triple for 64bit armv8.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -54,6 +54,9 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeDump =
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
+
+    val nativeTarget =
+      taskKey[String]("The LLVM target triple for cross compiling.")
   }
 
   @deprecated("use autoImport instead", "0.3.7")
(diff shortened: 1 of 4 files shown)


ekrich:
Based on @lolgab 's PR #1904 - I decided to experiment and do some changes for cross compiling too. This is a more conservative approach.

lolgab:
You need to add the `-target $targetTriple` option to the `compile` and `compileNativelibs` functions too, as I did in #1904 otherwise cross-compilation doesn't work because it's trying to link object files made for another target.

lolgab:
You're not invalidating the targets generated by `compileNativeLibs` when `nativeTarget` changes.

ekrich:
I don't see an easy way around the generating into different directories issue if you want to maintain the `probe` for target triple. I guess clean when you switch would be the only way. Will that fit the use cases you expect?

I added the other areas for target triple.

ekrich:
Related: https://github.com/scala-native/scala-native/issues/414

lolgab:
Thanks to the Zig compiler I could get a working cross-compiled Scala Native linux binary except for exception handling (there is some binary compatibility issue I'm still trying to figure out why happens).
So I would say this road is very promising!
Having to `clean` is better than not having the feature but I will dig more to see if it's possible to have the cache invalidation work without needing to `clean`.

ekrich:
If you are going to be using this in CI, then we may need the `env` var approach to allow the change of the target triple in which case you should have a clean setup for each build.

I think we should just look at this as just a first step towards cross building and see from experience what needs to change in the future.

ekrich:
Closed in favor of #2033

Issue #1193, opened 2018-03-08
Feature: consistent project names


Andrei-Pozolotin:
1. motivation: `sbt> eclipse`  produces short project names, 
which conflict with existing eclipse workspace projects with high probability:

2. solution: we suggest consistent project names with pattern based on `project.id`:
```
// Provide consistent project name pattern.
lazy val nameSettings = Seq(
  normalizedName := projectName(thisProject.value), // Maven <artifactId>
  name := s"Scala.native ${projectName(thisProject.value)}" // Maven <name>
)

// Generate project name from project id.
def projectName(project: sbt.ResolvedProject) : String = {
  convertCamelKebab(project.id)
}

// Convert "SomeName" to "some-name".
def convertCamelKebab(name: String) : String = {
  name.replaceAll( "([a-z])([A-Z]+)", "$1-$2" ).toLowerCase
}
```
3. for example:
* before

```
sbt> eclipse
[info] Successfully created Eclipse project files for project(s):
[info] nscplugin
[info] benchmarks
[info] util
[info] testingCompiler
[info] test-interface
[info] scalalib
[info] tests
[info] javalib
[info] test-interface-sbt-defs
[info] tools
[info] nir
[info] test-interface-serialization
[info] testingCompilerInterface
[info] auxlib
[info] nativelib
[info] sandbox
[info] sbtScalaNative
```
pom.xml
```xml
<name>scalalib</name>
<description>scalalib</description>
```

* after
```
sbt> eclipse
[info] Successfully created Eclipse project files for project(s):
[info] Scala.native nscplugin
[info] Scala.native benchmarks
[info] Scala.native util
[info] Scala.native testing-compiler
[info] Scala.native test-interface
[info] Scala.native scalalib
[info] Scala.native tests
[info] Scala.native javalib
[info] Scala.native test-interface-sbt-defs
[info] Scala.native tools
[info] Scala.native nir
[info] Scala.native test-interface-serialization
[info] Scala.native testing-compiler-interface
[info] Scala.native auxlib
[info] Scala.native nativelib
[info] Scala.native sandbox
[info] Scala.native sbt-scala-native
```
pom.xml
```xml
<name>Scala.native scalalib</name>
<description>Scala.native scalalib</description>
```

ekrich:
[...]

PR #1911, opened 2020-10-01
Revert to normal artifact names in the jar manifest

ekrich:
This reverts a portion of the a previous issue/change (https://github.com/scala-native/scala-native/issues/1193). The change prepended "Scala Native" to the artifact like `nativelib`. This was to allow all the Scala Native projects to be next to each other in an Eclipse workspace where you have other projects. 

This PR changes the behavior back, making the artifact name (Implementation-Title and Specification-Title) standard in the MANIFEST.MF in the Scala Native published jars. 

Here is a before and after look at this change.

Before:
```
-- listing properties --
Manifest-Version=1.0
Implementation-Title=Scala Native nativelib
Implementation-Vendor=org.scala-native
Specification-Version=0.4.0-SNAPSHOT
Specification-Title=Scala Native nativelib
Implementation-Vendor-Id=org.scala-native
Specification-Vendor=org.scala-native
Implementation-Version=0.4.0-SNAPSHOT
Implementation-URL=http://www.scala-native.org
```

After:
```
-- listing properties --
Manifest-Version=1.0
Implementation-Title=nativelib
Implementation-Vendor=org.scala-native
Specification-Version=0.4.0-SNAPSHOT
Specification-Title=nativelib
Implementation-Vendor-Id=org.scala-native
Specification-Vendor=org.scala-native
Implementation-Version=0.4.0-SNAPSHOT
Implementation-URL=http://www.scala-native.org
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,8 +16,8 @@ def projectName(project: sbt.ResolvedProject): String = {
 
 // Provide consistent project name pattern.
 lazy val nameSettings: Seq[Setting[_]] = Seq(
-  normalizedName := projectName(thisProject.value),         // Maven <artifactId>
-  name := s"Scala Native ${projectName(thisProject.value)}" // Maven <name>
+  normalizedName := projectName(thisProject.value), // Maven <artifactId>
+  name := projectName(thisProject.value)            // Maven <name>
 )
 
 // The previous releases of Scala Native with which this version is binary compatible.
--- a/build.sbt
+++ b/build.sbt
@@ -16,8 +16,7 @@ def projectName(project: sbt.ResolvedProject): String = {
 
 // Provide consistent project name pattern.
 lazy val nameSettings: Seq[Setting[_]] = Seq(
-  normalizedName := projectName(thisProject.value), // Maven <artifactId>
-  name := projectName(thisProject.value)            // Maven <name>
+  name := projectName(thisProject.value) // Maven <name>
 )
 
 // The previous releases of Scala Native with which this version is binary compatible.


ekrich:
@sjrd This one is ready for review.

build.sbt
```diff
@@ -19,8 +19,8 @@ def projectName(project: sbt.ResolvedProject): String = {
 
 // Provide consistent project name pattern.
 lazy val nameSettings: Seq[Setting[_]] = Seq(
-  normalizedName := projectName(thisProject.value),         // Maven <artifactId>
-  name := s"Scala Native ${projectName(thisProject.value)}" // Maven <name>
+  normalizedName := projectName(thisProject.value), // Maven <artifactId>
+  name := projectName(thisProject.value)            // Maven <name>
```
sjrd, line 23: Is this still any different from not having `nameSettings` at all?

ekrich: The `normalizedName` setting was to make sure that all the projects had lowercase names with dashes `-` connecting the words. I did try removing it but then some of the artifacts reverted to other forms. I think a more direct approach of naming the individual projects in `sbt` is more explicit. It is a little more work but could done if you prefer.

sjrd: Ah I see. Then I guess we can keep the setting for `name :=`.

`normalizedName` normally derives from `name`. So I believe that one could be removed.

ekrich: Yes, so this artifact would match the directory.
```scala
lazy val testInterface =
  project
    .in(file("test-interface"))
```
Ok, let me confirm that we can remove.

ekrich: Yes, it is the same before and after the change.
```
sbt:scala-native> testInterface/name
[info] test-interface
sbt:scala-native> testInterface/normalizedName
[info] test-interface
sbt:scala-native> 
```

sjrd (approved):

ekrich:
@sjrd This is ready for final review.

sjrd (approved):

PR #1924, opened 2020-10-07
Feature #Create dedicated method for class initializing in Generate

WojciechMazur:
This PR does not introduce any changes to plugin behavior other then making additional call in `main` method.
It's main purpose is making `Generate.genMain` more readable and easier to maintain and debug assembly by extracting class initializing logic to separate method.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/Generate.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Generate.scala
@@ -27,6 +27,7 @@ object Generate {
       genClassHasTrait()
       genTraitMetadata()
       genTraitHasTrait()
+      genClassInitializers()
       genTraitDispatchTables()
       genModuleAccessors()
       genModuleArray()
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/codegen/Generate.scala
```diff
@@ -115,6 +116,31 @@ object Generate {
...
+      implicit val fresh = Fresh()
+
+      insts += Inst.Label(fresh(), Seq())
+
+      defns.foreach {
+        case Defn.Define(_, name: Global.Member, _, _) if name.sig.isClinit =>
+          insts += Inst.Let(Op.Call(Type.Function(Seq(), Type.Unit),
+                                    Val.Global(name, Type.Ref(name)),
+                                    Seq()),
+                            Next.None)
```
sjrd, line 130: The previous code used an `unwind` handler here. I'm pretty sure we still need one. What if one of the static initializers throws an exception?

sjrd (changes requested):

PR #1929, opened 2020-10-08
Fix #1928: show file name for NIR version mismatch during linking

### LINKED ISSUE #1928: Incompatible NIR version assertion should print filename
When a file with an incompatible NIR version is encountered during linking, there should be some info about the file / dependency that actually triggered the issue. Currently there is no additional information:

```
[error] stack trace is suppressed; run last lib / Compile / nativeLink for the full output
[error] (lib / Compile / nativeLink) java.lang.AssertionError: assertion failed: Can't read binary-incompatible version of NIR.
```

jokade:
- pass a `bufferName` to `BinarySerializer` and use it to identify the file from which the NIR buffer was loaded
- add property `uri` to `VirtualDirectory` and use it to derive an absolute `bufferName` for each NIR buffer read
- print this `bufferName` when there is a NIR version mismatch in `Prelude.readFrom`


tools/src/main/scala/scala/scalanative/linker/ClassPath.scala
```diff
@@ -46,18 +46,24 @@ object ClassPath {
     def contains(name: Global) =
       files.contains(name.top)
 
+    private def makeBufferName(directory: VirtualDirectory, file: Path) =
+      directory.uri.toString + file
```
sjrd, line 50: Concatenating the string representation of a URI with the string representation of a Path does not seem right. Can we make that more semantic?

jokade: Done

sjrd (approved):

PR #1935, opened 2020-10-09
Avoid importing `runtime` in the java.lang package

catap:
Right now it is impossible to build `scala-native` on jdk14+ and it crashed with error:
```
imported `runtime' is permanently hidden by definition of package runtime in package lang
```

Fixed this issue.

### DIFF
--- a/nativelib/src/main/scala/java/lang/Object.scala
+++ b/nativelib/src/main/scala/java/lang/Object.scala
@@ -1,9 +1,7 @@
 package java.lang
 
 import scala.scalanative.unsafe._
-import scala.scalanative.runtime, runtime.ClassTypeOps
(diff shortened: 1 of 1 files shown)


ekrich (commented):
This change could use a comment why `_runtime` so someone won't go trying to change it back or may wonder in the future why it starts with an underscore.

lolgab:
If you import all the SN runtime members directly you don't need to intruduce the `_runtime` name.

catap:
@lolgab thanks, it makes code clear

ekrich (approved):
LGTM but will defer to others for a more definitive decision.

sjrd (approved):

Issue #1938, opened 2020-10-10
Can't build expression on scala-native via JDK12+


catap:
I've used bd19ebb2e and tried to run unit tests on openjdk 12.0.2:
```
OpenJDK Runtime Environment AdoptOpenJDK (build 12.0.2+10)
```

I run simple command: `sbt +tests/test` and before do it I run `find ./ -iname target | xargs rm -rf` to clean any artifacts from previous build.

This command fails with error tons of linking errors: https://gist.github.com/catap/26e9d9dae8ef57e0e55e119e6c481b63

catap:
Tests what cause this error:
 - `format(String, Array[Object]) for general conversion b/B`
 - `format(String, Array[Object]) for general conversion type 's' and 'S'`

From `FormatterUSSuite` and `DefaultFormatterSuite`

ekrich:
I am in the process of porting these tests to JUnit. I know that won't help your situation but I hope we can merge my work first if possible.

catap:
Ok, I simplified the case to have a error:
```
Seq(Double.box(1d), "")
```

If I try it to build via JDK 12+ I have an error
```
[info] Linking (1102 ms)
[warn] Found 2 errors on @"M5Test$D4mainLAL16java.lang.String_uEO" :
[warn]   0000  def @"M5Test$D4mainLAL16java.lang.String_uEO" : (@"T5Test$", array[@"T16java.lang.String"]) => unit {
[warn]   0001  %3(%1 : @"T5Test$", %2 : array[@"T16java.lang.String"]):
[warn]   0002    %4 = module @"T21scala.collection.Seq$"
[warn]   0003    %5 = module @"T13scala.Predef$"
[warn]   0004    %6 = box[@"T16java.lang.Double"] double 1.0
[warn]   0005    %7 = arrayalloc[@"T28java.lang.constant.Constable"] int 2
[warn]   0006    %8 = arraystore[@"T28java.lang.constant.Constable"] %7 : !?@"T37scala.scalanative.runtime.ObjectArray", int 0, %6 : !@"T16java.lang.Double"
[warn]   0007    %9 = arraystore[@"T28java.lang.constant.Constable"] %7 : !?@"T37scala.scalanative.runtime.ObjectArray", int 1, ""
[warn]   0008    %10 = as[array[@"T16java.lang.Object"]] %7 : !?@"T37scala.scalanative.runtime.ObjectArray"
[...]

PR #1941, opened 2020-10-10
Work around limitation for JDK12+ about j.l.constant.Constable

catap:
java.lang.Constable was introduced at JDK12 and it requires to link an application by scala-native via JDK12+.

It oversteps https://github.com/scala-native/scala-native/issues/1938

### DIFF
--- a/unit-tests/src/test/scala/java/util/DefaultFormatterTest.scala
+++ b/unit-tests/src/test/scala/java/util/DefaultFormatterTest.scala
@@ -830,7 +830,7 @@ class DefaultFormatterTest {
   }
 
   @Test def formatForGeneralConversionType_bB(): Unit = {
-    val triple = Array(
+    val triple: Array[Array[Object]] = Array(
       Array(Boolean.box(false), "%3.2b", " fa"),
       Array(Boolean.box(false), "%-4.6b", "false"),
       Array(Boolean.box(false), "%.2b", "fa"),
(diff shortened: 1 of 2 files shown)


sjrd:
If we're going that route, we should also add `ConstantDesc`, which can in theory expose the same issue.

However, I'm not a fan of adding those, because we will have to leave their methods undeclared, given that we cannot implement them. We're only pushing problems a bit further.

Can we fix the test suite so that those types are not inferred instead?

catap:
@sjrd yes, you raised a very good point.

Let split it to two separated topic:
1. fix building tests on JDK12+ and let me fix it by changing unit tests via this PR;
2. address an issue at least via #1938

sjrd (approved):

ekrich:
I could just add these changes to the JUnit PR so I don't have to rebase these.

ekrich:
Gosh, I was trying the latest with AdoptOpenJDK 15 and got this.
```
[info] Compiling 542 Scala sources to /Users/eric/workspace/scala-native/scalalib/target/scala-2.11/classes ...
[error] /Users/eric/workspace/scala-native/scalalib/target/scalaSources/2.11.12/scala/collection/mutable/StringBuilder.scala:31:13: overriding method isEmpty in trait CharSequence of type ()Boolean;
[error]  method isEmpty in trait IndexedSeqOptimized of type => Boolean cannot override a concrete member without a third member that's overridden by both (this rule is designed to prevent ``accidental overrides'')
[error] final class StringBuilder(private val underlying: JavaStringBuilder)
[error]             ^
[error] one error found
```

catap:
@errikos yes, scala doesn't compatible with JDK15+.

You can use this branch that fixes the issue: https://github.com/scala/scala/pull/9239

catap:
@sjrd let me rebase this branch

ekrich:
@catap Thanks for being patient with your change.

catap:
@sjrd maybe move forward on it?

PR #1961, opened 2020-10-21
Fix #1943: Support JUnit's @Ignore on test class

### LINKED ISSUE #1943: JUnit does not apply @Ignore on the class
Annotating `@Ignore class MyTest ...` allows individual tests to execute when it should ignore all the individual tests.

I nice to have feature would be for the annotation `@Ignore("#555")` to output as follows:
` Test java.util.concurrent.atomic.AtomicReferenceTest.updateAndGetUpdateFunction ignored #555`

This way the issues could be more visible.

WojciechMazur:
Resolves #1943 

This PR fixes behaviour of Native JUnit by allowing to annotate whole test class with `Ignore`. In such a scenario all tests within that class should be ignored. 

Proposal to include ignore reason was discarded - it's technically possible, but it violates Native/JVM JUnit outputs compatibility. In JVM version ignore reason is never included in the output

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -582,7 +582,6 @@ lazy val tests =
     )
     .settings(noPublishSettings)
     .settings(
-      nativeConfig ~= (_.withMode(scalanative.build.Mode.releaseFast)),
       // nativeOptimizerReporter := OptimizerReporter.toDirectory(
       //   crossTarget.value),
       // nativeLinkerReporter := LinkerReporter.toFile(
@@ -617,8 +616,6 @@ lazy val sandbox =
     .settings(scalacOptions -= "-Xfatal-warnings")
     .settings(noPublishSettings)
     .settings(
-      nativeConfig ~= (_.withMode(scalanative.build.Mode.releaseFast))
-
       // nativeOptimizerReporter := OptimizerReporter.toDirectory(
       //   crossTarget.value),
     )
--- a/build.sbt
+++ b/build.sbt
@@ -582,6 +582,7 @@ lazy val tests =
     )
     .settings(noPublishSettings)
     .settings(
+      nativeConfig ~= (_.withMode(scalanative.build.Mode.releaseFast)),
       // nativeOptimizerReporter := OptimizerReporter.toDirectory(
       //   crossTarget.value),
       // nativeLinkerReporter := LinkerReporter.toFile(
@@ -616,6 +617,8 @@ lazy val sandbox =
     .settings(scalacOptions -= "-Xfatal-warnings")
     .settings(noPublishSettings)
     .settings(
+      nativeConfig ~= (_.withMode(scalanative.build.Mode.releaseFast))
+
       // nativeOptimizerReporter := OptimizerReporter.toDirectory(
       //   crossTarget.value),
     )
--- a/junit-runtime/src/main/scala/scala/scalanative/junit/Bootstrapper.scala
+++ b/junit-runtime/src/main/scala/scala/scalanative/junit/Bootstrapper.scala
@@ -36,6 +36,7 @@ trait Bootstrapper {
  */
 final class TestMetadata(
     val name: String,
-    val ignored: Boolean,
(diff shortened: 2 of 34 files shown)


ekrich:
Thanks for this fix and looking into https://github.com/scala-native/scala-native/issues/1943

junit-runtime/src/main/scala/scala/scalanative/junit/Bootstrapper.scala
```diff
@@ -36,6 +36,7 @@ trait Bootstrapper {
  */
 final class TestMetadata(
     val name: String,
-    val ignored: Boolean,
+    val testClassIgnored: Boolean,
```
sjrd, line 39: IIUC, this will either be `true` for all tests in a class, or `false` for all tests in a class. If that's the case, it doesn't make sense to record it in the per-test metadata. Instead, we should have a `final class TestClassMetadata`, and a new `def classMetadata(): TestClassMetadata` in `Bootstrapper`.

junit-test/shared/src/test/scala/scala/scalanative/junit/IgnoreAllTest.scala
```diff
@@ -0,0 +1,14 @@
+package scala.scalanative.junit
+
+import org.junit._
+
+import scala.scalanative.junit.utils.JUnitTest
+
+@Ignore
+class IgnoreAllTest {
```
sjrd, line 8: Consider adding a `throw new Error("unreachable")` in the body of the class, to ensure that the test class is never instantiated if it has the `@Ignore` annotation.

sjrd (changes requested):

junit-runtime/src/main/scala/scala/scalanative/junit/Bootstrapper.scala
```diff
@@ -36,7 +37,17 @@ trait Bootstrapper {
...
     val annotation: org.junit.Test
 )
+
+/** Scala Native interal JUnit test class metadata
+ *
+ *  This class is public due to implementation details. Only the junit compiler
+ *  plugin may create instances of it.
+ *
+ *  Relying on this class directly is unspecified behavior.
+ *  */
```
sjrd, line 50:
```suggestion
 */
```

junit-runtime/src/main/scala/scala/scalanative/junit/Bootstrapper.scala
```diff
@@ -21,6 +21,7 @@ trait Bootstrapper {
   def before(instance: AnyRef): Unit
   def after(instance: AnyRef): Unit
 
+  def testMetadata(): TestClassMetadata
```
sjrd, line 24: Perhaps `testClassMetadata()` or `classMetadata()`, so that it's clear this applies to the class, not one test?

WojciechMazur: Good idea, I've changed it to testClassMetadata

build.sbt
```diff
@@ -582,6 +582,7 @@ lazy val tests =
     )
     .settings(noPublishSettings)
     .settings(
+      nativeConfig ~= (_.withMode(scalanative.build.Mode.releaseFast)),
```
sjrd, line 585: Hum, was this here before and I missed it? It doesn't seem like this should be part of this PR.

build.sbt
```diff
@@ -616,6 +617,8 @@ lazy val sandbox =
     .settings(scalacOptions -= "-Xfatal-warnings")
     .settings(noPublishSettings)
     .settings(
+      nativeConfig ~= (_.withMode(scalanative.build.Mode.releaseFast))
```
sjrd, line 620: Same here.

build.sbt
WojciechMazur: Good you've spotted it! I've probably added them by mistake after switching from other branch

sjrd (approved):

PR #1962, opened 2020-10-22
Feature - Remove redundant scala overrides

WojciechMazur:
This PR extracts changes from PR #1916 related to removing redundant overrides in `scalalib` module and moving common overrides to common directory.  It should allow for easier maintenance in the future.


sjrd:
I believe the following overrides should be kept, because they contain optimizations that really make sense for Scala Native as a runtime:
* `runtime/ScalaRunTime.scala`: optimizes the generic array operations with the knowledge that Arrays have a particular representation in Scala Native
* `scala/Predef.scala` and `scala/package.scala`, because they avoid to forcibly-reach most collection types in all applications.

sjrd (approved):

PR #1964, opened 2020-10-22
Feature/ CI - Move source checks to GitHub actions

WojciechMazur:
This PR changes environment of running source-checks from Travis CI to GitHub actions. It is motivated by recurring problems with this test. As @sjrd  suggested he never had problems when running the same checks in GitHub actions.

It also can be considered as first step in parallelization of tests - scripted tests can be run in GH actions when remaining tests would be run in existing Travis CI pipeline


WojciechMazur:
I cannot see started workflow for GH actions - when I've tested this on my SN fork there was no such problem. It's possible that this repo is configured to ignore workflows coming from PR and respects only those already existing or they're limited to some set of authors.

sjrd:
Only the workflows on master are actually applied, yes.

.github/workflows/check-sources.yml
```diff
@@ -0,0 +1,10 @@
+name: Sources Check
+on: [pull_request]
+jobs:
+  check-sources:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - run: ./scripts/check-cla
+      - run: ./scripts/check-lint
```
sjrd, line 9: `check-lint` should be in a separate workflow, because it should also run on pushes to master. `check-cla` however makes sense only for pull requests.

.github/workflows/check-sources.yml
```diff
@@ -0,0 +1,10 @@
...
+
```
sjrd, line 10: Spurious blank line.

scripts/check-cla
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
sjrd, line 1: I don't know how to verify this in the GitHub review UI, but make sure that this file is marked as Executable in git.

scripts/check-lint
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
sjrd, line 1: I don't know how to verify this in the GitHub review UI, but make sure that this file is marked as Executable in git.

scripts/check-cla
sjrd: Also, consider giving an explicit `.sh` extension to this file and `check-lint`.

scripts/source-checks
```diff
@@ -2,21 +2,19 @@
 
 set -e
```
sjrd, line 3: This file should be deleted.

sjrd (changes requested):

scripts/check-lint
WojciechMazur: Checking it with `git ls-files` gave following output: 
```
100755 bc378bf5f5ab7d384ab81f3ded1969e1802f8527 0       scripts/check-cla.sh
100755 513837e7aacd9cb7ce78a023ddeaa6fb87c5f4a5 0       scripts/check-lint.sh
```
Mode for both (renamed) files is includes executable flag

sjrd (approved):

PR #1971, opened 2020-10-23
Ignore .bsp folder

catap:
sbt since 1.4.0 created `.bsp/` folder inside root. Let ignore it.


catap:
Closed because it is duplicate of https://github.com/scala-native/scala-native/pull/1967

PR #1978, opened 2020-11-01
Port to JUnit scala tests

ekrich:
These are all the Scala tests except the ones in the `scalanative` package.

Here is a unified diff since the file names changed: https://gist.github.com/ekrich/f7779ad01a74b18c999c39d08b47fd55


ekrich:
@sjrd or @WojciechMazur - I was wondering if this could be reviewed. I am going to need to rebase already.

WojciechMazur (approved):
Looks good to me. The only thing I was thinking about was SymbolTest since the backtick symbol declaration is deprecated in 2.13. I was wondering about making two versions of these tests, one for 2.13+ and the other for older versions, but probably it should be part of the 2.13 branch, instead of this PR.

ekrich:
I think that is a good plan. I have one more batch of tests that I can finish this weekend. After that I guess we could remove the old test suite. 😢

sjrd (approved):

ekrich:
@WojciechMazur Do you mind putting a ticket in for the backtick issue? I can take care of that after I finish the rest of the tests. I guess this one could be merged too. 😄

PR #1980, opened 2020-11-02
Fix Build crashes in releaseFull mode

### LINKED ISSUE #1976: Tests do not compile with releaseFull mode

WojciechMazur:
Resolves #1976 
As described in the linked issue SN was crashing in tests when releaseFull mode was used. 
The combination of this PR and #1979 allows fixing this problem.
One of the root problems was the wrong calculation of the expected type when using Sub.lub methods. This PR allows passing bound as an optional argument to the mentioned method. Also in most of the usages of this method, we make sure to pass bound type basing on the current context. 

To be noted, in one of the recent tests I've discovered that the current Xmx setting of 5G is not sufficient in release full mode - it throws OutOfMemoryError when dumping lowered Defns (the content of all methods is stored in memory, sorted, and then written to file). It's recommended to disable the default nativeDump setting in `MyScalaNativePlugin` or increase Xmx (build passed with Xmx 15G, not tested with the lower value). This problem <del>should be addressed in later PR.</del> was fixed in #1982 

Interflow optimization in release full mode took for unit-tests took 51 minutes. Additional ~30 minutes are needed for compilation using clang.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
+++ b/tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
@@ -75,7 +75,7 @@ final class MergeProcessor(insts: Array[Inst],
         val mergeEmitted = mutable.Map.empty[Op, Val]
         val newEscapes   = mutable.Set.empty[Addr]
 
-        def mergePhi(values: Seq[Val], bound: Option[Type] = None): Val = {
+        def mergePhi(values: Seq[Val], bound: Option[Type]): Val = {
           if (values.distinct.size == 1) {
             values.head
           } else {
@@ -130,7 +130,7 @@ final class MergeProcessor(insts: Array[Inst],
                       case _                      => Val.Virtual(addr)
                     }
                   }
-                  mergeHeap(addr) = EscapedInstance(mergePhi(values))
+                  mergeHeap(addr) = EscapedInstance(mergePhi(values, None))
                 case VirtualInstance(headKind, headCls, headValues) =>
                   val mergeValues = headValues.zipWithIndex.map {
                     case (_, idx) =>
--- a/tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
+++ b/tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
@@ -75,7 +75,7 @@ final class MergeProcessor(insts: Array[Inst],
         val mergeEmitted = mutable.Map.empty[Op, Val]
         val newEscapes   = mutable.Set.empty[Addr]
 
-        def mergePhi(values: Seq[Val]): Val = {
+        def mergePhi(values: Seq[Val], bound: Option[Type] = None): Val = {
           if (values.distinct.size == 1) {
             values.head
           } else {
@@ -86,11 +86,7 @@ final class MergeProcessor(insts: Array[Inst],
               case (s, v) =>
                 s.materialize(v)
             }
-            val name = mergeFresh()
-            val bound = params.headOption match {
-              case Some(Val.Local(_, ty: Type.RefKind)) => ty
-              case _                                    => Type.Ref(Global.Top("java.lang.Object"))
-            }
+            val name    = mergeFresh()
             val paramty = Sub.lub(materialized.map(_.ty), bound)
             val param   = Val.Local(name, paramty)
             mergePhis += MergePhi(param, names.zip(materialized))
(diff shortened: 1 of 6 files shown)


tools/src/main/scala/scala/scalanative/interflow/Opt.scala
```diff
@@ -38,6 +38,11 @@ trait Opt { self: Interflow =>
     val fresh = Fresh(0)
     val state = new State(Local(0))
 
+    // Interflow usually infers better types on our erased type system
+    // than scalac, yet we live it a benefit of the doubt and make sure
```
ekrich, line 42: Maybe: `yet we leave it as a benefit`?

WojciechMazur: I've not seen this type when copy-pasting. Consider it fixed. Thanks for spotting this one!

tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
```diff
@@ -75,7 +75,7 @@ final class MergeProcessor(insts: Array[Inst],
         val mergeEmitted = mutable.Map.empty[Op, Val]
         val newEscapes   = mutable.Set.empty[Addr]
 
-        def mergePhi(values: Seq[Val]): Val = {
+        def mergePhi(values: Seq[Val], bound: Option[Type] = None): Val = {
```
sjrd, line 78: Consider removing the default value, so that it's painfully obvious at call site when we use `None` (which is the dangerous value, so it should be visible):
```suggestion
        def mergePhi(values: Seq[Val], bound: Option[Type]): Val = {
```

tools/src/main/scala/scala/scalanative/linker/Sub.scala
```diff
@@ -61,7 +61,8 @@ object Sub {
     }
   }
 
-  def lub(tys: Seq[Type], bound: Type)(implicit linked: linker.Result): Type = {
+  def lub(tys: Seq[Type], bound: Option[Type] = None)(
```
sjrd, line 64: Same here:
```suggestion
  def lub(tys: Seq[Type], bound: Option[Type])(
```

sjrd (changes requested):
Basically looks good. I only suggest to remove the default values.

sjrd (approved):

PR #1982, opened 2020-11-03
Lower memory usage in dump stage

WojciechMazur:
As mentioned in #1980 one of the remaining problems with a release full mode was failing while dumping lowered defns. 
This PR resolves the problem with dumping large `.hnir` files by writing directly chunked defns to separate files and merging them.

### DIFF
--- a/util/src/main/scala/scala/scalanative/util/ShowBuilder.scala
+++ b/util/src/main/scala/scala/scalanative/util/ShowBuilder.scala
@@ -44,5 +44,17 @@ object ShowBuilder {
   }
 
   final class FileShowBuilder(protected val out: java.io.Writer)
-      extends ShowBuilder
+      extends ShowBuilder {
+    override def str(value: Any): Unit = {
+      val body = value.toString
+      out.append {
+        /* Writer may have problems with strings containing surrogate pairs and throw exception,
+         * but such cases may be observed only when dumping linked and optimized defns.
+         * In lowered code strings are already transformed into array[byte] and effectively into safe string
+         */
+        if (!body.exists(_.isSurrogate)) body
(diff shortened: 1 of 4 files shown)


util/src/main/scala/scala/scalanative/util/ShowBuilder.scala
```diff
@@ -44,5 +44,17 @@ object ShowBuilder {
...
+      extends ShowBuilder {
+    override def str(value: Any): Unit = {
+      val body = value.toString
+      out.append {
+        /* Writer may have problems with strings containing surrogate pairs and throw exception,
+         * but such cases may be observed only when dumping linked and optimized defns.
+         * In lowered code strings are already transformed into array[byte] and effectively into safe string
+         */
+        if (!body.exists(_.isSurrogate)) body
+        else body.filterNot(_.isSurrogate)
```
sjrd, line 56: That is very suspicious. Instead of doing this at this level (which will actually drop stuff without notice), we should create the `Writer` so that invalid surrogates use a replacement instead of throwing an error.

sjrd (changes requested):
This needs a rebase. The logic in `dump` has since changed to use Java Streams, I believe.

sjrd:
The CI says this is broken in 2.13.

WojciechMazur:
Since storing generated code from each of the threads in buffers was removed in the past PR source of all evil in dumping code was already fixed.  The only positive feature this PR provides in theory faster generation of the output file. With complexity increased a lot. Dumping intermediate code is mainly used by us in the debugging process and even in a single thread, it's quite fast (few seconds for unit-tests). 
Probably this PR adds little or even no value to the project so maybe we should reject it?

PR #1984, opened 2020-11-05
Fix #805, extending class containing main does not include it in scope

### LINKED ISSUE #805: Extending a class which should bring main() into scope does not work properly
This can't run without overriding main and Scala Native, same app works fine on Scala.
https://github.com/schrepfler/scala-native-cli

```
import caseapp._

case class ExampleOptions(
  name: String
)

object Hello extends CaseApp[ExampleOptions] {

  //  needed as scala-native at the moment does not import properly the main from CaseApp
  override def main(args: Array[String]) = super.main(args)

  def run(options: ExampleOptions, arg: RemainingArgs): Unit = {
    println(s"Hello, ${options.name}!")
  }

}
```

WojciechMazur:
Resolves #805
Currently classes extending trait containing `main` method were correctly included to scope allowing for defining common SN startup definitions. However when extending class such main method was not included resulting in linking error. 
This PR  fixes this PR, by resolving real entry type signature.


tools/src/test/scala/scala/scalanative/linker/ClassReachabilitySuite.scala
```diff
@@ -323,4 +324,32 @@ class ClassReachabilitySuite extends ReachabilitySuite {
...
+    val source = """
+	  abstract class Parent {
+        def bar: Int
+		def main(): Unit = println(bar)
+      }
+      
+      object Test extends Parent {
+	    val bar: Int = 0
+      }
+	  """.stripMargin
```
lolgab, line 343: You have `stripMargin` here but no margin and strange indentation.

WojciechMazur: Oh, thanks, forgot about that. I've removed margin since all other sources don't contain it, but forget to remove strip margin...and changed indentation later when testing...

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -44,9 +44,22 @@ class Reach(
     // in reachUnavailable
     defns ++= done.valuesIterator.filter(_ != null)
 
+    val resultEntries = entries.head match {
```
sjrd, line 47: It seems quite dangerous to rely on `entries.head`. I wouldn't expect there to be any guarantee about the order or content of these entries.

sjrd (changes requested):

tools/src/main/scala/scala/scalanative/linker/Reach.scala
WojciechMazur: That's true, I've prepared a more generic solution that updates the original main with resolved one.
 Unfortunately, because of that, I needed to update all ReachabilitySuites - all of them were using `def main(): Unit` instead of `def main(_: Array[String]): Unit` which is not a real main method (sbt would not detect it as an entry point)

tools/src/test/scala/scala/scalanative/linker/TraitReachabilitySuite.scala
```diff
@@ -30,7 +30,8 @@ class TraitReachabilitySuite extends ReachabilitySuite {
   val ObjectInit: Global = g("java.lang.Object", Sig.Ctor(Seq.empty))
   val Test: Global = g("Test$")
```
ekrich, line 31: Should you refactor out `val TestClass = "Test$"` like you did above for `MainClass`?

sjrd (approved):

PR #1986, opened 2020-11-06
Update travis_setup.sh

xuwei-k:
I want to use this script in GitHub Actions `ubuntu-latest` but got following errors

```
sudo find /usr -name '*libunwind*' -delete
find: cannot delete ‘/usr/share/doc/libunwind8’: Directory not empty
Error: Process completed with exit code 1.
```

### DIFF
--- a/scripts/travis_setup.sh
+++ b/scripts/travis_setup.sh
@@ -21,7 +21,7 @@ else
     sudo apt-get update
 
     # Remove pre-bundled libunwind
-    sudo find /usr -name "*libunwind*" -delete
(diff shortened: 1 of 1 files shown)


ekrich:
If we are updating this for `0.4.0-M2` then I think `re2` is no longer required. I am not sure about the `libunwind` how it fits in exactly except we wanted version 8.

I think this file has fallen out of favor now that we use Docker containers for CI.

I still think it is useful for setup or CI if you are not using containers. Also, `jq` on the macOS side was for the CLA check if I recall correctly.

PR #1987, opened 2020-11-06
Fix #1711: Ignore non-jar non-directory elements on the classpath

### LINKED ISSUE #1711: Build fails if .so files are put in the "lib" folder
Compilation fails when .so files are copied in the "lib" folder (SBT's unmanaged dependencies folder, see https://www.scala-sbt.org/1.x/docs/Library-Dependencies.html).

The error is similar to the one described in this previous issue: https://github.com/scala-native/scala-native/issues/911

```java.lang.UnsupportedOperationException: Neither a jar, nor a directory```

The solution can consist in using a different directory or to change the SBT's "unmanaged dependencies folder". For instance this additional settings will allow to use the "lib" folder to contain .so files:
```unmanagedBase := baseDirectory.value / "jars"```

This behavior should be described in the SN FAQ or in a new troubleshooting section.

ekrich:
We could consider allowing the build to include external libraries to link in the future.

This fix avoids passing non-jar files in the classpath into the build interior.

It centralizes the classpath filters into one location making it easier to maintain but also moving it out of the plugin and into tools. Moving it out of the plugin means that tool integrators just pass the classpath and it is up to Scala Native to do sanity checks to avoid breaking the build. There was a previous filter for Issue #911 which was moved from the plugin. 

This was tested both for the `.so` file in the _sbt_ `lib` directory and the no files in main for the previous issue.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -117,11 +117,9 @@ object ScalaNativePluginInternal {
         val mainClass = selectMainClass.value.getOrElse {
           throw new MessageOnlyException("No main class detected.")
         }
-        val classpath = fullClasspath.value
-          .map(_.data.toPath)
-          .filter(f => Files.exists(f))
-        val maincls = mainClass + "$"
-        val cwd     = nativeWorkdir.value.toPath
+        val classpath = fullClasspath.value.map(_.data.toPath)
+        val maincls   = mainClass + "$"
+        val cwd       = nativeWorkdir.value.toPath
 
         val logger = streams.value.log.toLogger
         build.Config.empty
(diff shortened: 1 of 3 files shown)


WojciechMazur (approved):

sjrd:
The corresponding code in Scala.js throws for non-jar non-directory entries on the classpath:
https://github.com/scala-js/scala-js/blob/9348e8f3cafa3a02c9fd8b498968e7ffbe63ee33/linker/jvm/src/main/scala/org/scalajs/linker/PathIRContainer.scala#L46
This never caused any issue.

IMO those things should not end up on the classpath to begin with, and if they do, it's either a) the user's fault for putting them in `lib` in the first place or b) the build tool's fault for picking up non-jar files in the `lib` folder.

What makes Scala Native different in this regard? Why are the files ignored? Does the JVM ignore non-jar non-directory entries on the classpath?

sjrd:
> Does the JVM ignore non-jar non-directory entries on the classpath?

Yes, apparently it does.

Issue #2012, opened 2020-11-16
Optimizing in debug mode produces warnings on JUnit test


ekrich:
Remove the comments `/* */` around the imports and inside the test body and compile `scala.scalanative.regex.ExecTest`.

Run as shown in the following output:

```
sbt:scala-native> tests/testOnly scala.scalanative.regex.ExecTest
[info] compiling 7 Scala sources to /Users/eric/workspace/scala-native/unit-tests/target/scala-2.12/classes ...
[info] compiling 197 Scala sources to /Users/eric/workspace/scala-native/unit-tests/target/scala-2.12/test-classes ...
[info] Linking (2574 ms)
[info] Checking intermediate code (1322 ms)
[info] Dumping intermediate code (linked) (2921 ms)
[info] Discovered 4744 classes and 30631 methods
[info] Optimizing (debug mode) (8577 ms)
[warn] Found 1 errors on @"M19java.io.PrintStreamD6formatL16java.lang.StringLAL16java.lang.Object_L19java.io.PrintStreamEO" :
[warn]   0000  didopt def @"M19java.io.PrintStreamD6formatL16java.lang.StringLAL16java.lang.Object_L19java.io.PrintStreamEO" : (@"T19java.io.PrintStream", @"T16java.lang.String", array[@"T16java.lang.Object"]) => @"T19java.io.PrintStream" {
[warn]   0001  %40000(%1 : @"T19java.io.PrintStream", %2 : @"T16java.lang.String", %3 : array[@"T16java.lang.Object"]):
[warn]   0002    %50002 = ieq[@"T16java.lang.Object"] %1 : @"T19java.io.PrintStream", null
[warn]   0003    if %50002 : bool then %60000 else %70000
[warn]   0004  %70000:
[warn]   0005    jump %180000
[warn]   0006  %60000:
[warn]   0007    jump %160000
[warn]   0008  %160000:
[warn]   0009    %160001 = classalloc @"T30java.lang.NullPointerException"
[warn]   0010    %160002 = fieldstore[bool] %160001 : !?@"T30java.lang.NullPointerException", @"M19java.lang.ThrowableF18writableStackTracePT19java.lang.Throwable", true
[warn]   0011    %160003 = fieldstore[bool] %160001 : !?@"T30java.lang.NullPointerException", @"M19java.lang.ThrowableF17enableSuppressionPT19java.lang.Throwable", true
[...]

PR #2023, opened 2020-11-19
Restore ExecTest except slow test

ekrich:
I removed the output and restored all but the one slow test.

Issue: #2012 is confirmed fixed.

### DIFF
--- a/unit-tests/src/test/scala/scala/scalanative/regex/ExecTest.scala
+++ b/unit-tests/src/test/scala/scala/scalanative/regex/ExecTest.scala
@@ -451,7 +451,6 @@ class ExecTest {
 
         case 'C' | 'N' | 'T' | '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' |
             '8' | '9' =>
-          //System.err.println("skip: %s\n".format(line))
           break
         case _ =>
       }
--- a/unit-tests/src/test/scala/scala/scalanative/regex/ExecTest.scala
+++ b/unit-tests/src/test/scala/scala/scalanative/regex/ExecTest.scala
@@ -1,6 +1,6 @@
 package scala.scalanative
 package regex
-/*
+
 import java.io.InputStreamReader
 import java.io.File
 import java.io.FileInputStream
@@ -11,7 +11,7 @@ import java.util.{Arrays, Collections}
 
 import scala.util.control.NonFatal
 import scala.util.control.Breaks._
- */
+
 import org.junit.Ignore
 import org.junit.Test
 import org.junit.Assert._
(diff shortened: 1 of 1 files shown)


unit-tests/src/test/scala/scala/scalanative/regex/ExecTest.scala
```diff
@@ -463,7 +451,7 @@ class ExecTest {
 
         case 'C' | 'N' | 'T' | '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' |
             '8' | '9' =>
-          System.err.println("skip: %s\n".format(line))
+          //System.err.println("skip: %s\n".format(line))
```
sjrd, line 454: Consider removing the line instead. It was probably left there by mistake.

sjrd (changes requested):

sjrd (approved):

LeeTibbert (approved):
Eric,  LGTM

Thank you for rehabilitating & enabling the tests of this file.  Having the test of the published example working is a win for "Say what you do and then do it!" integrity.

The Fowler tests are a bonus.  

In the early stages of this work, Eric and I chatted about the exhaustive tests.
I concur with the approach of leaving the exhaustive tests in the file but Ignore'ing them.
They are too heavyweight for every CI run.

I ran those original tests from the RE2 project when I was porting the re2s code 
and some future devo will appreciate having the original available when they chase my
mistakes.

ekrich:
Hey guys. I am not really used to the multi Scala version in Scala Native. I needed to push another change for the `Array` comparison in `2.13`. Now using:
```
sbt:scala-native> +tests/testOnly scala.scalanative.regex.ExecTest
```
I also found one other print to console I need to hunt down so I'll need one more change.

unit-tests/src/test/scala/scala/scalanative/regex/ExecTest.scala
ekrich: I found another one that was outputting but I think it might be helpful for someone that has to work on this code again. The other one would leave an empty `if` - I could put a comment above `// left for debugging` or something. WDYT?

ekrich:
It runs fine but I am getting `[warning]`s on `2.11` as per Issue #2012 so I am not sure if we want to wait for that to be fixed or not.

WojciechMazur:
> It runs fine but I am getting `[warning]`s on `2.11` as per Issue #2012 so I am not sure if we want to wait for that to be fixed or not.

These warnings are present in all builds based on the current master. I think we can merge this one.

WojciechMazur:
@ekrich can you please rebase this branch onto master? Probably it would not cause any problems, but I would be calmer when I will see green CI outputs. There are no conflicts so it should be an easy fast-forward.

ekrich:
No problem. Hopefully it will be all green with no warnings.

PR #2027, opened 2020-11-22
Avoid j.l.Class.newInstance(), which is deprecated since JDK 9

catap:
`java.lang.Class.newInstance` was marked as deprecated since JDK9 and JDK15 removed almost all old deprecated methods.


ekrich (approved):
LGTM. I confirmed that the methods existed in JDK8 so it is backward compatible.

WojciechMazur (approved):

PR #2032, opened 2020-11-23
BoxedUnitArray should extend ObjectsArray

### LINKED ISSUE #2025: Unit IllegalArgumentException argument type mismatch at runtime
Running `scala-collection-compat` tests fail in a PR https://github.com/scala/scala-collection-compat/pull/387 to enable the library for the upcoming `0.4.0` release. Fails in `2.11, 2.12, 2.13`.

Minimal reproducer test for `2.13` only without the above mentioned library.
```Scala
import org.junit.Test
import org.junit.Assert._

import scala.collection.immutable.ArraySeq

// Collection Compat Issues
class UnitTest {

  // from ArraySeqTest.t6727_and_t6440_and_8627
  @Test def lazyListUnit(): Unit = {
    assertTrue(
      LazyList.continually(()).filter(_ => true).take(2) == Seq((), ())
    )
    assertTrue(
      LazyList.continually(()).filterNot(_ => false).take(2) == Seq((), ())
    )
  }

  // from LazyListTest.slice
  @Test def implicitFailsAtRuntime(): Unit = {
    // fails at runtime
    implicit def array2ArraySeq[T](array: Array[T]): ArraySeq[T] =
      ArraySeq.unsafeWrapArray(array)

    def unit1(): Unit = {}
    def unit2(): Unit = {}
    assertEquals("Units are equal", unit1, unit2)
    // unitArray is actually an instance of Immutable[BoxedUnit], the check to which is actually checked slice
    // implementation of ofRef
    val unitArray: ArraySeq[Unit] = Array(unit1, unit2, unit1, unit2)
    check(unitArray, Array(unit1, unit1), Array(unit1, unit1))

  }

  private def check[T](
      array: ArraySeq[T],
      expectedSliceResult1: ArraySeq[T],
      expectedSliceResult2: ArraySeq[T]
  ) {
    assertEquals(array, array.slice(-1, 4))
    assertEquals(array, array.slice(0, 5))
    assertEquals(array, array.slice(-1, 5))
    assertEquals(expectedSliceResult1, array.slice(0, 2))
    assertEquals(expectedSliceResult2, array.slice(1, 3))
    assertEquals(ArraySeq.empty[Nothing], array.slice(1, 1))
    assertEquals(ArraySeq.empty[Nothing], array.slice(2, 1))
  }

}

```

WojciechMazur:
When investigating #2025  I've seen that `BoxedUnitArray` is not correctly recognized as `Array[AnyRef] `
This PR fixes this issue and also contains few additional fixes:
- `Class[BoxedUnitArray].componentType` correctly returns `scala.runtime.BoxedUnit`
- `Class[BoxedUnitArray].isArray` correctly returns `true`
- `Array.copy` uses `isAssignableFrom` instead of directly comparing types using `eq`. It was needed to allow assigning coping `BoxedUnitArray` to `ObjectArray` as it is legal action on JVM. 

1 bug workaround was added: all class VirtualTables are enriched by `clone: Object` method signature in order to deal with newly discovered issue #2035 showing up in `RelectiveProxyTest` after applying changes from this PR. 

Resolves #2025


sjrd:
It seems the changes here are going in the right direction under the constraint to keep `Array[Unit]` being a separate thing.

However, on the JVM and in Scala.js, an `Array[Unit]` is simply encoded as an `Array[scala.runtime.BoxedUnit]` (which is therefore an array of references). Is there any reason that we can't do the same in Native? It seems that doing so would dramatically simplify a lot of things.

WojciechMazur:
> However, on the JVM and in Scala.js, an `Array[Unit]` is simply encoded as an `Array[scala.runtime.BoxedUnit]` (which is therefore an array of references). Is there any reason that we can't do the same in Native? It seems that doing so would dramatically simplify a lot of things.

In SN `Array[Unit]` is always encoded in the same way, however, all `Array[AnyRef]` are internally stored as `ObjectArray` making it impossible to distinguish two types of arrays. I believe that Scala.js has a richer implementation of classes, allowing them to get the correct component type.
But, yes, It might simplify things a lot - simply removing `BoxedUnitArray` and its usages from our code would fix issue #2025, but when testing this approach locally with 2.13.3 I've got failed tests in `jl.RuntimeTest`, `jl.ProcessTest`, and `ju.ArrayDequeueTest`

If we don't mind about the correctness of `Array[Unit]` class component type or distinguishing it from Array[Object] then we can try going with that approach.

sjrd:
> If we don't mind about the correctness of `Array[Unit]` class component type or distinguishing it from Array[Object] then we can try going with that approach.

No, I don't think we care about that. Or at least, we don't care more for `Unit` than for `String` or any other type.

WojciechMazur:
Closed in favor of #2036

Issue #1183, opened 2018-03-01
Proposal: add support cross-triple in native pipeline


Andrei-Pozolotin:
is seems that both current [ScalaNativePluginInternal.scala](https://github.com/scala-native/scala-native/blob/master/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala) and future #1177 PR #1143 lack support cross-triple build https://clang.llvm.org/docs/CrossCompilation.html

ekrich:
Scala Native now allows specifying the Target Triple via https://github.com/scala-native/scala-native/pull/2033

PR #2033, opened 2020-11-24
Remove target triple discovery code

ekrich:
After several rounds of discussion first around cross compiling and removing the target triple discovery and now removing the target triple, I am more convinced than ever that this is a prudent direction.

First @lolgab proposed this target triple PR https://github.com/scala-native/scala-native/pull/1904 when I was proposing this https://github.com/scala-native/scala-native/pull/1905

Now @catap is basically saying the same thing with this PR https://github.com/scala-native/scala-native/pull/2031 After some discussion I suggested that I put together a more conservative PR in hoping we could get some consensus.

This is a more conservative PR as it doesn't completely remove target triple code which is advertised as being needed here: http://clang.llvm.org/docs/CrossCompilation.html#general-cross-compilation-options-in-clang

A future PR can address exposing target triple and other related improvements. If the target triple is set correctly for the platform then there are no errors.

For full disclosure this also includes changing the internal plugin to use `sbt` slash syntax and swipes a super small change `.o` -> `oExt` as small clean up changes.

Thank you @lolgab and @catap for pushing for these changes and if this is ok then maybe we can add user support for target triple if the community agrees.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -54,7 +54,6 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeDump =
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
-
   }
 
   @deprecated("use autoImport instead", "0.3.7")
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -54,6 +54,9 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeDump =
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
+
+    val nativeTarget =
+      taskKey[String]("The LLVM target triple for cross compiling.")
   }
 
   @deprecated("use autoImport instead", "0.3.7")
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -55,8 +55,6 @@ object ScalaNativePlugin extends AutoPlugin {
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
 
-    val targetTriple =
-      taskKey[String]("The optional LLVM target triple for cross compiling.")
   }
 
   @deprecated("use autoImport instead", "0.3.7")
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -55,8 +55,8 @@ object ScalaNativePlugin extends AutoPlugin {
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
 
-    val nativeTarget =
-      taskKey[String]("The LLVM target triple for cross compiling.")
+    val targetTriple =
+      taskKey[String]("The optional LLVM target triple for cross compiling.")
   }
 
   @deprecated("use autoImport instead", "0.3.7")
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -43,6 +43,7 @@ object ScalaNativePluginInternal {
     nativeClangPP := nativeConfig.value.clangPP.toFile,
     nativeCompileOptions := nativeConfig.value.compileOptions,
     nativeLinkingOptions := nativeConfig.value.linkingOptions,
+    nativeTarget := nativeConfig.value.targetTriple,
     nativeMode := nativeConfig.value.mode.name,
     nativeGC := nativeConfig.value.gc.name,
     nativeLTO := nativeConfig.value.lto.name,
@@ -82,9 +83,6 @@ object ScalaNativePluginInternal {
   )
 
   lazy val scalaNativeConfigSettings: Seq[Setting[_]] = Seq(
-    nativeTarget := { // placeholder for future use
-      ""
-    },
     nativeLink / artifactPath := {
       crossTarget.value / (moduleName.value + "-out")
     },
@@ -101,6 +99,7 @@ object ScalaNativePluginInternal {
         .withClangPP(nativeClangPP.value.toPath)
         .withCompileOptions(nativeCompileOptions.value)
         .withLinkingOptions(nativeLinkingOptions.value)
+        .withTargetTriple(nativeTarget.value)
         .withGC(build.GC(nativeGC.value))
         .withMode(build.Mode(nativeMode.value))
         .withLTO(build.LTO(nativeLTO.value))
@@ -124,7 +123,6 @@ object ScalaNativePluginInternal {
           .withMainClass(maincls)
           .withClassPath(classpath)
           .withWorkdir(cwd)
-          .withTargetTriple(nativeTarget.value)
           .withCompilerConfig(nativeConfig.value)
       }
 
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -43,7 +43,6 @@ object ScalaNativePluginInternal {
     nativeClangPP := nativeConfig.value.clangPP.toFile,
     nativeCompileOptions := nativeConfig.value.compileOptions,
     nativeLinkingOptions := nativeConfig.value.linkingOptions,
-    targetTriple := nativeConfig.value.targetTriple.getOrElse(""),
     nativeMode := nativeConfig.value.mode.name,
     nativeGC := nativeConfig.value.gc.name,
     nativeLTO := nativeConfig.value.lto.name,
@@ -99,7 +98,6 @@ object ScalaNativePluginInternal {
         .withClangPP(nativeClangPP.value.toPath)
         .withCompileOptions(nativeCompileOptions.value)
         .withLinkingOptions(nativeLinkingOptions.value)
-        .withTargetTriple(toOption(targetTriple.value))
         .withGC(build.GC(nativeGC.value))
         .withMode(build.Mode(nativeMode.value))
         .withLTO(build.LTO(nativeLTO.value))
@@ -220,8 +218,5 @@ object ScalaNativePluginInternal {
     if (l.compareAndSet(prev, r :: prev)) r
     else registerResource(l, r)
   }
-
-  def toOption(s: String): Option[String] =
-    if (s.trim().isEmpty()) None else Some(s)
 
 }
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -21,9 +21,6 @@ object ScalaNativePluginInternal {
   val nativeWarnOldJVM =
     taskKey[Unit]("Warn if JVM 7 or older is used.")
 
-  val nativeTarget =
-    taskKey[String]("Target triple.")
-
   val nativeWorkdir =
     taskKey[File]("Working directory for intermediate build files.")
 
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -43,7 +43,7 @@ object ScalaNativePluginInternal {
     nativeClangPP := nativeConfig.value.clangPP.toFile,
     nativeCompileOptions := nativeConfig.value.compileOptions,
     nativeLinkingOptions := nativeConfig.value.linkingOptions,
-    nativeTarget := nativeConfig.value.targetTriple,
+    targetTriple := nativeConfig.value.targetTriple.getOrElse(""),
     nativeMode := nativeConfig.value.mode.name,
     nativeGC := nativeConfig.value.gc.name,
     nativeLTO := nativeConfig.value.lto.name,
@@ -99,7 +99,7 @@ object ScalaNativePluginInternal {
         .withClangPP(nativeClangPP.value.toPath)
         .withCompileOptions(nativeCompileOptions.value)
         .withLinkingOptions(nativeLinkingOptions.value)
-        .withTargetTriple(nativeTarget.value)
+        .withTargetTriple(toOption(targetTriple.value))
         .withGC(build.GC(nativeGC.value))
         .withMode(build.Mode(nativeMode.value))
         .withLTO(build.LTO(nativeLTO.value))
@@ -220,5 +220,8 @@ object ScalaNativePluginInternal {
     if (l.compareAndSet(prev, r :: prev)) r
     else registerResource(l, r)
   }
+
+  def toOption(s: String): Option[String] =
+    if (s.trim().isEmpty()) None else Some(s)
 
 }
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -85,12 +85,10 @@ object ScalaNativePluginInternal {
   )
 
   lazy val scalaNativeConfigSettings: Seq[Setting[_]] = Seq(
-    nativeTarget := interceptBuildException {
-      val cwd   = nativeWorkdir.value.toPath
-      val clang = nativeClang.value.toPath
-      Discover.targetTriple(clang, cwd)
+    nativeTarget := { // placeholder for future use
+      ""
     },
-    artifactPath in nativeLink := {
+    nativeLink / artifactPath := {
       crossTarget.value / (moduleName.value + "-out")
     },
     nativeWorkdir := {
@@ -114,7 +112,7 @@ object ScalaNativePluginInternal {
         .withDump(nativeDump.value)
     },
     nativeLink := {
-      val outpath = (artifactPath in nativeLink).value
+      val outpath = (nativeLink / artifactPath).value
       val config = {
         val mainClass = selectMainClass.value.getOrElse {
           throw new MessageOnlyException("No main class detected.")
@@ -140,7 +138,7 @@ object ScalaNativePluginInternal {
       outpath
     },
     run := {
-      val env    = (envVars in run).value.toSeq
+      val env    = (run / envVars).value.toSeq
       val logger = streams.value.log
       val binary = nativeLink.value.getAbsolutePath
       val args   = spaceDelimited("<arg>").parsed
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -63,22 +63,25 @@ object Build {
       ScalaNative.logLinked(fconfig, linked)
       val optimized = ScalaNative.optimize(fconfig, linked)
 
-      IO.getAll(workdir, "glob:**.ll").foreach(Files.delete)
-      ScalaNative.codegen(fconfig, optimized)
-      val generated = IO.getAll(workdir, "glob:**.ll")
+    // clean ll files
+    IO.getAll(workdir, "glob:**.ll").foreach(Files.delete)
+
+    val generated = ScalaNative.codegen(fconfig, optimized)
 
       val nativelibs   = NativeLib.findNativeLibs(fconfig.classPath, workdir)
       val nativelib    = NativeLib.findNativeLib(nativelibs)
       val unpackedLibs = nativelibs.map(LLVM.unpackNativeCode(_))
 
-      val objectFiles = config.logger.time("Compiling to native code") {
-        val nativelibConfig =
-          fconfig.withCompilerConfig(
-            _.withCompileOptions("-O2" +: fconfig.compileOptions))
+    val objectPaths = config.logger.time("Compiling to native code") {
+      val nativelibConfig =
+        fconfig.withCompilerConfig(
+          _.withCompileOptions("-O2" +: fconfig.compileOptions))
+      val libObjectPaths =
         LLVM.compileNativelibs(nativelibConfig, linked, unpackedLibs, nativelib)
-        LLVM.compile(fconfig, generated)
-      }
-
-      LLVM.link(config, linked, objectFiles, unpackedLibs, outpath)
+      val llObjectPaths = LLVM.compile(fconfig, generated)
+      libObjectPaths ++ llObjectPaths
     }
+
+    LLVM.link(config, linked, objectPaths, outpath)
+  }
 }
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -70,8 +70,7 @@ object Build {
       val nativelib    = NativeLib.findNativeLib(nativelibs)
       val unpackedLibs = nativelibs.map(LLVM.unpackNativeCode(_))
 
-      val msg = s"Compiling to native code (${targetTripleMsg(config)})"
-      val objectPaths = config.logger.time(msg) {
+      val objectPaths = config.logger.time("Compiling to native code") {
         val nativelibConfig =
           fconfig.withCompilerConfig(
             _.withCompileOptions("-O2" +: fconfig.compileOptions))
@@ -86,7 +85,4 @@ object Build {
 
       LLVM.link(config, linked, objectPaths, outpath)
     }
-
-  private def targetTripleMsg(config: Config): String =
-    config.compilerConfig.targetTriple.getOrElse("default target triple")
 }
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -72,7 +72,8 @@ object Build {
       val nativelib    = NativeLib.findNativeLib(nativelibs)
       val unpackedLibs = nativelibs.map(LLVM.unpackNativeCode(_))
 
-    val objectPaths = config.logger.time("Compiling to native code") {
+    val msg = s"Compiling to native code (${targetTripleMsg(config)})"
+    val objectPaths = config.logger.time(msg) {
       val nativelibConfig =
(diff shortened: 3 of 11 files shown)


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -83,12 +83,10 @@ object ScalaNativePluginInternal {
   )
 
   lazy val scalaNativeConfigSettings: Seq[Setting[_]] = Seq(
-    nativeTarget := interceptBuildException {
-      val cwd   = nativeWorkdir.value.toPath
-      val clang = nativeClang.value.toPath
-      Discover.targetTriple(clang, cwd)
+    nativeTarget := { // placeholder for future use
+      ""
```
catap, line 87: by default it is empty string, isn't it?

Also I feel that `Option[String]` for this case is more semantic.

ekrich: Totally agree but that should probably be addressed in a PR that allows cross compiling.

catap: I feel that you may remove it here. Can't you?

ekrich: I think people may want to use `-target` vs `-arch` but we don't want rip out `-target` if we may need it. If this PR does the job for you, it is much smaller and we are not preventing the other PRs either.

catap: Let close this discussion because it is style question.

ekrich:
If any one wants to play with this PR to prove that `-target` will work if supplied.

- Find target - https://stackoverflow.com/questions/15036909/clang-how-to-list-supported-target-architectures
- Put target in here - https://github.com/scala-native/scala-native/pull/2033/files#diff-a610de80dc17f463fea5440deffcec5b2b70a7dbeef002624b94f6fa894f4b77R87

For my machine I get the following: `x86_64-apple-darwin18.7.0`. Compiling produces the following warnings: 
`[error] warning: overriding the module target triple with x86_64-apple-macosx10.14.0 [-Woverride-module]`.

 If this triple is used, `x86_64-apple-macosx10.14.0` their are no errors.

So if you supply the target triple you get warnings or errors if you don't use the correct target triple. If you supply the correct `-target` it works fine. If you don't supply the `-target` the platform does the right thing.

Remember to `reload` in `sbt` to compile the tools.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
ekrich: You mean `-arch` versus `-target`?

catap: @ekrich I mean to remove `nativeTarget` with empty string because it is empty string by default.

But I might be wrong and it is the place where this key has it's default value.

ekrich: Oh ok, `sbt` will not allow the internal `nativeTarget` to be unassigned. I tried to remove that assignment/block and the build will not load. There might be some other way but this seemed the best for now and it leaves a placeholder and a place to experiment for now.

catap: @ekrich it is clear. Thanks.

catap (approved):
After thinking a while I'd like to close my PR and move forward with yours.

It is looks like a clean-clang-way and not apple's hack, that allows to make cross building very easy.

ekrich:
@sjrd @WojciechMazur I have converted this to a draft because I think we might want to consider exposing `nativeTarget` in the plugin to allow cross compiling using Target Triple. Is it preferred to move it from `Config` to `NativeConfig` since it will now be a user setting?

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -101,6 +105,7 @@ object CodeGen {
       }
 
       dir.merge(Seq(body), headers)
+      headers
```
catap, line 108: maybe make it clear and split herer to `-body.ll` and `-herader.ll` and merge them into `id.ll`?

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -43,29 +43,33 @@ object CodeGen {
...
       // This is an adhoc form of LTO. We use it in release mode if
       // Clang's LTO is not available.
-      def single(): Unit = {
+      def single(): Seq[Path] = {
         val sorted = assembly.sortBy(_.name.show)
-        new Impl(config.targetTriple, env, sorted)
-          .gen(id = "out", workdir)
+        Seq(
+          new Impl(config.targetTriple, env, sorted)
+            .gen(id = "out", workdir))
```
catap, line 72: I feel that add `targetTripple` for a name isn't bad idea here. Am I wrong?

tools/src/main/scala/scala/scalanative/build/ScalaNative.scala
```diff
@@ -87,12 +87,11 @@ private[scalanative] object ScalaNative {
...
   /** Given low-level assembly, emit LLVM IR for it to the buildDirectory. */
   def codegen(config: Config, linked: linker.Result): Seq[Path] = {
-    config.logger.time("Generating intermediate code") {
+    val llPaths = config.logger.time("Generating intermediate code") {
       CodeGen(config, linked)
     }
-    val produced = IO.getAll(config.workdir, "glob:**.ll")
-    config.logger.info(s"Produced ${produced.length} files")
-    produced
+    config.logger.info(s"Produced ${llPaths.length} files")
```
catap, line 93: let add a `target triple` here or use generic line if none.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
ekrich: I really didn't want to mess with that because we really just need the path to avoid all the look ups.

catap: ok, it was just an idea.

ekrich: I think right now we are just improving and hardening the code so we are in good territory.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
ekrich: The `nativeTarget` in `sbt` gets set to `targetTriple` in the config. I think that is fine.

ekrich:
This is ready for review.

**Summary**

We need to support cross compiling as a new feature for the following reasons.

1. Some hardware does not have a JVM and/or all the tools or doesn't have sufficient memory to run the tools.
2. In the past people have hacked the plugin and tools to support cross compiling.
3. This can open up the potential of embedded type platforms and hopefully 32 bit support via cross compiling.
4. As we gain more feedback and experience we can improve the build tools as needed.

**Highlights**

- We remove target triple probing as `clang` / `llvm` does a great job already of figuring out the target. Extensive searching online seems to suggest that adding the option `-Wno-override-module` to avoid warnings is fine if the target triple is not provided.
```
[info] Compiling to native code (default target triple) (1201 ms)
```

- Lorenzo @lolgab in his PR #1904 demonstrated that the probe code gets in the way of cross compiling. I was skeptical at first so my apologies. I don't add separate output directories for different targets like his PR but I concluded that if we need to build multiple targets we would need a `Seq(targetTriples)` so I didn't want to commit to more complexity until we have valid use cases. I guess you could have multiple projects using the same code with different options.

- Kirill @catap also went after this topic and had some really good optimizations that I used and improved that should make the builds faster. This also makes the build more understandable as the files that we generate and find to compile are passed along in the build rather than searched and discovered again and again. The special code for the GC and optional components is also more understandable now.

- When testing,  I used `llvm-config --host-target` which outputs `x86_64-apple-darwin18.7.0`. When adding the following to `sandbox`, `.settings(nativeTarget := "x86_64-apple-darwin18.7.0")` I end up getting the following warnings 8 times for each generated `.ll`. 
```
[error] warning: overriding the module target triple with x86_64-apple-macosx10.14.0 [-Woverride-module]
[error] 1 warning generated.
```
After replacing the target triple with `x86_64-apple-macosx10.14.0` the warnings go away. So the idea is that we only add the warning suppression when you use the default compiling. If you add the `nativeTarget` setting we will get warnings and errors so that makes it safe.
```
[info] Compiling to native code (x86_64-apple-macosx10.14.0) (958 ms)
```

Hopefully, we can merge this code and it can get exercised by people trying to cross compile before the `0.4.0` release.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -24,6 +24,9 @@ sealed trait NativeConfig {
   /** The compilation options passed to LLVM. */
   def compileOptions: Seq[String]
 
+  /** Target triple that defines current OS, ABI and CPU architecture. */
+  def targetTriple: String
```
WojciechMazur, line 28: Personally, I would go with `Option[String]` and `withTargetTriple(value: String)`. All `nativeX` settings including `nativeTargetTriple` exist only for backward compatibility and would be removed in the future. The only remaining way to interact with them would be done via `nativeConfig` in the same way as it's done in scala.js.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -54,6 +54,9 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeDump =
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
+
+    val nativeTarget =
+      taskKey[String]("The LLVM target triple for cross compiling.")
```
WojciechMazur, line 59: Personally, I would go with Option[String] and withTargetTriple(value: String). All `nativeX` settings including nativeTargetTriple exist only for backward compatibility and would be removed in the future. The only remaining way to interact with them would be done via nativeConfig in the same way as it's done in scala.js.

sjrd: I would add that, since `nativeTarget` didn't exist before, there is no backward compatibility to take care of. So we should *only* have it in the `nativeConfig`.

ekrich: I am a little confused. I looked at the Scala.js plugin and that is much more complex so it wasn't much help.

Right now `nativeTarget` is private in the internal plugin. We are moving it to be exposed in the plugin so the user can set it, correct? We can rename to `targetTriple` for the sbt users.

I do think it makes sense to make it an `Option[String]` in the `NativeConfig` as then the downstream processing is safer. If we make it `Option[String]` in the plugin then it looks weird when setting it.

Here is what it looks like added into `sandbox` for testing as an Option.
```scala
lazy val sandbox =
  project
    .in(file("sandbox"))
    .settings(scalacOptions -= "-Xfatal-warnings")
    .settings(targetTriple := Some("x86_64-apple-macosx10.14.0"))
    .settings(noPublishSettings)
    .dependsOn(nscplugin % "plugin", allCoreLibs, testInterface % Test)
```

Also, shouldn't it be a `settingKey` instead of a `taskKey`? Shouldn't all the "settings" be settings - not much of an sbt plugin expert? An internal task could be used to convert the setting from a `String` to an `Option[String]` maybe?

catap: @ekrich I definitely agree with you and I love idea "default" targetTriple.

Maybe call it `nativeTargetTriple`?

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -120,26 +120,29 @@ private[scalanative] object LLVM {
...
+    val targetOpt = target(config)
+
     // generate .o files for all included source files in parallel
-    paths.par.foreach { path =>
-      val opath = path + oExt
-      if (include(path) && !Files.exists(Paths.get(opath))) {
+    includePaths.par.map { path =>
+      val opath   = path + oExt
+      val objPath = Paths.get(opath)
+      if (!Files.exists(objPath)) {
```
lolgab, line 138: Keep in mind that when you change the target triple these files need to be invalidated.
This if there will be a target triple setting changeable by the users.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
ekrich: If I reading correctly, I think the suggestion was just to make it `targetTriple`.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
ekrich: As you can see I avoided different output directories but I thought it might be best to see how this goes. I think changing the target triple in the build and `reload`ing the build probably means you should `clean`.

The build currently removes the `.ll` on every build but I am not sure of that history. It is probably just because we have no timestamp checking between `.nir` and `.ll` and it is not a one to one mapping. https://github.com/scala-native/scala-native/pull/2033/files#diff-ba1b9a87401a482acad7eb0588b7cc93ffeabdd871c5abd747b7bd5a79cf6a4fR67-R68

Now that I think about it we may need a `Discover.targetTriple` so we can have an environment variable because other wise how could you use this in a CI build? Another option might be to have `Seq(targetTriple1, targetTriple2, ...)`. I hate to over engineer a feature on its first try especially since we need experience and feedback to find the sweet spot.

catap: @ekrich and somewhere here we jump back to our discussion :)

I feel that we should force user to use different sbt task for different targetTriple.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
catap: @ekrich I'm not sure that it is good idea to remove `native` prefix for only one setting.

ekrich: @catap I think I need to wait for more guidance. I am not exactly sure what @WojciechMazur and @sjrd are saying. See their comments above.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -54,6 +54,9 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeDump =
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
+
+    val targetTriple =
+      taskKey[String]("The optional LLVM target triple for cross compiling.")
```
sjrd, line 59: We shouldn't add this in the new public API. The new way to configure things is to transform `nativeConfig`, for all settings. All the other keys, like `nativeGC`, are there for compatibility reasons, but we shouldn't add *more* of them. Users should use the following config instead:
```scala
nativeConfig ~= { _.withTargetTriple("whatever") }
```

tools/src/main/scala/scala/scalanative/build/LLVM.scala
sjrd: I think the currently implemented solution is good: when changing the target triple, like any other config option, the user should `clean`. There is no difference between the target triple and, say, the GC configuration.

> Now that I think about it we may need a Discover.targetTriple so we can have an environment variable because other wise how could you use this in a CI build?

Using
```
$ sbt 'set nativeConfig ~= { _.withTargetTriple("whatever") }' test
```
There's no need for an environment variable.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -57,6 +60,9 @@ sealed trait NativeConfig {
   /** Create a new config with given compilation options. */
   def withCompileOptions(value: Seq[String]): NativeConfig
 
+  /** Create a new config given a target triple. */
+  def withTargetTriple(value: Option[String]): NativeConfig
```
sjrd, line 64: Consider adding an overload that takes a `String` and delegates to this one with `Some(value)`. That will be easier to use in most configs.

sjrd (changes requested):

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
ekrich: Thank you for the review, I understand now. I will clean this up and add some documentation as needed.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
ekrich: @catap If you see improvements in the code there maybe you should add an issue.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
ekrich: The sbt naming and use has been cleared up below - `nativeTarget` is not needed at all.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
ekrich: Good idea.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
ekrich: That is good.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
ekrich: Thanks, that looks much better in the code. Down in `GenCode` it gets converted back to `String` to avoid more changes.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
ekrich: A clean is required after changing settings at least the way things are now. A workaround is perhaps to have different projects for different targets or in CI I don't think that would be a problem because each build would be mostly clean.

tools/src/main/scala/scala/scalanative/build/ScalaNative.scala
ekrich: This is the number of LL files and the strategies are described in `GenCode`. The very next step in compilation prints the target triple so I think this is fine.

ekrich:
@sjrd This should be ready for re-review when you get a chance.

ekrich:
Related: https://github.com/scala-native/scala-native/issues/1183

tools/src/main/scala/scala/scalanative/build/LLVM.scala
lolgab: Personally I think this is a pity. One of the selling points of Scala Native compared to say GraalVM Native Image is that it could allow fast cross-compilation to generate programs for various operative systems. We're not yet there, this is sure, but I think that imposing a clean to change target is a waste of build time that will lower this benefit.
Compilation to NIR doesn't change between targets and this way if you want to cross compile you're asking to re-do this work over and over.
I hope that in the future \(when we will see usages of cross-compiling, for example [throw Zig](https://ziglang.org/#Zig-is-also-a-C-compiler)\) this decision will be revised.

ekrich: I agree that this is not ideal but we don't know when build settings change. Removing that check would slow down the normal use case. This check was there so that is not changing and I am not sure removing it is a great idea.

I am open to a `Seq["target triple"]` and individual build directories and lots of other stuff but I think it is better to have a nice change like this first and then when we prove things out we might find out that there are even better ways to do things.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -54,6 +54,7 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeDump =
       taskKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?")
+
```
sjrd, line 57: Spurious change here.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -63,22 +64,32 @@ object Build {
...
+
+      val generated = ScalaNative.codegen(fconfig, optimized)
 
       val nativelibs   = NativeLib.findNativeLibs(fconfig.classPath, workdir)
       val nativelib    = NativeLib.findNativeLib(nativelibs)
       val unpackedLibs = nativelibs.map(LLVM.unpackNativeCode(_))
 
-      val objectFiles = config.logger.time("Compiling to native code") {
+      val msg = s"Compiling to native code (${targetTripleMsg(config)})"
+      val objectPaths = config.logger.time(msg) {
```
sjrd, line 77: I don't see why we should display the target triple here, if it directly comes from the config anyway. Especially, why should we display that, but not all the other config values?

IMO this should be reverted:
```suggestion
      val objectPaths = config.logger.time("Compiling to native code") {
```

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -120,26 +120,29 @@ private[scalanative] object LLVM {
...
-    paths.foreach { path =>
-      if (!include(path)) {
-        val opath = Paths.get(path + oExt)
-        if (Files.exists(opath)) {
-          Files.delete(opath)
-        }
-      }
+    excludePaths.foreach { path =>
+      val opath = Paths.get(path + oExt)
+      if (Files.exists(opath)) Files.delete(opath)
```
sjrd, line 128: Side-effecting branches of if's should be on their own line, so that they stand out:
```suggestion
      if (Files.exists(opath))
        Files.delete(opath)
```

tools/src/main/scala/scala/scalanative/build/LLVM.scala
sjrd: If we do anything about this, it should be the same way that we invalidate the output of the linker if the config has changed in Scala.js:
https://github.com/scala-js/scala-js/blob/a27c22bc0bf15a84b14745231815ea0a4a12c90f/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L188-L205

But that is for a different PR.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -116,6 +127,14 @@ object NativeConfig {
     def withCompileOptions(value: Seq[String]): NativeConfig =
       copy(compileOptions = value)
 
+    def withTargetTriple(value: Option[String]): NativeConfig =
+      copy(targetTriple = value)
+
+    def withTargetTriple(value: String): NativeConfig = {
+      val optValue = if (value.trim().isEmpty()) None else Some(value)
+      withTargetTriple(optValue)
+    }
```
sjrd, line 136: This is too fancy. The overload with a `String` is just a helper to avoid the `Some()`. It should just be
```suggestion
    def withTargetTriple(value: String): NativeConfig =
      withTargetTriple(Some(value))
```
In a "configuration language", this would make sense. But here we are defining a configuration API.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -24,7 +24,9 @@ object Build {
    *  val clangpp  = Discover.clangpp()
    *  val linkopts = Discover.linkingOptions()
    *  val compopts = Discover.compileOptions()
-   *  val triple   = Discover.targetTriple(clang, workdir)
+   *  // Empty string or target triple if cross compiling
+   *  val triple   = ""
```
sjrd, line 28: Given my other comment, this shouldn't be an empty string. It can be `None`, or `Some(actualTriple)`, or `actualTriple`.

sjrd (changes requested):
All right. This basically looks good to me now. I just have a few tweaking comments.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
lolgab: Thank you @ekrich and @sjrd .
As you said, Eric, this is a good starting point to start playing with cross-compilation and it doesn't necessarily mean that can't be improved in the future when if we find better ways to do things.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
ekrich: Thanks.

tools/src/main/scala/scala/scalanative/build/Build.scala
ekrich: It is not just about config values but showing the ones that really matter. Currently, we show the `GC`, `lto`, optimization level like `debug` and `release-fast` and other `Dumping` messages. Target triple is very important because that is the native code that is being compiled. It also brings awareness to people using Scala Native and most Scala programmers probably know little about the Clang/llvm platform. The target triple will become even more important in the future if we can target smaller platforms or ones without a JVM that require cross compilation.

```
sbt:scala-native> sandbox/run
[info] Linking (1751 ms)
[info] Checking intermediate code (278 ms)
[info] Dumping intermediate code (linked) (740 ms)
[info] Discovered 924 classes and 6981 methods
[info] Optimizing (debug mode) (2677 ms)
[info] Checking intermediate code (92 ms)
[info] Dumping intermediate code (optimized) (528 ms)
[info] Dumping intermediate code (lowered) (943 ms)
[info] Generating intermediate code (2767 ms)
[info] Produced 8 files
[info] Compiling to native code (default target triple) (904 ms)
[info] Linking native code (immix gc, none lto) (184 ms)
[info] Total (10167 ms)
Hello, World!
[success] Total time: 14 s, completed Dec 20, 2020 7:50:43 AM
```

tools/src/main/scala/scala/scalanative/build/LLVM.scala
ekrich: Done, good to know.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
ekrich: Thanks, this is an oversight (copy) from a previous version where I had the `sbt` setting and when not set, `""` was the value passed from the setting.

tools/src/main/scala/scala/scalanative/build/Build.scala
ekrich: It is not required so I just removed it from the documentation altogether.

tools/src/main/scala/scala/scalanative/build/Build.scala
ekrich: @sjrd I just removed this as you recommended. We can always add it if we get requests and I asked around and it seems I was the most interested in this output. So this should be ready to go.

sjrd (approved):

PR #2034, opened 2020-11-24
Add ScalaRuntime 2.13 override

WojciechMazur:
This PR adds missing ScalaRuntime override resulting in `IllegalArgumentExceptions` in #2025. Since it was not overridden in 2.13 it tried to use the default JVM implementation using `reflect` to get array length. 

It only partially resolves the mentioned issue - it would fix `IllegalArgumentException`, but still would fail due to match errors in std lib. They're caused by `Array[BoxedUnit]` not being treated as an instance of `Array[AnyRef]` in all scala versions. They would be fixed in separate PR

### DIFF
--- a/scalalib/overrides-2.13/scala/runtime/ScalaRunTime.scala
+++ b/scalalib/overrides-2.13/scala/runtime/ScalaRunTime.scala
@@ -53,55 +53,38 @@ object ScalaRunTime {
 	classTag[T].runtimeClass.asInstanceOf[jClass[T]]
 
   /** Retrieve generic array element */
-  def array_apply(xs: AnyRef, idx: Int): Any = {
-	(xs: @unchecked) match {
-	  case x: Array[AnyRef]  => x(idx).asInstanceOf[Any]
-	  case x: Array[Int]     => x(idx).asInstanceOf[Any]
-	  case x: Array[Double]  => x(idx).asInstanceOf[Any]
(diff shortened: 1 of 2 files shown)


scalalib/overrides-2.13/scala/runtime/ScalaRunTime.scala
```diff
@@ -0,0 +1,298 @@
...
+import java.lang.{Class => jClass}
+import java.lang.reflect.{Method => JMethod}
+
+/** The object ScalaRunTime provides support methods required by
+ *  the scala runtime.  All these methods should be considered
+ *  outside the API and subject to change or removal without notice.
+ */
+object ScalaRunTime {
+  def isArray(x: Any, atLevel: Int = 1): Boolean =
+	x != null && isArrayClass(x.getClass, atLevel)
```
sjrd, line 31: This file seems to have a lot of tabs, which are not present in the original file.

sjrd (changes requested):
Looks like we should have a test for this. We might not be able to test with `Array[Unit]` in this PR, but we definitely test it for other types of arrays. Something like
```scala
def genericLength[T](a: Array[T]): Int = a.length

assertEquals(3, genericLength(new Array[Int](3)))
// same for a few other types (notably AnyRef)
```

sjrd (approved):

PR #2036, opened 2020-11-26
Fix #2025: Remove BoxedUnitArray

### LINKED ISSUE #2025: Unit IllegalArgumentException argument type mismatch at runtime

WojciechMazur:
This PR introduces another approach in order to resolve #2025. 
PR #2032 tried to change the inheritance of concrete array types, due to problems with casts and matching. 
This approach removes `BoxedUnitArray` from concrete array implementations. Its dictated by the fact that `Array[Unit]` is always translated to `Array[BoxedUnit]` which makes it an instance of `Array[AnyRef]`. 

Due to the limited information set of `Class[_]` we cannot get the correct Class.componentType. This may lead to problems at runtime. Only types with implemented component types were concreate `TArray` instances. One of them was removed `BoxedUnitArray`. This PR introduces regression as we no longer can distinguish `Array[BoxedUnit]` from other instances of `Array[_ <: AnyRef]`

Additional scalaib 2.13 override was added in order to fix the only existing usage of pattern matching on `Array[BoxedUnit]` in Scala stdlib.

### DIFF
--- a/scalalib/overrides-2.13/scala/Array.scala
+++ b/scalalib/overrides-2.13/scala/Array.scala
@@ -46,8 +46,8 @@ object Array {
   implicit def toFactory[A : ClassTag](dummy: Array.type): Factory[A, Array[A]] = new ArrayFactory(dummy)
   @SerialVersionUID(3L)
   private class ArrayFactory[A : ClassTag](dummy: Array.type) extends Factory[A, Array[A]] with Serializable {
-	def fromSpecific(it: IterableOnce[A]): Array[A] = Array.from[A](it)
-	def newBuilder: mutable.Builder[A, Array[A]] = Array.newBuilder[A]
+    def fromSpecific(it: IterableOnce[A]): Array[A] = Array.from[A](it)
+    def newBuilder: mutable.Builder[A, Array[A]] = Array.newBuilder[A]
   }
 
   /**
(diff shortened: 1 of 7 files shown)


scalalib/overrides-2.13/scala/Array.scala
```diff
@@ -0,0 +1,673 @@
...
+   *  @param src the source array.
+   *  @param srcPos  starting position in the source array.
+   *  @param dest destination array.
+   *  @param destPos starting position in the destination array.
+   *  @param length the number of array elements to be copied.
+   *
+   *  @see `java.lang.System#arraycopy`
+   */
+  def copy(src: AnyRef, srcPos: Int, dest: AnyRef, destPos: Int, length: Int): Unit = {
+	val srcClass = src.getClass
```
sjrd, line 107: There are some tabs in this file.

sjrd (changes requested):
Looks good. Only tabs to fix in the override file.

sjrd (approved):

PR #2037, opened 2020-11-26
Scala 2.13.4 support

WojciechMazur:
This PR adds support for Scala 2.13.4. `build.sbt` and CI would use this version to test 2.13 binary version

It fixes usages of no longer implicit `ArraySequenceOps` as well as pseudo-non-exhaustive match cases.

### DIFF
--- a/javalib/src/main/scala/java/math/BigDecimal.scala
+++ b/javalib/src/main/scala/java/math/BigDecimal.scala
@@ -398,7 +398,7 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
     }
 
     val (unscaled, bufLength) = {
-      val u = ArrayCharSequence(in).subSequence(begin, index).toString
+      val u = new String(in, begin, index - begin)
       val b = index - begin
       // A decimal point was found
       if ((index <= last) && (in(index) == '.')) {
@@ -413,8 +413,8 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
           index += 1
         }
         _scale = index - begin
-        val subSeq = ArrayCharSequence(in).subSequence(begin, begin + _scale)
-        (u + subSeq.toString, b + _scale)
+        val subString = new String(in, begin, _scale)
+        (u + subString, b + _scale)
       } else {
         _scale = 0
         (u, b)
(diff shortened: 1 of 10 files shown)


javalib/src/main/scala/java/math/BigDecimal.scala
```diff
@@ -398,7 +398,7 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
     }
 
     val (unscaled, bufLength) = {
-      val u = in.subSequence(begin, index).toString
+      val u = ArrayCharSequence(in).subSequence(begin, index).toString
```
sjrd, line 401: `ArrayCharSequence` will be deprecated in the next version. Let's avoid it altogether:
```suggestion
      val u = new String(in, begin, index - begin)
```

javalib/src/main/scala/java/math/BigDecimal.scala
```diff
@@ -413,7 +413,8 @@ class BigDecimal() extends Number with Comparable[BigDecimal] {
           index += 1
         }
         _scale = index - begin
-        (u + in.subSequence(begin, begin + _scale).toString, b + _scale)
+        val subSeq = ArrayCharSequence(in).subSequence(begin, begin + _scale)
```
sjrd, line 416: Same here:
```suggestion
        val subSeq = new String(in, begin, _scale)
```

javalib/src/main/scala/java/util/package.scala
```diff
@@ -5,9 +5,12 @@ package object util {
...
     @inline
-    def ===(that: Any): Boolean =
-      if (self.asInstanceOf[AnyRef] eq null) that.asInstanceOf[AnyRef] eq null
-      else self.equals(that)
+    def ===(that: Any): Boolean = {
+      val aself = self.asInstanceOf[AnyRef]
+      val athat = that.asInstanceOf[AnyRef]
+      if (aself eq null) athat eq null
+      else aself.equals(athat)
+    }
```
sjrd, line 13: We can advantageously replace all that by
```suggestion
    def ===(that: Any): Boolean = Objects.equals(self, that)
```

junit-runtime/src/main/scala/org/junit/Assert.scala
```diff
@@ -55,7 +55,7 @@ object Assert {
 
   @inline
   private def isEquals(expected: Any, actual: Any): Boolean =
-    expected.equals(actual)
+    expected == actual
```
sjrd, line 58: This is not a correct alternative. The semantics of JUnit assertions is to use Java's equality; not Scala's cooperative equality.

We can use `java.util.Objects.equals(expected, actual)` here.

sjrd (changes requested):

sjrd (approved):

PR #2038, opened 2020-11-26
CI / Integrate publishing snapshots with CI

WojciechMazur:
This PR includes a proposal to publish snapshot versions using `sbt-ci-release`. 
It changes the build to use the mentioned plugin and adds CI workflow responsible for publishing snapshots or tagged releases. 
The plugin would be published only when all tests in the `Run Tests` workflow would pass. 

It needs setting up 4 secrets in this repo setting. Instructions on how to do this can be found [here](https://github.com/olafurpg/sbt-ci-release#can-i-depend-on-maven-central-releases-immediately)  

I don't have experience with publishing either plugins or libraries, so I would welcome any expertise.


ekrich:
I do mine via Travis still so it is pretty much from the `sbt-ci-release` pages.
https://github.com/ekrich/sconfig/blob/master/.travis.yml

I have never published a plugin but it says the following: https://github.com/olafurpg/sbt-ci-release#how-do-i-publish-sbt-plugins

These are my notes, the first part is directly from the docs.
```
Ready for release
—————————————————
Always check for compatibility sbt> mimaReportBinaryIssues
git tag -a v0.1.0 -m "v0.1.0"
git push origin v0.1.0

Delete tags if ci does not work
————————————————————————————————
git tag --delete v0.1.0
git push --delete origin v0.1.0
go to sonatype and delete the staging repos - login to see them
redo release above
```

I really don't know much about the deleting tags thing - I had some problems and it seemed to work.

sideeffffect:
I'm sorry to see this closed? May I ask what was the reason? I thought that having access to snapshot builds would benefit the Scala Native community.

armanbilge:
@sideeffffect see:
- https://github.com/scala-native/scala-native/pull/3388

PR #2046, opened 2020-11-30
Do not allow running tests in root project

### LINKED ISSUE #2028: Some symbols are missed inside unit tests
```
[info] Dumping intermediate code (linked) (921 ms)
[error] missing symbols:
[error] * M41scala.scalanative.testinterface.TestMain$D4mainLAL16java.lang.String_uEO
[error] * T41scala.scalanative.testinterface.TestMain$
```

I get this via `sbt ++2.11.12 clean test` on `OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)` with 31b6733
 as local root

WojciechMazur:
Resolves #2028 

This PR introduces an explicit root project aggregating all other projects with disabled tests aggregation. When trying to run test in root project it would throw an exception.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -213,7 +213,7 @@ lazy val disabledTestsSettings: Seq[Setting[_]] = {
   def testsTaskUnsupported[T] = Def.task[T] {
     throw new MessageOnlyException(
       s"""Usage of this task in ${projectName(thisProject.value)} project is not supported in this build.
-         |To run tests use explicit syntax containing name of project: `<project_name>/<task>.
+         |To run tests use explicit syntax containing name of project: <project_name>/<task>.
          |You can also use one of predefined aliases: test-all, test-tools, test-runtime, test-scripted.
          |""".stripMargin
     )
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,6 @@ import scala.collection.mutable
 import scala.util.Try
 
 import build.ScalaVersions._
-import sbt.Def
 
 // Convert "SomeName" to "some-name".
 def convertCamelKebab(name: String): String = {
--- a/build.sbt
+++ b/build.sbt
@@ -3,6 +3,7 @@ import scala.collection.mutable
 import scala.util.Try
 
 import build.ScalaVersions._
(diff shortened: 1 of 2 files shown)


sjrd:
I don't think this addresses the issue in the correct way.

First, the proposed fix changes things in the sbt plugin, used by everyone, to solve an issue in our own build. That's not the appropriate place to do that. A fix should only apply in our build. What the sbt plugin currently does for Scala Native projects is consistent with what sbt does by default for JVM projects and what sbt-scalajs does for JS projects. There should be no reason to do things differently in Native.

Second, I'm not convinced that we *want* `test` to work in our own build. That touches a large amount of projects, and there is virtually no reason to test everything at once in a local build. If someone *really* wants to do that, there is `test-all`.

Instead of trying to make `test` work in our build, we should make it fail early and explicitly, telling the user to use `unit-tests/test` instead (in most cases) or `test-all` (if they really want to).

build.sbt
```diff
@@ -208,6 +209,52 @@ lazy val buildInfoSettings: Seq[Setting[_]] =
...
+             testingCompiler,
+             testInterfaceSbtDefs,
+             testInterface,
+             testRunner)
+  .aggregate(junitRuntime,
+             junitTestOutputsNative,
+             junitTestOutputsJVM,
+             junitAsyncNative,
+             junitAsyncJVM)
+  .aggregate(sandbox, tests)
```
sjrd, line 256: Instead of all this, you can simply put `noPublishSettings` and `disabledTestsSettings` at the top-level of the `build.sbt`:
```suggestion
noPublishSettings

disabledTestsSettings
```

sjrd (changes requested):

build.sbt
```diff
@@ -3,6 +3,7 @@ import scala.collection.mutable
 import scala.util.Try
 
 import build.ScalaVersions._
+import sbt.Def
```
sjrd, line 6: I'm pretty sure this is imported by default, and hence redundant.

sjrd (approved):

PR #2051, opened 2020-12-02
Fix #2035: Guard virtual lookup of non virtual methods

### LINKED ISSUE #2035: Investigate not found signatures in Lower.genMethodOp
When working with #2032 a new bug was spotted.  `Lower.genMethodOp` -> `genClassVirtualLookup` may generate invalid code which compiles, but fails in runtime.
There were several instances of Sig not being found in class `VirtualTable`. Since we're only looking for an index there when an element is not found it silently continues with result value `-1` This causes the generation of invalid `getElementPtr` LLVM IR. The resulting pointer may then lead to `dynamicCallsMap: PerfectHashMap` in RTTI instead of actual function leading to SeqFault.
Some examples of not existing Sigs were `ObjectArray.clone: Object`, or `StringBuilder.subString`. 

For most of the time, this bug seems to be unharmful, however, it may show up when using reflective calls.
When testing locally code for `ReflectivProxyTest` in the sandbox it behaved differently than code in unit tests and did not fail in runtime.

WojciechMazur:
This PR fixes silent, yet potentially deadly bugs in `Lower` phase. Due to not exhaustive enough checks, it could silently generate invalid pointer arithmetic operation, leading to a segmentation fault in the runtime, eg. in the case when the non-virtual method was dynamically dispatched.
Changes of this PR include determinating in advance how the given `Sig.Method` should be dispatched: 
-  If `Sig` in a given class is non-virtual (e.g constructor,  extern call) it would always be statically dispatched
- If the result of statistical analysis proves that `Sig` of given `Class` does not contain virtual calls it would be also statically dispatched
- If target of method invocation is `null` (eg. `null.asInstanceOf[T].method`) it always would end-up with `NullPointerOperationException`
- If any other case virtual method lookup would be used to determine the correct method target based on traits dispatch table or class virtual methods table. 

Additionally code that would previously silently generate errors would now be checked and will throw an exception.  

Resolves #2035

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
@@ -476,7 +476,10 @@ object Lower {
       val Op.Method(v, sig) = op
       val obj               = genVal(buf, v)
       def genClassVirtualLookup(cls: Class): Unit = {
-        val vindex  = vtable(cls).index(sig)
+        val vindex = vtable(cls).index(sig)
+        assert(vindex != -1,
+               s"Virtual table of ${cls.name} have not contained $sig ")
+
         val typeptr = let(Op.Load(Type.Ptr, obj), unwind)
         val methptrptr = let(
           Op.Elem(rtti(cls).struct,
@@ -522,8 +525,19 @@ object Lower {
             let(n, Op.Copy(Val.Global(impl, Type.Ptr)), unwind)
           case _ =>
             obj.ty match {
-              case ClassRef(cls) =>
+              // It may happen that method exists in multiple targets, but is not a virtual call
+              // In such cases we handle it as statically resolvable method with single target
+              case ClassRef(cls) if cls.calls.contains(sig) =>
                 genClassVirtualLookup(cls)
+              case ClassRef(cls) =>
+                val methodName = Global.Member(cls.name, sig)
+                val methodRef  = Val.Global(methodName, Type.Ptr)
+                assert(
+                  linked.infos.contains(methodName),
+                  s"Method $methodName is neither virtual or statically resolvable"
+                )
+
+                let(n, Op.Copy(methodRef), unwind)
               case TraitRef(_) if Object.calls.contains(sig) =>
                 genClassVirtualLookup(Object)
               case TraitRef(trt) =>
--- a/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
@@ -2,7 +2,7 @@ package scala.scalanative
 package codegen
 
 import scala.collection.mutable
-import scalanative.util.ScopedVar
+import scalanative.util.{ScopedVar, unsupported}
 import scalanative.nir._
 import scalanative.linker.{
   Class,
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -522,8 +525,19 @@ object Lower {
             let(n, Op.Copy(Val.Global(impl, Type.Ptr)), unwind)
           case _ =>
             obj.ty match {
-              case ClassRef(cls) =>
+              // It may happen that method exists in multiple targets, but is not a virtual call
+              // In such cases we handle it as statically resolvable method with single target
+              case ClassRef(cls) if cls.calls.contains(sig) =>
```
sjrd, line 530: I think the proposed fix is correct when everything else is correct. However, I fear that it will too easily "tolerate" other errors without complaining, and instead generate wrong code.

We should be able to determine in advance if a method *should* be a virtual call or not. And if it should not be a virtual call (because it's a constructor, for example), then don't even collect the `targets`. For a non-virtual call, we should be able to find with certainty a *single* `target`, and generate a call to that. Only if we have a virtual call should we collect the multiple `targets`, then generate the proper code for `Seq()`, `Seq(impl)` or more.

sjrd (changes requested):

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
WojciechMazur: I've changed logic to explicitly check if method is virtual or not with additional handling Null (short-circuit throwing null)

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -475,8 +475,12 @@ object Lower {
 
       val Op.Method(v, sig) = op
       val obj               = genVal(buf, v)
+
       def genClassVirtualLookup(cls: Class): Unit = {
-        val vindex  = vtable(cls).index(sig)
+        val vindex = vtable(cls).index(sig)
+        assert(vindex != -1,
+               s"Virtual table of ${cls.name} have not contained $sig ")
```
sjrd, line 482:
```suggestion
               s"The virtual table of ${cls.name} does not contain $sig")
```

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -532,8 +525,39 @@ object Lower {
...
       }
 
-      genGuardNotNull(buf, obj)
-      genMethodLookup()
+      def genStaticMethod(cls: Class): Unit = {
+        val method = cls
+          .resolve(sig)
+          .getOrElse {
+            unsupported(
+              s"Did not found signature of method $sig in ${cls.name}")
```
sjrd, line 533:
```suggestion
              s"Did not find the signature of method $sig in ${cls.name}")
```

sjrd (changes requested):
Only typos.

sjrd (approved):

sjrd:
@WojciechMazur Could you write an updated commit message for this one? I think the PR description does not reflect the latest state.

PR #2058, opened 2020-12-07
Fix #1121: Allow ScalaDoc generation not to emit warnings

### LINKED ISSUE #1121: `packageDoc`: Scaladoc generation failed

WojciechMazur:
This PR resolves #1121. Due to common issue when generating docs we  could get the warning `dropping dependency on node with no phase object: mixin` causing failure due to scalacOption `-Xfatal-warnings`

PR makes components list in NativePlugin and JUnitNativePlugin dynamic, based on global. This fix removes sproucious warnings. Additionally, it adds `docsSettings` containing extra `apiMappings` for Java. They're added only to projects using them, and throwing errors when generating docs. 
Generation of ScalaDoc was disabled in the following projects:
 * `javalib` - ends up with `MalformedInputException: Input length = 1`
* `scalalib` - ScalaDoc should not be generated
* projects using `noPublish` settings, eg: `tests`, `sandbox`, `junitTestOutputsX`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -768,7 +768,7 @@ lazy val junitPlugin =
 
 val commonJUnitTestOutputsSettings = Def.settings(
   nameSettings,
-  disabledDocsSettings,
+  noPublishSettings,
   Compile / publishArtifact := false,
   Test / parallelExecution := false,
   Test / unmanagedSourceDirectories +=
--- a/build.sbt
+++ b/build.sbt
@@ -19,6 +19,54 @@ lazy val nameSettings: Seq[Setting[_]] = Seq(
   name := projectName(thisProject.value) // Maven <name>
 )
 
+lazy val disabledDocsSettings: Seq[Setting[_]] = Def.settings(
+  sources in (Compile, doc) := Nil
+)
+
+lazy val docsSettings: Seq[Setting[_]] = {
+  val javaDocBaseURL: String = "http://docs.oracle.com/javase/8/docs/api/"
+  // partially ported from Scala.js
+  Def.settings(
+    autoAPIMappings := true,
+    exportJars := true,                                    // required so ScalaDoc linking works
+    scalacOptions in (Compile, doc) -= "-Xfatal-warnings", // needed only for 2.11 docs to mitigate not found java.lang members
+    // Add Java Scaladoc mapping
+    apiMappings ++= {
+      val optRTJar = {
+        val bootClasspath = System.getProperty("sun.boot.class.path")
+        if (bootClasspath != null) {
+          // JDK <= 8, there is an rt.jar (or classes.jar) on the boot classpath
+          val jars = bootClasspath.split(java.io.File.pathSeparator)
+
+          def matches(path: String, name: String): Boolean =
+            path.endsWith(s"${java.io.File.separator}$name.jar")
+
+          val jar = jars
+            .find(matches(_, "rt"))                   // most JREs
+            .orElse(jars.find(matches(_, "classes"))) // Java 6 on Mac OS X
+            .get
+          Some(file(jar))
+        } else {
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -19,6 +19,54 @@ lazy val nameSettings: Seq[Setting[_]] = Seq(
   name := projectName(thisProject.value) // Maven <name>
 )
 
+lazy val disabledDocsSettings: Seq[Setting[_]] = Def.settings(
+  sources in (Compile, doc) := Nil
+)
+
+lazy val docsSettings: Seq[Setting[_]] = {
+  val javaDocBaseURL: String = "http://docs.oracle.com/javase/8/docs/api/"
```
sjrd, line 27:
```suggestion
  val javaDocBaseURL: String = "https://docs.oracle.com/javase/8/docs/api/"
```

build.sbt
```diff
@@ -19,6 +19,54 @@ lazy val nameSettings: Seq[Setting[_]] = Seq(
...
+  // partially ported from Scala.js
+  Def.settings(
+    autoAPIMappings := true,
+    exportJars := true,                                    // required so ScalaDoc linking works
+    scalacOptions in (Compile, doc) -= "-Xfatal-warnings", // needed only for 2.11 docs to mitigate not found java.lang members
```
sjrd, line 32: Can we try without this? It might work in this build.

build.sbt
```diff
@@ -19,6 +19,54 @@ lazy val nameSettings: Seq[Setting[_]] = Seq(
...
+        if (bootClasspath != null) {
+          // JDK <= 8, there is an rt.jar (or classes.jar) on the boot classpath
+          val jars = bootClasspath.split(java.io.File.pathSeparator)
+
+          def matches(path: String, name: String): Boolean =
+            path.endsWith(s"${java.io.File.separator}$name.jar")
+
+          val jar = jars
+            .find(matches(_, "rt"))                   // most JREs
+            .orElse(jars.find(matches(_, "classes"))) // Java 6 on Mac OS X
```
sjrd, line 46: I think we can get rid of this. We don't support Java 6.

sjrd (changes requested):

build.sbt
WojciechMazur: As pointed in a comment there is a single error existing only in 2.11 which caused adding this option. We can remove it if/when we would drop 2.11 support. 
```
[error] nativelib/src/main/scala/scala/scalanative/reflect/Reflect.scala:19:3: Could not find any member to link for "java.lang.InstantiationException".
[error]   /** Instantiates a new instance of this class using the zero-argument
[error]   ^
[error] one error found
```
I've tried to put the usage of this class inside commont in double-quotes but it does not fix the problem

sjrd: If it's only in 2.11, we should add the second only in 2.11, then:
```suggestion
    scalacOptions in (Compile, doc) := {
      val prev = (scalacOptions in (Compile, doc)).value
      if (scalaVersion.value.startsWith("2.11.") prev.filter(_ != "-Xfatal-warnings")
      else prev
    }
```

sjrd (approved):

Issue #897, opened 2017-07-06
Proposal: Interop 2.0


densh:
# Interop 2.0

Scala Native 0.4 is going to contain a revamped interop layer based on feedback we've gather so far. Here are some rough overview of the updated features:

## Extern members

* `@extern object` no more.

* `def $name(..$args): $T = extern` can appear in any class, module or trait. 

* `val $name: $T = extern` can appear in any class, module or trait.

* `var $name: $T = extern` can appear in any class, module or trait.

* Extern members must provide explicit return type (enforced by the compiler.)

* Binds to native symbols with `$name` or to a custom name given via `@name("...")`.

* `@link("...")` can appear on any class, module or trait.

## Structs

* `CStructN` no more.

* `@struct class $name(...$params) extends ..$traits { ..$methods }`

* Can have arbitrary number of method and fields, both mutable and immutable.

* Can extend traits.

* Can be a case-class.

* Implicitly stack-allocated.
 
  ```
  @struct class Point(var x: Int, var y: Int)

  val point = new Point(10, 20) // implicitly allocates a stack slot
  point.x = 20                  // updates stack-allocated memory
  point.y = 30
  ```

* Resides on stack, inside managed objects, inside arrays and other structs in unboxed form.

  ```
  val array = new Array[Point](10) 
  array(0).x = 10                  // updated array-allocated memory
  array(0).y = 10
  ```

* Boxes whenever passed to generic erased context (i.e. upcast to Any.)

* Transperantly copied by-value across different storage locations.

## Pointers

* `Ptr[T]` with most operations unchanged.

* `_N` methods are gone, new implicit conversion `implicit def inner[T <: AnyRef](ptr: Ptr[T]): T`
  together with extra checks in the compiler to make sure that it only applies to structs. This
  enables calling methods and reading fields directly through pointer:

  ```
  var points: Ptr[T] = stackalloc[Point](100)
  points.x = 10 // updates points(0).x
  points.y = 20 // updates points(0).y
  points += 1
  points.x = 30 // updates points(1).x
  points.y = 40 // updates points(1).y
  ```

* Boxes whenever passed to generic context.

## Function pointers

* CFunctionPtr as before, with known bugs fixed.

## C numeric types

* Type aliases in the `scalanative.native.*` as before.

## Unsigned numeric types

* They become `@struct` classes instead of scalac-style value classes. This enables storing them unboxed inside of the arrays.

* Otherwise unchanged as before

## Generic intrinsics methods 

[...]

PR #2062, opened 2020-12-10
Allow extern definitions defined outside extern objects

### LINKED ISSUE #1991: Declaration of external functions with multiple signatures not working as top level objects
During the Port of `object ExternSuite` to `class ExternTest` the `extern` definitions were lifted from inside the `object` to outside the `class` (the top level). The original issue was here. https://github.com/scala-native/scala-native/issues/1652

As noted in the following, definitions are currently not allowed inside a class. https://github.com/scala-native/scala-native/issues/897

Note that the `ExternTest` uses the `snprintf` C function which is a `CVarArg*` method and is normally supported via an `implicit` because only `CVarArgList` is supported by Scala Native. 
https://github.com/scala-native/scala-native/blob/master/clib/src/main/scala/scala/scalanative/libc/package.scala

As a mediation so the PR https://github.com/scala-native/scala-native/pull/1989 can be merged while this issue is fixed, one test is ignored with the offending code commented out.

Note 2: This seems to work on macOS but not in CI in Linux.
https://travis-ci.com/github/scala-native/scala-native/builds/199498581

Output:
```
[info] Generating intermediate code (23491 ms)
[info] Produced 2 files
[error] /home/scala-native/scala-native/unit-tests/target/scala-2.11/native/1.ll:480068:19: error: '@snprintf' defined with type 'i32 (i8*, i64, i8*, i32)*'
[error]   %_35 = call i32 @"snprintf"(i8* %_30, i64 10, i8* %_32, i8* %_34)
[error]                   ^
[error] 1 error generated.
[info] Compiling to native code (9759 ms)
[error] clang: error: no such file or directory: '/home/scala-native/scala-native/unit-tests/target/scala-2.11/native/1.ll.o'
```

WojciechMazur:
This PR implements #897 for extern definitions. It also resolves #1991. 
It includes all not yet implemented features from the proposal, that is:

>* `@extern object` no more.
>* `def $name(..$args): $T = extern` can appear in any class, module or trait. 
>* `val $name: $T = extern` can appear in any class, module or trait.
>* `var $name: $T = extern` can appear in any class, module or trait.
>* `@link("...")` can appear on any class, module, trait, method or field. If the given definition is reachable at link-time, linker >would link with the corresponding native library automatically.

This PR should be followed by changes to #2060 including changes to syntax rules if it would be merged earlier, or should be rebased and adjusted in otherwise.

### DIFF
--- a/unit-tests/src/test/scala/scala/scalanative/unsafe/ExternTest.scala
+++ b/unit-tests/src/test/scala/scala/scalanative/unsafe/ExternTest.scala
@@ -81,6 +81,7 @@ class ExternTest {
     }
 
     case class Ext() extends ExtClass() with ExtTrait
+
   }
 
   @Test def allowsToUseExternNotDefineInObject(): Unit = {
@@ -97,6 +98,5 @@ class ExternTest {
     assertEquals(
       expectedLength,
       ext.extClass(stackalloc[Byte](bufsize), bufsize, testFormat, testString))
-    )
   }
 }
--- a/unit-tests/src/test/scala/scala/scalanative/unsafe/ExternTest.scala
+++ b/unit-tests/src/test/scala/scala/scalanative/unsafe/ExternTest.scala
@@ -12,12 +12,10 @@ object ExternTest {
    * These also are having problems at the top level with our CI
    * with the current Docker configuration, see #1991
    */
-  @extern
   object Ext1 {
     def snprintf(buf: CString, size: CSize, format: CString, l: CString): Int =
       extern
   }
-  @extern
   object Ext2 {
     @name("snprintf")
     def p(buf: CString, size: CSize, format: CString, i: Int): Int = extern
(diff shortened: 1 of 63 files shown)


ekrich (commented):
Please remember to update the documents as this get closer to being final.

unit-tests/src/test/scala/scala/scalanative/unsafe/ExternTest.scala
```diff
@@ -64,15 +32,71 @@ class ExternTest {
...
+  }
+
   @Test def sameExternNameInTwoDifferentObjects_Issue1652(): Unit = {
-    ExternTest.runTest()
+    import scalanative.libc.string
+    val bufsize = 10L
+    val buf1    = stackalloc[Byte](bufsize)
+    val buf2    = stackalloc[Byte](bufsize)
+    Ext1.snprintf(buf1, bufsize, c"%s", c"hello")
+    assertTrue(string.strcmp(buf1, c"hello") == 0)
```
lolgab, line 51: Did you try to change to `assertEquals` here (and two lines below)?

PR #2068, opened 2020-12-14
Concurrent Immix & Commix

catap:
This commit is rebased version of https://github.com/scala-native/scala-native/pull/1726

Well, it contains a lot of changes during rebase that allows to make it works again on 24e0983e3c478325dacad44d5a7e4a7146c5cbe5.

This is the first step to split https://github.com/catap/scala-native/tree/multithreading to some trivial changes that can be reviews and tested.

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/immix/ThreadManager.c
+++ b/nativelib/src/main/resources/scala-native/gc/immix/ThreadManager.c
@@ -1,0 +1,99 @@
+#include <setjmp.h>
+#include <signal.h>
+#include <stdio.h>
+
+#include "ThreadManager.h"
+#include "GCTypes.h"
+#include "semaphore/Semaphore.h"
+
+#define SIG_SUSPEND SIGXFSZ
+#define SIG_RESUME SIGXCPU
+
+// Signal handlers can't receive parameters.
+// These variables need to be accessed as global variables.
+
+void *suspendingThreadStackTop;
+Semaphore semaphore;
+
+void ThreadManager_suspendHandler(int sig) {
+    sigset_t sigset;
+
+    jmp_buf regs;
+    setjmp(regs);
+    word_t *dummy;
+
+    suspendingThreadStackTop = &dummy;
+
+    sigfillset(&sigset);
+    sigdelset(&sigset, SIG_RESUME);
+    Semaphore_Post(&semaphore);
+    sigsuspend(&sigset);
+}
+
+void ThreadManager_resumeHandler(int sig) {}
+
+void ThreadManager_Init(ThreadManager *threadManager) {
+    Semaphore_Init(&semaphore, 0);
+    threadManager->threadList = NULL;
+
+    struct sigaction suspend_action;
+    sigemptyset(&suspend_action.sa_mask);
+    sigaddset(&suspend_action.sa_mask, SIG_RESUME);
+    suspend_action.sa_flags = 0;
+    suspend_action.sa_handler = ThreadManager_suspendHandler;
+    sigaction(SIG_SUSPEND, &suspend_action, NULL);
+
+    struct sigaction resume_action;
+    sigemptyset(&resume_action.sa_mask);
+    resume_action.sa_flags = 0;
+    resume_action.sa_handler = ThreadManager_resumeHandler;
+    sigaction(SIG_RESUME, &resume_action, NULL);
+}
+
+void ThreadManager_RegisterThread(ThreadManager *threadManager,
+                                  void *stackBottom) {
+    threadManager->threadList =
+        ThreadList_Cons(pthread_self(), stackBottom, threadManager->threadList);
+}
+
+void ThreadManager_suspendThread(ThreadManager *threadManager,
+                                 pthread_t thread) {
+    if (pthread_equal(pthread_self(), thread))
+        return;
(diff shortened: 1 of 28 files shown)


catap:
@sjrd / @WojciechMazur can I ask you to review it?

catap:
@lolgab raised very good point that mutex can slow down allocation on single threaded application.

Let me share quite naive benchmark: https://gist.github.com/catap/5f9078198073d5330cece538c518173e

As far as I can see impact of mutex is near to negligible.

ekrich:
Since Immix is default maybe you should only change Commix if that would work.

catap:
@ekrich without this changes enabling multithreading very-very-very bad idea.

ekrich:
@catap I don't think you understand what I was saying. We could have Boehm, and Immix (default GC) to support only single threaded for now. Then you could update Commix only, add a define for `SN_MULTI_THREADED` or something like that which makes Commix multi-threaded only when you select it in the build. The define would get set for conditional compilation and Commix would be automatically selected.

This way you only modify one GC and it should be the same as the current when not in multi-threaded mode. This makes the changes smaller and maintains the status quo for the current setup.

catap:
@ekrich Boehm supports multithreads ;)

Also, I see significant performance degradation (like a few times) on Commix vs Immix on 0.4.0; I haven't created a test case but I try to minimize it.

So, here my point (after thinking a while): I agree that we definitely need a switch to enable multithreading because it is very experimental and very unstable.

and this PR is good place to add such switch that will be disabled by default.

lolgab:
> So, here my point (after thinking a while): I agree that we definitely need a switch to enable multithreading because it is very experimental and very unstable.

Agree 100%,
I was reading about signals and how dangerous they are, and they don't play well with epoll.. So I guess that if the GC tries to stop the thread waiting on epoll from native-loop it will break it.
https://unixism.net/2021/02/making-signals-less-painful-under-linux

nativelib/src/main/resources/scala-native/gc/immix/ThreadManager.c
```diff
@@ -0,0 +1,99 @@
...
+#include <stdio.h>
+
+#include "ThreadManager.h"
+#include "GCTypes.h"
+#include "semaphore/Semaphore.h"
+
+#define SIG_SUSPEND SIGXFSZ
+#define SIG_RESUME SIGXCPU
+
+// Signal handlers can't receive parameters.
```
WojciechMazur, line 12: Yes, however we can replace usage of `sa_handler`  with `sa_sigaction`. We can pass pointer to struct containing semaphore and pointer for stack_top in `siginfo_t.si_ptr`, however we would need to replace `pthread_kill` with`pthread_sigqueue`. I'm not very experienced in C signals but such approach might remove this global  variables and  help to create easier to understand structure of thread manager workflow.

nativelib/src/main/resources/scala-native/gc/immix/ThreadManager.h
```diff
@@ -0,0 +1,21 @@
+#ifndef IMMIX_THREADMANAGER_H
+#define IMMIX_THREADMANAGER_H
+
+#include <pthread.h>
+#include "datastructures/ThreadList.h"
+
+typedef struct {
+    ThreadList *threadList;
+    pthread_mutex_t mutex;
+} ThreadManager;
```
WojciechMazur, line 10: Can we try to deduplicate at least `ThreadManager`, `ThreadList` and `Semaphore`? They're exactly the same in commix and immix

WojciechMazur (changes requested):
We should conditionally enable multithreading as it was stated in previous conversations. Most of changes introduced here could be deduplicated since they're exactly the same. It might make maintaince easier. I'm not very experienced with signals but we can probably get rid of global variables, correct me please if I'm wrong. Also some additional documentation might be nice.

#edit: We need to make sure that changes would be done in both immix and commix!

nativelib/src/main/resources/scala-native/gc/immix/ThreadManager.h
ekrich: I was thinking about this but now the way the plugin deals with the GCs would make it hard to share code as we compile all the code and then remove everything not selected.
https://github.com/scala-native/scala-native/blob/master/tools/src/main/scala/scala/scalanative/build/LLVM.scala#L127-L131

I think the way to improve the compile this is to only compile the code we are interested in. The only downside would be that we don't compile the code until the `GC` or `optional` code is selected and used. Not everything gets compiled. This also would require a PR and I would hate to slow progress here.

This is partly why I was suggesting that we totally leave `immix` alone since it is default so we can't mess up normal single threaded code and only enhance `commix` for now for multi-threaded. Since `commix is optional, we can mess people up if they are using it with single threaded but at least it would have less impact.

This would also simplify this PR as well with less changes.

ekrich: This is not correct what I say above. We only delete the not included `.o` files so that if we switch to another GC or remove a previous optional package the old code will be deleted. It still affects how we could share code between `immix` and `commix` however as it would be not really supported with the current code.

WojciechMazur:
Closing this PR as the concurrency awareness was already introduced to Immix/Commix in seperate PR. Thank you for your work which did helped with that.

PR #2069, opened 2020-12-14
Fix #415: Report usage positions of missing definitions when linking

### LINKED ISSUE #415: Position of unresolved links
It would be super helpful to find all positions (Source file, line, column) of unresolved links.

For example:

```
@java.lang.util.Formatter_class_uint_uint
├── /home/gui/native/src/main/scala/A.scala:10:10
└── /home/gui/native/src/main/scala/B.scala:20:20
@foo.bar_class_uint_uint
└── /home/gui/native/src/main/scala/Z.scala:20:20
```

WojciechMazur:
This PR adds aggregation of missing definitions with their usage positions when performing static reachability phase. 
Currently, when some definition is missing, we would result in a single exception due to usage of `Map.apply` which does not contain any greater context. 
Thanks to this PR we would get list of all missing definitions with places of their usage, eg.:
```
    [error] Found 2 missing definitions while linking
    [error] Not found Top(java.time.Instant$)
    [error] 	at <project_dir>/javalib/src/main/scala/java/nio/file/attribute/FileTime.scala:9
    [error] Not found Top(java.util.Locale$)
    [error] 	at <project_dir>/javalib/src/main/scala/java/lang/String.scala:11
    [error] 	at <project_dir>/javalib/src/main/scala/java/lang/String.scala:43
    [error] 	at <project_dir>/javalib/src/main/scala/java/text/DateFormatSymbols.scala:28
    [error] 	at <project_dir>/javalib/src/main/scala/java/text/DecimalFormat.scala:60
    [error] 	at <project_dir>/javalib/src/main/scala/java/util/Formatter.scala:6
    [error] 	at <project_dir>/javalib/src/main/scala/java/util/Formatter.scala:45
    [error] 	at <project_dir>/javalib/src/main/scala/java/util/Formatter.scala:53
    [error] stack trace is suppressed; run last tests / Test / nativeLink for the full output
    [error] (tests / Test / nativeLink) scala.scalanative.linker.Reach$ReachabilityFailureException: Undefined definitions found in reachability phase

```
Resolves #415

### DIFF
--- a/tools/src/main/scala/scala/scalanative/linker/Reach.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/Reach.scala
@@ -1,10 +1,14 @@
 package scala.scalanative
 package linker
 
+import java.nio.file.{Path, Paths}
 import scala.collection.mutable
+import scala.util.control.NoStackTrace
 import scalanative.nir._
 
 class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
+  import Reach._
+
   val unavailable = mutable.Set.empty[Global]
   val loaded      = mutable.Map.empty[Global, mutable.Map[Global, Defn]]
   val enqueued    = mutable.Set.empty[Global]
@@ -14,6 +18,7 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
   val links       = mutable.Set.empty[Attr.Link]
   val infos       = mutable.Map.empty[Global, Info]
   val from        = mutable.Map.empty[Global, Global]
+  val missing     = mutable.Map.empty[Global, Set[NonReachablePosition]]
 
   val dyncandidates = mutable.Map.empty[Sig, mutable.Set[Global]]
   val dynsigs       = mutable.Set.empty[Sig]
@@ -23,6 +28,7 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
   loader.classesWithEntryPoints.foreach(reachClinit)
 
   def result(): Result = {
+    reportMissing()
     cleanup()
 
     val defns = mutable.UnrolledBuffer.empty[Defn]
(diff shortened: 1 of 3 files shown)


tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -738,4 +772,10 @@ object Reach {
     reachability.process()
     reachability.result()
   }
+
+  private[scalanative] case class NonReachablePosition(path: Path, column: Int)
+
+  case class ReachabilityFailureException(msg: String)
+      extends RuntimeException(msg)
+      with NoStackTrace
```
sjrd, line 780: I don't see a reason for `NoStackTrace` here. `NoStackTrace` is useful for exceptions thrown in non-failure paths, which must therefore be fast to initialize. This is not the case here.

Perhaps you wanted to avoid the stack trace to be displayed in the sbt shell? If that's the case, then there is a better solution: the sbt plugin should catch the exception and wrap it in a `MessageOnlyException`.

sjrd: Also, consider extending `Exception` rather than `RuntimeException`. Or better: don't we have a generic `LinkingException` somewhere? If not, maybe we should define one.

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -728,6 +736,32 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
...
+    if (missing.nonEmpty) {
+      val log = config.logger
+      log.error(s"Found ${missing.size} missing definitions while linking")
+      missing.foreach {
+        case (global, positions) =>
+          log.error(s"Not found $global")
+          positions.toList
+            .sortBy(p => (p.path, p.column))
+            .foreach { pos =>
+              log.error(s"\tat ${pos.path.toString}:${pos.column}")
```
sjrd, line 757: The column? Shouldn't this be the line?

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -728,6 +736,32 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
...
+            }
+
```
sjrd, line 759: Spurious blank line.

sjrd (changes requested):
This is cool :)

sjrd (approved):

PR #2070, opened 2020-12-14
NIRCompiler: moved to the right place

catap:
`NIRCompiler` inside package `scala.scalanative` this mean that it should be at `scala/scalanative`.


ekrich:
Given that it is in a compiler directory, I would think the package should be `scala.scalanative.compiler`.

catap:
@ekrich it was the first idea. After that I discovered https://github.com/scala-native/scala-native/blob/master/tools/src/test/scala/scala/scalanative/NIRCompiler.scala

WojciechMazur (approved):

PR #2071, opened 2020-12-15
Support boehm installed from macports

catap:
Right now the code base contains hardcoded path to boehm that is installed by brew.

MacOS world has an alternative that is called macports and it installs everything inside `/opt/local`.

This commit fixed build scala native application with boehm GC on system where boehm is installed via macports.

It also introduced two more environment variables `SCALANATIVE_INCLUDE_DIRS` and `SCALANATIVE_INCLUDE_DIRS` that allows to control which dirs should be used without hardcode inside `build.sbt`.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/Discover.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Discover.scala
@@ -1,6 +1,7 @@
 package scala.scalanative
 package build
 
+import java.io.File
 import java.nio.file.{Files, Path, Paths}
 import scala.util.Try
 import scala.sys.process._
@@ -43,10 +44,16 @@ object Discover {
   /** Find default clang compilation options. */
   def compileOptions(): Seq[String] = {
     val includes = {
-      val includedir =
+      val llvmIncludeDir =
         Try(Process("llvm-config --includedir").lineStream_!.toSeq)
           .getOrElse(Seq.empty)
-      ("/usr/local/include" +: includedir).map(s => s"-I$s")
+
+      val includeDirs =
+        getenv("SCALANATIVE_INCLUDE_DIRS")
+          .map(_.split(File.pathSeparatorChar).toSeq)
+          .getOrElse(Seq("/usr/local/include", "/opt/local/include"))
+
+      (includeDirs ++ llvmIncludeDir).map(s => s"-I$s")
     }
     includes :+ "-Qunused-arguments"
   }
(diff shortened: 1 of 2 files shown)


tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -43,21 +43,33 @@ object Discover {
   /** Find default clang compilation options. */
   def compileOptions(): Seq[String] = {
     val includes = {
-      val includedir =
-        Try(Process("llvm-config --includedir").lineStream_!.toSeq)
+      val llvmIncludeDir =
+        Try(Process("llvm-config --includedir").lineStream_!)
```
sjrd, line 47: The `.toSeq` must stay. Otherwise exceptions thrown during reading will not be caught by the `Try`.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -43,21 +43,33 @@ object Discover {
...
     includes :+ "-Qunused-arguments"
   }
 
   /** Find default options passed to the system's native linker. */
   def linkingOptions(): Seq[String] = {
     val libs = {
-      val libdir =
-        Try(Process("llvm-config --libdir").lineStream_!.toSeq)
+      val llvmLibDir =
+        Try(Process("llvm-config --libdir").lineStream_!)
```
sjrd, line 64: The `.toSeq` must stay. Otherwise exceptions thrown during reading will not be caught by the `Try`.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -43,21 +43,33 @@ object Discover {
...
-      val includedir =
-        Try(Process("llvm-config --includedir").lineStream_!.toSeq)
+      val llvmIncludeDir =
+        Try(Process("llvm-config --includedir").lineStream_!)
           .getOrElse(Seq.empty)
-      ("/usr/local/include" +: includedir).map(s => s"-I$s")
+
+      val includeDirs =
+        getenv("SCALANATIVE_INCLUDE_DIRS")
+          .map(_.split(" ").toSeq)
```
sjrd, line 52: A space in an env variable? That doesn't seem like the best choice. I would expect `java.io.File.pathSeparatorChar`.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -43,21 +43,33 @@ object Discover {
...
-      val libdir =
-        Try(Process("llvm-config --libdir").lineStream_!.toSeq)
+      val llvmLibDir =
+        Try(Process("llvm-config --libdir").lineStream_!)
           .getOrElse(Seq.empty)
-      ("/usr/local/lib" +: libdir).map(s => s"-L$s")
+
+      val libDirs =
+        getenv("SCALANATIVE_LIB_DIRS")
+          .map(_.split(" ").toSeq)
```
sjrd, line 69: Same here.

sjrd (changes requested):

tools/src/main/scala/scala/scalanative/build/Discover.scala
catap: @sjrd I love it. Thanks!

catap: @sjrd technically it works because env is null terminated :) but I agree that it quite odd and can broke something. A lot of things.

ekrich:
I think this is going to be annoying for people not using MacPort. Can we check to see if the directory exists before we add it?
```
[error] ld: warning: directory not found for option '-L/opt/local/lib'
```

catap:
@ekrich very good point. Thanks.

PR #2073, opened 2020-12-16
Throw an OutOfMemoryError if malloc cannot allocate

catap:
Malloc may returns `NULL` when it can't allocate required memory.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/unsafe/Zone.scala
+++ b/nativelib/src/main/scala/scala/scalanative/unsafe/Zone.scala
@@ -51,6 +51,9 @@ object Zone {
         throw new IllegalStateException("zone allocator is closed")
       }
       val rawptr = libc.malloc(size)
+      if (rawptr == null) {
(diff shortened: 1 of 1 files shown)


ekrich:
You were seeing this problem or is this just defense - good programming practice?

Also, should this be an `OutOfMemoryError` or something? I am assuming that the exception should not be caught vs `NullPointerException` which normally can be?

This has a pretty good discussion if anyone is interested - https://stackoverflow.com/questions/9101597/under-what-circumstances-can-malloc-return-null

catap:
@ekrich I don't know to be honest.

Modern linux allows to allocate huge amount of memory and it uses OOM Killer when the kernel hasn't got enough :) So, I feel that it can't be achieved easily, but it can be achieved at same cases.

Regarding `OutOfMemoryError` I guess it is better exception. Much better.

catap:
@ekrich yes, it is more human friendly. Thanks!

sjrd (approved):

PR #2075, opened 2020-12-16
Fix #2074: Compare unsigned types by value in equals/hashCode

### LINKED ISSUE #2074: Unsigned types should be compared by their underlying value
Current implementation of unsigned types (`ULong`, `UInt`, etc.) does not override `Object.equals` and `Object.hashCode` methods what  leads comparing them using reference equality. 
This behaviour leads to multiple unexpected bugs, eg. when comparing to unsigned values with same underlying value results in `false`

WojciechMazur:
This PR adds overrides for `Object.equals` and `Object.hashCode` methods in unsigned types. 

Resolves #2074

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/unsigned/UByte.scala
+++ b/nativelib/src/main/scala/scala/scalanative/unsigned/UByte.scala
@@ -266,7 +266,7 @@ final class UByte private[scalanative] (
   @inline override def hashCode(): Int = underlying.##
 
   @inline override def equals(obj: Any): Boolean = obj match {
-    case that: UByte => this == that
+    case that: UByte => this.underlying == that.underlying
     case _           => false
   }
 
--- a/nativelib/src/main/scala/scala/scalanative/unsigned/UByte.scala
+++ b/nativelib/src/main/scala/scala/scalanative/unsigned/UByte.scala
@@ -263,6 +263,13 @@ final class UByte private[scalanative] (
 
   @inline final override def toString(): String = toInt.toString()
 
+  @inline override def hashCode(): Int = underlying.##
+
+  @inline override def equals(obj: Any): Boolean = obj match {
+    case that: UByte => this == that
+    case _           => false
+  }
+
   // "Rich" API
 
   @inline final def max(that: UByte): UByte =
--- a/unit-tests/src/test/scala/scala/scalanative/unsigned/UnsignedEqualityTest.scala
+++ b/unit-tests/src/test/scala/scala/scalanative/unsigned/UnsignedEqualityTest.scala
@@ -5,7 +5,7 @@ import org.junit.Assert._
 
 class UnsignedEqualityTest {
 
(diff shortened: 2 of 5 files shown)


ekrich:
So different types are not comparable without promoting the smaller type to the larger type?

nativelib/src/main/scala/scala/scalanative/unsigned/UByte.scala
```diff
@@ -263,6 +263,13 @@ final class UByte private[scalanative] (
 
   @inline final override def toString(): String = toInt.toString()
 
+  @inline override def hashCode(): Int = underlying.##
+
+  @inline override def equals(obj: Any): Boolean = obj match {
+    case that: UByte => this == that
```
sjrd, line 269: We should compare the `underlying`s, otherwise we'll have to go again through the dispatch of `__scala_==`, although we already know that we have a specific type:
```suggestion
    case that: UByte => this.underlying == that.underlying
```
The same applies for all the other types, of course.

unit-tests/src/test/scala/scala/scalanative/unsigned/UnsignedEqualityTest.scala
```diff
@@ -0,0 +1,41 @@
+package scala.scalanative.unsigned
+
+import org.junit.Test
+import org.junit.Assert._
+
+class UnsignedEqualityTest {
+
+  def testEquality[T <: AnyRef](u1: T, u2: T, u3: T): Unit = {
```
sjrd, line 8: `T` doesn't seem to be used. You could use `AnyRef` for the parameters instead.

sjrd (changes requested):

sjrd:
> So different types are not comparable without promoting the smaller type to the larger type?

Not through `equals`. They are comparable through `==`, which uses a different dispatch for those types. This behavior is consistent with regular (signed) numeric types.

WojciechMazur:
This change for some unknown reason lead to loggin of test duration in JUnit, mesured in seconds, using scientific/exponent notation. Needs to be rebased onto fix in #2076

LeeTibbert (commented):
@WojciechMazur  Thank you for the change of this PR.

Pardon me for not understanding, but j.u.Map.scala seems identical
to a recently merged PR.  I suspect this  file may disappear as a changed
file when this PR is rebased on the Reporter PR which it requires.

WojciechMazur:
> @WojciechMazur Thank you for the change of this PR.
> 
> Pardon me for not understanding, but j.u.Map.scala seems identical
> to a recently merged PR. I suspect this file may disappear as a changed
> file when this PR is rebased on the Reporter PR which it requires.

It's probably just some artifact after rebase + force push.

sjrd:
The related changes LGTM now, but those other commits and their changes shouldn't be here. Perhaps another interactive rebase on top of master is necessary.

sjrd (approved):

PR #2076, opened 2020-12-16
Enforce usage of fixed notation for time in JUnit reporter

WojciechMazur:
After appling changes in #2074 JUnit Reporter, for some unknown reason, started to log test duration using scientific / exponent notation which lead to failures in `junitTestOutputsNative/test`. 

This PR enforces usage of fixed notation with precision limited to 3 digits.

### DIFF
--- a/junit-runtime/src/main/scala/scala/scalanative/junit/Reporter.scala
+++ b/junit-runtime/src/main/scala/scala/scalanative/junit/Reporter.scala
@@ -24,7 +24,7 @@ private[junit] final class Reporter(eventHandler: EventHandler,
         Ansi.c(s", ", Ansi.BLUE) +
         Ansi.c(s"$ignored ignored",
                if (ignored == 0) Ansi.BLUE else Ansi.YELLOW) +
-        Ansi.c(s", $total total, ${timeInSeconds}s", Ansi.BLUE)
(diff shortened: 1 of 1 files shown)


junit-runtime/src/main/scala/scala/scalanative/junit/Reporter.scala
```diff
@@ -102,9 +102,10 @@ private[junit] final class Reporter(eventHandler: EventHandler,
       ""
     }
 
-    val m = formatTest(method, Ansi.RED)
+    val m             = formatTest(method, Ansi.RED)
+    val timeFormatted = f"$timeInSeconds%.3f"
     val msg =
-      s"$prefix$m failed: $fmtName${ex.getMessage}, took $timeInSeconds sec"
+      s"$prefix$m failed: $fmtName${ex.getMessage}, took $timeFormatted sec"
```
sjrd, line 108: Why not
```suggestion
      f"$prefix$m failed: $fmtName${ex.getMessage}, took $timeInSeconds%.3f sec"
```
?

WojciechMazur: I'm not too familiar with `f' Interpolator and thought I'd need to add `%s` for each other usage. Great to know it can be used that way. Thank you for showing me that.

sjrd (approved):

PR #2077, opened 2020-12-16
Prevent a memory overflow in unit tests

catap:
This test allocated on stack an array with 32 elements but walking for 100 elements.

Nobody knows that it touches in memory and which side effects it creates.

Technically it should be safe, but nobody knows :)

### DIFF
--- a/unit-tests/src/test/scala/scala/scalanative/unsafe/CArrayOpsTest.scala
+++ b/unit-tests/src/test/scala/scala/scalanative/unsafe/CArrayOpsTest.scala
@@ -14,7 +14,7 @@ class CArrayOpsTest {
     val ptr   = alloc.asInstanceOf[Ptr[Int]]
 
     assertTrue(ptr == carr.at(0))
-    (1 to 100).foreach { i => assertTrue((ptr + i) == carr.at(i)) }
+    (1 to 31).foreach { i => assertTrue((ptr + i) == carr.at(i)) }
   }
 
   @Test def applyUpdate(): Unit = {


unit-tests/src/test/scala/scala/scalanative/unsafe/CArrayOpsTest.scala
```diff
@@ -14,7 +14,7 @@ class CArrayOpsTest {
     val ptr   = alloc.asInstanceOf[Ptr[Int]]
 
     assertTrue(ptr == carr.at(0))
-    (1 to 100).foreach { i => assertTrue((ptr + i) == carr.at(i)) }
+    (1 to 31).foreach { i => assertTrue((ptr + i) == carr.at(i)) }
```
ekrich, line 17: Shouldn't 32 elements be from 0 to 31?

catap: @ekrich `0` is tested at line 16 :)

ekrich: Ok thanks. Did not look at the whole file.

sjrd:
Where does this 32 come from? Could we add a comment at the beginning of this test class that explains why we're testing for 32 elements?

catap:
@sjrd the tests started as:
```
@Test def atN(): Unit = {
    val alloc = stackalloc[CArray[Int, Digit2[_3, _2]]]
    val carr  = !alloc
     val ptr   = alloc.asInstanceOf[Ptr[Int]]
...
```

As you can see it allocated an array with 32 elements. Why 32? I have no idea :) Just a number.

Which comment should I add?

sjrd:
Oh it's the `Digit2[_3, _2]` that literally means 32! I was not seeing that. Never mind, then.

sjrd (approved):

PR #2087, opened 2020-12-23
Remove dummy java.time implementation from core repo.

WojciechMazur:
This PR continues the topic of removal of dummy implementations from SN core repository started in #2079, by removal of `java.time` package, or exactly `Instant` and `Duration`. 
It adopts only usage of Instant of core repo inside `FileTime` to be not directly dependent from `Instant`. 

After merging this PR SN users should use a custom implementation of `java.time`, eg. [scala-java-time](https://github.com/cquiroz/scala-java-time). Alternatively, dummy implementation should be moved to a separate repo and published.

This PR needs to be rebased on top of merged #2079, as it bootstraps `javalib-ext-dummies` and `testsExt` projects.


javalib/src/main/scala/java/nio/file/attribute/FileTime.scala
```diff
@@ -3,37 +3,43 @@ package java.nio.file.attribute
 import java.util.concurrent.TimeUnit
 import java.time.Instant
 
-final class FileTime private (instant: Instant) extends Comparable[FileTime] {
-  def compareTo(other: FileTime) =
-    instant.compareTo(other.toInstant())
+final class FileTime private (private val epochMillis: Long,
+                              private val nanos: Int)
```
sjrd, line 7: If we store `epochMillis` and `nanos`, we will not be able to represent all the values that `FileTime` is supposed to hold. From looking at [the JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/FileTime.html), it's not clear, but I got the intuition that the canonical representation is a `value: Long` and `unit: TimeUnit`. Experimentation in a REPL suggests that my intuition is correct:
```scala
scala> import java.nio.file.attribute.FileTime
import java.nio.file.attribute.FileTime

scala> import java.util.concurrent.TimeUnit
import java.util.concurrent.TimeUnit

scala> FileTime.from(Long.MaxValue - 10L, TimeUnit.DAYS)
val res10: java.nio.file.attribute.FileTime = 219250468-12-04T15:30:07Z

scala> res10.to(TimeUnit.DAYS)
val res11: Long = 9223372036854775797 // exact number of days kept

scala> res10.to(TimeUnit.MILLISECONDS)
val res12: Long = 9223372036854775807 // overflowed to Long.MaxValue

scala> FileTime.from(Long.MaxValue - 10L, TimeUnit.NANOSECONDS)
val res13: java.nio.file.attribute.FileTime = 2262-04-11T23:47:16.854775797Z

scala> res13.to(TimeUnit.DAYS)
val res15: Long = 106751

scala> res13.to(TimeUnit.NANOSECONDS)
val res16: Long = 9223372036854775797 // exact number of nanoseconds kept
```
Therefore, I strongly believe that we should store `value: Long` and `unit: TimeUnit`, and adapt everything in terms of that canonical representation.

The above REPL session can be a good source of additional test cases.

sjrd (changes requested):
Do we still need `java.time.Duration` and `java.time.temporal.TemporalAmount` at all? Can't we entirely remove them, instead of moving them to javalib-ext-dummies?

javalib/src/main/scala/java/nio/file/attribute/FileTime.scala
```diff
@@ -3,15 +3,17 @@ package java.nio.file.attribute
...
   def compareTo(other: FileTime) = {
-    val compareMillis = epochMillis.compareTo(other.epochMillis)
-    if (compareMillis == 0) {
-      nanos.compareTo(other.nanos)
-    } else compareMillis
+    if (this.unit == other.unit) {
+      this.value.compareTo(other.value)
+    } else {
+      val compareUnit = TimeUnit.NANOSECONDS
+      this.to(compareUnit).compareTo(other.to(compareUnit))
```
sjrd, line 15: I don't think this will work in all cases. If one is stored in DAYS, its value in ns wouldn't contain all the information, and so it might compare equal to a `FileTime` that is stored in NANOSECONDS with that same value.

I think the easiest way to get this right is to compare *both* at the level of DAYS and NANOSECONDS. That would be enough, because 1 Day can be represented as a Long value in ns. Since DAYS and NANOSECONDS are the two extreme units, we would cover the whole range. So it would be something like
```suggestion
      val daysComp = this.to(TimeUnit.DAYS).compareTo(other.to(TimeUnit.DAYS))
      if (daysComp != 0)
        daysComp
      else
        this.to(TimeUnit.NANOSECONDS).compareTo(other.to(TimeUnit.NANOSECONDS))
```

javalib/src/main/scala/java/nio/file/attribute/FileTime.scala
```diff
@@ -21,25 +23,31 @@ final class FileTime private (private val epochMillis: Long,
...
-  def toMillis(): Long = epochMillis
+  def toMillis(): Long = to(TimeUnit.MILLISECONDS)
 
-  def toInstant(): Instant = Instant.ofEpochMilli(epochMillis).plusNanos(nanos)
+  def toInstant(): Instant = {
+    val asNanos = to(TimeUnit.NANOSECONDS)
+    val seconds = to(TimeUnit.SECONDS)
+    Instant.ofEpochSecond(
+      seconds,
+      asNanos - TimeUnit.NANOSECONDS.convert(seconds, TimeUnit.SECONDS))
```
sjrd, line 37: Easier and more clearly correct:
```suggestion
      Math.floorMod(asNanos, 1000000000)
```

javalib/src/main/scala/java/nio/file/attribute/FileTime.scala
```diff
@@ -21,25 +23,31 @@ final class FileTime private (private val epochMillis: Long,
...
-  def fromMillis(value: Long): FileTime =
-    new FileTime(value, 0)
+  def fromMillis(value: Long): FileTime = from(value, TimeUnit.MILLISECONDS)
 
-  def from(instant: Instant): FileTime =
-    new FileTime(instant.toEpochMilli, instant.getNano)
+  def from(instant: Instant): FileTime = {
+    val secondsAsNanos =
+      TimeUnit.NANOSECONDS.convert(instant.getEpochSecond, TimeUnit.SECONDS)
+    new FileTime(secondsAsNanos + instant.getNano, TimeUnit.NANOSECONDS)
```
sjrd, line 51: This will not work for large `Instant`s that cannot be represented in a Long value of nanoseconds. We must start with the Seconds, and only keep more precision (in milliseconds or nanoseconds) if the seconds are <= `Long.MaxValue / 1000` (or `1000000` for nanoseconds).

sjrd (changes requested):

unit-tests-ext/src/test/scala/java/nio/file/attribute/FileTimeExtTest.scala
```diff
@@ -0,0 +1,106 @@
...
+    val maxSecondsWithMillis = Long.MaxValue / 1e3.toLong - 1L
+
+    def toUnit(seconds: Long, nanos: Long, unit: TimeUnit) = {
+      val fromSeconds = unit.convert(seconds, TimeUnit.SECONDS)
+      val fromNanos   = unit.convert(nanos, TimeUnit.NANOSECONDS)
+      fromSeconds + fromNanos
+    }
+
+    def fromSeconds(seconds: Long, unit: TimeUnit) =
+      FileTime.from(toUnit(seconds, Instant.MAX.getNano, unit), unit)
```
sjrd, line 50: What does `Instant.MAX.getNano` represent? Why is this better than using `0L`? This deserves a comment.

unit-tests-ext/src/test/scala/java/nio/file/attribute/FileTimeExtTest.scala
```diff
@@ -0,0 +1,106 @@
...
+      Instant.ofEpochSecond(timestampAsSeconds))
+
+    fileTimeFromInstantEquals(
+      FileTime.from(timestampAsNanos + nanosAdjustment, TimeUnit.NANOSECONDS),
+      Instant.ofEpochSecond(timestampAsSeconds, nanosAdjustment))
+  }
+
+  @Test def fromInstantMaxPrecision(): Unit = {
+    val maxSecondsWithNanos  = Long.MaxValue / 1e9.toLong - 1L
+    val naxSecondsWithMicros = Long.MaxValue / 1e6.toLong - 1L
```
sjrd, line 40:
```suggestion
    val maxSecondsWithMicros = Long.MaxValue / 1e6.toLong - 1L
```

unit-tests-ext/src/test/scala/java/nio/file/attribute/FileTimeExtTest.scala
```diff
@@ -0,0 +1,106 @@
...
+                  TimeUnit.MICROSECONDS)
+
+    testPrecision(naxSecondsWithMicros,
+                  TimeUnit.MICROSECONDS,
+                  TimeUnit.MILLISECONDS)
+
+    testPrecision(maxSecondsWithMillis, TimeUnit.MILLISECONDS, TimeUnit.SECONDS)
+
+  }
+  @Test def toInstant(): Unit = {
```
sjrd, line 92:
```suggestion
    testPrecision(maxSecondsWithMillis, TimeUnit.MILLISECONDS, TimeUnit.SECONDS)
  }

  @Test def toInstant(): Unit = {
```

unit-tests-ext/src/test/scala/java/nio/file/attribute/FileTimeExtTest.scala
```diff
@@ -0,0 +1,106 @@
...
+    fileTimeFromInstantEquals(
+      FileTime.from(timestampAsSeconds, TimeUnit.SECONDS),
+      Instant.ofEpochSecond(timestampAsSeconds))
+
+    fileTimeFromInstantEquals(
+      FileTime.from(timestampAsNanos + nanosAdjustment, TimeUnit.NANOSECONDS),
+      Instant.ofEpochSecond(timestampAsSeconds, nanosAdjustment))
+  }
+
+  @Test def fromInstantMaxPrecision(): Unit = {
```
sjrd, line 38: It seems to me that this test is overengineered, to the point where it is difficult to understand what it tests, and how. Consider using "dumb" tests instead, using known input/output pairs and relying on `FileTime.equals`:
```scala
assertEquals(FileTime.from(1L, DAYS), FileTime.from(Instant.ofEpochSecond(86400L, 0)))
```

sjrd: Playing more with that, I think I was wrong before. Storing `value` and `unit` is not enough. Apparently `FileTime` has strictly more precision than `Instant`. I could not find an `Instant` for which `FileTime.from(instant).toInstant() == instant` was false.

So, in fact, it seems the reality is much simpler: a `TimeUnit` stores a pair of `days: Long` and `nanos: Long`. Probably with `nanos` capped at `86400000000000L - 1L` (so working with a `floorMod(_, 86400000000000L)`), the number of nanoseconds in a day, to make things easier.

That would make the implementation of all methods, in particular `equals` and `compareTo`, much easier.

sjrd (changes requested):

unit-tests-ext/src/test/scala/java/nio/file/attribute/FileTimeExtTest.scala
WojciechMazur: Actually `FileTime.from(instant).toInstant() eq instant` also would be true. I belive JDK stores original Instant or created new one based on other parameters.

sjrd: Right, but we can't do that (because a field of type `Instant` would not link when we don't have `Instant` in the codebase), and there's certainly no requirement to offer `eq`. We should be able to guarantee `==`, though.

javalib/src/main/scala/java/nio/file/attribute/FileTime.scala
WojciechMazur: Guess it's outdated, as now, after conversion to (epochDays, dayNanos) we are having a greater range than Instant. Now we need to make sure conversion toInstant does not overexcites Instant limits. To check this additional tests were added to `FileTimeExtTest`

unit-tests-ext/src/test/scala/java/nio/file/attribute/FileTimeExtTest.scala
```diff
@@ -0,0 +1,88 @@
...
+                 fromInstantMax.to(TimeUnit.SECONDS))
+    assertEquals(Instant.MAX.getNano, fromInstantMax.toInstant.getNano)
+
+    val fromInstantMin = FileTime.from(Instant.MIN)
+    assertEquals(Instant.MIN.getEpochSecond,
+                 fromInstantMin.to(TimeUnit.SECONDS))
+    assertEquals(Instant.MIN.getNano, fromInstantMin.toInstant.getNano)
+  }
+
+  @Test def toInstantWhenLargerThenInstantRange(): Unit = {
```
sjrd, line 68:
```suggestion
  @Test def toInstantWhenLargerThanInstantRange(): Unit = {
```

unit-tests-ext/src/test/scala/java/nio/file/attribute/FileTimeExtTest.scala
```diff
@@ -0,0 +1,88 @@
...
+    assume(Instant.MAX.getEpochSecond != Long.MaxValue)
+    assume(Instant.MIN.getEpochSecond != Long.MinValue)
```
sjrd, line 70: `assume` here is misleading. I read that as a JUnit assumption, which would simply *ignore* the test if it is not true. Consider using `assert` instead.

sjrd (approved):
LGTM, except two tiny details below.

sjrd (approved):

PR #2088, opened 2020-12-23
Support `java.util.Date` methods using `java.time.Instant`.

WojciechMazur:
This PR adds methods added in Java8 to `java.util.Date` that is method `toInstant` and `Date.from(_: Instant)`
It should be rebased on top of merged #2079 as it contains bootstrap of `testsExt` project and on top of #2087 where `Instant.ofEpochSeconds` method used in tests was implemented.

### DIFF
--- a/javalib/src/main/scala/java/util/Date.scala
+++ b/javalib/src/main/scala/java/util/Date.scala
@@ -1,5 +1,7 @@
 package java.util
 
+import java.time.Instant
+
 /** Ported from Scala JS and Apache Harmony
  * - omits deprecated methods
  * - toString not jdk compatible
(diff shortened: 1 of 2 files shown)


ekrich:
You have a typo in the title - should be `java.util.Date`.

sjrd (changes requested):
This should be accompanied by a port of [`InstantTest`](https://github.com/scala-js/scala-js/blob/master/test-suite-ex/shared/src/test/scala/org/scalajs/testsuite/javalib/time/InstantTest.scala) in the unit-test-ext. This is important to make sure that relying on the `ArithmeticException` is indeed correct.

Otherwise, LGTM. (Probably shouldn't be Draft anymore.)

WojciechMazur:
`InstantTest` was ported in #2087, when it's merged I'll rebase this branch on top of it and convert from draft.

sjrd (approved):

PR #2091, opened 2020-12-24
Never use default path that doesn't exists

catap:
This commit is following up for c27fa07f757b0a83f54bdf817514db3920c64ad4.

When a system hasn't got `/opt/local/lib` it complains like:
```
[error] ld: warning: directory not found for option '-L/opt/local/lib'
```

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/Discover.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Discover.scala
@@ -41,6 +41,9 @@ object Discover {
     path
   }
 
+  private def filterExisting(paths: Seq[String]): Seq[String] =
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -41,6 +41,9 @@ object Discover {
     path
   }
 
+  private def existedOnly(paths: Seq[String]): Seq[String] =
```
sjrd, line 44:
```suggestion
  private def filterExisting(paths: Seq[String]): Seq[String] =
```
?

sjrd (approved):

PR #2095, opened 2020-12-29
Port localized String.to{Lower, Upper}Case from Scala.js

WojciechMazur:
This PR ports localized methods `String.toLowerCase` and `String.toUpperCase` from Scala.js along with needed changes in `java.lang.Character`. 
It was additionally extended to handle 2 corner cases inside `toLowerCase`: 
* Greek capital letter sigma  &#x03A3; `\u03A3` being last cased letter in word - normally it's mapped to &#x03c3;  `\u03C3`, but if it's last cased letter it should be &#x03c2; `\u03C2`
* Lating capital letter I wth dot above &#x0130;  `\u0130` - it should be mapped to two chars: &#x0069; `\u0069` and  &#x0307;  `\u0307`. 
This handling was needed only in ScalaNative, as it is nativly supported in JS

### DIFF
--- a/javalib/src/main/scala/java/lang/Character.scala
+++ b/javalib/src/main/scala/java/lang/Character.scala
@@ -1191,7 +1191,6 @@ object Character {
   private[lang] final val CombiningClassIsNone = 0
   private[lang] final val CombiningClassIsAbove = 1
   private[lang] final val CombiningClassIsOther = 2
- 
   
   /* Ported from Scala.js, commit: ac38a148, dated: 2020-09-25
    * Indices representing the start of ranges of codePoint that have the same
--- a/javalib/src/main/scala/java/lang/Character.scala
+++ b/javalib/src/main/scala/java/lang/Character.scala
@@ -410,12 +410,10 @@ object Character {
   private[this] def getTypeLT256(codePoint: Int): scala.Byte =
     charTypesFirst256(codePoint)
 
+  // Ported from Scala.js, commit: ac38a148, dated: 2020-09-25
   private[this] def getTypeGE256(codePoint: Int): scala.Byte = {
-    // the idx is increased by 1 due to the differences in indexing
-    // between charTypeIndices and charType
-    val idx = Arrays.binarySearch(charTypeIndices, codePoint) + 1
-    // in the case where idx is negative (-insertionPoint - 1)
-    charTypes(Math.abs(idx))
+    charTypes(
+      findIndexOfRange(charTypeIndices, codePoint, hasEmptyRanges = false))
   }
 
   // Two digit() constructors, digitWithValidRadix(), and
@@ -749,8 +747,9 @@ object Character {
     isMirrored(c.toInt)
 
   def isMirrored(codePoint: Int): scala.Boolean = {
-    val idx = Arrays.binarySearch(isMirroredIndices, codePoint) + 1
-    (Math.abs(idx) & 1) != 0
+    val indexOfRange =
+      findIndexOfRange(isMirroredIndices, codePoint, hasEmptyRanges = false)
+    (indexOfRange & 1) != 0
   }
 
   /* Conversions */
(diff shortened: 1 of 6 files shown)


sjrd:
Should the handling of Σ really be part of this PR? Is it really the last issue with those methods? I suspect that there are still other things from SpecialCasing.txt that are not handled. For example, uppercasing ß, which must become SS. I think those cases should all be handled together, in a separate PR.

javalib/src/main/scala/java/lang/Character.scala
```diff
@@ -1189,6 +1188,118 @@ object Character {
   private[this] lazy val isMirroredIndices =
     uncompressDeltas(isMirroredIndicesDeltas)
 
+  private[lang] final val CombiningClassIsNone = 0
+  private[lang] final val CombiningClassIsAbove = 1
+  private[lang] final val CombiningClassIsOther = 2
+ 
+
```
sjrd, line 1195: Double blank line.

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -667,22 +668,259 @@ final class _String()
     buffer
   }
 
-  def toLowerCase(locale: Locale): _String =
-    toCase(Character.toLowerCase)
+  /*  Ported from Scala.js, commit: ac38a148, dated: 2020-09-25
```
sjrd, line 671:
```suggestion
  /* Ported from Scala.js, commit: ac38a148, dated: 2020-09-25
```

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -667,22 +668,259 @@ final class _String()
...
+        case '\u0307' if afterI(i)     => ""
+        case '\u0049' if !beforeDot(i) => "\u0131"
+        case _                         => null
+      }
+    }
+
+    preprocessed.toLowerCase()
+  }
+
+  @inline def toLowerCase(): _String = toCase(Character.toLowerCase)
```
sjrd, line 834: To reduce the overall diff:
```suggestion
  @inline
  def toLowerCase(): _String = toCase(Character.toLowerCase)
```

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -667,22 +668,259 @@ final class _String()
...
+    val preprocessed = replaceCharsAtIndex { i =>
+      (this.charAt(i): @switch) match {
+        case '\u0069' => "\u0130"
+        case _        => null
+      }
+    }
+
+    preprocessed.toUpperCase()
+  }
+  @inline def toUpperCase(): _String = toCase(Character.toUpperCase)
```
sjrd, line 919: Missing blank line. Also, to reduce the overall diff:
```suggestion
  }
  
  @inline
  def toUpperCase(): _String = toCase(Character.toUpperCase)
```

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -711,6 +949,76 @@ final class _String()
...
+   *  - `toUpperCaseTurkishAndAzeri()`
+   *
+   *  @param replacementAtIndex
+   *    A function from index to `String | Null`, which should return a special
+   *    replacement string for the character at the given index, or `null` if
+   *    the character at the given index is not special.
+   */
+  @inline
+  private def replaceCharsAtIndex(replacementAtIndex: Int => String): String = {
+
```
sjrd, line 969: Spurious blank line.

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -711,6 +949,76 @@ final class _String()
...
+    var prep           = ""
```
sjrd, line 970: Using a `var` of type `String` is efficient in Scala.js, but not ideal in Scala Native. We should use a `java.lang.StringBuilder`. Ideally we would only allocate it lazily the first time `replacement != null`. Something like:
```suggestion
    var prep: java.lang.StringBuilder = null
```
and later
```scala
...
      if (replacement != null) {
        if (prep == null)
          prep = new java.lang.StringBuilder(len * 2)
        prep.append(this, startOfSegment, i)
        ...
```

sjrd (changes requested):

sjrd (approved):

PR #2097, opened 2020-12-29
Fix #1327 and #1415: Enforce C formatting via clangfmt with minimum version

### LINKED ISSUE #1327: Update clangfmt script to use clang-format version 6.0

ekrich:
This PR addresses C and C++ formatting
1. Enables the `scripts/clangfmt` check to run in CI
2. Checks the ~~minimum~~ `clang-format` version to be version 6 (like clang minimum version) ~~but allow newer formatters~~. Risk: It is unknown whether newer formatters will differ from older ones forcing developers to install an older version. If this PR passes CI then a newer one should work fine. **_Edit: both the command line options and the formatter itself won't allow this. Usage documentation added._**
3. The `CLANG_FORMAT_PATH` env var can be set to mitigate the risk above so as not to conflict with a newer LLVM.
4. The command `scripts/clangfmt --test` now does a dry run and does not format and with the error option returns non-zero when formatting needs to be applied. **_Edit: this only works when we get to version 10 when the options are added and the script it edited - working code and directions are left in the file._**
5. Formatting is not 100% deterministic so checking via `--test` and repeating could be needed - I saw this behavior in this PR due to trailing spaces and a multi-line comment. **_Edit: hand editing for wrapping (80 chars) can mitigate this and also allow for nicer formatting._**
6. Current formatting style could be tweaked if desired.


ekrich:
Nice try but no luck - command line options have changed.

nativelib/src/main/resources/scala-native/libunwind/assembly.h
```diff
@@ -18,26 +18,26 @@
...
+#define PPC64_OFFS_FP 312
+#define PPC64_OFFS_V 824
 #ifdef _ARCH_PWR8
 #define PPC64_HAS_VMX
 #endif
 #elif defined(__POWERPC__) || defined(__powerpc__) || defined(__ppc__)
 #define SEPARATOR @
 #elif defined(__arm64__)
-#define SEPARATOR %%
+#define SEPARATOR % %
```
lolgab, line 35: Is this correct?

ekrich: I had to fix the script first but now I can fix this. I need to revert that file and then do not format that file. I am not sure why everything works ok on `macOS` - it must not use the assembly.
```
// clang-format off
header in here
// clang-format on
```

ekrich: Ok, got that fixed. I had found that culprit but you found it faster.

ekrich:
@WojciechMazur This is ready for review. Most of the format changes are in our `libunwind` - it is an older version maybe before they used `clang-format` but was really poorly formatted. Some hand formatting can help out this version of `clang-format` if people work on the C code and don't like the way things wrap or something.

Pretty excited this is actually done.

Ignored: `libunwind/assembly.h` otherwise things get messed up and won't compile.

nativelib/src/main/resources/scala-native/libunwind/assembly.h
ekrich: Builds and runs now.

scripts/clangfmt
```diff
@@ -12,9 +12,8 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-# The required version of clang-format.
-CLANG_FORMAT_VERSION=5.0
-CLANG_FORMAT_VERSION_STRING="clang-format version $CLANG_FORMAT_VERSION"
+# The required version of clang-format (matches minimum clang version) for CI
```
WojciechMazur, line 15: Header of this file contains the comment `clang-format is from the most recent version of LLVM supported by Scala Native`, but we're using the opposite. We should fix it as it might be misleading in the future.

ekrich: Thanks for the review. I have fixed up the comment. I was really surprised too that it would be so inconsistent between versions. I was hopeful we wouldn't need special setup and was surprised that older versions don't have the `--dry-run` option.

scripts/clangfmt
```diff
@@ -61,14 +61,31 @@ while [ "$#" -gt 0 ]; do
...
+# Remove for version 10 and above
+opts="-i"
+err=
+
 if [ "$#" -gt 0 ]; then
-  "$clang_format" --style=file -i "$@"
+  "$clang_format" --style=file "$opts" "$@"
 else
-  find . -name "*.[ch]" -or -name "*.cpp" | xargs "$clang_format" --style=file -i
+  find . -name "*.[ch]" -or -name "*.cpp" | \
```
WojciechMazur, line 80: What about `.hpp` extensions? I think some people are also using them. We even have few of such files in `libunwind`.

scripts/clangfmt
```diff
@@ -26,28 +25,29 @@ die() {
...
-  done
+  if check_clang_format_version clang-format; then
+    clang_format=clang-format
+  fi
 fi
 
 if [ -z "$clang_format" ]; then
   die "clang-format version $CLANG_FORMAT_VERSION not found" \
-      "Install LLVM version $CLANG_FORMAT_VERSION and rerun."
+      "clang-format version $version found" \
```
WojciechMazur, line 48: Maybe let's change it to `clang-format version $CLANG_FORMAT_VERSION expected, but version $version found`?

nativelib/src/main/resources/scala-native/libunwind/include-libunwind/__libunwind_config.h
```diff
@@ -10,99 +10,102 @@
...
-#    error "Unsupported MIPS ABI and/or environment"
-#  endif
-#  define _LIBUNWIND_HIGHEST_DWARF_REGISTER _LIBUNWIND_HIGHEST_DWARF_REGISTER_MIPS
-# else
-#  error "Unsupported architecture."
-# endif
+#if defined(__i386__)
+#define _LIBUNWIND_TARGET_I386
+#define _LIBUNWIND_CONTEXT_SIZE 8
+#define _LIBUNWIND_CURSOR_SIZE 15
```
WojciechMazur, line 31: Is there a rule for clang-format to allow indented definitions that were existing before? Not a big deal if there is no such thing, but it better as easier to understand when it was structured

WojciechMazur (commented):
Just two proposals, or rather questions.  The most important one is formating `.hpp` file from seperate comment, they should also be formatted correctly.

scripts/clangfmt
ekrich: Yes, this has been added now.

scripts/clangfmt
ekrich: Very good idea, thanks.

nativelib/src/main/resources/scala-native/libunwind/include-libunwind/__libunwind_config.h
ekrich: I know this can be done, will find the option.

ekrich: Refer to https://clang.llvm.org/docs/ClangFormatStyleOptions.html
Search for `IndentPPDirectives `
I sort of like the last option better than the second: `PPDIS_BeforeHash`

What do you think?

ekrich: Gosh, after poking around on the Mac, most are just flush. I did see one file with the following `clapack.h` but it is not 4 spaces.
```
#include <stdint.h>
#if __has_include(<os/availability.h>)
# include <os/availability.h>
#else // __has_include(<os/availability.h>)
# undef API_AVAILABLE
# define API_AVAILABLE(...) /* Nothing */
#endif // __has_include(<os/availability.h>)
```
This doesn't seem to common so I'm wondering if we should just go with the defaults.

WojciechMazur: I think we can go with the defaults with this one. 

The only problem is clang-format 6.0 does not support BeforeHash indent https://releases.llvm.org/6.0.0/tools/clang/docs/ClangFormatStyleOptions.html

ekrich: I think that is fine for now. It was late last night and I was looking at the latest.

ekrich:
@WojciechMazur I rebased this due to doc conflicts. I made the docs more generic so they don't encode the versions in the text hopefully to make them easier to update. This should be ready to go now.

WojciechMazur (approved):

PR #2098, opened 2020-12-30
String to{Lower, Upper}Case handles Unicode special, non-locale dependent, casing rules

WojciechMazur:
This PR continues work originated in #2095 Scala.js port, becouse it have not contained handling for Unicode special handling for characters not being locale dpendent, as it is probably implemented natively in JS. 
Changes include handling of greek sigma letter, latin I with dot, and other characters specified in [SpecialCasing.txt](https://unicode.org/Public/UNIDATA/SpecialCasing.txt)

Before converting from draft PR it needs to be rebased onto changes in #2095.


javalib/src/main/scala/java/lang/String.scala
```diff
@@ -667,22 +668,598 @@ final class _String()
...
+       * # 03C2; 03C2; 03A3; 03A3; # GREEK SMALL LETTER FINAL SIGMA
+       *
+       * # Note: the following cases are not included, since they would case-fold in lowercasing
+       *
+       * # 03C3; 03C2; 03A3; 03A3; Final_Sigma; # GREEK SMALL LETTER SIGMA
+       * # 03C2; 03C3; 03A3; 03A3; Not_Final_Sigma; # GREEK SMALL LETTER FINAL SIGMA
+
+       */
+      (charAt(i): @switch) match {
+        case '\u03A3' if isFinalCased(i) => "\u03C2"
```
sjrd, line 906: Could you call this `if finalSigma(i)`, and quote in the comment of `finalSigma` the definition of `Final_Sigma` in Section 3.13 of the Unicode standard, like is done for `afterI` and `beforeDot`, for example?

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -667,22 +668,598 @@ final class _String()
...
+          ((c >= 0x24B6) && (c <= 0x24E9))    // Circled latin capital or small letters A to Z
+        }
+
+        @tailrec
+        def followsCased(i: Int): scala.Boolean = {
+          if (i < offset) false // Not found non cased in whole string
+          else {
+            val c = charAt(i)
+            // Outside word boundary and not found cased
+            if (c.isWhitespace) false
```
sjrd, line 865: Where does `isWhitespace` come from? I see no reference to whitespace in the Unicode spec for `Final_Sigma`. However I see the concept of "case-ignorable", which I don't see in the code.

We should stick 100% to the Unicode spec of `Final_Sigma`, and comment everything with the references that we're using to make the code directly comparable to the spec.

sjrd (changes requested):

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -667,22 +668,598 @@ final class _String()
...
+     * 1FF4; 1FF4; 038F 0345; 038F 0399; # GREEK SMALL LETTER OMEGA WITH OXIA AND YPOGEGRAMMENI
+     *
+     * 1FB7; 1FB7; 0391 0342 0345; 0391 0342 0399; # GREEK SMALL LETTER ALPHA WITH PERISPOMENI AND YPOGEGRAMMENI
+     * 1FC7; 1FC7; 0397 0342 0345; 0397 0342 0399; # GREEK SMALL LETTER ETA WITH PERISPOMENI AND YPOGEGRAMMENI
+     * 1FF7; 1FF7; 03A9 0342 0345; 03A9 0342 0399; # GREEK SMALL LETTER OMEGA WITH PERISPOMENI AND YPOGEGRAMMENI
+     */
+    replaceCharsAtIndex { i =>
+      val c = this.charAt(i)
+      if (c < 0x80) null // fast-forward ASCII characters
+      else if (c >= 0x1E96 && c <= 0x1E9A) (c: @switch) match {
```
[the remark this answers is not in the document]
WojciechMazur: The only reason behind this approach was to limit number of checks that needs to be done, so instead of ~100 checks we would need to perform only 6 range checks + remaining ungroupped cases matches in worst case (when character does not contain any special handling). Altough when I'm looking at this code now, maybe we should populate HashMap with this special cases and use it instead? I'm not sure which approach might be better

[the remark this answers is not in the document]
sjrd: Given the large number of cases, a `java.util.HashMap[Char, String]` would probably be better, indeed.

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -667,22 +668,598 @@ final class _String()
...
+      (this.charAt(i): @switch) match {
+        case '\u0069' => "\u0130"
+        case _        => null
+      }
+    }
+
+    preprocessed.toUpperCase()
+  }
+  @inline def toUpperCase(): _String = {
+    /* Generated based on unconditional mappings in [SpecialCasing.txt](https://unicode.org/Public/UNIDATA/SpecialCasing.txt)
```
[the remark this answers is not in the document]
ekrich: Should we change the `java.lang.Character` `toUppercase` and `toLowercase` to use Unicode 13.0.0 as well? I think that was the version for JDK 15+. https://www.oracle.com/java/technologies/javase/15-relnote-issues.html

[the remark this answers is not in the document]
sjrd: Not in this PR. Note that all the logic here comes from SpecialCasing.txt, which hasn't changed since at least as far back as Unicode 4. So even if we use a specific link to 13, we're not in contradictions with using an older version of Unicode elsewhere.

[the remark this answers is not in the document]
ekrich: Thanks for the clarification. It is not hard if we need it in the future.

javalib/src/main/scala/java/lang/String.scala
WojciechMazur: You're right, it was a kind of simplification used to match with JVM, eg. matching only characters within word boundary. I've replaced it with the correct implementation based on Unicode specification. To make it possible I had to add two additional methods to `Characters`  and associated arrays of range indices.

sjrd: > You're right, it was a kind of simplification used to match with JVM, eg. matching only characters within word boundary.

Hum, wait. Do you mean that the JVM's `toLowerCase()` behaves differently than the Unicode spec? Do you have an example input string where it's not the same?

WojciechMazur: Not really, by simplification I've meant simplification on our side, eg. test ```assertEquals("dσς aσς bσc", "DΣΣ AΣΣ BΣC".toLowerCase())`` was created as I've seen that only last sigma in each word should be changed and didn't know or interpretted incorectlly `caseIgnorable` rule from specification. 

In each case I've tested on JVM (for Final_Sigma rule) incompatibilites with Unicode spec comes from implementation of different Unicode version in given JDK. For example there are 181 more "cased" symbols in Unicode 13.0.0 when comparing to 10.0.0 used in JDK 11. I belive it's fixed in newer versions.

sjrd (changes requested):
This is much better now, thank you. I still have one concern over supplementary code points.

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -831,8 +831,61 @@ final class _String()
     preprocessed.toLowerCase()
   }
 
-  @inline
-  def toLowerCase(): _String = toCase(Character.toLowerCase)
+  @inline def toLowerCase(): _String = {
```
sjrd, line 834: This should probably not be `@inline` anymore.
```suggestion
  def toLowerCase(): _String = {
```

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -831,8 +831,61 @@ final class _String()
...
+       *     before C: \p{cased}(\p{case-ignorable})*
+       *     after C:  !((\p{case-ignorable})*\p{cased})
+       */
+      def isFinalSigma(idx: Int): scala.Boolean = {
+        import Character._
+
+        val hasCasedBefore = {
+          val j = skipCaseIgnorableCharsBackwards(idx)
+          j > 0 && isCased(this.codePointBefore(j)) ||
+          j == 0 && isCased(this.codePointAt(j))
```
sjrd, line 850: I think the last condition will always return false. If `this.codePointAt(0)` were `isCased`, it cannot be `isCaseIgnorable` at the same time, and therefore `skipCaseIgnorableCharsBackwards` would not have skipped it. So it would have returned a `j > 0`.
```suggestion
          val j = skipCaseIgnorableCharsBackwards(idx)
          j > 0 && isCased(this.codePointBefore(j))
```

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -993,37 +1054,63 @@ for (cp <- 0 to Character.MAX_CODE_POINT) {
...
 
-  private def skipCharsWithCombiningClassOtherThanNoneOrAboveForwards(
-      i: Int): Int = {
+  private def skipConditionalCharsForwards(i: Int)(
+      shouldSkip: Int => scala.Boolean): Int = {
     // scalastyle:off return
-    import Character._
-    val len = length()
+    // use offset + count instead of length to handle Strings created using subString method
+    val len = offset + count
```
sjrd, line 1061: `length()`, and `codePointAt()` should take care of `offset`. We shouldn't have to do this here. If we were manipulating `value` directly, we would have to do this; but when we're using `length()`, `charAt`, etc., we must not take `offset` into account, because those methods already do that. Perhaps `codePointAt` is wrong?

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -875,16 +928,17 @@ for (cp <- 0 to Character.MAX_CODE_POINT) {
...
-      j > 0 && (codePointBefore(j) match {
+      def isSoftDotted(codePoint: Int): scala.Boolean = codePoint match {
         case 0x0069 | 0x006a | 0x012f | 0x0268 | 0x0456 | 0x0458 | 0x1e2d |
             0x1ecb =>
           true
         case _ => false
-      })
+      }
+      j > 0 && isSoftDotted(codePointBefore(j)) ||
+      j == 0 && isSoftDotted(codePointAt(j))
```
sjrd, line 941: Same thing here. The last condition should always be false.
```suggestion
      j > 0 && isSoftDotted(codePointBefore(j))
```

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -993,37 +1054,63 @@ for (cp <- 0 to Character.MAX_CODE_POINT) {
...
+      if (!shouldSkip(cp))
+        return j - offset // Subtract offset so resulting index can be passed to codePointAt
       j -= Character.charCount(cp)
     }
     0
     // scalastyle:on return
   }
 
+  private def skipCharsWithCombiningClassOtherThanNoneOrAboveForwards(
+      i: Int): Int =
```
sjrd, line 1090: Please put `{}` around the body of methods that are not one-liners.

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -993,37 +1054,63 @@ for (cp <- 0 to Character.MAX_CODE_POINT) {
...
+    skipConditionalCharsForwards(i) { cp =>
+      import Character._
+      combiningClassNoneOrAboveOrOther(cp) == CombiningClassIsOther
+    }
+
+  private def skipCharsWithCombiningClassOtherThanNoneOrAboveBackwards(
+      i: Int): Int = skipConditionalCharsBackwards(i) { cp =>
+    import Character._
+    combiningClassNoneOrAboveOrOther(cp) == CombiningClassIsOther
+  }
```
sjrd, line 1100:
```suggestion
      i: Int): Int = {
    skipConditionalCharsBackwards(i) { cp =>
      import Character._
      combiningClassNoneOrAboveOrOther(cp) == CombiningClassIsOther
    }
  }
```

sjrd (changes requested):

javalib/src/main/scala/java/lang/String.scala
WojciechMazur: I've reverted changes related to `offset` and `count`, I looks that problem with operations on substring  was occouring due to some other bug which was already fixed in a meantime. Thanks you for for spotting this.

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -875,7 +927,6 @@ for (cp <- 0 to Character.MAX_CODE_POINT) {
   if (!output.contains('\u0307'))
     println(cp.toHexString)
 }
-
      */
```
sjrd, line 930: Spurious change.

javalib/src/main/scala/java/lang/String.scala
```diff
@@ -919,9 +970,16 @@ for (cp <- 0 to Character.MAX_CODE_POINT) {
   }
 
   @inline
-  def toUpperCase(): _String = toCase(Character.toUpperCase)
+  def toUpperCase(): _String = {
```
sjrd, line 973: This method should also lose its `@inline`.

sjrd (approved):
Only two trivial comments below. Otherwise LGTM.

sjrd (approved):

PR #2099, opened 2021-01-04
Fix #1849: Streamline clang version detection

### LINKED ISSUE #1849: Streamline clang version detection

ekrich:


ekrich:
I am actually really surprised this doesn't work - I will look into the configuration. On macOS, looking a `brew` install where links are set, you find `clang-n`, `clang`, and `clang++` all linked to `clang`. I also had difficulty using a non-symlinked `brew link` version of clang because the platform `ld` was found in the path and did not work correctly.

ekrich:
The previous detection did a `which` with all the listed versions and plain `clang` and `clang++` and grabbed the latest version so it would pickup `clang-6.0` for example. I think is is reasonable to have a `clang` on the path if one is installed but I don't think Ubuntu agrees so either a path needs to be added or the `CLANG_PATH` and `CLANGPP_PATH` needs to be added at least to CI.

https://ubuntu.pkgs.org/18.04/ubuntu-universe-amd64/clang-6.0_6.0-1ubuntu2_amd64.deb.html

ekrich:
Ready for review.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -80,90 +80,111 @@ object Discover {
...
-    def maybePath(p: Path) = p match {
-      case path if Files.exists(path) => Some(path.abs)
-      case none                       => None
-    }
 
-    def definesBuiltIn(
-        pathToClangBinary: Option[String]): Option[Seq[String]] = {
+    // this could be obsolete with our current minimum version
+    def checkDefinesBuiltIn(clang: String): Unit = {
+      val defineStrs = "__DECIMAL_DIG__ __LDBL_DECIMAL_DIG__"
```
WojciechMazur, line 92: Inded it might be obsolete check.`__DECIMAL_DIG__` preprocessor definition exists since clang 3.7
https://github.com/llvm/llvm-project/commit/4992099b42e9bb6faceb61ed093155bf2ebe7157

Probably we can remove this

WojciechMazur (changes requested):
In overall it looks good to me, we can probably only remove obsolete check and make passing env variable more explicit.

tools/src/main/scala/scala/scalanative/build/Discover.scala
ekrich: Ok, I just removed that in the first commit.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -27,16 +27,16 @@ object Discover {
   def GC(): GC =
     getenv("SCALANATIVE_GC").map(build.GC(_)).getOrElse(build.GC.default)
 
-  /** Find the newest compatible clang binary. */
+  /** Use the clang binary on the path or via CLANG_PATH env var. */
   def clang(): Path = {
-    val path = discover("clang", clangVersions)
```
ekrich, line 32: So should I just call it like this and remove all the internal code to simplify things?
```scala
val path = discover("clang", "CLANG_PATH")
// and
val path = discover("clang++", "CLANGPP_PATH")
```

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -80,90 +80,111 @@ object Discover {
...
 
+  /** Link to setup documentation */
+  private[scalanative] val docSetup =
+    "http://www.scala-native.org/en/latest/user/setup.html"
+
+  /** Discover the binary path using environment
+   *  variables or the command from the path.
+   */
+  private[scalanative] def discover(binaryName: String): Path = {
     val envName =
```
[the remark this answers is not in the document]
ekrich: So should I just call it like this and remove all the internal code to simplify things?
```scala
val path = discover("clang", "CLANG_PATH")
// and
val path = discover("clang++", "CLANGPP_PATH")
```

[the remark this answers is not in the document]
ekrich: Ok, I think that looks much better.

tools/src/main/scala/scala/scalanative/build/Discover.scala
ekrich: Thanks for checking. I thought we didn't need it but I also didn't know anything about it.

WojciechMazur (approved):

sjrd (approved):

PR #2106, opened 2021-01-12
Fix wrong positions of class fields

WojciechMazur:
This PR fixes bug occouring when generating position metadata for Class fields. Currently position of fields would always point to owner `ClassDef` position. 

After this fix position of created `nir.Defn.Var` would be based on position of `Tree` within `ClassDef` by finding `ValDef` such that its symbol matches symbol known form list of class decls.

### DIFF
--- a/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
+++ b/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
@@ -360,8 +360,9 @@ trait NirGenStat[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
                           Seq(_1, _2))
       }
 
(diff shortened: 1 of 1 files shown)


ekrich:
Not part of this PR but `genClassConstructorsInfo` is a beast.

WojciechMazur:
Yeah, `genClassConstructorsInfo` is huge. Also probably we can fix one bug also there. To `genClassConstructorsInfo` we're passing list of constructor symbols and implicit position of ClassDef. 
@sjrd do you think we use their symbol? I remember you once told me to always use `Tree` position instead, but maybe it should be only according to other usages, for example in usages `Op`. In such case we can also simplify logic fo class fields and also use field symbol position instead.

sjrd:
For `genClassFields` we can use the symbol's position, yes. That's OK because we know that symbol is always compiled in the current run.

sjrd (approved):

PR #2109, opened 2021-01-13
Support for Scala 2.12.13

WojciechMazur:
This PR fixes errors when compiling against 2.12.13. 

Changes in this PR include: 
 * Bumping up scala 2.12 version used in the build
 * Synchronization of scalalib override for `scala.reflect.ClassTag` - it has two new lazy values: `emptyArray` and `emptyWrappedArray`
 * Override of new `@nowarn` annotation in order to suppress self-warnings in compile
 * Fixed problems with wrong syntax


scalalib/overrides-2.12/scala/annotation/nowarn.scala
```diff
@@ -0,0 +1,37 @@
...
+ * Licensed under Apache License 2.0
+ * (http://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.annotation
+
+// This file was overidden to add @nowarn annotation removing compile time warnings
```
sjrd, line 15: Why do we need this?

scalalib/overrides-2.12/scala/reflect/ClassTag.scala
```diff
@@ -48,6 +49,13 @@ import java.lang.{ Class => jClass }
  */
 @scala.annotation.implicitNotFound(msg = "No ClassTag available for ${T}")
 trait ClassTag[T] extends ClassManifestDeprecatedApis[T] with Equals with Serializable {
+
+  @transient private[scala] lazy val emptyArray: Array[T] =
+    newArray(0)
+
+    @transient private[scala] lazy val emptyWrappedArray: mutable.WrappedArray[T] =
```
sjrd, line 56:
```suggestion
  @transient private[scala] lazy val emptyWrappedArray: mutable.WrappedArray[T] =
```

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -943,7 +943,7 @@ final class URI private () extends Comparable[URI] with Serializable {
     }
     if (!path.endsWith("/") && seglist.length > 0 &&
         include(seglist.length - 1)) {
-      newpath.deleteCharAt(newpath.length() - 1)
+      newpath.deleteCharAt(newpath.size - 1)
```
sjrd, line 946: I think the real issue here is that `newpath` uses a Scala `StringBuilder` instead of a `java.lang.StringBuilder`. This is an opportunity to address the root issue.

nativelib/src/main/scala/scala/scalanative/regex/Prog.scala
```diff
@@ -159,13 +159,13 @@ class Prog {
     val out = new StringBuilder()
     var pc  = 0
     while (pc < inst.size()) {
-      val len = out.length()
+      val len = out.length
```
sjrd, line 162: Same here: `out` should be a `java.lang.StringBuilder`.

scalalib/overrides-2.12/scala/annotation/nowarn.scala
WojciechMazur: When compiling in 2.12.13 we would get this warning, treated as an error in our build
```
[error] scala-native/scalalib/target/scalaSources/2.12.13/scala/annotation/nowarn.scala:34:7: Implementation restriction: subclassing Classfile does not
[error] make your annotation visible at runtime.  If that is what
[error] you want, you must write the annotation class in Java.
[error] class nowarn(value: String = "") extends ClassfileAnnotation
[error]       ^
[error] one error found

```
Logic in Scala 2.12.x compiler causing this problem can be seen [here](https://github.com/scala/scala/blob/e5580a6f616eeee0680c3a167d86f4bb755fc951/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L1837)
Definition of `nowarn` is defined [here](https://github.com/scala/scala/blob/2.12.x/src/library/scala/annotation/nowarn.scala)

Actually I've made mistake in this override. I've commited change where `class nowarn`  extends `StaticAnnotation` instead of original `ClassfileAnnotation`, as I thought that for some adding missing usage of `@nowarn` on class definition caused this problems. 

Probably we should conditionally remove treating warnings as errors for 2.12.13 in scalalib project.

sjrd: If it's in the `scalalib`, it's not our problem. We should simply disable fatal warnings in the scalalib rather than chaning the sources just to get rid of warnings. Warnings should be fatal for those writing the code in scala/scala, not for us who are just reusing and recompiling that code.

build.sbt
```diff
@@ -512,7 +519,19 @@ lazy val scalalib =
       scalacOptions += "-language:postfixOps",
       // The option below is needed since Scala 2.13.0.
       scalacOptions += "-language:implicitConversions",
-      scalacOptions += "-language:higherKinds"
+      scalacOptions += "-language:higherKinds",
+      /* Used to disable fatal warnings due to problems with compilation of `@nowarn` annotation */
+      scalacOptions := Def.task {
+        val prev = scalacOptions.value
+        val shouldSkipFatalWarnings = {
+          matchesScalaVersion(scalaVersion.value) {
```
sjrd, line 527: There is already `CrossVersion.partialVersion` for this in sbt. See an example here:
https://github.com/sbt/sbt/issues/6080#issuecomment-758194996

nativelib/src/main/scala/scala/scalanative/regex/Prog.scala
```diff
@@ -165,7 +164,7 @@ class Prog {
         out.append('*')
       }
       out
-        .append("        ".substring(out.length() - len))
+        .append("        ".substring(out.length - len))
```
sjrd, line 167: This shouldn't change anymore.

sjrd (changes requested):
Two small comments. Otherwise LGTM.

sjrd (approved):

PR #2111, opened 2021-01-14
Pass always to lto config

lolgab:
Currently the LTO setting is passed to the compiler only when the mode is not debug.
This is strange as a user because after setting the LTO in a project, you `nativeLink` 
and read in the logs that the LTO is `none` when you have set `thin`.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -215,10 +215,9 @@ private[scalanative] object LLVM {
       flto(config) ++ Seq("-rdynamic", "-o", outpath.abs) ++ target(config)
     val paths   = objectsPaths.map(_.abs)
     val compile = config.clangPP.abs +: (flags ++ paths ++ linkopts)
-    val ltoName = lto(config).getOrElse("none")
 
     config.logger.time(
-      s"Linking native code (${config.gc.name} gc, $ltoName lto)") {
+      s"Linking native code (${config.gc.name} gc, ${config.LTO.name} lto)") {
       config.logger.running(compile)
       Process(compile, config.workdir.toFile) ! Logger.toProcessLogger(
         config.logger)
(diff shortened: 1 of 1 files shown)


WojciechMazur:
As @densh stated in [PR introducing the LTO](https://github.com/scala-native/scala-native/pull/1281#issuecomment-405951397) he didn't passed it in debug mode, to optimize its compilation speed. The question is much how time overhead this change would produce?

lolgab:
@WojciechMazur 
Then should it be a "parameter" of the release mode, so you can set it only when you are in release mode? Because what is the point of allowing to set a parameter if then it doesn't get set?
We already have problem with the `optimize` flag. Debug mode is optimized for compilation speed, but the `optimize` flag is set to `true` in debug mode, which ( I imagine ) it consumes compilation time, right?

lolgab:
If you decide to merge this PR please amend the commit message, because it is wrong.
I improved the PR title but didn't amend the commit.

ekrich:
I agree with this change. We should allow `LTO` with `debug` as the `LTO` optimizations can cause problems and debugging is desired. I did some research and it seems in the `clang` `3.8` time frame there was a thread where `-O0` did not work with `LTO` but that was early days. Also, we have control of `NIR` optimizations but `LTO` is clearly out of our hands and in the hands of the LLVM tool chain. 

`LTO` `thin` should be relatively fast anyway as it uses parallelism. We could also introduce Incremental and Cache Pruning as well. https://clang.llvm.org/docs/ThinLTO.html

ekrich:
@lolgab Did you compare compile times in debug before and after this change to see if it was noticeable?

ekrich:
Here is an old article about LTO thin: http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html 
I would suspect that the performance has gotten better since the time this article was written. It would be nice to have some recent metrics to see the difference in speed especially on a larger project.

I also think there is some onus on the user to know what they are doing to some extent with different options. This PR makes things more transparent and less likely that we get asked the question why LTO thin is ignored.

ekrich:
There is a more substantial paper here but still a little old - https://www.researchgate.net/publication/314105281_ThinLTO_Scalable_and_incremental_LTO

lolgab:
> @lolgab Did you compare compile times in debug before and after this change to see if it was noticeable?

No, I didn't.

ekrich:
Well, after the change it can be checked if people are motivated.

ekrich:
@WojciechMazur Given that the default LTO is None, we are not forcing the user into a slower compile in debug mode, we are just allowing them to use LTO if they wish. I did find some references online where LTO thin was not working correctly with optimizing and they changed to debug with LTO to try and find out the cause (it was a pretty old version of LLVM). I think that makes the case for this change stronger in that at least diagnosing via debug can be done easier.

This needs a rebase. What do you think?

WojciechMazur (approved):

PR #2133, opened 2021-01-26
Fix: Type.isPtrBox always returns false

WojciechMazur:
`Type.isPtrBox` had a bug leading to returning false in any case, caused by using invalid Map of types. 

This method is important in terms of performance, as it used to determine if boxing/unboxing can be delayed or skipped entirely in the optimizer.

As a result of fixing this issue, in my - currently private - branch when I've experimented with CStructs support, it was possible to skip boxing when accessing or assigning a value of struct.

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Types.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Types.scala
@@ -121,7 +121,7 @@ object Type {
 
   def isPtrBox(ty: Type): Boolean = ty match {
     case refty: Type.RefKind =>
-      box.get(Type.Ref(refty.className)).contains(Type.Ptr)
+      unbox.get(Type.Ref(refty.className)).contains(Type.Ptr)
     case _ =>
       false
   }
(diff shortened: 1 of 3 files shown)


ekrich:
Boy, that sounds like really good news, minus the bug.

WojciechMazur:
Looks like there was some hidden bug in the Interflow, Now it's getting rid of boxing so good, that it removed while condition from the following code and created an infinite loop!  :D 
```
    type Cons = CStruct2[Int, Ptr[Byte]]
    Zone { implicit z =>
      val out  = collection.mutable.ListBuffer.empty[Int]
      var head = cons(10, cons(20, cons(30, null)))
      while (head != null) {
        out += head._1
        head = head._2.asInstanceOf[Ptr[Cons]]
      }
   }   
```

If we replace condition with `head != null && head.toLong != 0` there would be only the second check remaining. 
It's probably some issue in `Combine` phase

sjrd (approved):

PR #2139, opened 2021-01-31
Reduce indirection for the implementation of getClass

densh:
This PR adds an additional pointer field in the beginning of the rtti with the value of `classOf[Class[_]]`, thus making it usable an instance of `java.lang.Class[_]`. A few benefits of this change:

* Less indirection in the implementation of `object.getClass` (it's a single load now!)
* No need for caching, since classes are now allocated in readonly memory of the binary, similarly to constant strings
* Simpler runtime code without separation into `runtime.Type` and `java.lang.Class` (they are the same thing now).


densh:
This needs a bit more work to be fully baked:

- [ ] Drop RawType and replace it with `j.l.Class` everywhere
- [ ] Make sure that interflow's partial evaluation intrinsics are in agreement with new implementation of `j.l.Class`

lolgab:
Does this break backward compatibility?

sjrd:
How will we fit a proper handling of array `Class`es in this scheme? Currently all reference array classes share a unique `Class` (that of `Array[Object]`, but that is incorrect. At some point we'll have to fix that, and then we will *have to* lazily create classes for arrays (since we can have arrays of arrays, etc.). That means we will at least need a writeable field in `Class` (or `Type`) to store the lazily create Class of arrays of this class.

densh:
@lolgab It doesn't change any of the binary stable APIs. `scalanative.runtime` is not considered stable and can break at any time. 

@sjrd Class ids are dense and can be used to index into a hashmap or even an array stored at the top level. Then we can apply any mutations to the `tableForX[cls.id] = value`, instead of changing the instance of `cls.x = value`. As long as the fast path of `.getClass` is still efficient (no allocation, single load) I see it as an improvement from the current state of affairs.

densh:
Update: all the TODOs have been addressed.

nativelib/src/main/scala/scala/scalanative/runtime/Arrays.scala
```diff
@@ -153,11 +153,11 @@ object Array {
   }
 }
 
-// ###sourceLocation(file: "nativelib/src/main/scala/scala/scalanative/runtime/Arrays.scala.gyb", line: 168)
+// ###sourceLocation(file: "/home/denys/src/scala-native/nativelib/src/main/scala/scala/scalanative/runtime/Arrays.scala.gyb", line: 168)
```
WojciechMazur, line 156: This and the following should be replaced with a relative path.

tools/src/main/scala/scala/scalanative/interflow/PolyInline.scala
```diff
@@ -51,6 +51,7 @@ trait PolyInline { self: Interflow =>
       if (mode == build.Mode.ReleaseFast) {
         classCount <= 8 && implCount == 2
       } else {
+
```
WojciechMazur, line 54: Unnecessary new line

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -186,11 +199,8 @@ object Lower {
       case _                             => super.onVal(value)
     }
 
-    def genClassOf(buf: Buffer, node: ScopeInfo)(
-        implicit pos: Position): Val = {
-      val tpePtr = rtti(node).const
-      buf.call(toClassTy, toClassVal, Seq(Val.Null, tpePtr), Next.None)
-    }
+    def genClassOf(buf: Buffer, node: ScopeInfo)(implicit pos: Position): Val =
```
WojciechMazur, line 202: This method no longer needs neither `Buffer` nor `Position` argument. There is a single usage of this method in `genVal`  that can now be inlined.
 
Also, usages of `genVal` can now be reverted to the previous `onVal` which does not need `Buffer` argument, however, it can be addressed is separate PR.

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -4,6 +4,19 @@ package codegen
...
+  ClassTraitIdName,
+  ClassNameName,
+  ClassSizeName,
+  ClassIdRangeUntilName,
+  StringName,
+  StringValueName,
+  StringOffsetName,
+  StringCountName,
+  StringCachedHashCodeName
+}
```
WojciechMazur, line 19: Is such import really needed? Can't we import all of them or use qualified names like we used to do?

WojciechMazur (changes requested):
Looks good, I only have some minor styling adjustments

densh:
@WojciechMazur Thanks for the comments, I've addressed them in follow-up commits.

nativelib/src/main/scala/java/lang/Class.scala
```diff
@@ -19,10 +19,15 @@ object rtti {
...
-  @alwaysinline private def ty: Ptr[Type] =
-    fromRawPtr[Type](rawty)
+final class _Class[A] {
+  var id: Int           = _
+  var traitId: Int      = _
+  var name: String      = _
+  var size: Int         = _
+  var idRangeUntil: Int = _
+
+  def isClass: Boolean =
```
sjrd, line 29:
```suggestion
  private def isClass: Boolean =
```

sjrd: Even better, consider direction using `id < 0` in `isInterface()`, and use `!isInterface()` at the only other call site of `isClass`.

sjrd (changes requested):
One small comment. Otherwise LGTM.

densh:
@sjrd Good call, I've addressed the suggestion in e0459ae.

sjrd (approved):

WojciechMazur (approved):

PR #2141, opened 2021-02-01
Fix: #2135, segmentation-faults when working with file channels

WojciechMazur:
This PR provides minimal fix of issue #2135 . It adds missing method to javalib interface`SeekableByteChannel: postition(Long): SeekableByteChannel`. 
It also adds basic implementation of `Channels.new{Input/Output}Stream` (reuses implementation used in `java.nio.file.spi.FileSystemProvider`) allowing to run reproduction project linked in the issue.

In the following updates we should provide implementation for other stub methods in `java.nio.channels.Channels`.


javalib/src/main/scala/java/nio/channels/Channels.scala
```diff
@@ -1,8 +1,89 @@
...
-      is: java.io.InputStream): java.nio.channels.ReadableByteChannel = ???
+  def newInputStream(ch: AsynchronousByteChannel): InputStream = ???
+
+  @scalanative.annotation.stub
+  def newOutputStream(ch: AsynchronousByteChannel): OutputStream = ???
+
+  @scalanative.annotation.stub
+  def newChannel(in: InputStream): ReadableByteChannel = ???
+
+  @scalanative.annotation.stub
```
ekrich, line 56: Do you prefer to use fully qualify name so there is no import?

javalib/src/main/scala/java/nio/channels/Channels.scala
```diff
@@ -1,9 +1,94 @@
...
+      override def write(b: Array[Byte], off: Int, len: Int): Unit = {
+        channel.write(ByteBuffer.wrap(b, off, len))
+      }
+      override def close(): Unit =
+        channel.close()
+    }
+  }
+
+  @stub
+  def newInputStream(ch: AsynchronousByteChannel): InputStream = ???
```
sjrd, line 49: Is this, and other `@stub`s, actually required? It would be better not to declare them at all.

Overall, we should get to a state where we have no more `@stub` method left anywhere. Either it's implemented correctly, or it's not there.

sjrd (changes requested):

javalib/src/main/scala/java/nio/channels/Channels.scala
WojciechMazur: When methods will be declared as stubs and `linkStubs` in NativeConfig is set to `false` (default) user can get information about missing methods at compile-time, instead of getting unpleasant surprises in the runtime. Based on that it might be better to keep the annotations until we will provide the actual implementation of async channels.

sjrd: Ah, I meant to ask whether declaring those methods is required in the first place. Can't we just remove `newInputStream` altogether? As well as all the other things annotated with `@stub`.

WojciechMazur: Since https://github.com/scala-native/scala-native/pull/2162 was merged we should be able to remove them. They should be implicitly treated as stubs in such case.

sjrd (approved):

PR #2145, opened 2021-02-03
Allow to link as dynamic library

### LINKED ISSUE #155: Add support for linking Scala code as a native library

WojciechMazur:
This PR adds a feature allowing to link code as a shared library. All statically accessible methods annotated with `@export` annotation would be treated as entry points of the created library. 

Currently, only methods can be exported and used in a static library. 

With this in place, we can easily add additional support for static libraries in the next PRs 
Resolves #155 

Example of usage: 

``` sbt
// build.sbt
nativeConfig ~= { _
 .withBuildTarget(BuildTarget.libraryDynamic)
 .withLinkingOptions(Seq("-o", "libmylib.so" )) 
}
```

``` scala 
// mylib.scala

object myLib{
  @exportAccessors("mylib_current_count", "mylib_set_counter")
  var counter: Int = 0

  @exportAccessors("error_message")
  val ErrorMessage: CString = c"Something bad just happend!"

  @exported
  def addLongs(l: Long, r: Long): Long = l + r

  @exported("mylib_addInts")
  def addInts(l: Int, r: Int): Int = l + r
}
```

``` c
# libmylib.h

# This function needs to be called before invoking any methods defined in Scala Native.
# Returns 0 if successfull or non-zero in the otherwise
# It is used to setup GC, initialize instanstinable classes, etc.
int ScalaNativeInit(void);

int ScalaNativeInit(void);
long addLongs(long, long);
int addInts(int, int);
int mylib_current_count();
void mylib_set_counter(int);
```

``` c
# test.c
#include "libmylib.h"
#include <assert.h>

int main(int argc, char** argv){
  // assert(ScalaNativeInit() == 0); // Would be called automatically in dynamic libs by default
  addLongs(0L, 4L);
  mylib_addInts(4, 0);
  printf("Current count %d\n", mylib_current_count());
  mylib_set_counter(42);
}
```

Additional changes: 
* Logic of `ClassInstancesRegistry` needed to be changed. `scalanative.runtime.toClass` method was getting into an infinite loop. New minimal representation is all that we need to have to prevent the garbage collector from collecting these instances. 

Things that might be addressed in the following PRs:
* Hidding library symbols: Currently most of the symbols created from Scala methods are treated as globally accessible. Why might try to find a way to hide them from library users to prevent their usage in an uncontrolled manner. This step however might result in breaking stack trace printing - hidden symbols would be replaced with placeholder text. 
* ExternZone? - if we want to return heap-allocated pointer in the exported method we need to do this using low-level methods, eg, libc.malloc. Given the recent discussion about migration to `calloc` in Zone allocation, it might be worth thinking about Zone implementation that allows remaining some allocations (returned one / marked using the special method?) 
* Should we allow to export module fields - values and/or variables. Implementation of the latter might be problematic since we need to make sure that they were already initialized. Although global variables in C are tricky and hard to maintain.


docs/user/interop.rst
```diff
@@ -154,6 +154,48 @@ And then call it just like a regular Scala function:
 
    myprintf(c"2 + 3 = %d, 4 + 5 = %d", 2 + 3, 4 + 5)
 
+Exported methods
+----------------
+
+When linking Scala Native as library, you can mark its entry point functions with ``@export`` annotation,
+to make these definition C ABI compatible. Name of resulting function would match name of the method or it can be enforced
```
hepin1989, line 161: definitions

docs/user/sbt.rst
```diff
@@ -183,6 +183,25 @@ Use the following approach in `sbt` to set the target triple:
...
+    nativeConfig ~= { _.withBuildTarget(BuildTarget.libraryDynamic) }
+
+1. **application** (default)
+
+   Results in creating ready to use executable program.
+
+2. **libraryDynamic**
+
+   Results in dynamic library being built based on entry point methods annotated with `@export`,
+   for details see :ref:`interop`.
```
hepin1989, line 203: how about producing a header file too.

WojciechMazur: We can do this in either following PR or in seperate project outside core repo. I opt for the latter one. All informations needed to do this are contained in NIR format so we can e.q.: make it a part of `scala-native-bindgen`.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -243,4 +261,16 @@ private[scalanative] object LLVM {
       case Some(tt) => Seq("-target", tt)
       case None     => Seq("-Wno-override-module")
     }
+
+  private def buildCompileOpts(config: Config): Seq[String] =
+    config.compilerConfig.buildTarget match {
+      case BuildTarget.Application    => Seq()
+      case BuildTarget.LibraryDynamic => Seq("-fPIC")
```
ekrich, line 268: In the Clang command line docs this shows as `-fpic`. This is to avoid recompiling pre-compiled headers is I am reading this correctly?

WojciechMazur: I based it on this [thread](https://stackoverflow.com/questions/3544035/what-is-the-difference-between-fpic-and-fpic-gcc-parameters). So basically `-fPIC` is safer to use, altought maybe we should try to use `-fpic` when possible.

ekrich: Wow, somehow when I searched this the first time I didn't find `-fPIC` and only found `-fpic`. These are not the same things at all I would guess. The first one have no explanation.
https://clang.llvm.org/docs/ClangCommandLineReference.html

catap:
@WojciechMazur hey, any news on this one? Is it possible to expect on the next release?

WojciechMazur:
> @WojciechMazur hey, any news on this one? Is it possible to expect on the next release?

@catap Besides rebase it might need some additional tests before review. If there would not be any big issues it might be included in the next release.

catap:
@WojciechMazur thanks. If you need any help feel free to ask. Looks like I'll have a bit time soon ;)

adampauls:
Curious about the status here. Any sense of a timeline on when this ability might be available? Thanks!

WojciechMazur:
@adampauls Recently that feature request was not maintained, we're preparing to make a release 0.4.4 this week. I think it would be a good candidate to include in the next release 0.4.5 (in about 4-6 weeks from now), also since we would like to combine it with initial rust native sources integration.

armanbilge:
👍 to getting this in the next release with the Rust interop, it would be very helpful for some of the ideas I have for that :) happy to help try and revive this PR as well 😅

scripted-tests/run/build-library/src/main/c/testlib.cpp
```diff
@@ -0,0 +1,57 @@
...
+    }
+    assert(exceptionCaught);
+
+    exceptionCaught = false;
+    try {
+        fail();
+    } catch (const std::exception &e) {
+        exceptionCaught = true;
+    }
+    assert(exceptionCaught);
```
WojciechMazur, line 54: @adampauls On Linux, both of these variants seemed to work correctly, if it would still be failing (in the CI) I would try to investigate it more.

adampauls: Hmm, I'm unable to run the tests now, they fail for other reasons (as they do in CI). Should I assume you're fixing that up now, or should I take a look?

WojciechMazur: Don't worry about that, I have it fixed already. Would push the changes tomorrow in the morning

scripted-tests/run/build-library/build.sbt
```diff
@@ -15,46 +15,52 @@ scalaVersion := {
 
 nativeConfig := {
   val prev = nativeConfig.value
-  val outPath = crossTarget.value / "libtest.so"
+  val filename = if (Platform.isWindows) "test.dll" else "libtest.so"
```
adampauls, line 18: You probably already know this, but the name on macs is typically `.dylib`. Now that you're doing something platform-specific, maybe worth adding here too.

ekrich: Yeah, I am working on a PR to internalize those changes inside the tools API - too much of this stuff has leaked into the plugin.

adampauls:
Just a friendly check in on the status here?

ekrich:
I have done some preliminary review but haven't commented yet but am also hoping to help this and the Rust integration https://github.com/scala-native/scala-native/pull/2351 with this PR - https://github.com/scala-native/scala-native/pull/2605

armanbilge:
For those following along here: [Anton](https://github.com/keynmol/) asked on Discord what the next steps for this PR are. [Wojciech](https://github.com/WojciechMazur) responds:

> As already mentioned it might be also added to 0.4.6 if needed. The only blocker for that was lack of broader tests, especially in terms of GC correctness and ergonomics of the usage (initialization of the SN integration in foreign code). Unfortunetlly I've never had enough time to perform some decent tests in this area.  I would welcome any extended tests for that PR

https://discord.com/channels/632150470000902164/635668881951686686/993513586557202522

If anyone here is interested, sounds like a good opportunity for contribution! :)

ekrich:
This thread for Android - https://discord.com/channels/632150470000902164/635668881951686686/1017347650594545664
https://github.com/cnlohr/rawdrawandroid

rcano:
As I'm using this PR to integrate SN with godot, here are some ergonomics that I find lacking (as in, they force me to write C which IMO should be undesirable):
* Having to write C to invoke 1 line `ScalaNativeInit()`, when my shared library already has a method that should be the one to do it (I think an annotation in the same vein as `@exported` could be used to signal this)
* There's no way to add attributes to exported methods, if you must add them in order to comply with some loader, you have to do a C wrapper of the function. Attributes can be hard as they are platform dependent.

catap:
Thanks for amazing work!

Issue #2144, opened 2021-02-03
Clang linking failure with `.withOptimize(false)`


lolgab:
Trying out zio I had this linking error:
```
[error] /Users/lorenzo/zio-repro/target/scala-2.13/native/5.ll:25631:7: error: value doesn't match function result type 'i32'
[error]   ret i8* %_9
[error]       ^
[error] 1 error generated.
```

It only happens with `withOptimize(false)` in the settings.

I couldn't find a minimal reproduction.
You can find, however a reproduction using Zio here: https://github.com/lolgab/zio-repro

WojciechMazur:
The problem is in this lines of `ZScope.unsafeRelease`
``` scala
  java.util.Arrays.sort(
    array,
    (l: OrderedFinalizer, r: OrderedFinalizer) =>
      if (l eq null) -1 else if (r eq null) 1 else l.order - r.order
  )
```

We expect this SAM Comperator should return `Int` value, but it returns `Integer`. After calling `compare` method it boxes result for some reason.

lolgab:
@WojciechMazur You're right!
A smaller reproduction is:
```scala
java.util.Arrays.sort(Array.empty[Object], (l: Object, r: Object) => 0)
```

PR #2147, opened 2021-02-04
Fix: #2144, Lambda function generation should check if boxing is needed. 

WojciechMazur:
This PR fixes issue #2144. 
Up until now, Scala Native generation was always trying to box a result of the lambda method, in most cases, it was correct behavior. When defining `java.util.Comparator` using lambda generation was creating unnecessary boxing of `Int`, that would be discarded in the Interflow, however with disabled optimization, it would lead to LLVM compile-time error. 
With this fix, we apply boxing only when it's really needed.


nezosb:
As it turns out, this PR fixes #2316 as well. After investigating it myself, the problem there ended up being the same i.e. unnecessary boxings done in SAMs. Type mismatch causes unspecified behavior on returned values, resulting in the issue. 

I am leaving this here for posterity, and to link nicely from the aforementioned issue.

ekrich (approved):
From a logical standpoint the changes here look good.

sjrd (approved):

PR #2154, opened 2021-02-10
Allow testing against Scala partest and JUnit tests

WojciechMazur:
This PR adds new sbt projects allowing to test Scala Native against external tests defined in Scala repository. It adds support for JUnit tests and partests ported from Scala.js 
It defines two partest interface implementations: legacy `scala/scala-partest` used for tests in Scala 2.11.12, and latest version of this interface existing in `scala/scala` repository used for Scala 2.12.13 and 2.13.4 (it can be used as reference for backporting it back to Scala.js)

Defined blacklists are not fully exhaustive, they're mainly copy-pasted from Scala.js and not yet checked (I estimate ~30-40 CPU hours on my device to run all `run` tests for single Scala version). Probably we should provide some caching mechanism as it is done in Scala.js)
Also Scala.js uses in-memory file system library `jimfs`, we might add support for this in the future to speed-up build, however, it currently it would lead to exceptions combined with our `VirtualDirectory` logic

Remaining blocker issues:
- [x] Allow for cross-build of `tools` project to allow testing Scala 2.11 and 2.13


.github/workflows/run-tests.yml
```diff
@@ -40,6 +40,15 @@ jobs:
       - name: Test tools
         run: sbt "++ ${{ matrix.scala }} -v" "-no-colors" "-J-Xmx3G" test-tools
 
+        # Running all partests would take ~2h for each Scala version, run only single test of each kind
+        # to make sure that infrastructure works correctly.
+      - name: Run subset of partest tests
```
sjrd, line 45: We should also have a step to run the JUnit tests. I believe for those we can run all the tests.

docs/contrib/build.rst
```diff
@@ -204,6 +204,33 @@ The build has roughly five groups of sub-projects as follows:
 
     - ``(scripted-tests)`` (JVM project)
 
+6. External tests and its dependencies. Sources of these tests are not stored
+   in this project, but fetched from external sources, e.g.: Scala compiler repository.
+   Sources in this project define interface used by Scala Native and tests filters.
+
+    - ``scalaPartest (scala-partest)`` (JVM project, uses Scala Native artifacts)
+
+		- ``scalaPartestRuntime (scala-partest-runtime)`` (Scala native project)
```
sjrd, line 213: Tabs.

scripts/partest-check-files.sc
```diff
@@ -0,0 +1,77 @@
+import $ivy.`com.lihaoyi::ammonite-ops:2.3.8`, ammonite.ops._, mainargs._
+import java.io.File
+
+@main(
+  doc = """" +
+  "Tool used to check integrity of files defined in partest tests and 
+  actually defined in partest repo. Allow to check which blacklisted files
+  are not existing and can suggest correct blacklisted item name.
+  Also checks for duplicates in blacklisted items.""")
```
sjrd, line 9: This script should be executed in the CI, to make sure we don't commit corrupted lists.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -64,19 +63,15 @@ object Build {
...
-          .map { destPath =>
-            val paths = NativeLib.findNativePaths(workdir, destPath)
-            val (projPaths, projConfig) =
-              Filter.filterNativelib(fconfig, linked, destPath, paths)
-            LLVM.compile(projConfig, projPaths)
+        /* Used to pass alternative paths of compiled native (lib) sources,
+         * eg: reused native sources used in partests.
+         */
+        val libObjectPaths = scala.util.Properties
+          .propOrNone("scalanative.build.paths.libObj")
```
sjrd, line 70:
```suggestion
          .propOrNone("scalanative.build.paths.libobj")
```
?

tools/src/main/scala/scala/scalanative/build/Filter.scala
```diff
@@ -23,7 +23,7 @@ private[scalanative] object Filter {
    * @return The paths filtered to be included in the compile.
    */
   def filterNativelib(config: Config,
-                      linkerResult: linker.Result,
+                      linked: Seq[nir.Attr.Link],
```
sjrd, line 26: Spurious change.

tools/src/main/scala/scala/scalanative/build/Filter.scala
```diff
@@ -44,7 +44,7 @@ private[scalanative] object Filter {
       def include(path: String) = {
         if (path.contains(optPath)) {
           val name = Paths.get(path).toFile.getName.split("\\.").head
-          linkerResult.links.map(_.name).contains(name)
+          linked.map(_.name).contains(name)
```
sjrd, line 47: Spurious change.

sjrd (changes requested):
Mostly comments about the CI.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -54,7 +54,6 @@ object Build {
     config.logger.time("Total") {
```
ekrich, line 54: I am not super excited about this because we are changing the build for such a specific purpose unless there are other specific use cases in mind.

Refactoring out the "findAndCompileNativeSources" is good but perhaps instead of using `fold` for the special case you could have `match` / `case` with `None` being first as the normal case and then hiding the body of the `fold` in another method.

I think this would be more readable and show intent better. Also, I think that the `// compile generated ll` should go first in that block before the "findAndCompileNativeSources". It logically follows closer to the code just above.

WDYT?

.github/workflows/run-tests.yml
WojciechMazur: JUnit tests were added to `test-runtime` alias used in the CI. Should I move them to some other place?

sjrd: Ah no, that's fine.

sjrd (approved):

tools/src/main/scala/scala/scalanative/build/Build.scala
WojciechMazur: You're right, I've replaced it with a proper match.
 I think this change might be useful for us in other aspects, eg. we can try to add some elements of incremental compilation to our build, for now, it can be just native sources, not much, but it still might save a few seconds.

PR #2156, opened 2021-02-12
Allow for cross building tools

WojciechMazur:
This PR adopts our codebase to be compatible with Scala 2.13 and cross builts most of the remaining projects with 2 exclusions:
 - `nirParser` which is not published or used by any other of our components
 - `sbtScalaNative` is sbt plugin project
 
Currently, tools are only used within sbt plugin, which can only be published in Scala 2.12 and therefore not cross-built.  
Incoming support for Scala partests (#2154) needs `tools` to run tests against Scala 2.11 and 2.13.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -2,4 +2,3 @@
 -J-Xms1024M
 # Needed for Scala 2.11 tools compilation. Failing in CI for test-scripted
 -J-Xss2m
-
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,2 +1,5 @@
 -J-Xmx5G
 -J-Xms1024M
+# Needed for Scala 2.11 tools compilation. Failing in CI for test-scripted
+-J-Xss2m
+
--- a/build.sbt
+++ b/build.sbt
@@ -79,14 +79,6 @@ lazy val mimaSettings: Seq[Setting[_]] = Seq(
   }
 )
 
-// Sbt 1.4.0 introduced mandatory key linting.
-// The val crossSbtVersions is used in toolsSettings below.
-// toolsSettings are used by nir, test-runner, util, etc.
-// but Sbt 1.4.0 still complained about crossSbtVersions.
-// Disable the linting for it, rather than upsetting the apple cart by
-// deleting the probably essential crossSbeVersions. Minimal change.
-Global / excludeLintKeys += crossSbtVersions
-
 // Common start but individual sub-projects may add or remove scalacOptions.
 // project/build.sbt uses a less stringent set to bootstrap.
 inThisBuild(
@@ -242,7 +234,6 @@ lazy val noPublishSettings: Seq[Setting[_]] = Seq(
 lazy val toolSettings: Seq[Setting[_]] =
   Def.settings(
     sbtVersion := sbt10Version,
-    crossSbtVersions := List(sbt10Version),
     crossScalaVersions := Seq(sbt10ScalaVersion),
     javacOptions ++= Seq("-encoding", "utf8")
   )
@@ -363,7 +354,6 @@ lazy val nscplugin =
     .in(file("nscplugin"))
     .settings(mavenPublishSettings)
     .settings(
-      crossScalaVersions := libCrossScalaVersions,
       crossVersion := CrossVersion.full,
       Compile / unmanagedSourceDirectories ++= Seq(
         (nir / Compile / scalaSource).value,
--- a/build.sbt
+++ b/build.sbt
@@ -415,7 +415,7 @@ lazy val sbtScalaNative =
           .value
       }
     )
-    .dependsOn(testRunner, tools)
+    .dependsOn(tools, testRunner)
 
 lazy val nativelib =
   project
--- a/build.sbt
+++ b/build.sbt
@@ -413,16 +413,9 @@ lazy val sbtScalaNative =
             testRunner / publishLocal
           )
           .value
-      },
-      // Depending on projects defining `crossScalaVersion` other then this `scalaVersion` would result in build failure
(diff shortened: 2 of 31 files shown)


ekrich (changes requested):
https://github.com/scala-native/scala-native/pull/2156/files/563ce78502398ea4990c847e276a595d44e77a74#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R305

The spelling of fastparse  needs correcting in a comment. There is a WIP PR for this upgrade but the author got stuck. https://github.com/scala-native/scala-native/pull/1789

sjrd:
Have you considered removing nirparser altogether?

ekrich:
Do we have a way to view the text version or NIR without it? If so, that is a really good idea.

WojciechMazur:
> Do we have a way to view the text version or NIR without it? If so, that is a really good idea.

NirParser is not used to view text. It's more like a textual deserializer. We can always view NIR as text (eg. when nativeDump is enabled) using `NirShowBuilder` / `Show`. I guess that some long day ago it was convenient to store NIR in textual format, but now its mainly just a legacy code.

.sbtopts
```diff
@@ -1,2 +1,5 @@
 -J-Xmx5G
 -J-Xms1024M
+# Needed for Scala 2.11 tools compilation. Failing in CI for test-scripted
+-J-Xss2m
+
```
sjrd, line 5: Spurious blank line.

build.sbt
```diff
@@ -400,9 +413,16 @@ lazy val sbtScalaNative =
...
           .value
-      }
+      },
+      // Depending on projects defining `crossScalaVersion` other then this `scalaVersion` would result in build failure
+      unmanagedSourceDirectories in Compile ++= Seq(
+        (testRunner / Compile / scalaSource).value,
+        (tools / Compile / scalaSource).value,
+        (nir / Compile / scalaSource).value,
+        (util / Compile / scalaSource).value,
+        baseDirectory.value.getParentFile / "test-interface-common/src/main/scala"
```
sjrd, line 423: We must not do this. That would mean that the published artifacts have two copies of those things.

We should keep the regular `dependsOn` dependency on the other projects. Instead, we should avoid running any task in `sbtScalaNative/*` when we have done `++2.11.*` or `++2.13.*`.

sjrd (changes requested):
Aren't we supposed to change something in the CI scripts somewhere, so that we actually test this?

build.sbt
WojciechMazur: Yes, but then we would not be able to run `sbtScalaNative/scripted` for 2.11 and 2.13.
It would end up with something like this when trying to compile 
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:scala-native/"), "sbtScalaNative"):
[error]    org.scala-native:test-runner _2.12, _2.11
[error]    org.scala-native:nir _2.12, _2.11
[error]    org.scala-native:tools _2.12, _2.11
```

It might be a bug or design flow in sbt. 
How about duplicating these 3 projects, eg. making project `pluginTools` using the same directory as `tools` but with other configurations? May it actually work? In Scala.js thanks to duplication of projects there is no such issue because you explicitly command to depend on `*v2_12` project.

WojciechMazur:
> Aren't we supposed to change something in the CI scripts somewhere, so that we actually test this?

Actually, we don't have to. Until now every time we run `test-tools` in most underlying tests we used Scala 2.12, because it was only defined cross version. When running `++ <version>` we are not forcing using that version, so it was skipped.
With set cross versions we are actually running these tests with specified versions.

build.sbt
sjrd: > Yes, but then we would not be able to run `sbtScalaNative/scripted` for 2.11 and 2.13.

Right. I think we will have to live with that. Anyway, it doesn't really make sense; not in the way `++` is supposed to be used. If I do `; ++ 2.13.4 ; sbtScalaNative/scripted`, I would *expect* that sbt-scala-native itself be compiled with 2.13.4, and then we run `scripted` on that. Of course that doesn't work because an sbt plugin cannot build on 2.13.4. It seems that so far we've abused `++` and `scripted` to mean "compile the sbt plugin with 2.12 anyway, but inside the scripted tests, use projects with version 2.13.4".

> How about duplicating these 3 projects, eg. making project `pluginTools` using the same directory as `tools` but with other configurations? May it actually work? In Scala.js thanks to duplication of projects there is no such issue because you explicitly command to depend on `*v2_12` project.

Eventually, we'll probably have to decompose into `v2_*` like in Scala.js. But that has a significant cost on build load time. The ability to run scripted tests for 2.11 and 2.13 is not worth that additional time. In Scala.js we had to do it to be able to remove the `linker` from the *static* dependencies of `sbt-scalajs`, and instead dynamically class-load it. But we're not there yet for Scala Native.

WojciechMazur: Actually, I've found a way around that would satisfy both using `dependsOn` and remaining scriptedTests for 2.11 and 2.13. 
I've extracted scripted tests to separate project with enabled `ScriptedPlugin` and locked scalaVersion to 2.12. 
This allows publishing artifacts based on `scalaVersion` defined in `ThisBuild`. The remaining problem was publishing plugin artifact itself - it is done in alias `test-scripted` - in first command we run `+ sbtScalaNative/publishLocal` - this way it would publish for all (so 2.12 only) plugin versions. 

I needed to move scripted sources to the inner directory to make sure that they don't interfere with `target` which would lead to failure, so I may add this change in follow-up PR.

sjrd (approved):

PR #2158, opened 2021-02-15
Remove legacy textual NIR parser

WojciechMazur:
This PR removes legacy project `nirParser` and its sources as it's no longer relevant for Scala Native workflow. 

In the past, NIR could have been represented in both binary and textual format, however now it's always serialized as a binary. 
This project is not published or used anywhere, also since it was written using Fastparse 1.x.x it's more difficult to maintain and eventually would need to be rewritten to be used with Fastparse 2.x.x. 
The only usage of this module was in `test-tools` where `nirParser` self-tests were run. 
This change does not include any additional regression. It's still possible to create NIR dumps in textual form - this logic contained in `scalanative.nir.Show` module.


sjrd (changes requested):
You should also remove the reference to `nirparser` in https://github.com/scala-native/scala-native/blob/master/docs/contrib/build.rst
Tip: `git grep nirparser` is useful here, to make sure that no stray reference exists anywhere.

Otherwise LGTM.

sjrd (approved):

PR #2161, opened 2021-02-17
Scalanative now uses a scala object rather than an extern object for GC

markehammons:
This PR is the first in a set I'm wanting to do to port the GC code from C to scala. That is, my plan is to eventually have immix, commix, and none implemented as scalanative code instead of C. 

Likewise, I would like to change the design of the GC so that nativelib depends on a provided GC lib so that users can switch GC through a `libraryDependencies +=` rather than a special GC flag that in scala native.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ bin/
 /.bloop/
 /.metals/
 /project/**/metals.sbt
+/project/**/.bloop
 
 # Build Server Protocol, used by sbt
 /.bsp/
--- a/nativelib/src/main/scala/scala/scalanative/runtime/GC.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/GC.scala
@@ -2,6 +2,7 @@ package scala.scalanative
 package runtime
 
 import scalanative.unsafe._
+import scalanative.unsigned._
 
 /**
  * The Boehm GC conservative garbage collector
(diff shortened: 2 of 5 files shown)


sjrd:
That sounds like a very ambitious goal. If the GC is itself a Scala object, it would have to somehow deal with *itself* and its dependencies in a very very special way. That sounds like a recipe for disaster.

ekrich:
Looks like you need to run `scripts/scalafmt` on the code.

markehammons:
> That sounds like a very ambitious goal. If the GC is itself a Scala object, it would have to somehow deal with _itself_ and its dependencies in a very very special way. That sounds like a recipe for disaster.

It's difficult, but it can be done without the GC doing allocations. Please look at [this branch](https://github.com/markehammons/scala-native/tree/GCNone_demo), which recreates the `None` gc in pure scalanative.

sjrd:
That still sounds like a lot of effort, and resulting in pretty fragile code.

What is the point, here? What actual benefit is this supposed to provide?

markehammons:
> That still sounds like a lot of effort, and resulting in pretty fragile code.
> 
> What is the point, here? What actual benefit is this supposed to provide?

Dogfooding for one. Also, I don't believe the current implementation is any more fragile than the c implementation, and I believe I can build tools that can aid in writing things.

ekrich:
@sjrd Do you think if we supported non-GC stack allocated objects or structs that this approach might become viable? This feature would be really useful for interopt for the C complex routines and other library routines that pass structs in and out.

Also, maybe if we had a different way to package/integrate GCs then it might allow easier experimentation. This could be good, especially if it could be done with no performance or integration downsides.

WojciechMazur:
The separation of managed (Scala) code and unmanaged GC implementation dramatically simplifies the latter. At the current state of Scala Native we shall not try to change it as it would not bring big enough improvement, until we would be able to allocate classes and their inner fields without the need of GC.

PR #2162, opened 2021-02-17
Fix reporting undefined symbols not marked with stub

WojciechMazur:
Recently it could have been possible to observe multiple runtime exceptions that should actually fail when linking. 
These problems could have been spotted when enabling `nativeCheck`, although they would be reported as warnings. Additionally by default checks are disabled, so users in multiple cases were unaware of an existing problem. 

An example of the mentioned problem can be using Java standard library methods not defined in the `javalib` and not marked with `@stub` annotation, eg. `String::repeat`. In such case, the method targets list was always empty, but there was no special handling for such cases. 

This PR adds additional handling to `Reach` adding an additional list with a method whose targets should be checked again at the very end of the `Reach` pipeline. This way we can be sure that all needed classes were already loaded. If the method still does not contain any targets we can be sure that the method should be treated as a missing symbol and reported accordingly. 

Since missing symbols found using this approach were never contained in the scala-native code (they may come from dependencies not compiled with SN) they would always be leading to breaking the build, even if `linkStubs` is enabled. The only way to mitigate this problem is to explicitly define the method stub by the user.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/linker/Reach.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/Reach.scala
@@ -23,6 +23,9 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
   val dynsigs       = mutable.Set.empty[Sig]
   val dynimpls      = mutable.Set.empty[Global]
 
+  private case class DeleyedMethod(owner: Global.Top, sig: Sig, pos: Position)
+  private val delayedMethods = mutable.Set.empty[DeleyedMethod]
+
   entries.foreach(reachEntry)
   loader.classesWithEntryPoints.foreach(reachClinit)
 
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -23,6 +23,9 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
   val dynsigs       = mutable.Set.empty[Sig]
   val dynimpls      = mutable.Set.empty[Global]
 
+  private case class DeleyedMethod(owner: Global.Top, sig: Sig, pos: Position)
```
ekrich, line 26: This, and associated use sites.
```suggestion
  private case class DelayedMethod(owner: Global.Top, sig: Sig, pos: Position)
```

tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -99,6 +102,28 @@ class Reach(config: build.Config, entries: Seq[Global], loader: ClassLoader) {
...
+          }
+          val targets = info.targets(sig)
+          if (targets.isEmpty && wasAllocated) {
+            addMissing(top.member(sig), position)
+          } else {
+            todo ++= targets
+          }
+        }
+    }
+    process()
```
sjrd, line 124: That second `process()` could cause new stuff to be reached, hence added to `delayedMethods`, and those would still not be treated.

It looks like we need either
* a loop, doing `process()` and `processDelayed()` until there is nothing left, or
* when finding a new method, immediately check whether it is in the `delayedMethods` set, and extract it to put in the todo

sjrd (changes requested):

sjrd (approved):

PR #2167, opened 2021-02-21
move process_monitor.cpp to javalib

markehammons:
UnixProcess lives in javalib as per #1352. process_monitor.cpp, which only serves UnixProcess, was not moved for some reason. This PR puts process_monitor.cpp where it belongs, as well as updating process_monitor.cpp to use the C++ stdlib's thread, mutex, and conditional_variable libs, making it cross platform to windows.


ekrich:
@markehammons I did the bulk of the moving C code around after the capability to move was in place. The C/C++ code belongs in a library but is not enforced other than by convention. It should be with the matching Scala file and that portion is enforced via the build.

The motivation is to have `posixlib` optional so we could support Windows for example. See https://github.com/scala-native/scala-native/pull/1885 for a discussion.

Since this is Unix process, we should have POSIX and `pthread` available so I am not sure it make since to bring in C++ threads especially since we have a `pthread` version of `java.lang.Thread`s in a decent state and for the most part. I am not sure there has been a big migration to C++ threads from pthreads. Also, the GCs use `pthreads`.

I think it makes sense to move it to `javalib` as the C interface class `UnixProcess` and `ProcessMonitor` is in the `javalib`. I am not sure why I didn't move it unless it is tied into `nativelib` in someway but it could have been an oversight.

@sjrd From a build perspective, would it be ok to just move `clib` and `posixlib` tests directly into the `test` directories of those projects?

sjrd:
> From a build perspective, would it be ok to just move `clib` and `posixlib` tests directly into the `test` directories of those projects?

No, that is going to be very difficult, because it will create a circular dependency between projects (though not between configurations) and not every tool is prepared to deal with that. We had that kind of thing in Scala.js very early on, and it caused several issues (although I don't remember exactly what; it was 7 years ago).

ekrich:
>No, that is going to be very difficult, because it will create a circular dependency between projects (though not between configurations) and not every tool is prepared to deal with that. We had that kind of thing in Scala.js very early on, and it caused several issues (although I don't remember exactly what; it was 7 years ago).

I thought there might be a problem since I didn't see that sort of thing in Scala.js. We can make them separate projects then.

markehammons:
> Since this is Unix process, we should have POSIX and pthread available so I am not sure it make since to bring in C++ threads especially since we have a pthread version of java.lang.Threads in a decent state and for the most part. I am not sure there has been a big migration to C++ threads from pthreads. Also, the GCs use pthreads.

I did research on <thread> and the c equivalent <threads.h> in the lead up to this PR, and it seems that C++ stdlib threads are seeing significant adoption in new code. I made the decision to transition this code to the stdlib threads because they are cross platform (the only thing blocking process monitor from being used for WinProcess and UnixProcess is the signal and wait code). You mentioned the GCs use posix threads, and I’d suggest we make the same changes there to make the GCs more easily adapted to a Windows environment.

Another reason to move to the c++ standard threads is that the code is more explicit wrt time (time spec was being used as a time point rather than a duration, which was not at all obvious from scanning through the code) as well as handling mutexes more safely (unique lock locks the mutex and unlocks it upon unique lock’s destruction, as opposed to the manual unlocking we were doing with posixlib).

Considering these libraries are readily available in clang 6 and above, I think it’s good for us to use them in order to improve the cross-platform capabilities of our C++ code such as the GCs.

markehammons:
@ekrich One option is to have the process manager code be as it currently is, and have the result tracking code in posixlib (and winlib when it exists) as a function like wait result or something.

markehammons:
@ekrich looking at how things are done in windows, I think only minor modifications to the wait_loop in process_manager would be needed to support windows. This could be handled with having a function waitresult that is defined for windows and unix, and returns the resultcode of the latest exited subprocess. What do you think?

ekrich:
@markehammons If you have some ideas, I think the best way is to have a proposal issue with steps to accomplish with one PR for each small step so it is manageable. The hardest part is to get consensus and then you need small PRs that are easy enough to review. The Scala Center folks are the project leaders and final arbitrators but it is a community based project so we try and balance the needs of the overall community.

@LeeTibbert and I have been around awhile and we try and help out and guide the best we can. The threads thing is tough because real time platforms probably use real time `pthreads` so we should probably stick with them for the Unix/Linux type platforms. We also have lots of working code, and the `java.lang.Thread` in the pipeline.

We are also trying to support 32bit led by @shadaj . Boehm supports 32bit, Windows, and tons of platforms so that is our first line on the cross platform GC front. The other could be ported once alternative platforms are working.

This is what we use for the C conformance: https://en.cppreference.com/w/c/header
This is what we use for POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/
Click on number 13, headers. Basically, I look as POSIX as an extension to Standard C. Where macros and functions are in both places they should be in the C lib.

If you would like to work on the windows port, we believe we can make small changes to not affect the mainline Unix type platform and allow progress towards a native Windows experimental type platform. @WojciechMazur did some work on separating the code but I never got the final word on that effort. Is this something that can be pushed for review, study and discussion?

I hope this helps explain things a bit. I have a few ideas and maybe others do as well.

WojciechMazur (changes requested):
Good idea about switching to c++ threads! Sorry that you needed to wait. I only see few details that can be changed.

PR #2168, opened 2021-02-22
Restore `scala.collection` optimization overrides as patch files

WojciechMazur:
Thir PR restores `scala.collection` optimizations that were removed in #1962, however now instead of storing the whole Scala file, we're only storing a diff that would be applied when compiling. 
With this approach, it would be easier to know which parts of the Scala standard library were overridden and maintain it when upstream changes occur. 
Some of the other override files that were existing were also replaced with overrides. 

Additionally, it adds an Ammonite script that can be used to created diffs, recreates overrides bases on diffs, and clean the overrides directory.


ekrich (commented):
Does this library need to loaded before the build starts?

WojciechMazur:
> Does this library need to loaded before the build starts?
> 
> Not sure the review worked correctly, this one.
> `libraryDependencies += "io.github.java-diff-utils" % "java-diff-utils" % "4.9"`

We could use `patch` binary instead, although it might be more problematic in terms of portability, eg. in someone would want to build SN on Windows. In fact, I was using this approach when prototyping. 
Also in #2154 we also add an additional library (`jGit`) used to fetch Scala sources at a specified tag. I tried to found some commons between them, but I've failed. (maybe I've missed something, git patch works a bit differently I guess)

ekrich:
I was really just asking if the dependency needed to be in `project/build.sbt` or just in `build.sbt`?

scalalib/overrides-2.12/scala/Array.scala
```diff
@@ -0,0 +1,717 @@
...
+ *  val b = Array.ofDim[Int](2)
+ *  val c = Array.concat(a, b)
+ *  }}}
+ *  where the array objects `a`, `b` and `c` have respectively the values
+ *  `Array(1, 2)`, `Array(0, 0)` and `Array(1, 2, 0, 0)`.
+ *
+ *  @author Martin Odersky
+ *  @since  1.0
+ */
+object Array extends FallbackArrayBuilding {
```
sjrd, line 54: Looks like this file should not exist.

build.sbt
```diff
@@ -614,18 +618,47 @@ lazy val scalalib =
...
+              sources += source
             else
-              s.log.debug(s"not including $src")
+              s.log.debug(s"not including $path")
+          }
+
+          if (patchGlob.matches(sourcePath)) {
+            val sourceName      = path.stripSuffix(".patch")
+            val scalaSourcePath = scalaSrcDir / sourceName
+            val outputFile      = scalaSrcDir / (sourceName + ".patched")
```
sjrd, line 637: The output file should be somewhere in a `target/` directory, since it will contain generated files. Additionally, that way it can have the extension `.scala` instead of `.patched`.

scalalib/overrides-2.12/scala/Array.scala
WojciechMazur: Actually `Array` in 2.12 was missing, it exists in both 2.11 and 2.13. Probably was removed by mistake in the past. 
It also contains optimizations existing in 2.11 and 2.13. 

However, maybe we should store them as diffs as well? Patch for Array takes ~250 lines where the original file has ~720 lines. What do you think? Same about other not included files like `Predef`, `Symbol`, `package.scala`

WojciechMazur:
> I'm surprised to see many overrides for 2.11 and 2.12, but none for 2.13. Were they forgotten?

At this stage, I wanted to restore previously existing overrides, they were mainly changing old Scala Collections, eg. `ReusableCBF` class, it does not exist in 2.13, that why it was not included. However, we might try to find other possible optimizations and include them in the following PRs

> Also, I was hoping that, even when the sources are different in 2.11/2.12/2.13 are different, often the `.patch` files would end up being the same, and hence we would only have to commit them once under `overrides/`. Is that not the case?

It should be possible, I may try to check if it's possible, probably with more lines of context in diff

WojciechMazur:
@sjrd Follow up on the question about reusing overrides. We can do this for most of the diffs in collections (only 3 raises conflicts), but we would need to use the `patch` program instead of Java library - in all of them I've tried they fail to apply the patch, probably they're not checking context from unified diff. When using GNU `patch` there is no problems. 

We can switch to this method, although build Scala Native on non-Unix platforms, eg. Windows might be problematic I guess.

ekrich:
Can a library like `maven diff-match-patch` work?

WojciechMazur:
> Can a library like `maven diff-match-patch` work?

Yes, it can work, when we create and apply patches with it, however, it does not create unified diffs, but makes tokenized diffs instead. They're good in terms of storage needed, but they're not good in a human-readable format (no context, only characters that changed)
I've got changes ready to use it instead if we decide to use it anyway. 

Below an example of how diff created using diff-match-patch typically looks like (I've applied `diffCleanupSemantic` it to make it more human-readable as its documentation suggested)
```
@@ -1066,16 +1066,247 @@
 eq%5D %7B%0A  
+@inline override def ReusableCBF: GenericCanBuildFrom%5BNothing%5D = ReusableCBFInstance%0A  private object ReusableCBFInstance extends GenericCanBuildFrom%5BNothing%5D %7B%0A    @inline override def apply() = newBuilder%5BNothing%5D%0A  %7D%0A%0A  @inline 
 implicit
@@ -1364,16 +1364,20 @@
 eq%5BA%5D%5D =
+%0A   
  Reusabl
@@ -1418,18 +1418,27 @@
 rom%5BA%5D%5D%0A
+%0A
   
+@inline 
 def newB
@@ -1474,16 +1474,20 @@
 eq%5BA%5D%5D =
+%0A   
  new mut

```

sjrd:
> @sjrd Follow up on the question about reusing overrides. We can do this for most of the diffs in collections (only 3 raises conflicts), but we would need to use the `patch` program instead of Java library - in all of them I've tried they fail to apply the patch, probably they're not checking context from unified diff. When using GNU `patch` there is no problems.
> 
> We can switch to this method, although build Scala Native on non-Unix platforms, eg. Windows might be problematic I guess.

That is unfortunate. :( And it's very surprising that there does not exist a Java library to do this correctly. What libraries have you tried?

WojciechMazur:
@sjrd I've used java-diff-utils. Although now, I think it might need another try - I've tried to apply patches created using GNU `diff`. Maybe there is some slight difference in the implementation of these two resulting in failures. Although that should not be the case.

WojciechMazur:
I've replaced line-based GNU based diffs, with character-based diff-match-patch diffs. Although it makes it less readable, it would allow us to use on Windows in future.

sjrd (changes requested):
I think the only thing that's missing here is to document in https://github.com/scala-native/scala-native/blob/master/docs/contrib/build.rst (or somewhere in that area) how to work with the override files. I.e., what's the workflow of `create`/`recreate`/`edit`/etc. to work effectively with those?

PR #2170, opened 2021-02-23
Support Scala 2.13.5

WojciechMazur:
This PR adds support for Scala 2.13.5 and this version to cross-build and CI tests matrix. 

Scala 2.13.5 includes changed name for one `typer` methods that we're using -  `checkClassType`  was renamed intro `checkClassTypeOrModule`. Special handling for this was added to `NirCompat`.
Additionally, definitions of `NirGenPhase` and `PrepNativeInterop` were changed to allow usage of `NirCompat` in both of them. 
This also allowed us to define `nirAddons` only once.


WojciechMazur:
To be noted - GitHub actions do not run tests when the workflow file was changed. I've run `tests-all` locally and they passed everything in 2.13.5. I'm waiting for 2.13.4 results but until now it runs without problems.

sjrd (approved):

PR #2175, opened 2021-02-24
Implement MD5 and SHA-* digests in java.security.MessageDigest

lolgab:


sjrd:
Hum ... looks like this class should be taken out of the core, and implemented in a library that focuses on the crypto digests.

lolgab:
Does so little code deserve a library on its own?
Consider that `MessageDigest` should implement, [according to the javadoc](https://docs.oracle.com/javase/7/docs/api/java/security/MessageDigest.html), only MD5, SHA-1 and SHA-256.
I saw `java.util.zip._` implemented on top of `zlib` on the same fashion, does it apply to it too?

ekrich:
Does Harmony use a native library as well?

sjrd:
The problem is not so much the dependency on native code, but rather that we can't really implement this class correctly, because of the SPI stuff that is built inside it. So if we can get away without having it in the core, it would be better. I wonder why it was even added in the core with this dummy implementation? 🤔

lolgab:
> Does Harmony use a native library as well?

I can't find the actual code. It can be that the Harmony implementation is a dummy one too.

> The problem is not so much the dependency on native code, but rather that we can't really implement this class correctly, because of the SPI stuff that is built inside it.

I don't know what SPI is and why it doesn't allow us to have a correct implementation, I will check it out.
Anyway, I'll try to implement the other two algorithms and see if I can achieve feature parity with the original JVM. With MD5 it seems to behave correctly, it consumes constant memory and doesn't leak any memory (I'm using an array allocated by the GC to store the native struct).

catap:
Honestly I don't think that make dependency from OpenSSL is such good idea.

javalib/src/main/scala/java/security/MessageDigest.scala
```diff
@@ -27,3 +40,44 @@ private class DummyMessageDigest(algorithm: String)
...
+  def MD5_Update(c: Ptr[Byte], data: Ptr[Byte], len: CSize): CInt = extern
+  def MD5_Final(md: CString, c: Ptr[Byte]): CInt                  = extern
+}
+
+private final class Md5MessageDigest(algorithm: String)
+    extends MessageDigest(algorithm) {
+  import crypto._
+
+  // Array with length equals to sizeof(MD5_CTX)
+  private val c = new Array[Byte](92).asInstanceOf[ByteArray].at(0)
```
catap, line 57: Are you sure that all `libssl` implementation has this size?

lolgab:
@catap The dependency is optional as the `zlib` dependency is. If you don't use `MessageDigest` it will not be linked in the binary and you'll not depend on it.

catap:
@lolgab when if it doesn't leak it seems as easy and quick win.

Because implement hash is a week and make it fast enough it is a lot of work, and idea reuse libssl/libcrypto is very excites here :)

javalib/src/main/scala/java/security/MessageDigest.scala
lolgab: I have to check, I'm not sure.
I first tried to fetch it with a C file but that creates a dependency to Openssl (all the files in `/resources/scala-native` regardless if you use them or not), so I checked this number on my Mac OS machine with OpenSSL.

catap:
@lolgab do not forget about `clone`.

lolgab:
This was ported to a standalone library: https://github.com/lolgab/scala-native-crypto

ekrich:
cc/ @catap

PR #2159, opened 2021-02-16
Normalize names in clib and posixlib

ekrich:
The have been previous discussions about naming and this PR brings the code in line with the styles and standards of those past discussion. Documentation has been added.

Hopefully, much of this glue code will become a thing of the past when C interopt improves. None of the code changes should change the external APIs.

PR #2182, opened 2021-02-26
Introduced NativeThread

catap:
This commit introduced `NativeThread` which is a part of #947

This is a step to split https://github.com/catap/scala-native/tree/multithreading

### DIFF
--- a/posixlib/src/main/resources/scala-native/nativethread.c
+++ b/posixlib/src/main/resources/scala-native/nativethread.c
@@ -1,0 +1,66 @@
+#include <sys/types.h>
+#include <stdint.h>
+#include <pthread.h>
+#include <sched.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+pthread_attr_t PTHREAD_DEFAULT_ATTR;
+struct sched_param PTHREAD_DEFAULT_SCHED_PARAM;
+int PTHREAD_DEFAULT_POLICY;
+size_t PTHREAD_DEFAULT_STACK_SIZE;
+int initialized = 0;
+
+void init() {
+    pthread_attr_init(&PTHREAD_DEFAULT_ATTR);
+    pthread_attr_getschedparam(&PTHREAD_DEFAULT_ATTR,
+                               &PTHREAD_DEFAULT_SCHED_PARAM);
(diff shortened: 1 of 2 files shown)


catap:
@sjrd / @WojciechMazur can I ask you to review it?

densh (approved):
LGTM!

WojciechMazur:
Can't we implement this using Scala Native? We already have bindings for `pthread` and `sched`. All structures are passed by reference so there should be no problem with this.

posixlib/src/main/resources/scala-native/nativethread.c
```diff
@@ -0,0 +1,66 @@
...
+    pthread_attr_getschedparam(&PTHREAD_DEFAULT_ATTR,
+                               &PTHREAD_DEFAULT_SCHED_PARAM);
+    pthread_attr_getschedpolicy(&PTHREAD_DEFAULT_ATTR, &PTHREAD_DEFAULT_POLICY);
+    pthread_attr_getstacksize(&PTHREAD_DEFAULT_ATTR,
+                              &PTHREAD_DEFAULT_STACK_SIZE);
+
+    initialized = 1;
+}
+
+int get_max_priority() {
```
ekrich, line 25: For helper methods. the standard is to prefix the function or define with `scalanative_`. Unfortunately, this was just documented and has not made it in the latest docs yet.
See commit: https://github.com/scala-native/scala-native/pull/2159/commits/4f5859a64e2d5dbcd9b5ade5ab2fd6525552c24a

ekrich (changes requested):
Everything else looks good but I am not a `pthread` expert so I can't comment on how it works.

ekrich:
@catap or @densh, now that I am looking at this can't these additional functions and helpers just be added to `posix/pthread.scala` and `pthread.c`?

posixlib/src/main/resources/scala-native/nativethread.c
ekrich: @WojciechMazur Is it easy enough to publish the latest? I am wondering if a link somewhere pointing to the latest might be good for contributors and people that are curious?

WojciechMazur:
Thank for your contribution, however we have end up with a different implementation of native thread in current main branch

catap:
np, thanks!

Issue #2127, opened 2021-01-23
Enhance clang discovery, compile, and linking options


ekrich:
In a previous PR we enhanced the clang discovery code: https://github.com/scala-native/scala-native/pull/2099

This was an improvement but it has been discovered that the method could be overly flexible allowing different competing tools in the tool chain. We currently have `CLANG_PATH` and `CLANGPP_PATH` which means that we could have 2 different versions of `clang` and `clang++`. Do we need to support multiple paths or could we just have one path?

It was also determined that `llvm-config` which is called for getting compile and linking options can be named with the version number like `llvm-config-8` which means that our discovery code is too simplistic. I believe we should use the platform options and not use hardcoded paths and options but this is up for debate. A twist here is that `macOS` Xcode doesn't have `llvm-config` so if we want to support the platform installed tools rather that `brew` installed we would need to find a solution.

Not relying on `llvm-config` means that we may miss or need to add different paths based on the distributions installed locations or different package managers. This was manifested in this PR that added support for `macports`: https://github.com/scala-native/scala-native/pull/2071 It has not been verified that if `llvm` was also installed from `macports` whether or not this PR would have been needed if the `llvm-config` usage was working. Note that we had to add a patch for other platforms. https://github.com/scala-native/scala-native/pull/2091

Comments and experience welcomed.

ekrich:
Also see Issue #666

ekrich:
Right now, our llvm-config handling doesn't work for compile and link options.
I would like to unify `clang`, `clang++` and `llvm-config` under one call but our Discover doesn't hold any state.

They are all called independently and you can't really count on the order - It would be good to unify detecting `clang` on the path and the other binaries in one shot or via a `LLVM_BIN` set by the user.

If we fix this it should help with a windows port too.

ekrich:
I noticed when looking at this more carefully that using `llvm-config` only adds the LLVM includes and libs to the compile and link options respectively. Even if this works, I don't think it is being used or what was the original intent of the feature unless there was a plan to use the LLVM infrastructure.

PR #2189, opened 2021-03-03
Fix sbt plugin incompatibilities on Windows

WojciechMazur:
This PR fixes fatal errors in the sbt plugin when trying to use it on Windows. Changes include: 
* detection of os used in sbt plugin
* using `where` command instead of `which` 
* passing paths in process commands in sequences, to correctly handle spaces
* fixes to regex and globing filters including usage of the correct path separator

Some methods with single usage were inlined. 

There are still some pending sbt plugin related issues that would need to resolve in the following PRs, e.g.:
* Handling `llvm-config` dependent settings - `llvm-config` is not included in prebuilt binaries targeted for Windows. We don't want to force users to build it locally. Probably for `include` and `lib` dir we can use fixed relative paths based on the location of `clang` or `clang++`.  I have such change prepared, however I decided to not include it now, because of next problem
* Adding proper handling and detection of `include` and `lib` directories, however might finally decide to leave this part for users to configure.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/Discover.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Discover.scala
@@ -84,15 +84,15 @@ object Discover {
    */
   private[scalanative] def checkClangVersion(pathToClangBinary: Path): Unit = {
     def versionMajorFull(clang: String): (Int, String) = {
-      val versionCommand = s"$clang --version"
-      // Pass command as seq of string to handle spaces in path
-      val versionString = Process(Seq(clang, "--version"))
+      val versionCommand = Seq(clang, "--version")
+      val versionString = Process(versionCommand)
         .lineStream_!(silentLogger())
         .headOption
         .getOrElse {
           throw new BuildException(
-            s"""Problem running '$versionCommand'. Please check clang setup.
-                 |Refer to ($docSetup)""".stripMargin)
+            s"""Problem running '${versionCommand
+                 .mkString(" ")}'. Please check clang setup.
+               |Refer to ($docSetup)""".stripMargin)
         }
       // Apple macOS clang is different vs brew installed or Linux
       // Apple LLVM version 10.0.1 (clang-1001.0.46.4)
@@ -136,7 +136,7 @@ object Discover {
                                     envPath: String): Path = {
     val binaryNameOrPath = sys.env.getOrElse(envPath, binaryName)
     val locateCmd        = if (Platform.isWindows) "where" else "which"
-    val path = Process(s"$locateCmd $binaryNameOrPath")
+    val path = Process(Seq(locateCmd, binaryNameOrPath))
       .lineStream_!(silentLogger())
       .map { p => Paths.get(p) }
       .headOption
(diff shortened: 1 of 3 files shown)


ekrich:
I had put in an issue https://github.com/scala-native/scala-native/issues/2127 and had at least a plan sketch to use the `llvm-config` command to get compiling and linking options and only adding defaults if it didn't work. Using Mac tools instead of brew is strange because there is no installed `llvm-config` for example.

I found that `llvm-config` as it is now didn't work as it normally is `llvm-config-6` for example if version 6. I also found that `clang++` is normally aliased to `clang` so maybe it would make more sense to default to `CLANG_PATH` instead of the current. Maybe there is a use case to have a different version of C and C++ compilers but I am not sure.

I think PR right now looks good.

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -185,4 +178,12 @@ private[scalanative] object NativeLib {
       case true  => Some(path)
       case false => None
     }
+
+  private def makeDirPath(path: Path, elems: String*): String = {
+    val pathSep = if (Platform.isWindows) raw"\\" else File.separator
```
ekrich, line 183: This is needed because it is inside a `glob` which is like a regex?

WojciechMazur: Yes, I think that glob in java is also implemented based on regex. 
Without this, we would not be able to correctly find all native libs.

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -85,7 +85,8 @@ object Discover {
   private[scalanative] def checkClangVersion(pathToClangBinary: Path): Unit = {
     def versionMajorFull(clang: String): (Int, String) = {
       val versionCommand = s"$clang --version"
```
sjrd, line 87: Unused?

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -133,8 +134,9 @@ object Discover {
    */
   private[scalanative] def discover(binaryName: String,
                                     envPath: String): Path = {
-    val binaryNameOrPath = sys.env.get(envPath).getOrElse(binaryName)
-    val path = Process(s"which $binaryNameOrPath")
+    val binaryNameOrPath = sys.env.getOrElse(envPath, binaryName)
+    val locateCmd        = if (Platform.isWindows) "where" else "which"
+    val path = Process(s"$locateCmd $binaryNameOrPath")
```
sjrd, line 139: Should this also use a `Seq` for the arguments?

tools/src/main/scala/scala/scalanative/build/Platform.scala
```diff
@@ -0,0 +1,9 @@
+package scala.scalanative.build
+
+object Platform {
+  private lazy val osUsed = System.getProperty("os.name", "unknown").toLowerCase
```
sjrd, line 4: This won't work on a Turkish/Lithuanian/Azeri system.
```suggestion
  private lazy val osUsed = System.getProperty("os.name", "unknown").toLowerCase(Locale.ROOT)
```

sjrd (changes requested):

sjrd (approved):

PR #2201, opened 2021-03-15
Fix to allow 32bit branch to compile and run on Raspberry PI

ekrich:
This change allows the Scala Native `sandbox` to compile and run on 32bit Raspberry Pi OS (Debian variant). The hardware is Raspberry Pi 4 64bit reported as `armv7l` via `uname -mpi` or via `clang-7` as `armv6k-unknown-linux-gnuabihf`.

The changes in this PR are needed just because certain untested paths in the code were not exercised prior to this test/configuration.

These are some additional steps needed to repeat the action.

1. Use the following branch - https://github.com/shadaj/scala-native/tree/32-bit-linker-2
2. Install Boehm GC - `sudo apt install libgc-dev`
3. Edit `build.sbt` add import and then add to `sandbox` settings.
```Scala
lazy val sandbox =
...
.settings(nativeConfig ~= {_.withGC(GC.boehm)})
...
```
4. Add the following to `tools/src/main/scala/scalanative/build/config.scala` on line 85.
```Scala
case "armv6k" => true
```
5. Raspberry Pi 32bit has a maximum process size of 3gb so edit `.sbtopts`. Make max memory 2gb
```
-J-Xmx2G
```
Here is the proof.
```sh
$ file /home/pi/workspace/scala-native/sandbox/target/scala-2.12/sandbox-out
/home/pi/workspace/scala-native/sandbox/target/scala-2.12/sandbox-out: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=ee72073415a19429103a3ba40aa45ad40bd60378, not stripped
```
And ...
```sh
$ /home/pi/workspace/scala-native/sandbox/target/scala-2.12/sandbox-out
Hello, World!
```
```

### DIFF
--- a/nativelib/src/main/resources/scala-native/libunwind/Unwind-EHABI.h
+++ b/nativelib/src/main/resources/scala-native/libunwind/Unwind-EHABI.h
@@ -16,7 +16,9 @@
 #if defined(_LIBUNWIND_ARM_EHABI)
 
 #include <stdint.h>
-#include <unwind.h>
+// #include <unwind.h>
+// Change from original, point to local header file
+#include "include-libunwind/unwind.h"
 
 // Unable to unwind in the ARM index table (section 5 EHABI).
 #define UNW_EXIDX_CANTUNWIND 0x1
--- a/nativelib/src/main/resources/scala-native/libunwind/Unwind-EHABI.h
+++ b/nativelib/src/main/resources/scala-native/libunwind/Unwind-EHABI.h
@@ -16,8 +16,6 @@
 #if defined(_LIBUNWIND_ARM_EHABI)
 
 #include <stdint.h>
-// #include <unwind.h>
-// Change from original, point to local header file
 #include "include-libunwind/unwind.h"
 
 // Unable to unwind in the ARM index table (section 5 EHABI).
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: nativelib/src/main/resources/scala-native/libunwind/Unwind-EHABI.h [hunk 1]
```diff
@@ -16,7 +16,9 @@
 #if defined(_LIBUNWIND_ARM_EHABI)
 
 #include <stdint.h>
-#include <unwind.h>
+// #include <unwind.h>
+// Change from original, point to local header file
+#include "include-libunwind/unwind.h"
```
WojciechMazur, line 21:
```suggestion
#include "include-libunwind/unwind.h"
```
If think this comment is not really needed. It's the last place in the whole codebase where `include <unwind.h>` existed. In all other places, the local header file was always included alredy.

ekrich: Sure, that is fair. Plus, if we go to upgrade the changes will be pretty obvious.

WojciechMazur (approved):

Issue #2202, opened 2021-03-16
Improve GC.none and add sbt setting for how much memory to preallocate


bjornregnell:
## Version
addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.0")
scalaVersion := "2.13.4"

## Problem
Currently the performance when using GC.none is degraded due to page faults in some cases, see e.g.
https://github.com/lunduniversity/scalades/tree/main/benchmark

## Expectation
1. GC.none should be at least as efficient as GC.commix as long as it can cope using initial memory allocated

2. It would be good if GC.none could preallocate memory according to a setting in build.sbt.

ekrich:
Thanks for the issue.

Currently, this should be controlled via ENV vars but I am not sure it is implemented. See #1826

bjornregnell:
Here is a flamegraph of GC.none where page faults are happening
![image](https://user-images.githubusercontent.com/3601817/111381379-ec78d600-86a5-11eb-95ca-b83d4db541bf.png)
and the same program with GC.commix that does some minor GC work in the beginning and then runs unrestricted:
![image](https://user-images.githubusercontent.com/3601817/111381570-2b0e9080-86a6-11eb-8e3b-4e35826bd2e2.png)
Original SVG Graphs are here (but SVG cannot be uploaded to github):
https://github.com/lunduniversity/scalades/tree/main/benchmark/scala-native

WojciechMazur:
Fixed in #2205

PR #2205, opened 2021-03-17
Add a configurable env flag that controls memory preallocation when G…

teodimoff:
…C.none is selected.
Motivated by https://github.com/scala-native/scala-native/issues/2202, we add a configurable flag to adjust the behavior of mmap. An example will be 
```bash 
export GC_NONE_PREALLOC_SIZE=1500M
``` 
which will preallocate 1500 megabytes of memory. If no GC_NONE_PREALLOC_SIZE is provided the default allocation policy of 4GB  chunks will be used. This preallocation lasts only 1 cycle. If the specified preallocation is not enough for the application the next time Scala-Native asks for memory it will be allocated by the default policy - incremental chunks of the os page size.

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/none/gc.c
+++ b/nativelib/src/main/resources/scala-native/gc/none/gc.c
@@ -1,16 +1,54 @@
 #include <stdlib.h>
 #include "MemoryMap.h"
+#include "MemoryInfo.h"
+#include "Parsing.h"
 
 // Dummy GC that maps chunks of 4GB and allocates but never frees.
-
-// Map 4GB
-#define CHUNK (4 * 1024 * 1024 * 1024L)
 
 void *current = 0;
 void *end = 0;
 
+static size_t DEFAULT_CHUNK;
+static size_t PREALLOC_CHUNK;
+static size_t CHUNK;
+static size_t TO_NORMAL_MMAP = 1L;
+static size_t DO_PREALLOC = 0L; // No Preallocation.
+
+void Prealloc_Or_Default() {
+
+    if (TO_NORMAL_MMAP == 1L) { // Check if we have prealloc env varible
+                                // or execute default mmap settings
+        size_t memorySize = getMemorySize();
+
+        DEFAULT_CHUNK = // Default Maximum allocation Map 4GB
+            Choose_IF(Parse_Env_Or_Default_String("GC_MAXIMUM_HEAP_SIZE", "4G"),
+                      Less_OR_Equal, memorySize);
+
+        PREALLOC_CHUNK = // Preallocation
+            Choose_IF(Parse_Env_Or_Default("GC_INITIAL_HEAP_SIZE", 0L),
+                      Less_OR_Equal, DEFAULT_CHUNK);
+
+        if (PREALLOC_CHUNK == 0L) { // no prealloc settings.
+            CHUNK = DEFAULT_CHUNK;
+            TO_NORMAL_MMAP = 0L;
+
+        } else { // config prealloc settings and the flag to reset the
+                 // mmap settings the next iteration.
+            CHUNK = PREALLOC_CHUNK;
+            DO_PREALLOC = 1L;    // Do Preallocate.
+            TO_NORMAL_MMAP = 2L; // Return settings to normal on next iteration.
+        }
+    } else if (TO_NORMAL_MMAP == 2L) {
+        DO_PREALLOC = 0L;
+        CHUNK = DEFAULT_CHUNK;
+        TO_NORMAL_MMAP = 0L; // break the cycle and return to normal mmap alloc
+    } else {
+    }
+}
+
 void scalanative_init() {
-    current = memoryMap(CHUNK);
+    Prealloc_Or_Default();
+    current = memoryMapPrealloc(CHUNK, DO_PREALLOC);
     end = current + CHUNK;
 }
 
--- a/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
+++ b/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
@@ -25,6 +25,25 @@
 #define HEAP_MEM_FLAGS (MAP_PRIVATE | MAP_ANONYMOUS)
 #endif
 
+#ifdef MAP_NORESERVE
+#ifndef __linux__
+// MAP_POPULATE is linux exclusive. We will use madvice.
+#define HEAP_MEM_FLAGS_PREALLOC (MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS)
+#else
+#define HEAP_MEM_FLAGS_PREALLOC                                                \
+    (MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS | MAP_POPULATE)
+#endif
+
+#else
+#ifndef __linux__
+// MAP_POPULATE is linux exclusive. We will use madvice.
+#define HEAP_MEM_FLAGS_PREALLOC (MAP_PRIVATE | MAP_ANONYMOUS)
+#else
+#define HEAP_MEM_FLAGS_PREALLOC (MAP_PRIVATE | MAP_ANONYMOUS | MAP_POPULATE)
+#endif
+
+#endif
+
 // Map anonymous memory (not a file)
 #define HEAP_MEM_FD -1
 #define HEAP_MEM_FD_OFFSET 0
@@ -52,5 +71,44 @@ word_t *memoryMap(size_t memorySize) {
 #else // Unix
     return mmap(NULL, memorySize, HEAP_MEM_PROT, HEAP_MEM_FLAGS, HEAP_MEM_FD,
                 HEAP_MEM_FD_OFFSET);
+#endif
+}
+
+word_t *memoryMapPrealloc(size_t memorySize, size_t doPrealloc) {
+#ifdef _WIN32
+    HANDLE hMapFile;
+    ULARGE_INTEGER memSize;
+    memSize.QuadPart = memorySize;
+
+    hMapFile = CreateFileMappingW(
+        INVALID_HANDLE_VALUE, // use paging file
+        NULL,                 // default security
+        PAGE_READWRITE,       // read/write access
+        memSize.u.HighPart,   // maximum object size (high-order DWORD)
+        memSize.u.LowPart,    // maximum object size (low-order DWORD)
+        NULL);                // name of mapping object
+
+    if (hMapFile == NULL) {
+        return NULL;
+    }
+    return (word_t *)(MapViewOfFile(hMapFile, FILE_MAP_ALL_ACCESS, 0, 0,
+                                    memorySize));
+#else // Unix
+    word_t res;
+    if (doPrealloc) {
+        res = mmap(NULL, memorySize, HEAP_MEM_PROT, HEAP_MEM_FLAGS_PREALLOC,
+                   HEAP_MEM_FD, HEAP_MEM_FD_OFFSET);
+    } else {
+        res = mmap(NULL, memorySize, HEAP_MEM_PROT, HEAP_MEM_FLAGS, HEAP_MEM_FD,
+                   HEAP_MEM_FD_OFFSET);
+    }
+
+#ifndef __linux__
+    // if we are not on linux the next best thing we can do is to mark the pages
+    // as MADV_WILLNEED.
+    madvise(res, memorySize, MADV_WILLNEED);
+#endif
+
+    return res;
 #endif
 }
--- a/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
+++ b/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
@@ -94,7 +94,7 @@ word_t *memoryMapPrealloc(size_t memorySize, size_t doPrealloc) {
     return (word_t *)(MapViewOfFile(hMapFile, FILE_MAP_ALL_ACCESS, 0, 0,
                                     memorySize));
 #else // Unix
-    word_t res;
+    word_t *res;
     if (doPrealloc) {
         res = mmap(NULL, memorySize, HEAP_MEM_PROT, HEAP_MEM_FLAGS_PREALLOC,
                    HEAP_MEM_FD, HEAP_MEM_FD_OFFSET);
--- a/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.h
+++ b/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.h
@@ -6,4 +6,6 @@
 
 word_t *memoryMap(size_t memorySize);
 
+word_t *memoryMapPrealloc(size_t memorySize, size_t doPrealloc);
+
 #endif // MEMORYMAP_H
--- a/nativelib/src/main/resources/scala-native/gc/shared/Parsing.h
+++ b/nativelib/src/main/resources/scala-native/gc/shared/Parsing.h
@@ -1,0 +1,97 @@
+#include <string.h>
+#include <stdio.h>
+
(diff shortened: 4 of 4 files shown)


bjornregnell:
Cool! 
* Should not this PR also include docs?
* Could it include an sbt setting that can be used in build.sbt or this a separate PR?

teodimoff:
No... the are not misplaced at all. First we have to agree on the principle change, but i will add a test some time later today/tomorrow maybe i have the spare time. I will check how to do this in sbt bit someone could also chim in.

bjornregnell:
After publishing a build of this locally (thanks @teodimoff for guidance on how to); with `addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.1-SNAPSHOT")` in `plugins.sbt` and this build.sbt it works fine:
```
scalaVersion := "2.13.4"

// Set to false or remove if you want to show stubs as linking errors
nativeLinkStubs := true

resolvers in Global += "Local Maven Repository" at "file://" + Path.userHome + "/.m2/repository"

envVars in ThisBuild := Map(
  "GC_NONE_PREALLOC_SIZE" -> sys.env.getOrElse("GC_NONE_PREALLOC_SIZE", "1500M"))

nativeConfig ~= { 
  _.withLTO(LTO.thin)
    .withMode(Mode.releaseFast)  // change to releaseFull to get optimized binary in target
    .withGC(GC.none)
}
```
Execution of [scalades/benchmark/scala-bative](https://github.com/lunduniversity/scalades/tree/main/benchmark/scala-native) with more or less mem:
```
$ export GC_NONE_PREALLOC_SIZE=1500M
$ ./target/scala-2.13/scala-native-out
Execution time: 0.384 seconds
Mean number of customers in queuing system: 3.5352454287968054
$ export GC_NONE_PREALLOC_SIZE=1M
$ ./target/scala-2.13/scala-native-out
Execution time: 0.801 seconds
Mean number of customers in queuing system: 3.5258284673466553
$ export GC_NONE_PREALLOC_SIZE=800M
$ ./target/scala-2.13/scala-native-out
Execution time: 0.544 seconds
Mean number of customers in queuing system: 3.5190386794594697
```

```

**Comment by ekrich** [comment]
Related: https://github.com/scala-native/scala-native/issues/1826

**Comment by ekrich** [comment]
From the Boehm docs - https://github.com/ivmai/bdwgc/blob/master/doc/README.environment

Perhaps we should use
```
GC_INITIAL_HEAP_SIZE=<bytes> -  Initial heap size in bytes.  May speed up
                                process start-up.  Optionally, may be
                                specified with a multiplier ('k', 'M' or 'G')
                                suffix.
```

**Comment by teodimoff** [comment]
@ekrich Suggested changes addressed. We added a new function getAvailableMemory() which uses 
 ```c
 sysconf  _SC_AVPHYS_PAGES and _SC_PAGE_SIZE
 ```
 and  to calculate the 'current' available memory. If GC_INITIAL_HEAP_SIZE requires more memory than available it will preallocate getAvailableMemory() Only. As usual when the pre-allocated memory is exhausted we delegate to the default 4GB mmap with page os size chunks.

**Comment by teodimoff** [comment]
How to investigate: Run tests / Test runtime (2.12.13, debug, commix) (pull_request) Failing after 7m — Test runtime (2.12.13, debug, commix) 
```Error:  scala.scalanative.testinterface.common.RPCCore$ClosedException``` and ```caused by: java.lang.RuntimeException: Process /home/scala-native/scala-native/unit-tests/target/scala-2.12/tests-out finished with non-zero value 139``` does not tell me much.

**Comment by ekrich** [comment]
I was doing some checking on macOS and found these.
```c
    printf("_SC_PHYS_PAGES: %d\n", _SC_PHYS_PAGES);
    printf("_SC_PAGE_SIZE: %d\n", _SC_PAGE_SIZE);
```
Result:
```
os.name=Darwin
os.version=18.7.0
version=Darwin Kernel Version 18.7.0: Tue Jan 12 22:04:47 PST 2021; root:xnu-4903.278.56~1/RELEASE_X86_64
_SC_PHYS_PAGES: 200
_SC_PAGE_SIZE: 29
```

WojciechMazur:
Exit code 139 in most cases means segmentation fault. There is currently no good way of debugging this. In most cases I was trying to add some additional logging somewhere in testrunner that would print last N messages in case of unexpected exception and based on that to find which test actually failed.  
With known failing test I try to reproduce it in sandbox, this way it's easier to find cause using lldb or other debuger. If it does not fail in sandbox then I attach inside debugger to actual process by pid (to make it easier I set breakpoint in ProcessRunner, otherwise we might miss actual test execution)

teodimoff:
@ekrich Okay i will check them tomorrow and add ifdef macro as needed. Have to see how  portable sysconf because i get the impression that is POSIX compliant, but a quick check shows it is not. I will try to find the most portable solution. Just have to fix this error first.  For macos the proper way to find out free memory is vm_stat.free_count - vm_stat.speculative_count. I will look tomorrow because it getting late here.

teodimoff:
@ekrich Okay our native code should get a good clean up (which i may get to eventually), because we keep copying the platform specific code (immix/Memory.h, commix/Memory etc etc), but this is for another PR. For this PR the smallest viable change is to use the commix Memory.h to get the memory size with the caveat that this is total amount of memory(the _SC_PHYS_PAGES are also total amount of physical pages), not current available memory. For this PR  this should suffice for now, because to support this properly  (functions like getMemorySize, getAvailMemory, etc), there is no getting around Platform code (Platform.h,Platform.c etc), and we should think what we want to support in a future Platform.h(which will take more effort to get right). 

Anyway this PR is good for review.

ekrich:
@teodimoff I think the code used in Immix and Commix could be used in None - https://github.com/scala-native/scala-native/blob/master/nativelib/src/main/resources/scala-native/gc/commix/Memory.h

teodimoff:
@ekrich This is what i`ve done. To keep PR changes minimal i deleted the AvailMem files and used the commix Memory.h getMemorySize().

ekrich:
Sorry, I missed those commits. I am just a step behind here.

nativelib/src/main/resources/scala-native/gc/none/gc.c
```diff
@@ -9,19 +10,58 @@
...
+    if (TO_NORMAL_MMAP == 1L) { // Check if we have prealloc env varible or
+                                // execute default mmap settings
+        size_t PreallocSize = Parse_Env_Or_Default("GC_INITIAL_HEAP_SIZE", 0);
+
+        if (PreallocSize == 0) { // no prealloc settings.
+            TO_NORMAL_MMAP = 0L;
+
+        } else { // config prealloc settings and the flag to reset the mmap
+                 // settings the next iteration.
+            if (PreallocSize <= getMemorySize()) {
```
bjornregnell, line 43: I'm not a C-programmer but will try to give my review input anyway.
Would it not be better to call `getMemorySize()` once before the `if` and save the result in a local variable?

nativelib/src/main/resources/scala-native/gc/none/gc.c
```diff
@@ -9,19 +10,58 @@
...
+                CHUNK = PreallocSize;
+                DUMMY_GC_FLAGS = set_prealloc();
+                TO_NORMAL_MMAP = 2L;
+
+            } else if (PreallocSize > getMemorySize() && getMemorySize() > 0) {
```
bjornregnell, line 48: `getMemorySize()` is called her two more times...

nativelib/src/main/resources/scala-native/gc/none/gc.c
```diff
@@ -9,19 +10,58 @@
...
+
+                CHUNK = getMemorySize();
```
bjornregnell, line 50: ...and here once more.

bjornregnell (changes requested):
I'm not a C-programmer but will boldly ;) try to give my review input below in the code anyway.
Suggestion: factor out call to `getMemorySize()` and save it in a local var. Or is there any special reason to call it multiple times?

nativelib/src/main/resources/scala-native/gc/none/gc.c
teodimoff: The intention was deliberate, because using getAvailSize()(the previous version of getMemorySize function) was getting the current available memory. Using getMemorySize() gives a static answer (it gets the total amount of memory). This is not critical, we only call this three times as you pointed, but for the sake of good code style i will factor it to use only one getMemorySize() call.

bjornregnell: OK good. Yes, I couldn't tell if there was some OS side-effect or some temporal consideration, but if not it is easier to grok if changed like so and I agree also better coding style.

teodimoff:
@ekrich No worries. I think this is done for a review now.

bjornregnell (approved):
LGTM

ekrich:
@WojciechMazur There is the `Memory.h` file (from Immix/Commix) that contains the C implementation and there is also the `Parsing.h` that is being added/sort of copied from`Settings.c/.h` that does the parsing of the env vars that contain `k K m M g G` for kilobytes and bigger. I encouraged @teodimoff to use the Boehm name for the `INITIAL` and `MAXIMUM` heap which is think makes allot of sense because then those env vars would work for Boehm too.

What I am wondering is if we should consider adding `common` at the top level for `Memory.h` and `Parsing.h` with the C code so they are header only and won't mess up the `.o` removal code and then the Immix and Commix could be also refactored to use? 

It could contain the basic shared items like here in Immix: https://github.com/scala-native/scala-native/blob/master/nativelib/src/main/resources/scala-native/gc/immix/Settings.h

The Boehm values are shown here https://github.com/ivmai/bdwgc/blob/master/doc/README.environment and here
https://github.com/ivmai/bdwgc/blob/master/misc.c  Misc also contains the `GC_parse_mem_size_arg` function.
These are applicable: `GC_INITIAL_HEAP_SIZE`, `GC_MAXIMUM_HEAP_SIZE`, and `GC_USE_ENTIRE_HEAP`.

We could defer the `GC_LOG_` and `GC_PRINT_` stuff for a future PR but there is a bunch of those probably include `DEFINE`s to leave the code in the compiled version. That could be used for our current `STATS` setup and shared across GCs.

I know this kind of expands this PR a bit but it also takes quite awhile to think about and review this sort of thing. What do you think?

WojciechMazur:
@ekrich 
I agree that including logic from other GC looks like a dirty hack and we should do move some crucial parts to common sources. I've got some pending work in my develop branch for windows that might be used for fixing deduplication that might be used later  https://github.com/WojciechMazur/scala-native/commit/6e1b06d92a8f185f9d23d3fc3cf234209bcea9c3. 
I've got there some logic that allows us to decide which files should be compiled based on defined filters as well as provides some support for sharing sources based on their relative path (from resources/scala-native/*), eg. 
* `gc/shared/*` would be compiled and included for all GCs
* `gc/immix-commix/*` would be compiled only for those two implementations
* `gc/immix/*` sources only for immix
* `posix/*` would only be compiled when detected compilation for posix compatible os 

Of course, naming can change, eg. to `scalanative_gc` or `_gc_` to not clash with user-defined directories, for `posix` we could stay with that name to allow for conditional compilation of posix/windows native sources.

I think, for now, we can pass this PR as it is with added duplication and fix this issue later in separate follow-up PR.

ekrich:
@WojciechMazur We could add a shared dir and put Memory.h and Settings.h in that directory. Since there is no .c files it would work as it does now. We could make the #define SHARED_MEMORY_H to make sure it doesn’t conflict. This way the files are in place and a follow on PR could have the other GCs use the code. I think that would be a good middle ground.

ekrich:
Just to capture some key agreement points from Gitter, reworded to be more directive.

We should putting the Memory.h and Settings.h in the `shared` directory at the top level. We can make the defines as follows:
```c
#ifndef SHARED_MEMORY_H
#define SHARED_MEMORY_H
// and
#ifndef SHARED_SETTINGS_H
#define SHARED_SETTINGS_H

// Since the current are immix and commix like this.

#ifndef IMMIX_MEMORY_H
#define IMMIX_MEMORY_H
```

And then after the `None` gets merged the other `GC`s could follow and any changes made during that refactoring?

@WojciechMazur
Mar 26 17:56
If I remember correctly current settings were depending on BlockMeta size which is different in `immix` and `commix`, was that handled? I think it was used for min or max default heap size. I'm fine with that change overall.

teodimoff:
Yes this should work. A common header. Most of the functions we will support as common are even more generic than gc related. In my opinion, their place should be in a common or util at top level so they can be reused without a gc prefix .

nativelib/src/main/resources/scala-native/gc/none/Parsing.h
```diff
@@ -0,0 +1,97 @@
+#include <string.h>
```
WojciechMazur, line 1: We can now move it to `gc/shared/` and use it also in Immix/Commix

nativelib/src/main/resources/scala-native/gc/none/gc.c
```diff
@@ -1,30 +1,83 @@
 #include <stdlib.h>
 #include <sys/mman.h>
-
+#include "Parsing.h"
+#include "../commix/Memory.h"
```
WojciechMazur, line 4: Location and name of this file has changed after adding Windows support (since it is not case-sensitived and clashed with other header files)
```suggestion
#include "MemoryInfo.h"
```

nativelib/src/main/resources/scala-native/gc/none/gc.c
```diff
@@ -1,30 +1,83 @@
...
+static size_t CHUNK;
+static size_t DUMMY_GC_FLAGS = DEFAULT_DUMMY_GC_FLAGS;
+static size_t TO_NORMAL_MMAP = 1L;
+
+size_t set_prealloc() {
+#ifdef __linux__
+    return (MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS | MAP_POPULATE);
+#else
+    // MAP_POPULATE is linux exclusive
+    return DEFAULT_DUMMY_GC_FLAGS(MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS);
```
WojciechMazur, line 34: Is it correct? Should it not be just `DEFAULT_DUMMY_GC_FLAGS`? 
Also, we might think about removing the `DUMMY_` part from all names, since in most cases those are actual/final values.

nativelib/src/main/resources/scala-native/gc/none/gc.c
```diff
@@ -1,30 +1,83 @@
...
 void scalanative_init() {
+    Prealloc_Or_Default();
     current = mmap(NULL, CHUNK, DUMMY_GC_PROT, DUMMY_GC_FLAGS, DUMMY_GC_FD,
                    DUMMY_GC_FD_OFFSET);
     end = current + CHUNK;
+#ifndef __linux__
+    // if we are not on linux the next best thing we can do is to mark the pages
+    // as MADV_WILLNEED.
+    madvise(current, CHUNK, MADV_WILLNEED);
+#endif
```
WojciechMazur, line 80: Can we move it now to `shared/MemoryMap.c` (util with shared memory mapping function for both Unix/Windows platforms)?

WojciechMazur (changes requested):
After the merge of sharing sources in GC and adding support for Windows, we would need to rebase. Can I please ask you do that? I've added some comments about files that were moved or renamed.

teodimoff:
Force pushed rebased branch on top of master. Duplicated memoryMap into memoryMapPrealloc with extra argument to control if (size_t memorySize) has to be preallocated with the account of not being on linux system tring madvice with WILL_NEED. It is supposed to work on BSD and OSX. On windows it will do nothing. At some point Windows will need to handle this properly(preallocating/prefaulting the memory pages). VirtualAlloc with MEM_COMMIT | MEM_RESERVE seems worth checking out.

WojciechMazur (approved):

PR #2206, opened 2021-03-17
Allow for linktime conditional branching

WojciechMazur:
This PR adds a new concept to Scala Native by allowing to define values resolved at link-time. Based on that we're able to define link-time conditions allowing us to discard unused execution branches while performing class loading, before performing static reachability analysis.
When performing linking values can be resolved based on values defined in `NativeConfig` in form of `Map[String, Any]` with runtime checks limiting allowed values to non-empty strings and non-null primitives. In case if the property is not defined linking would fail. Link time property needs to be non-null literal. It also needs to define the explicit return type and keep `resolved` at its right-hand side in the same manner as it's done with extern defintions.

Link time conditions are based on `if-else` clauses. Runtime branching is replaced with link-time if the value used in the condition is annotated with `@resolvedAtLinktime` annotation. In the case of usage of mixed runtime and link-time resolved values, a compile-time error is being thrown.

Example of usage:
```
import scalantive.unsafe.{resolvedAtLinktime, resolved}

object Test {

  // linktime value resolved from property `Test.boolProp` 
  @resolvedAtLinktime
  def boolProp: Boolean = resolved

  // linktime value resolved from property `my.string.prop`
  @resolvedAtLinktime(withName = "my.string.prop")
  def stringProp: String = resolved

  @resolvedAtLinktime()
  def intProp: Int = resolved

   val x = if(!boolProp) path1()
    else if(intProp > 2 && intProp <= 10) path2()
    else path3()

    if(stringProp == "") ()
    else println(stringProp)
  }
}
```

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/unsafe/resolvedAtLinktime.scala
+++ b/nativelib/src/main/scala/scala/scalanative/unsafe/resolvedAtLinktime.scala
@@ -5,7 +5,7 @@ import scala.annotation.meta.{field, getter}
 import scala.scalanative.runtime.intrinsic
 
 /** Used to annotate that given value should be resolved at link-time,
- *  based on provided `withName` parameter or on fully-qualified name otherwise */
+ *  based on provided `withName` parameter */
 @field @getter
-private[scalanative] class resolvedAtLinktime(withName: String = intrinsic)
+private[scalanative] class resolvedAtLinktime(withName: String)
     extends StaticAnnotation
--- a/nativelib/src/main/scala/scala/scalanative/unsafe/resolvedAtLinktime.scala
+++ b/nativelib/src/main/scala/scala/scalanative/unsafe/resolvedAtLinktime.scala
@@ -1,0 +1,11 @@
+package scala.scalanative.unsafe
+
+import scala.annotation.StaticAnnotation
+import scala.annotation.meta.{field, getter}
+import scala.scalanative.runtime.intrinsic
+
+/** Used to annotate that given value should be resolved at link-time,
+ *  based on provided `withName` parameter or on fully-qualified name otherwise */
+@field @getter
+private[scalanative] class resolvedAtLinktime(withName: String = intrinsic)
+    extends StaticAnnotation
--- a/nir/src/main/scala/scala/scalanative/nir/Linktime.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Linktime.scala
@@ -6,7 +6,7 @@ sealed trait LinktimeCondition {
 
 object LinktimeCondition {
 
-  case class SimpleCondition(name: Global, comparison: Comp, value: Val)(
+  case class SimpleCondition(propertyName: String, comparison: Comp, value: Val)(
       implicit val position: Position)
       extends LinktimeCondition
 
@@ -31,19 +31,11 @@ object Linktime {
   // Takes Global for constant struct describing linktime property.
   // Replaced with resolved value at link-time.
   final val PropertyResolveFunctionName: Global.Member =
-    Linktime.member(Sig.Method("resolveProperty", Seq(Type.Ptr, Type.Nothing)))
+    Linktime.member(Sig.Method("resolveProperty", Seq(Rt.String, Type.Nothing)))
 
   final def PropertyResolveFunctionTy(retty: Type): Type.Function =
-    Type.Function(Seq(Type.Ptr), retty)
+    Type.Function(Seq(Rt.String), retty)
 
   final def PropertyResolveFunction(retty: Type): Val.Global =
     Val.Global(PropertyResolveFunctionName, PropertyResolveFunctionTy(retty))
-
-  def nameToLinktimePropertyName(name: Global): Global.Member = {
-    val Global.Member(owner, sig) = name
-    require(sig.isMethod, s"Expected method, but received ${sig.unmangled}")
-
-    val Sig.Method(ident, _, _) = sig.unmangled
-    owner.member(Sig.Generated(s"${ident}_property"))
-  }
 }
--- a/nir/src/main/scala/scala/scalanative/nir/Linktime.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Linktime.scala
@@ -1,21 +1,26 @@
 package scala.scalanative.nir
 
-sealed trait LinktimeCondition
+sealed trait LinktimeCondition {
+  def position: Position
+}
 
 object LinktimeCondition {
 
-  case class SimpleCondition(name: Global, comparison: Comp, value: Val)
+  case class SimpleCondition(name: Global, comparison: Comp, value: Val)(
+      implicit val position: Position)
       extends LinktimeCondition
 
-  case class ComplexCondition(op: Bin,
-                              left: LinktimeCondition,
-                              right: LinktimeCondition)
+  case class ComplexCondition(
+      op: Bin,
+      left: LinktimeCondition,
+      right: LinktimeCondition)(implicit val position: Position)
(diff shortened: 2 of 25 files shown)


sjrd:
As a first comment on the public API, I don't think using env vars or system properties is the best way to go. They're too sensitive to the environment. We should use a dedicated field of `nativeConfig` instead.

Now whether that field should be an arbitrary `Map[String, Any]` or something more precise, I am not sure. In Scala.js for example we have something similar with `linkingInfo`, and it is limited a fixed set of fields that are known from the linker. The values can be set by the users, but the users cannot define their own fields. Before we open that to custom keys, we should know what use cases it fills.

WojciechMazur:
I've replaced system properties and env variables with `Map[String,Any]` in nativeConfig. If we decide to add fixed keys we can pass them in this map along with possible user-defined settings.

nir/src/main/scala/scala/scalanative/nir/Linktime.scala
```diff
@@ -0,0 +1,49 @@
+package scala.scalanative.nir
+
+sealed trait LinktimeCondition {
+  def position: Position
+}
+
+object LinktimeCondition {
+
+  case class SimpleCondition(name: Global, comparison: Comp, value: Val)(
```
sjrd, line 9: Instead of storing a reference to the definition of the condition here, consider directly storing the string literal that is in its annotation. For example if calling
```scala
@linkTimeResolved("foo")
def bar: String = ...

if (bar == "hello")
```
instead of storing
1. a member of `bar` that references `"foo"`, and
2. a `SimpleCondition(Global(bar), ...)`

consider storing
1. nothing for the *definition* of `bar` (not present in the NIR at all)
2. a `SimpleCondition("foo", ...)`

This would further simplify several things, IMO, and notably the fact that we do not need to enhance arbitrary `Global`s in the NIR with the possibility of being a link-time-resolved value.

nativelib/src/main/scala/scala/scalanative/unsafe/resolvedAtLinktime.scala
```diff
@@ -0,0 +1,11 @@
+package scala.scalanative.unsafe
+
+import scala.annotation.StaticAnnotation
+import scala.annotation.meta.{field, getter}
+import scala.scalanative.runtime.intrinsic
+
+/** Used to annotate that given value should be resolved at link-time,
+ *  based on provided `withName` parameter or on fully-qualified name otherwise */
+@field @getter
+private[scalanative] class resolvedAtLinktime(withName: String = intrinsic)
```
sjrd, line 10: Consider making the parameter mandatory. This annotation should be used rarely enough that it is not worth trying to guess what a good default should be.
```suggestion
private[scalanative] class resolvedAtLinktime(withName: String)
```

tools/src/test/scala/scala/scalanative/linker/LinktimeConditionsSpec.scala
```diff
@@ -0,0 +1,418 @@
...
+       |  @resolvedAtLinktime()
+       |  def float: Float = resolved
+       |  
+       |  object inner{
+       |   @resolvedAtLinktime()
+       |   def countFrom: Long = resolved
+       |   
+       |   @resolvedAtLinktime("secret.performance.multiplier")
+       |   def performanceMultiplier: Double = resolved
+       |  }
```
sjrd, line 35:
```suggestion
       |  object inner{
       |    @resolvedAtLinktime()
       |    def countFrom: Long = resolved
       |
       |    @resolvedAtLinktime("secret.performance.multiplier")
       |    def performanceMultiplier: Double = resolved
       |  }
```

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -155,7 +163,56 @@ object NativeConfig {
     def withOptimize(value: Boolean): NativeConfig =
       copy(optimize = value)
 
-    override def toString: String =
+    override def withLinktimeProperties(v: Map[String, Any]): NativeConfig = {
+      def isNumberOrString(value: Any) = {
+        def isNonUnitPrimitive = value match {
+          case _: java.lang.Boolean | Boolean => true
```
sjrd, line 169: This seems extremely weird. It would match any instance of `java.lang.Boolean` (which also means any `Boolean`), or *the companion object `scala.Boolean`*. I'm pretty sure that is not what was intended.

I think what you intended was
```scala
def isNonUnitPrimitive = value match {
  case _: Byte | _: Char | _: Short | _: Int | _: Long | _: Float | _: Double => true
  case                                                                        => false
}
```

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -155,7 +163,56 @@ object NativeConfig {
...
+          case _: java.lang.Long | Long       => true
+          case _: java.lang.Float | Float     => true
+          case _: java.lang.Double | Double   => true
+          case _                              => false
+        }
+
+        def isNonEmptyString = value match {
+          case s: String => s.nonEmpty
+          case _         => false
+        }
```
sjrd, line 183: Why reject empty strings?

sjrd (changes requested):
Overall it looks good. I identified two more things that can be simplified: no default value for `@linkTimeResolve("foo")`, and directly storing the `"foo"` at call site rather than a reference to the member.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
WojciechMazur: I thought it might be better for the consistency of data, but on the other hand, I can now think of some usages where the empty string would be useful. I'll remove this condition

nir/src/main/scala/scala/scalanative/nir/Linktime.scala
WojciechMazur: Good idea, it's a much better approach, since NIR global struct consistent of only a single field. 
Fixed that and PR is ready for next review round

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
```diff
@@ -606,6 +606,10 @@ trait NirGenStat[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
           // Have a concrete method with JavaDefaultMethodAnnotation; a blivet.
           // Do not emit, not even as abstract.
 
+          case _
+              if sym.annotations.exists(_.symbol == ResolvedAtLinktimeClass) =>
```
sjrd, line 610:
```suggestion
              if sym.hasAnnotation(ResolvedAtLinktimeClass) =>
```

nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala
```diff
@@ -787,14 +839,33 @@ trait NirGenStat[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
...
+      case _ => ()
+    }
+  }
+
+  protected object LinktimeProperty {
+    def unapply(tree: Tree): Option[(String, nir.Position)] = {
+      if (tree.symbol == null) None
+      else {
+        tree.symbol.annotations
+          .find(_.symbol == ResolvedAtLinktimeClass)
```
sjrd, line 858:
```suggestion
        tree.symbol
          .getAnnotation(ResolvedAtLinktimeClass)
```

sjrd (approved):
Just two tiny improvements below. Otherwise LGTM!

PR #2207, opened 2021-03-17
Fix resolving binary version in ScalaNativeCrossVersion for snapshots

WojciechMazur:
This fix replaces invalid resolving of binary version. Until now it used regex expression that was not able to handle snapshot and rc versions. This leads to problems when testing snapshot versions of libraries. Fix includes aligning logic used to extract cross version with Scala.js behaviour.
Eg. version `0.4.1-SNAPSHOT` was used as a binary version instead of `0.4`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -312,7 +312,7 @@ lazy val tools =
     .settings(
       libraryDependencies ++= Seq(
         "org.scalacheck" %% "scalacheck" % "1.14.3" % "test",
-        "org.scalatest"  %% "scalatest"  % "3.1.1"  % "test"
+        "org.scalatest" %% "scalatest" % "3.1.1" % "test"
       ),
       Test / fork := true,
       Test / javaOptions ++= {
--- a/build.sbt
+++ b/build.sbt
@@ -312,7 +312,7 @@ lazy val tools =
     .settings(
       libraryDependencies ++= Seq(
         "org.scalacheck" %% "scalacheck" % "1.14.3" % "test",
-        "org.scalatest" %% "scalatest" % "3.1.1" % "test"
+        "org.scalatest"  %% "scalatest"  % "3.1.1"  % "test"
       ),
       Test / fork := true,
       Test / javaOptions ++= {
--- a/nir/src/main/scala/scala/scalanative/nir/Versions.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Versions.scala
@@ -25,7 +25,7 @@ object Versions {
   final val revision: Int = 8 // a.k.a. MINOR version
 
   /* Current public release version of Scala Native. */
-  final val current: String              = "0.4.1-SNAPSHOT"
+  final val current: String = "0.4.1-SNAPSHOT"
   final val currentBinaryVersion: String = binaryVersion(current)
 
   private object FullVersion {
@@ -37,10 +37,13 @@ object Versions {
       version match {
         case FullVersionRE(major, minor, patch, preReleaseString) =>
           Some(
-            (major.toInt,
-             minor.toInt,
-             patch.toInt,
-             preRelease(preReleaseString)))
+            (
+              major.toInt,
+              minor.toInt,
+              patch.toInt,
+              preRelease(preReleaseString)
+            )
+          )
         case _ => None
       }
     }
--- a/nir/src/main/scala/scala/scalanative/nir/Versions.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Versions.scala
@@ -25,5 +25,32 @@ object Versions {
   final val revision: Int = 8 // a.k.a. MINOR version
 
   /* Current public release version of Scala Native. */
-  final val current: String = "0.4.1-SNAPSHOT"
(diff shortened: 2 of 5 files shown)


sjrd:
Is this in line with https://github.com/scala-js/scala-js/blob/d09ce384d35e3310272e8099dda5f2d3fee8dac1/ir/shared/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala#L37-L52 ?

WojciechMazur:
> Is this in line with https://github.com/scala-js/scala-js/blob/d09ce384d35e3310272e8099dda5f2d3fee8dac1/ir/shared/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala#L37-L52 ?

Partially yes. I mean there is no logic for major versions, but for `0.x.x` versions it is correct.
In ScalaNative we only use a single `current` version without an explicit binary version.
Until now if pre-release version was used eg. `0.4.1-SNAPSHOT` it would result in an artefact with a cross-native part set to `_native0.4.1-SNAPSHOT_`  instead of `_native0.4_`. This PR fixes this issue. 

Do you think we should port the whole logic from scala-js now?

sjrd:
The important aspect is that we must still make sure that the binary version for 0.5.0-SNAPSHOT will be 0.5.0-SNAPSHOT, and *not* 0.5. We can only reuse the short binary version once we're at 0.x.y-SNAPSHOT for y > 0.

WojciechMazur:
Usage of `CrossVersion.scalaBinaryVersion` would be compatible for that case, however, since >=1.0.0 it's no longer correct. According to test cases in Scala.js for 1.0.0 cross-version suffix should be `native1` but it results would result in `native1.0.0` 
In such a case, I guess it might be a good idea to port it right now.

WojciechMazur:
I've aligned the current logic with Scala.js and added unit tests. As currently, Scala Native uses NIR compat/revision versioning I have not tried to port logic from Scala.js based on `binaryEmitted` version and it's additional validations. 

Also, I think that we should switch to Scala.js IR versioning in the next major release or since 1.x.x as it seems a lot more readable and informative - NIR in version 5.8 seems ridiculous when the latest version of SN is 0.4 and does not bring any additional information allowing to easily check when given binary was created.

avdv:
Hi.

Am I correct that this means that the current 0.4.1-SNAPSHOT should be compatible with 0.4.0 since both would be mapped to the same binary version?

I just tried to use a locally build scala-native off the master branch, where I set `currentVersion` to 0.4.0 in place of the official release (because I wanted to avoid rebuilding all my dependencies) and it failed with:
```
[error] missing symbols:
[error] * M34scala.scalanative.runtime.package$D7toClassR_L15java.lang.ClassEO
```
Probably because of commit 2f5b2a0780cbc4c0cf5c5da0624b8e517275da3e ?

What's the plan on binary compat for scala-native? Thanks!

ekrich:
@avdv There are no compatibility guarantees when using functions from the runtime API as they are not part of the public API. Is there a way to use public APIs for you use case?

WojciechMazur:
@avdv They should be compatible. After merging PR from the commit you correctly typed we removed some internal methods. I guess we will try to add some fixes to allow compatibility with 0.4.0.

ekrich:
@WojciechMazur I checked and it did indeed remove the method. My claim that that is in `runtime` which I thought was internal.

avdv:
Thanks guys!

> They should be compatible. After merging PR from the commit you correctly typed we removed some internal methods. It looks like it calls to them are still present in the NIR format. I guess we will try to add some fixes to allow compatibility with 0.4.0 since I believe these calls are not direct, but transitive due to `Object.getClass` calls.

That sound good.

BTW, is there a tool to demangle those names back to a package / class / method name, like c++filt?

ekrich:
@avdv I must be missing something - sorry about the noise.

avdv:
FTR, I applied this small patch:
```diff
diff --git a/nativelib/src/main/scala/scala/scalanative/runtime/package.scala b/nativelib/src/main/scala/scala/scalanative/runtime/package.scala
index 2a16e306..ae107d93 100644
--- a/nativelib/src/main/scala/scala/scalanative/runtime/package.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/package.scala
@@ -6,6 +6,12 @@ import scalanative.runtime.Intrinsics._
 
 package object runtime {
 
+  @deprecated("compatibility for 0.4.0")
+  def toClass(raw: RawPtr): _Class[_] = {
+    val rtti = loadRawPtr(raw)
+    castRawPtrToObject(rtti).asInstanceOf[_Class[_]]
+  }
+
   /** Used as a stub right hand of intrinsified methods. */
   def intrinsic: Nothing = throwUndefined()
 
```
... and I can use scala-native from master as a drop-in replacement for 0.4.0 :tada:

sjrd (approved):

avdv:
Nice work @WojciechMazur ! After merging this (and #2356) it works like a charm using 0.4.0 libraries with 0.4.1-SNAPSHOT and `%%%` :heart: !

PR #2225, opened 2021-03-27
Introduced a way to construct user's event loop.

catap:
Here a tricky commit that allows to enduser to construct his own event loop which can includes `epoll`, `poll`, libuv's `poll` or whatever he would like to use.

Right now such event loop can be constructed via direct access to `queue` inside `scala.scalanative.runtime.ExecutionContext` via magic of pointers because scala-native hasn't got any reflection (yet?).

Thus, after this changes `scala-native-loop` event loop will be something like:
```
  def run(): Unit = {
    while ({
      val hasNext = scala.scalanative.runtime.once()
      if (hasNext) {
        uv_run(loop, UV_RUN_NOWAIT)
      } else {
        uv_run(loop, UV_RUN_ONCE)
      }
      uv_loop_alive(loop) || hasNext
    })()
  }
```

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/ExecutionContext.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/ExecutionContext.scala
@@ -14,8 +14,8 @@ object ExecutionContext {
 
   private val queue: ListBuffer[Runnable] = new ListBuffer
 
-  private[runtime] def loop(): Unit = {
-    while (queue.nonEmpty) {
+  private[runtime] def once(): Boolean = {
+    if (queue.nonEmpty) {
       val runnable = queue.remove(0)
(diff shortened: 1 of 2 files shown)


lolgab:
Isn't this the same as [this pr](https://github.com/scala-native/scala-native/pull/1852)?

catap:
@lolgab well, I haven't seen this PR yet :) Thanks.

Anyway, here you may notice that I also added `Boolean` to `once` that allows to know shall it be used for the next turn and it allows to make this call safe, without thrown an exception.

Why? Because I have no idea how to rewrite `scala-native-loop` event loop via your pr.

So, here the question: which one we keep opens? I can incorporate your tests here and include your as co-author. Is it fine for you?

PR #2230, opened 2021-03-29
Allow for filtering native sources based on relative path

WojciechMazur:
This PR is a cherry-pick of my changes in branch with initial Windows support and present my proposal 
for the restructuration of native sources directories.

Currently, our method of including native sources quite simple, however not enough flexible, eg. in terms of sharing sources and limited to a single GC directory and optional directories. 
Based on the fact that all native use the same naming schema, we are able, to extract paths relative to `.../resources/scala-native/` in each project. Based on that define set of filter taking into account elements of this relative path. As an example relative path `gc/immmix/...` does always point to our ImmixGC implementation. 

I propose to introduce a more structured naming schema for our directories. As an example, by moving posix sources to the inner directory `posix` we can easily create a filter that would include these sources to compilation only when we're using POSIX compliant system. In the future, we can do the same for Windows sources. Thanks to that we can include an additional `shared` directory that would be always compiled on each platform. 
I've found out that some posix headers are in most parts compilable on Windows, eg. `errno.h` with only 4 missing codes. We can place such header in top-level (`/resources/scala-native/errno.h`) allowing them to be shared. Such an approach should always work, as we assume that `posixlib` and new `windowslib` would be always existing since they both would be defined as library dependency of `javalib`
As a step towards sharing common sources between our GC implementation we can define a directory that would be shared between specified implementations, eg. sources in `shared_immix-commix` would be compiled for each of these GCs, but not for NoneGC. 

I open for further discussion about this idea. I think that we need to enforce some better directories structure, posix case right now is a blocker for windows support since with the current structure we're not able to detect which directories should be compiled and which should not.


ekrich:
I took a quick look at this PR - I also submitted a PR based on just sharing GC code which is just based on the current practice.

I have a couple of general comment about deficiencies in the current build system.
1. Linking is global so there is no concept of this C code is associated with this Scala stub code. This also means that the C code and Scala stub code is not associated via the projects or the C code compiled independently as part of the project. There is also no concept of building the Scala and C together and thus the dependencies of the C code.
2. All the include / exclude code in `LLVM.compileNativeLibs` is just for the `nativelib` project itself.
3. There are some good things because the current code does one directory traversal to gather all the C/C++ code which is more efficient. This also led to some nice optimizations that passes the output to the next processing step to avoid looking up the inputs more than once.
4. The C code can be in the `clib` or the `posixlib` and the build system doesn't care.

I think if we look at the shortfalls of 1-4 above, we might what to look from that point of view first. I am not sure there is a better solution but I think it is worth thinking about. It might make sense to look and see if we can compile projects and libraries separately and leave the complexity of compiling the `nativelib` to itself somehow. Looking upstream to prior to where `LLVM.compileNativelibs` is called might yield some different thoughts. We could even think about making the GC code itself into a separate project or even our curated `libunwind`. Just throwing out some ideas.

ekrich:
I have a way we could potential do this safer and in a more general way in the second comment rather than the suggest just above. https://github.com/scala-native/scala-native/pull/2234/

WojciechMazur:
Closed in favour of approach filtering native sources on compiler pre-processor conditions merged in #2276

PR #2232, opened 2021-03-29
Removes magic constants from the commix sempahore length code

ncordon:
## What
Removes the magic constants from the code that limits the length of the semaphore names in commix GC

## Why
There are system dependent variables that limit the maximum length of semaphores for `sem_open`, both in POSIX compliant and in OSX systems. In OSX it should be limited by `PSEMNAMLEN`, in rest of the systems, by `_POSIX_PATH_MAX`. Source docs: [POSIX.1-2017](https://pubs.opengroup.org/onlinepubs/9699919799/).  See #2220 for discussion around `PSEMNAMLEN`.

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/commix/Phase.c
+++ b/nativelib/src/main/resources/scala-native/gc/commix/Phase.c
@@ -9,15 +9,30 @@
 #include <errno.h>
 #include <stdlib.h>
 
+/*
+If in OSX, sem_open cannot create a semaphore whose name is longer than
+PSEMNAMLEN characters. Else, we stick to POSIX compliance:
+
+  The length of the name argument cannot exceed {_POSIX_PATH_MAX} on systems
+  that do not support the XSI option or exceed {_XOPEN_PATH_MAX} on XSI systems.
+  We restrict to _POSIX_PATH_MAX, since it is the most restrictive and
+  _XOPEN_PATH_MAX does not seem to be available in all Linux distros
+
+The +1 accounts for the null char at the end of the name
+*/
+#ifdef __APPLE__
+#include <sys/posix_sem.h>
+#define SEM_MAX_LENGTH PSEMNAMLEN + 1
+#else
+#define SEM_MAX_LENGTH _POSIX_PATH_MAX + 1
+#endif
+
 void Phase_Init(Heap *heap, uint32_t initialBlockCount) {
     pid_t pid = getpid();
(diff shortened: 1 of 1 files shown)


nativelib/src/main/resources/scala-native/gc/commix/Phase.c
```diff
@@ -9,15 +9,34 @@
 #include <errno.h>
 #include <stdlib.h>
 
+/*
+If in OSX, sem_open cannot create a semaphore whose name is longer than
+PSEMNAMLEN characters. Else, we stick to POSIX compliance:
+
+  For POSIX.1-2017, XSI-conforming implementations shall set the
+  symbolic constant _XOPEN_UNIX to a value other than -1 and shall set the
+  symbolic constant _XOPEN_VERSION to the value 700.
```
ncordon, line 18: This is defined here: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap02.html

nativelib/src/main/resources/scala-native/gc/commix/Phase.c
```diff
@@ -9,15 +9,34 @@
...
-    char startWorkersName[workerNameSize];
-    char startMasterName[masterNameSize];
-    snprintf(startWorkersName, workerNameSize, "commix_mt_%d", pid);
-    snprintf(startMasterName, masterNameSize, "commix_wk_%d", pid);
+    // size = platform dependent max semaphore length +
+    //        null character at the end
+    char startWorkersName[SEM_MAX_LENGTH + 1];
+    char startMasterName[SEM_MAX_LENGTH + 1];
+    snprintf(startWorkersName, SEM_MAX_LENGTH + 1, "mt_%d_commix", pid);
+    snprintf(startMasterName, SEM_MAX_LENGTH + 1, "wk_%d_commix", pid);
```
ncordon, line 39: I've moved the variable part to the beginning of the names, in case they need to be truncated at the end

ekrich:
One weird thing I noticed. I used the following in Xcode to test:
```c
#include <semaphore.h>
```
But after your code was pushed I looked for `PSEMNAMLEN` in the above trying to follow more `#include`s etc. I couldn't find it but did find it in the one you used.
```c
#include <sys/posix_sem.h>
```

ekrich (approved):
LGTM. I can't reconcile the `#include` referenced in the comment but it is definitely where you say it is on my system.

WojciechMazur (approved):

PR #2236, opened 2021-04-01
WIP Re-enables some tests that were ignored for identity based maps

ncordon:
## What

Re-enables tests that were producing these warns (because they were disabled):

```
[warn] Test assumption in test java.util.IdentityHashMapTest.shouldRemoveStoredElementsInDoubleCornerCases failed: org.junit.internal.AssumptionViolatedException: JVM vs Native cache differences, took 0.001 sec
[warn] Test assumption in test java.util.IdentityHashMapTest.shouldStoreDoublesAlsoInCornerCases failed: org.junit.internal.AssumptionViolatedException: JVM vs Native cache differences, took 0.000 sec
[warn] Test assumption in test java.util.IdentityHashMapTest.shouldStoreIntegers failed: org.junit.internal.AssumptionViolatedException: JVM vs Native cache differences, took 0.000 sec
```

## Why

There seemed to be a comment that was never addressed, because those maps were in fact changed to scalajs similar implementations? Probably @ekrich can clarify that better?

Addressed #2233 partly.

### DIFF
--- a/unit-tests/src/test/scala/java/util/MapTest.scala
+++ b/unit-tests/src/test/scala/java/util/MapTest.scala
@@ -23,14 +23,9 @@ trait MapTest {
   // temporary until Platform is introduced
   val executingInJVM = false
 
-  // replace when new IdentityHashMap from Scala.js
-  // private def assumeNotIdentityHashMapOnJVM(): Unit =
-  //   assumeFalse("JVM vs Native cache differences",
-  //               executingInJVM && factory.isIdentityBased)
-
-  // temp implementation
   private def assumeNotIdentityHashMapOnJVM(): Unit =
-    assumeFalse("JVM vs Native cache differences", factory.isIdentityBased)
+    assumeFalse("JVM vs Native cache differences",
+                executingInJVM && factory.isIdentityBased)
 
   @Test def shouldStoreStrings(): Unit = {
     val mp = factory.empty[String, String]


ekrich:
The original version of `IdentityHashMap` was built for Scala Native. When implementing the map on Scala.js we used the built from scratch versions of `java.util` collections and the implementation is much better. The collections have not been ported to Scala Native so we can't use the Scala.js implementation or the tests. I remember the comment - maybe all the tests were ported but those tests just fail on Scala Native so we `assumeFalse` to avoid running until the collection classes are ported from Scala.js.

It looked like those tests were just warnings (reminders that we have future work)

That would be good project but we would need direction from @sjrd as to the base efficiency of those implementations if ported to Scala Native. The collections also use `java.util.ScalaOps` to avoid any dependency on the Scala library. Hopefully I got this all correct.

sjrd:
> That would be good project but we would need direction from @sjrd as to the base efficiency of those implementations if ported to Scala Native.

I believe the implementations don't rely on any performance aspects that are specific to JavaScript. So they should have good performance characteristics when ported to Scala Native.

ncordon:
I thought the tests could be in fact re-enabled, because the CI is in fact passing except when we disable the optimize and we run with `SCALANATIVE_OPTIMIZE=false`. 

I've come with a reproducible example of the problem for the sandbox:

```scala
object Test {
    val map = new java.util.IdentityHashMap[Double, String]
    map.put(1.2345, "test")
    println(map.get(1.2345))
  }
}
```

When we set `.withOptimize(false)` in the sandbox plugin configuration, `sandbox/run` prints a `null`. When we enable the optimization, we get the correct output by `sandbox/run`: `test`.

Another interesting bit is that when we get the key directly from the map, once created, and use it to retrieve the object, it works fine with or without optimization:

```scala
object Test {
    val map = new java.util.IdentityHashMap[Double, String]
    map.put(1.2345, "test")
    val key = map.keySet().toArray().apply(0)
    println(map.get(key))
  }
}
```

* What does the optimization do? 
* Could the no-optimization be boxing the Doubles somehow instead of considering it primitive types?

sjrd:
Ah no but wait. The fact that those tests are ignored is correct, and they should not be re enabled. They truly only make sense for non-identity maps. (They also make sense on JS, for obscure reasons.) They pass with optimizations because the optimizations can hide double boxing, but that is not at all a guarantee.

PR #2276, opened 2021-04-29
Allow for compilation of non-gc native sources on Windows

WojciechMazur:
This PR continues work done in direction of compiling Scala Native on Windows. It addresses all remaining native sources in clib and nativelib to be compilable on the Windows platform. It also adds compat layer to posixlib since most of this binding can be reused in Windows. 

* Added pre-processor conditions to ignore sources not compiled on Windows
* Created os specific folders in nativelib/platform
* Moved POSIX specific  `libunwind` directory and `process_monitor.c` to `platform/posix`
* Added Windows specific implementation of `unwind.h` based in WinDBG.lib instead of libunwind
* Replaced usages of some deprecated methods on Windows platform
* Added typedefintions and macros to for unsupported methods, constants and types in Windows
* Extended CI to test compilation of all native sources


.github/workflows/run-tests-windows.yml
```diff
@@ -45,7 +46,84 @@ jobs:
...
-          Get-ChildItem -Path $compiledDirs -Include *.c -Recurse -File `
+          Get-ChildItem -Path $compiledDirs -Filter *.c -Recurse -File `
           | ForEach-Object {clang -std=gnu11 -Wno-override-module $includes -c "$_"}
-          Get-ChildItem -Path $compiledDirs -Include *.cpp -Recurse -File `
-          | ForEach-Object {clang++ -std=c++14 -Wno-override-module $includes -c "$_"}
\ No newline at end of file
+          Get-ChildItem -Path $compiledDirs -Filter *.cpp -Recurse -File `
+          | ForEach-Object {clang++ -std=c++14 -Wno-override-module $includes -c "$_"}
+
+  compile-non-gc:
```
sjrd, line 54: Is there any reason why these are not simply more steps in the first job?

nativelib/src/main/resources/scala-native/time_millis.cpp
```diff
@@ -0,0 +1,44 @@
+#if defined(_WIN32)
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#else
+#include <stdio.h>
+#include <sys/time.h>
+#endif
+
+extern "C" {
+long long scalanative_current_time_millis() {
```
sjrd, line 10: Does this file really need to be C++?

sjrd (changes requested):
Only a few comments.

nativelib/src/main/resources/scala-native/time_millis.cpp
WojciechMazur: Good point, it probably can be regular C

.github/workflows/run-tests-windows.yml
WojciechMazur: First job defined matrix for different GCs, this way we compile another source only once.

sjrd (approved):

sjrd (approved):

PR #2281, opened 2021-05-03
Allow to compile simple programs on Windows

WojciechMazur:
This PR continues topic of Windows support. 
Based on #2206 it defines platform-dependent code paths needed to run simple programs, like `hello world`, which includes reimplementing `FileDescriptor` and `File{Input, Output}Stream`, as well as setting env variables and system properties at start of the program. 
New project `windowslib` containing bindings was created with minimal set of needed methods, at this moment some of them look like a good candidate to be included in other files, however, additional methods would be included in the future. Since most of the existing bindings come from `xapi.h` files in Scala Native they're included in classes following the given naming convention:
`XApi` - external method and type definitions, 
`XApiExt`- constants defined as preprocessor definitions, but safe enough to be replaced with Scala values. 
`XApiOps` - helper methods, implicit conversions/classes, 
All Constants and Enum values are having the same names as in the C header files.

* Added `wchar.h` bindings
* Added `{to, from}CWideString` methods needed for compatibility with several Windows API methods
* Bootstrapeed `windowslib` project and added minimal needed bindings to Win32 API
* Added `scala.scalanative.meta.LinktimeInfo` object containing linktime property `isWindows` used for conditional branching
* Adapted tests in `tools` to work with newly added linktime properties (they're always needed, as they're used in `java.lang.System`
* Refactored `CodeGen`, moved generic part to separate file, added OS (compiler) compat layer
* Forced using `c++14` standard and other specific flags on Windows (minimal supported) as well as LLVM 10 in the CI
* Output files targeting windows contain `.exe` extension
* Extended CI to run `sandbox/run` (HelloWorld)

This PR nees #2206 to actually run, results of  successful Windows CI containing mentioned branch can be found [here](https://github.com/WojciechMazur/scala-native/actions/runs/806960062)


ekrich:
@WojciechMazur I can extract `wchar` from this PR but it pulls a few other things along as well. It requires `CWideString` and the conversation methods and some other small things. If this is pretty close I could just review this instead.

After merging I could work on moving the C time stuff from POSIX for cleanup.

WojciechMazur:
@ekrich If have enough spare time then go ahead, thank you! 
I think we can leave conversion methods for separate PR. I believe CWideString was already defined inside `scalanative.unsafe` package object as 4-byte type. There was also defined CWideString (or CWString, I don't remember) inside `windowslib` since it uses only 2-byte wide chars.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -69,9 +66,8 @@ object CodeGen {
       // Clang's LTO is not available.
       def single(): Seq[Path] = {
         val sorted = assembly.sortBy(_.name.show)
-        Seq(
-          new Impl(targetTriple, env, sorted)
-            .gen(id = "out", workdir))
+        Impl(config, env, sorted)
+          .gen(id = "out", workdir) :: Nil
```
sjrd, line 70:
```suggestion
        Impl(config, env, sorted).gen(id = "out", workdir) :: Nil
```

tools/src/test/scala/scala/scalanative/linker/LinktimeConditionsSpec.scala
```diff
@@ -51,15 +51,16 @@ class LinktimeConditionsSpec extends LinkerSpec with Matchers {
                        |}""".stripMargin
 
   case class Entry[T](propertyName: String, value: T, lintimeValue: Val)
-
   val defaultEntries = Seq(
```
sjrd, line 54: This blank line should probably be kept.

windowslib/src/main/resources/scala-native/windows/console.c
```diff
@@ -0,0 +1,8 @@
+#if defined(_WIN32) || defined(WIN32)
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+
+DWORD scalanative_win32_console_std_in_handle() { return STD_INPUT_HANDLE; }
+DWORD scalanative_win32_console_std_out_handle() { return STD_OUTPUT_HANDLE; }
+DWORD scalanative_win32_console_std_err_handle() { return STD_ERROR_HANDLE; }
+#endif // defined(Win32)
```
sjrd, line 8: Missing EOL.

sjrd: Also in a number of files below.

windowslib/src/main/scala/scala/scalanative/windows/package.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package object windows {
+  type Word      = UShort //uint_16
+  type DWord     = UInt   //uint_32
+  type DWordLong = ULong  //uint_64
+  /* Actually large integers are union types with size of 64-bits
+    When compiler does not support large integer types it allows to store
+    low and hight parts of number in its structure fields. In all other cases
+    its recommended to use its QuadPart field directly.
+   */
```
sjrd, line 14:
```suggestion
  /* Actually large integers are union types with size of 64-bits
   * When the compiler does not support large integer types it allows to store
   * low and high parts of number in its structure fields. In all other cases
   * its recommended to use its QuadPart field directly.
   */
```

windowslib/src/main/scala/scala/scalanative/windows/package.scala
```diff
@@ -0,0 +1,33 @@
+package scala.scalanative
+
+import scala.scalanative.unsafe._
+import scala.scalanative.unsigned._
+
+package object windows {
+  type Word      = UShort //uint_16
```
sjrd, line 7: Do these definitions really make sense as public members of the `windows` package? Do they correspond to types that are defined in Windows APIs and used everywhere?

windowslib/src/main/scala/scala/scalanative/windows/util/HelperMethods.scala
```diff
@@ -0,0 +1,28 @@
+package scala.scalanative.windows.util
+
+import scala.scalanative.unsafe._
+import scala.scalanative.windows.HandleApi.Handle
+import scala.scalanative.windows.ProcessThreadsApi._
+import scala.scalanative.windows._
+
+object HelperMethods {
+  def withUserToken[T](desiredAccess: DWord)(fn: Handle => T): T = {
```
sjrd, line 9: Shouldn't this be private to some scope?

sjrd (changes requested):

clib/src/main/scala/scala/scalanative/libc/wchar.scala
```diff
@@ -0,0 +1,127 @@
...
+  type wint_t    = CInt
+  type wctype_t  = CInt
+  type wchar_t   = CWideChar
+  type WString   = CWideString
+  type mbstate_t = Byte
+  type mbstate   = Ptr[mbstate_t]
+
+  def btowc(c: Int): wint_t                                     = extern
+  def fgetwc(file: Ptr[FILE]): wint_t                           = extern
+  def fgetws(ws: WString, num: Int, stream: Ptr[FILE]): WString = extern
```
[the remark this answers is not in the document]
ekrich: I am suppose to be reworking the `wchar` API as a separate PR to make this PR smaller and there is some code in `posixlib` that should be in `clib` but we can rework after this is merged. It looks like they are suppose to be defined in `wchar` - https://en.cppreference.com/w/c/io - note they are list in the IO section. Strange indeed.

javalib/src/main/scala/java/io/FileDescriptor.scala
```diff
@@ -1,43 +1,145 @@
...
+  override def toString(): String =
+    s"FileDescriptor(${fd}, readonly=$readOnly"
+
+  /* Unix file descriptor underlying value */
+  @alwaysinline
+  private[java] def fd: Int = fileHandle.toInt
+
+  /* Windows file handler underlying value */
+  private[java] lazy val handle: Handle =
+    fromRawPtr[Byte](Intrinsics.castLongToRawPtr(fileHandle))
```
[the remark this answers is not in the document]
WojciechMazur: Mainly to cache conversion to Ptr, however `def` with `@alwaysinline` annotation should fine, especially since it's going to be optimized anyway. I've replaced it with `def`

nativelib/src/main/scala/scala/scalanative/unsafe/package.scala
```diff
@@ -213,6 +215,86 @@ package object unsafe {
...
+  /** Convert a java.lang.String to a CWideString using given UTF-16 LE charset.*/
+  @alwaysinline
+  def toCWideStringUTF16LE(str: String)(implicit z: Zone): Ptr[CChar16] = {
+    toCWideStringImpl(str, StandardCharsets.UTF_16LE, 2)
+      .asInstanceOf[Ptr[CChar16]]
+  }
+
+  private def toCWideStringImpl(str: String,
+                                charset: Charset,
+                                charSize: => CInt)(implicit z: Zone) = {
```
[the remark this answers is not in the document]
WojciechMazur: I've forgotten to remove that - now `WideCharSize` is calculated only once, in the past we needed to call the C helper function to get the correct size.

windowslib/src/main/scala/scala/scalanative/windows/util/HelperMethods.scala
WojciechMazur: Those helper methods are mostly used in javalib. Should I move it there and made it `private[java]`?

sjrd: Yes, please.

windowslib/src/main/scala/scala/scalanative/windows/package.scala
WojciechMazur: In my opinion, they do make sense. Windows API uses only its self-defined types everywhere, they're listed on [this page](https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types#pcwstr) 
Sticking to their naming convention allows us to make our API more consistent and eliminates possible type-based bugs. 
However, if placing them in `windows` package is the main concern, we might move them to some object, eg. `windows.WinDef` (since they're defined in  file`WinDef.h`)

sjrd: > In my opinion, they do make sense. Windows API uses only its self-defined types everywhere, they're listed on [this page](https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types#pcwstr)
> Sticking to their naming convention allows us to make our API more consistent and eliminates possible type-based bugs.

OK, sounds good.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -40,9 +40,11 @@ private[scalanative] object LLVM {
       if (isLl || !Files.exists(objPath)) {
         val compiler = if (isCpp) config.clangPP.abs else config.clang.abs
         val stdflag =
-          if (isLl) Seq()
-          else if (isCpp) Seq("-std=c++11")
-          else Seq("-std=gnu11")
+          if (isLl) None
+          else if (isCpp) {
+            if (config.targetsWindows) Some("-std=c++14")
```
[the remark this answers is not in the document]
WojciechMazur: C++14 or newer standard is needed to compile code using Windows API shipped with Windows 10 (used locally for development) or Server 2016 / 2019 used in the CI. 
I don't think we would be willing to support the older distributions, since their usage is now rather marginal.

As an example, header files included in our glue layer are using return type deduction (`auto`) which is not supported in std++11. There might have been other incompatibilities as well

javalib/src/main/scala/java/io/FileOutputStream.scala
```diff
@@ -64,15 +77,34 @@ class FileOutputStream(fd: FileDescriptor, file: Option[File] = None)
...
-      val mode  = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH
-      val fd    = open(toCString(file.getPath()), flags, mode)
-      if (fd == -1)
-        throw new FileNotFoundException(
-          s"$file (${fromCString(string.strerror(errno.errno))})")
-      else
-        new FileDescriptor(fd)
+      val fileName = toCString(file.getPath())
+      if (isWindows) {
+        val handle = CreateFileA(
```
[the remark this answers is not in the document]
WojciechMazur: I think it's the only usage of `xxxA` function. If I remember correctly for some reason it was failing with `CreateFileW`. I'll check if it's still the case.

tools/src/test/scala/scala/scalanative/linker/package.scala
```diff
@@ -0,0 +1,13 @@
+package scala.scalanative
+
+import scala.scalanative.nir._
+
+package object linker {
+  val linktimeInfoProperties = "scalanative.meta.linktimeinfo"
```
sjrd, line 6:
```suggestion
  val linktimeInfoProperties = "scala.scalanative.meta.linktimeinfo"
```
should fix the build errors.

clib/src/main/scala/scala/scalanative/libc/wchar.scala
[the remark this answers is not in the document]
WojciechMazur: I've limited it to the only needed method from wchar. The second one - `wcslen` is kept in `runtime.libc` in order to not create cyclic dependency. 
PR should be ready for another review cycle

ekrich:
I was wondering if you used `scala-native-bindgen` for the `wchar` or by hand?

WojciechMazur:
> I was wondering if you used `scala-native-bindgen` for the `wchar` or by hand?

Actually, at the moment when I was writing them, I've forgotten about the existence of the bindgen. I've rewrote them by hand + using regexpr

sjrd (approved):

PR #2287, opened 2021-05-06
Allow to run unit-tests on Windows

WojciechMazur:
This PR continues topic of Scala Native support for Windows. It adds platform-specific compact layer for `java.net` Sockets allowing to execute unit tests. It also adds proper implementation for required `java.io` methods. To limit the number of changes currently only smaller subset of tests `testsExt` was added to CI

* Included `testsExt` tests in the CI
* Added minimal `java.net` implementation
* Added binding for `time.h` not POSIX compliment `*_s`  methods used in Windows
* Added minimal `java.io.File` implementation required for tests
* Added bindings for required  WinSock2 and FileAPI functions
* Fixed parsing symbols in Stacktrace on Windows
* Added `-g` to compilation and linking step on Windows to allow gathering stack traces. 
* Fixed `java.util.Date` to compile on Windows

This PR needs to be rebased onto #2281 to actually work.


windowslib/src/main/resources/scala-native/windows/time.c
```diff
@@ -0,0 +1,133 @@
+// Windows specific copy of posixlib/time.h
+// Uses *_s variants of methods instead of *_r
+#if defined(_WIN32)
+#define _CRT_SECURE_NO_WARNINGS
```
ekrich, line 4: I really think we should consider just doing a `clib` time API - We should have virtually all the time stuff in `clib` and then just the posix specific in the `posixlib`.

Both time APIs can access the same C code and we could just deprecate most of the Scala in `posixlib`. You could then add just windows specific stuff in another interface as needed.

Do you think we should just address that in another PR and try it get it merged first?

.github/workflows/run-tests-windows.yml
```diff
@@ -121,11 +121,19 @@ jobs:
...
 
       - name: Run sandbox
-        # Set heap limits manually, there seams to be bug in detection of avaiable memory
-        # which results in undefined behavior at runtime
+        # Set heap limits manually, there seams to be bug in detection of available memory
+        # which results in undefined behaviour at runtime
         env:
-          SCALANATIVE_MIN_SIZE: 128M
-          SCALANATIVE_MAX_SIZE: 1G
+          SCALANATIVE_MAX_SIZE: 128M
```
sjrd, line 127: I guess all of this will disappear once rebased on top of #2293 ?

javalib/src/main/scala/java/io/File.scala
```diff
@@ -318,7 +372,10 @@ class File(_path: String) extends Serializable with Comparable[File] {
 
   // Ported from Apache Harmony
   def toURI(): URI = {
-    val path = getAbsolutePath()
+    val path = getAbsolutePath().map {
+      case '\\' => '/'
+      case c    => c
+    }
```
sjrd, line 378:
```suggestion
    val path = getAbsolutePath().replace('\\', '/')
```

javalib/src/main/scala/java/io/FileOutputStream.scala
```diff
@@ -16,6 +16,7 @@ import scala.scalanative.windows.winnt.AccessRights._
 
 class FileOutputStream(fd: FileDescriptor, file: Option[File] = None)
```
sjrd, line 17: The default parameter is now redundant with the overload, and should be removed.

javalib/src/main/scala/java/lang/StackTraceElement.scala
```diff
@@ -50,7 +51,7 @@ private[lang] object StackTraceElement {
     var methodName = ""
 
     def readSymbol(): Boolean = {
-      if (read() != '_') {
+      if (!isWindows() && read() != '_') {
```
sjrd, line 54: This definitely deserves a comment.

javalib/src/main/scala/java/net/AbstractPlainSocketImpl.scala
```diff
@@ -318,14 +218,14 @@ private[net] class PlainSocketImpl extends SocketImpl {
   }
 
   override def close(): Unit = {
-    if (fd.fd != -1) {
-      cClose(fd.fd)
-      fd = new FileDescriptor
+    if (!isClosed) {
+      fd.close()
+      fd = InvalidSocketDescriptor
     }
```
sjrd, line 224: It looks like this should also do
```scala
isClosed = true
```

javalib/src/main/scala/java/net/AbstractPlainSocketImpl.scala
```diff
@@ -553,5 +454,14 @@ private[net] class PlainSocketImpl extends SocketImpl {
           + optID + ", errno: " + errno.errno)
     }
   }
+}
 
+object AbstractPlainSocketImpl {
```
sjrd, line 459:
```suggestion
private[net] object AbstractPlainSocketImpl {
```

windowslib/src/main/resources/scala-native/windows/time.c
```diff
@@ -0,0 +1,129 @@
...
+char **scalanative_tzname() { return tzname; }
+
+long scalanative_timezone() { return _timezone; }
+
+int scalanative_daylight() { return _daylight; }
+
+// Windows compat
+void tzset() { return _tzset(); }
+
+#endif // defined(_WIN32)
```
sjrd, line 129: Missing EOL

windowslib/src/main/resources/scala-native/windows/winSocket.c
```diff
@@ -0,0 +1,14 @@
...
+
+#pragma comment(lib, "Ws2_32.lib")
+
+DWORD scalanative_winsock_wsadata_size() { return sizeof(WSADATA); }
+
+DWORD scalanative_winsocket_fionbio() { return FIONBIO; }
+
+SOCKET scalanative_winsock_invalid_socket() { return INVALID_SOCKET; }
+
+#endif // defined(_WIN32)
```
sjrd, line 14: Missing EOL

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -52,8 +52,9 @@ private[scalanative] object LLVM {
           if (config.targetsWindows) Seq("-g")
           else Nil
         }
-        val flags = opt(config) +: stdflag ++: platformFlags ++: "-fvisibility=hidden" +:
-          config.compileOptions
+        val flags =
+          opt(config) +: stdflag ++: platformFlags ++: "-fvisibility=hidden" +:
```
ekrich, line 56: You could probably avoid the extra line here by wrapping the "-fvisibility=hidden" +: onto the next line so the two lines are more equal and the first line won't wrap at 80.

windowslib/src/main/scala/scala/scalanative/windows/util/Conversion.scala
```diff
@@ -0,0 +1,16 @@
+package scala.scalanative.windows.util
+
+import scala.scalanative.unsigned._
+import scala.scalanative.windows.{Word => WinWord}
+
+object Conversion {
+  def wordToBytes(word: WinWord): (Byte, Byte) = {
+    val lowByte  = (word & 0xFF.toUShort).toByte
+    val highByte = (word >> 8 & 0xFF.toUShort).toByte
```
sjrd, line 9:
```suggestion
    val lowByte  = (word & 0xFF.toUShort).toByte
    val highByte = ((word >> 8) & 0xFF.toUShort).toByte
```
or even
```suggestion
    val lowByte  = (word.toInt & 0xFF).toByte
    val highByte = ((word.toInt >> 8) & 0xFF).toByte
```

windowslib/src/main/scala/scala/scalanative/windows/util/Conversion.scala
```diff
@@ -0,0 +1,16 @@
+package scala.scalanative.windows.util
+
+import scala.scalanative.unsigned._
+import scala.scalanative.windows.{Word => WinWord}
+
+object Conversion {
```
sjrd, line 6:
```suggestion
private[windows] object Conversion {
```
?

windowslib/src/main/scala/scala/scalanative/windows/util/HelperMethods.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.IOException
+import scala.scalanative.unsafe._
+import scala.scalanative.windows.HandleApi._
+import scala.scalanative.windows.HandleApiExt._
+import scala.scalanative.windows.FileApiExt._
+import scala.scalanative.windows.ProcessThreadsApi._
+import scala.scalanative.windows._
+
+object HelperMethods {
```
sjrd, line 11:
```suggestion
private[windows] object HelperMethods {
```

sjrd (changes requested):

.github/workflows/run-tests-windows.yml
WojciechMazur: Yes, that's correct. I've create that fix separately so it can be applied at any time (eg. before merging PR with CodeGen support for Windows)

windowslib/src/main/scala/scala/scalanative/windows/util/HelperMethods.scala
WojciechMazur: Actually, this should be moved to javalib and made `private[java]` as discussed in the previous PR. It was not moved after rebasing.

sjrd (approved):

Issue #893, opened 2017-07-03
[windows support] setting a garbage collector currently causes compiled programs to crash on startup


Keshi:
Garbage collection doesn't seem to be working on windows at the moment.  Empty programs built with scala native crash with "Could not create file mapping object (1455)" printed to console at some point upon start up if nativeGC:="immix" or nativeGC:="boehm" is selected.   MSDN says that 1455 is ERROR_COMMITMENT_LIMIT, which means "The paging file is too small for this operation to complete".  
However, if built with nativeGC:="none", things run fine.

muxanick:
what windows version do you use?

Keshi:
Oh, this is quite cool!  I'm able to run this in the visual studio debugger and see the allocator source. :)  

I'm using windows 7.  It seems to be trying to mmap my entire currently allocated virtual memory (16GB at the time, system managed), which just fails.  However, if I do some shenanigans (adding another 16gb paging file on a different drive, then removing it again, to result in 32gb allocated) to expand the allocated virtual memory in windows, while in a breakpoint after Heap_getMemoryLimit(), then allocating the large heap seems to succeed.    It seems that we can't ask windows to mmap the heap memory limit?

muxanick:
It works for me in windows 10, I can ask for 16Gb twice, so seems Windows 7 has different behavior with that.

Keshi:
The call to CreateFileMappingW seems to commit the entire amount right away.  I guess the difference is that I have 12GB of commit already from other programs.  The first call brings it to 26/32 Commit GB, and the second call fails.

I think you're right about the different behaviour.  According to https://blogs.msdn.microsoft.com/oldnewthing/20150130-00/?p=44793, it is "significantly improved in Windows 8.1, which defers committing the page tables until you actually use them."

muxanick:
SEC_RESERVE is default I think on windows 10, but we can try to explicitly use it.
But you are right, GC shouln't ask for all available memory, it should have start up settings for that value

Keshi:
According to https://blogs.technet.microsoft.com/markrussinovich/2008/11/17/pushing-the-limits-of-windows-virtual-memory/, it seems in at least one version of Windows, the behaviour of Window's commit limit expansion (when in system managed mode) is to expand it over time when the commit limit is approached.  I guess this explains why requesting so large a commit fails, as it doesn't approach the limit (just zooms past it in one go), and Windows will only succeed in raising the limit some time after the failure so it just rejects first.

[...]

PR #2293, opened 2021-05-13
Incrementally commit memory in GC on Windows

WojciechMazur:
PR #2264 has added support for comping GCs on Windows, however, it contained few bugs. One of them was caused by invalid memory allocation, which behaves differently when compared with Unix. 
In case if Immix/Commix heap is not limited (no ScalaNative_Max_Size env variable) it tries to map memory chunk equal to all physical memory available. On Unix, we can use `mmap` with total size of physical memory and the system will handle the actual handling of committing memory until there is no more left, which would lead to OutOfMemory error
On Windows however we need to make sure that the allocated chunk of memory is less or equal to available memory, otherwise, it would fail at call site (in our case at GC init). This behaviour was described in issue #893. Also in case if we would use all actually available memory resource we would have very little change to start any other process, eg. TestSuite worker node, as there would be no more available memory left. 

To handle this issue strategy for allocating memory was changed for Windows. `memoryMap` helper function now does not automatically commit a given chunk of memory, instead it needs to be explicitly committed. Thanks to this change we're able to commit only a minimal amount of memory at startup, and commit additional chunks every time it is needed (on-call to Heap_grow in Immix/Commix)
Additionally, a method used to allocate memory was changed - instead of the usage of `CreateFileMapping`, we're using now `VirtualAlloc` which is more suitable for this task, and potentially faster.

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
+++ b/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
@@ -25,8 +25,7 @@
 #endif
 
 // Map private anonymous memory, and prevent from reserving swap
-#define HEAP_MEM_FLAGS
-(MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS)
+#define HEAP_MEM_FLAGS (MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS)
 #define HEAP_MEM_FLAGS_PREALLOC                                                \
     (MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS | MAP_POPULATE)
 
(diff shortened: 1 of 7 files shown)


nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
```diff
@@ -4,45 +4,31 @@
...
-#define HEAP_MEM_FLAGS_PREALLOC (MAP_PRIVATE | MAP_ANONYMOUS)
-#else
-#define HEAP_MEM_FLAGS_PREALLOC (MAP_PRIVATE | MAP_ANONYMOUS | MAP_POPULATE)
+#define MAP_POPULATE 0
 #endif
 
-#endif
+// Map private anonymous memory, and prevent from reserving swap
+#define HEAP_MEM_FLAGS
+(MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS)
```
sjrd, line 29:
```suggestion
    (MAP_NORESERVE | MAP_PRIVATE | MAP_ANONYMOUS)
```

nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
```diff
@@ -51,66 +37,44 @@
...
-        PAGE_READWRITE,       // read/write access
-        memSize.u.HighPart,   // maximum object size (high-order DWORD)
-        memSize.u.LowPart,    // maximum object size (low-order DWORD)
-        NULL);                // name of mapping object
-
-    if (hMapFile == NULL) {
-        return NULL;
+    if (!doPrealloc) {
+        return memoryMap(memorySize);
     }
```
sjrd, line 54: It seems this could be moved to the `#else // Unix` branch, no?

nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
```diff
@@ -51,66 +37,44 @@
...
 }
+
+bool memoryCommit(void *ref, size_t memorySize) {
+#ifdef _WIN32
+    return VirtualAlloc(ref, memorySize, MEM_COMMIT, PAGE_READWRITE) != NULL;
+#else
+    // No need for committing on UNIX
+    return true;
+#endif
+}
```
sjrd, line 80: Missing EOL.

sjrd (approved):
Only a few comments, but otherwise LGTM.

PR #2294, opened 2021-05-13
Allow to use CommixGC on Windows

WojciechMazur:
Merged PR #2264 allowed for compilation of GCs on Windows however it contained few bugs left. One of them was a problem with an endless loop inside CommixGC. 
It was caused by the difference in sizes of some structs, eg. `GreyPacketsRef` between Unix and Windows and the assumption that this struct should be equal to exactly 8-bytes. Such difference was caused by not portable enough code - MSVC does not support bitfields inside packed structs as expected - it seems that on the whole size of the variable is counted into the size of the struct, or some additional alignment was added.

To handle this issue, usages of 24-bit fields based on `uint32_t` were replaced with a dedicated `UInt24` struct which should fully portable across platforms. 
Additionally, there were added compile-time assertions guarding for the correct size of memory-layout fragile structs.

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.c
+++ b/nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.c
@@ -34,7 +34,7 @@ void GreyList_Init(GreyList *list) {
     list->head.atom = GreyPacketRef_Empty();
 }
 
-uint32_t GreyList_Size(GreyList *list) {
+UInt24 GreyList_Size(GreyList *list) {
     GreyPacketRef head;
     head.atom = list->head.atom;
     return head.sep.size;
@@ -42,7 +42,7 @@ uint32_t GreyList_Size(GreyList *list) {
 
 void GreyList_Push(GreyList *list, word_t *greyPacketsStart,
                    GreyPacket *packet) {
-    uint32_t packetIdx = GreyPacket_IndexOf(greyPacketsStart, packet);
+    UInt24 packetIdx = GreyPacket_IndexOf(greyPacketsStart, packet);
     GreyPacketRef newHead;
     newHead.sep.idx = packetIdx;
     newHead.sep.timesPoped = (uint16_t)packet->timesPoped;
@@ -51,9 +51,9 @@ void GreyList_Push(GreyList *list, word_t *greyPacketsStart,
     do {
         // head will be replaced with actual value if
         // atomic_compare_exchange_strong fails
-        newHead.sep.size = head.sep.size + 1;
-        uint32_t nextIdx = head.sep.idx;
-        if (nextIdx == GREYLIST_LAST) {
+        newHead.sep.size = UInt24_plus(head.sep.size, 1);
+        UInt24 nextIdx = head.sep.idx;
+        if (UInt24_equals(nextIdx, GREYLIST_LAST)) {
             packet->next.atom = GreyPacketRef_Empty();
         } else {
             packet->next.atom = head.atom;
@@ -64,7 +64,7 @@ void GreyList_Push(GreyList *list, word_t *greyPacketsStart,
 
 void GreyList_PushAll(GreyList *list, word_t *greyPacketsStart,
                       GreyPacket *first, uint_fast32_t size) {
-    uint32_t packetIdx = GreyPacket_IndexOf(greyPacketsStart, first);
+    UInt24 packetIdx = GreyPacket_IndexOf(greyPacketsStart, first);
     GreyPacketRef newHead;
     newHead.sep.idx = packetIdx;
     newHead.sep.timesPoped = (uint16_t)first->timesPoped;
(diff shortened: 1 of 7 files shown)


nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.h
```diff
@@ -9,25 +9,32 @@
...
+#include "UInt24.h"
 
 typedef Object *Stack_Type;
 
-// No support for 3-byte alignment in MSVC
+// UInt24 used instead of uint_32 bitset for cross-platform compatiblity, mainly
+// due the lack of support for 3-byte alignment in MSVC
 // https://docs.microsoft.com/en-us/cpp/c-language/padding-and-alignment-of-structure-members?redirectedfrom=MSDN&view=msvc-160
-// Would result in size of 16 bytes on Windows and 8 on Unix
+// It would would result in size of 16 bytes on Windows and 8 on Unix
```
ekrich, line 19: Looks like you have an extra `would` here.

nativelib/src/main/resources/scala-native/gc/immix_commix/UInt24.h
```diff
@@ -0,0 +1,30 @@
...
+static inline UInt24 UInt24_plus(UInt24 value, uint32_t arg) {
+    uint32_t v = UInt24_toUInt32(value);
+    return UInt24_fromUInt32(v + arg);
+}
+
+static inline bool UInt24_equals(UInt24 v1, UInt24 v2) {
+    return UInt24_toUInt32(v1) == UInt24_toUInt32(v2);
+}
+
+#endif // UINT24_H
```
sjrd, line 30: Missing EOL

nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.c
```diff
@@ -105,8 +105,8 @@ GreyPacket *GreyList_Pop(GreyList *list, word_t *greyPacketsStart) {
         if (next.atom != 0) {
             nextValue.atom = next.atom;
         } else {
-            nextValue.sep.idx = headIdx + 1;
-            nextValue.sep.size = head.sep.size - 1;
+            nextValue.sep.idx = UInt24_plus(headIdx, 1);
+            nextValue.sep.size = UInt24_plus(head.sep.size, -1);
```
sjrd, line 109: `UInt24_plus` takes a `uint32_t` as second argument, but here we pass a negative number. Should be an `int32_t` instead?

nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.c
```diff
@@ -34,15 +34,15 @@ void GreyList_Init(GreyList *list) {
     list->head.atom = GreyPacketRef_Empty();
 }
 
-uint32_t GreyList_Size(GreyList *list) {
+UInt24 GreyList_Size(GreyList *list) {
```
sjrd, line 37: It seems all the callers of `GreyList_Size` immediately convert it to `uint32_t`. I suggest keeping the result type of `uint32_t` here, and doing the conversion in this function.

nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.h
```diff
@@ -54,30 +61,31 @@ bool GreyPacket_IsEmpty(GreyPacket *packet);
...
+    uint32_t idxValue = UInt24_toUInt32(idx);
+    assert(idxValue >= 2);
+    return (GreyPacket *)greyPacketsStart + (idxValue - 2);
 }
 
 static inline uint64_t GreyPacketRef_Empty() {
     GreyPacketRef initial;
     initial.sep.idx = GREYLIST_LAST;
-    initial.sep.size = 0;
+    initial.sep.size = (UInt24){0};
```
sjrd, line 88: Is there any reason not to do
```suggestion
    initial.sep.size = UInt24_fromUInt32(0);
```
here?

sjrd (changes requested):

nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.c
WojciechMazur: Yes, you're right. I've fixed that.

nativelib/src/main/resources/scala-native/gc/commix/datastructures/GreyPacket.h
WojciechMazur: Not really,  only possible possive value from that might be replacing method call with constant literal. However, after all we have optimizers and this method Is probably not called very often. 
Switched to your proposal as it's more readable and less error prone.

ScalaWilliam:
Just wondering if there's anything else for this PR? Very keen to try it out in a release :-)

sjrd (approved):
This should be added to the CI script. Otherwise LGTM.

PR #2298, opened 2021-05-17
Remove link to pthread_create missing C helper function

ekrich:
As reported by @errikos, `pthread_create` didn't have a helper method causing a link error. The POSIX function should work without helper functions.


WojciechMazur (approved):

errikos:
Sorry to be late to the party, but this should also be done for `pthread_join` (probably even more functions).

ekrich:
@errikos I didn't look closely at all the code, but cleaning this up is much appreciated. @LeeTibbert has pointed out that some of the code that returns `#define` constants should just return the constant rather than having an intermediate value.

PR #2304, opened 2021-05-18
Respect doPrealloc if we are not on linux.

teodimoff:
i have made small fix and a test for the gc preallocation:
https://github.com/teodimoff/scala-native/tree/gc-prealloc-test/staging/gc-prealloc
My first question is where the test should go? If in scripted test i will need some help with the yml file. The test need to run two times(be it either the result binary or sbt run). Before the second run we need to export the export GC_INITIAL_HEAP_SIZE=700M to see the difference. Also the results is printing the execution time for both runs if preallocation works the second the will take less time. Is this enough or we want to actually compare both execution times to see if the second is actually less? If so how?

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
+++ b/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
@@ -105,8 +105,10 @@ word_t *memoryMapPrealloc(size_t memorySize, size_t doPrealloc) {
 
 #ifndef __linux__
     // if we are not on linux the next best thing we can do is to mark the pages
-    // as MADV_WILLNEED.
-    madvise(res, memorySize, MADV_WILLNEED);
+    // as MADV_WILLNEED but only if doPrealloc is enabled.
+    if (doPrealloc) {
+        madvise(res, memorySize, MADV_WILLNEED);
+    }
 #endif
 
     return res;


WojciechMazur (approved):

WojciechMazur:
We should compare actual execution times, as long as we're sure that it always should be faster. 

When it comes to test implementation it's a much more difficult topic. 
I don't think we're able to modify env variables in scripted tests, I'm not sure the `exec` command from [the support syntax list](https://www.scala-sbt.org/1.x/docs/Testing-sbt-plugins.html#step+4%3A+write+a+script) would allow for that. 
We can however run another process with custom env from inside of ScalaNative. Maybe we can run the same program with different execution paths based on arguments. We can start main program, which starts sequentially two additional processes (using the same executable), one with pre-alloc, another without. Both `worker` nodes would run the same code. The manager process might collect execution time for both programs and compare them, if the values are not expected we would fail there. 
Since we're running scripted tests only in Scala 2.12 we can expect that path to executable would be always the same, that is `target/scala-2,12/<test-name>-out`


#Off topic
It would be best to not add any additional tests into the scripted tests unless there is no other way since they're already taking a lot of time. In my opinion, scripted tests should be mainly limited to testing sbt plugin, and not underlying GC/runtime implementation. 
Unfortunately, we do not have any better infrastructure suitable for such a test right now. Inside unit tests, we currently can run other processes, but would not be able to build a testing program from inside of that module. 
A single project where we can build and run some code, from inside of tests is `tools`, e.g. based on `ReachabilitySuite` trait we can check that given code can be linked using ScalaNative. In the same manner, it could be extended to additionally build executable used for further testing. However, I don't think that GC tests should be placed there. I'd rather propose an additional project `runtime-tests` based on `tools` tests concepts. I believe many of the scripted tests could be ported to such a new framework. 
At this moment I don't think we would have enough time to prepare the new framework mentioned before, so I guess the best place for your test would be `scripted-tests`, at least for now.

teodimoff:
Hi Wojciech, thanks for the pointers. I Will poke here and there to see what should be done. The start two new process with ./gc-prealloc-out seems a good development path. Currently we ran the scripted tests with gc immix. How can i(or can i) tell it to run gc-prealloc test with none gc? Yesterday i skimmed through the yml file i think i got the basic idea. Will this be a separe PR? If so , i will open another PR to ask questions and move forward when i have some progress.

teodimoff:
Okay lets merge this, because i want this finished and i will proceed with another PR for the test, because i may delay some time.

PR #2306, opened 2021-05-20
Change C compiler standard from gnu11 to c11

ekrich:
Please include a release note.

Currently, Scala Native supports clang 6 and uses the `-std=c++11` and `-std=c11` standards while compiling C++ and C respectively. So Scala Native can potentially support more platforms this is quite conservative. This change is because the initial selection of `gnu11` supports GNU extensions beyond `c11` so it is potentially not 100% C standard compliant. See https://gcc.gnu.org/onlinedocs/gcc/Standards.html

>By default, GCC provides some extensions to the C language that, on rare occasions conflict with the C standard. See Extensions to the C Language Family. Some features that are part of the C99 standard are accepted as extensions in C90 mode, and some features that are part of the C11 standard are accepted as extensions in C90 and C99 modes. Use of the -std options listed above disables these extensions where they conflict with the C standard version selected. You may also select an extended version of the C language explicitly with -std=gnu90 (for C90 with GNU extensions), -std=gnu99 (for C99 with GNU extensions) or -std=gnu11 (for C11 with GNU extensions).

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
+++ b/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
@@ -7,7 +7,7 @@
 #else // Unix
 // For Linux
 // https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html
-#define _BSD_SOURCE
+#define _XOPEN_SOURCE 700
 #include <sys/mman.h>
 // Darwin defines MAP_ANON instead of MAP_ANONYMOUS
 #if !defined(MAP_ANONYMOUS) && defined(MAP_ANON)
--- a/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
+++ b/nativelib/src/main/resources/scala-native/gc/shared/MemoryMap.c
@@ -5,6 +5,9 @@
 #define NOMINMAX
 #include <windows.h>
 #else // Unix
+// For Linux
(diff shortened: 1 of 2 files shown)


ekrich:
See https://github.com/scala-native/scala-native/pull/1273

ekrich:
Guess I will have to try this on a Linux install Ubuntu 18.04 would be a good choice. Looking at this below it seems weird the macOS works ok.

See https://man7.org/linux/man-pages/man2/mmap.2.html
>Certain flags constants are defined only if suitable feature test
       macros are defined (possibly by default): _DEFAULT_SOURCE with
       glibc 2.19 or later; or _BSD_SOURCE or _SVID_SOURCE in glibc 2.19
       and earlier.  (Employing _GNU_SOURCE also suffices, and requiring
       that macro specifically would have been more logical, since these
       flags are all Linux-specific.)  The relevant flags are:
       MAP_32BIT, MAP_ANONYMOUS (and the synonym MAP_ANON),
       MAP_DENYWRITE, MAP_EXECUTABLE, MAP_FILE, MAP_GROWSDOWN,
       MAP_HUGETLB, MAP_LOCKED, MAP_NONBLOCK, MAP_NORESERVE,
       MAP_POPULATE, and MAP_STACK.

ekrich:
@LeeTibbert if you could take this over and add a commit or two, that would be great.

PR #2318, opened 2021-05-31
In sandbox, put Test.scala in src/main/scala

errikos:
Some IDEs (ehm, IntelliJ) do not provide good support for files being outside the standard directories.

In the `sandbox` module, move `Test.scala` to `src/main/scala`, so that it is indexed and all the intelligent IDE features work.

This also allows producing more complex programs in `sandbox` more easily.


ekrich:
@densh Was there any reason for this historically? Can it be moved now?

WojciechMazur (approved):

PR #2323, opened 2021-06-14
Add support for @volatile annotation

WojciechMazur:
Blocked by #2327 

This PR adds proper support for `@volatile` annotation to match behaviour known from the JVM. For each store/load operation on fields containing this annotation, there would be generated `<op> atomic volatile` using sequential consistent memory order (in the manner as it's done on the JVM). 

Since `Boolean` is represented as `i1` it needed additional handling to allow for atomic operations - atomic operation would be executed on `i8` type, while actual value would be truncated to zero-extended to match type expected by the LLVM. 

This feature also creates changes in the NIR format by introducing a new `Defn` attribute (`Attr.Volatile`) and changing the `Op.Load`/`Op.Store` object by adding the `isAtomic` boolean field. Due to this change, the NIR version was bumped from 5.8 to 5.9.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
@@ -626,7 +626,7 @@ private[codegen] abstract class AbstractCodeGen(
         }
         genCall(genBind, callDef, unwind)
 
-      // Special case of load needing additional converstion, since atomic operations do not support i1
+      // Special case of load needing additional conversion, since atomic operations do not support i1
       case Op.Load(Type.Bool, ptr, true) =>
         val pointee = fresh()
         val loaded  = fresh()
@@ -694,7 +694,7 @@ private[codegen] abstract class AbstractCodeGen(
           }
         }
 
-      // Special case of store needing additional converstion, since atomic operations do not support i1
+      // Special case of store needing additional conversion, since atomic operations do not support i1
       case Op.Store(Type.Bool, ptr, value, true) =>
         val pointee  = fresh()
         val extended = fresh()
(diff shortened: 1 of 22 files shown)


WojciechMazur:
This change does not include any new special behaviour for method local variables (handled in the NIR as `Type.Var` and accessed using `Op.VarLoad`/`Op.VarStore` high-level operations.) 
Is there any use-case where it should be needed (since they do never leave the scope)? 
The only case I can think of is creating a new thread from the inside of the method, but in such case, the field should be translated into the class field (capture field) of the underlying Lambda-class (which at some point would be needed to create `Runnable`)

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
```diff
@@ -626,7 +626,32 @@ private[codegen] abstract class AbstractCodeGen(
         }
         genCall(genBind, callDef, unwind)
 
-      case Op.Load(ty, ptr) =>
+      // Special case of load needing additional converstion, since atomic operations do not support i1
```
ekrich, line 629: converstion -> conversion

lolgab:
@WojciechMazur Shouldn't this feature be put under a feature-flag related to multithreading?
I'm interested in using SN in a single threaded fashion without paying the cost of the `@volatile` annotations that the stdlib puts for thread safety purposes. It is especially interesting now that multithreading is not yet supported.

WojciechMazur:
@lolgab Good point, we might actually create a switch in NativeConfig to enable experimental multithreading features, currently it would need to be explicitly enabled, but it might change in the future when it would be stable enough.
 This will allow us to decide should given load/store operation be atomic or not, since it's happening at generating LLVM IR. At the same time, we would be ready for multithreaded environments since information about `@volatile` annotation would be already present in the NIR.

PR #2325, opened 2021-06-17
Introduce a build plugin

ekrich:
In a recent PR, we improved the build to build each native project individually, gathering up all the `.o` files for linking.

See _Refactor build to compile each project and ll separately_: https://github.com/scala-native/scala-native/pull/2234

This isolated the special code needed to compile the `nativelib` which used to have a marker file but was changed to be a `scala-native.properties` file. Currently this special code, is used to select only the code needed for the current GC and the `optional` dir that has `z.c` that links with the optionally installed `zlib` used to support the `java.util.{jar, zip}` code.

By adding the following to the `scala-native.properties`, `plugin=scala.scalanative.build.nativelib.FilterPlugin` we can use basic reflection to invoke the plugin.

Benefits:
1. Allow us to move the GC code to its own project and allow it to do its own filtering.
2. Move the `optional` code to the `javalib` and let it handle its own filtering.
3. All this code gets removed from the nativelib.

Downsides:
1. Reflection code
2. Additional artifact if number 1 above is done.

Future ideas:
1. Cross compile the tools code so the plugins could be moved to the respective library.
2. Add features like `nativeCompilerOptions` and `nativeLinkingOptions` or other `NativeConfig` options to be supplied by this mechanism in the `scala-native.properties` file. Recently, we experimented with trying to use `-std=c11` instead of `gnu11` and this failed miserably because clearly the `posixlib` and maybe other libs need special defines like `-D_POSIX_C_SOURCE` or other project specific options to compile like provided by `gnu11`. Clearly that is not a great option for windows.
3. Invoke the native build as an executable like we do `clang` so we can do number 1 above.
4. This ability to have a plugin could be extended to third party libs because the basic build can only supply basic options and we have no idea of special flags or options that may be needed.
5. We initially started down this path of this PR and the previous so we could filter code for the Windows support but I am not sure we need it any more but there are some simple rules that could be implemented if needed. See: https://github.com/scala-native/scala-native/pull/2230

If we can find a design that would work I would like to move forward maybe in a series of PRs. We would need to come up with a bootstrap plan because tools would be needed be compiled as JVM initially I would guess but it is hard for me to keep this all straight. We would also need to supply a shim or some way to use `portable-scala-reflect` on the JVM side.


sjrd:
Use of reflection has a number of downsides. Why not use a factory function in the Config instead?

It seems the functionality of this "plugin" is very specific, given its extremely vague name `Plugin`. It also seems to significantly constrain the kinds of changes we could do in the future to this part of the code.

Quoting the advantages that you write:

> * Allow us to move the GC code to its own project and allow it to do its own filtering.
> * Move the `optional` code to the `javalib` and let it handle its own filtering.
> * All this code gets removed from the nativelib.

It seems to me that we can get all of that just by decomposing the GC and the java.util.zip handling in separate jars, that would contain their appropriate .c code. Then we would select those jars when appropriate, and not do any filtering after that.

ekrich:
Let me step back for a second before I answer the specific comments.

Originally, we compiled all the native libs at once except LL. @WojciechMazur in one of his Windows PR was using filtering. I knew this wasn't going to scale as any library could perhaps have a `windows` directory so we could do things we weren't expecting to third party libs. The build was improved to compile each native lib so that the only special lib was `nativelib` that needed filtering. We replaced the marker file with `scala-native.properties` file thinking we could then use that for future configuration or filtering purposes.

On a second front we wanted shared code for the GCs so additional items where added to the GC objects so they could be inspected to select the correct paths slightly altering the current `nativelib` filter.

On further inspection of both the GC filtering and the Java ZIP filtering made me think, simple filter rules wouldn't work in a properties file. We could do `windows:include=...` or `unix:exclude=...` but not things that require looking into generated code like the Java ZIP thing or the Config based things like GC that are dynamic. Thinking about it we could perhaps do something for the GC, but it would probably take way more code than the direct "filter" code we have whether in a plugin or something else.

As I said in the initial comment - it may be a good idea to allow libraries to add options or filter or even have a plugin that is needed for their library. Any thing added to config applies across the entire build so having `scala-native.properties` might be a really good thing for library authors to add compiler options. I like to consider the future in design but not really code for it as we tend to be wrong so it is wasted code and complexity. Right now I would like to split the GC and put the native code for `java.util.zip` in the `javalib`.

>Use of reflection has a number of downsides. Why not use a factory function in the Config instead? 

I had not thought of this but conceptually understand how this would work. I wanted to have the code in the library which is a native lib and the "plugin" `trait` in the `nativelib` so if we decided to extend this to user libs we could do so but this requires that the `tools` or a portion of it be a native lib. But I think this would require reflection and we would have to be darn sure so overall I think a factory is a good idea. Going the whole bootstrap and native tools is probably too much especially for now.

>It seems the functionality of this "plugin" is very specific, given its extremely vague name Plugin. It also seems to significantly constrain the kinds of changes we could do in the future to this part of the code.

I agree. I think as long as the code is internal we could certainly change things however we see fit. If we wanted to use the factory method we could just put `factory=gc` or `factory=javazip` or `factory=windows` ... in the properties file or something like that. We could leave the code in tools or whatever you think is best.

>It seems to me that we can get all of that just by decomposing the GC and the java.util.zip handling in separate jars, that would contain their appropriate .c code. Then we would select those jars when appropriate, and not do any filtering after that.

For the GC, I am not sure how we would do that without some code to select the code needed and delete the previous as it currently does. For the java.util.zip, a separate jar would be possible but we would also have to make that an optional jar so the user would have to add that in their build - that would solve it if we wanted to go that route. The current method works pretty well but maybe the error message is obscure if you didn't install `libz`.

Hope this helps and maybe @WojciechMazur can weigh in because perhaps he has found reasons we need or don't need something like this. I think for now a simple factory like `BuildFilterFactory` with a `BuildFilter` being like the current `Plugin` and putting that in the tools would work well.

ekrich:
Tested code movement in Sandbox. Didn't want to add scripted.
```Scala
import java.io._
import java.util.zip._

object Test {
    println("Hello, World!")
    println(System.getProperty("user.dir"))
    val sourceFile = "sandbox/src/main/scala/test1.txt"
    val fos        = new FileOutputStream("compressed.zip")
    val zipOut     = new ZipOutputStream(fos)
    val fileToZip  = new File(sourceFile)
    val fis        = new FileInputStream(fileToZip)
    val zipEntry   = new ZipEntry(fileToZip.getName())
    zipOut.putNextEntry(zipEntry)
    val bytes  = new Array[Byte](1024)
    var length = 0
    while ({ length = fis.read(bytes); length >= 0 }) {
      zipOut.write(bytes, 0, length)
    }
    zipOut.close()
    fis.close()
    fos.close()
  }
}
```

ekrich:
@sjrd I addressed the comments you had so far.

@WojciechMazur Maybe you could take a look at the PR and discussion and comment as well.

ekrich:
@sjrd I think I addressed your concerns if you would like to take another review pass.

ekrich:
We still have some issues that need to be solved. It is unclear to me at this point the best solution.

1. Each project in needs to contain both Scala and Native code needed. If those parts are optional they need to be selected as part of the `tools` build process.
2. Third party libraries need a way to supply any build arguments needed.

These two things may not be solved with the same method. It might be a good idea to close this PR and then create 2 issues for 1 and 2 above.

ekrich:
The consideration here would be to use property files like a project descriptor where the library developer could add options etc. that could be picked up by the build. We currently have the one that is used as a marker for `nativelib`. The might help us be able to publish Scala Native libraries to Maven Central and then they could be compiled on the host box similar to how we handle C glue code.

PR #2331, opened 2021-07-06
Update junit-runtime to match JUnit 4.13.2

### LINKED ISSUE #1881: Update JUnit runtime to version 4.13

nezosb:
Changes to junit-runtime were ported over from Scala.js. `assertThrows` was merged with `expectThrows` and added to junit-runtime. Generated output logs from junit-runtime were also adjusted to match the ones from JUnit 4.13.2.

I also added two tests to 'junit-tests' concerning newly added `assertThrows`.

As far as I understand (correct me if I'm wrong), the other additions in JUnit 4.13 concern fixes and additions to features which SN does not already support, so this PR should resolve #1881.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -799,14 +799,17 @@ lazy val testRunner =
     .settings(mavenPublishSettings)
     .settings(testInterfaceCommonSourcesSettings)
     .settings(
-      libraryDependencies ++= Seq(
-        "org.scala-sbt" % "test-interface"  % "1.0",
-        "com.novocode"  % "junit-interface" % "0.11" % "test"
-      )
+      libraryDependencies += "org.scala-sbt" % "test-interface" % "1.0",
+      libraryDependencies ++= jUnitJVMDependencies
     )
     .dependsOn(tools, junitAsyncJVM % "test")
 
 // JUnit modules and settings ------------------------------------------------
+
+lazy val jUnitJVMDependencies = Seq(
+  "com.novocode" % "junit-interface" % "0.11"   % "test",
+  "junit"        % "junit"           % "4.13.2" % "test"
+)
 
 lazy val junitRuntime =
   project
@@ -867,9 +870,7 @@ lazy val junitTestOutputsJVM =
     .in(file("junit-test/output-jvm"))
     .settings(
       commonJUnitTestOutputsSettings,
-      libraryDependencies ++= Seq(
-        "com.novocode" % "junit-interface" % "0.11" % "test"
-      )
+      libraryDependencies ++= jUnitJVMDependencies
     )
     .dependsOn(junitAsyncJVM % "test")
 
--- a/unit-tests-ext/src/test/scala/scala/scalanative/junit/utils/ThrowsHelper.scala
+++ b/unit-tests-ext/src/test/scala/scala/scalanative/junit/utils/ThrowsHelper.scala
@@ -1,15 +1,15 @@
 package scala.scalanative.junit.utils
 
-import AssertThrows._
+import AssertThrows.assertThrows
 
-// Calls to this should probably be changed to expectThrows.
+// Calls to this should probably be changed to assertThrows.
 // This was added as it was all over the place in the pre
 // JUnit code.
 object ThrowsHelper {
   def assertThrowsAnd[T <: Throwable, U](
       expectedThrowable: Class[T],
       code: => U)(cond: T => Boolean): Unit = {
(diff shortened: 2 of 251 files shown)


build.sbt
```diff
@@ -868,7 +869,8 @@ lazy val junitTestOutputsJVM =
     .settings(
       commonJUnitTestOutputsSettings,
       libraryDependencies ++= Seq(
-        "com.novocode" % "junit-interface" % "0.11" % "test"
+        "com.novocode" % "junit-interface" % "0.11"   % "test",
+        "junit"        % "junit"           % "4.13.2" % "test"
```
WojciechMazur, line 873: Since there are at least two usages of these dependencies, we might think about moving them to some `val JUnitDependencies`, eg and use it instead. This way when we will update versions in the future, they would need edit in only a single place.

unit-tests-ext/src/test/scala/scala/scalanative/junit/utils/ThrowsHelper.scala
```diff
@@ -1,15 +1,15 @@
 package scala.scalanative.junit.utils
 
-import AssertThrows._
+import AssertThrows.assertThrows
 
-// Calls to this should probably be changed to expectThrows.
+// Calls to this should probably be changed to assertThrows.
```
WojciechMazur, line 5: This comment can probably be removed now, since it was already changed

unit-tests/src/test/scala/utils/ThrowsHelper.scala
```diff
@@ -1,15 +1,15 @@
 package scala.scalanative.junit.utils
 
-import AssertThrows._
+import AssertThrows.assertThrows
 
-// Calls to this should probably be changed to expectThrows.
+// Calls to this should probably be changed to assertThrows.
```
WojciechMazur, line 5: This comment is now also obsolete

WojciechMazur (commented):
Overall looks good, there're just a few minor issues to fix.

build.sbt
nezosb: Good idea. I'll force-push the changes when the CI will finish its work, in case there is anything else I need to fix.

unit-tests-ext/src/test/scala/scala/scalanative/junit/utils/ThrowsHelper.scala
nezosb: The thing is, I have already renamed the comment itself to remain internally consistent with the code. Before the merger the object looked like this:
```
// Calls to this should probably be changed to expectThrows.
// This was added as it was all over the place in the pre
// JUnit code.
object ThrowsHelper {
  def assertThrowsAnd[T <: Throwable, U](
      expectedThrowable: Class[T],
      code: => U)(cond: T => Boolean): Unit = {
    val c = cond(expectThrows(expectedThrowable, code))
    assert(c)
  }
}
``` 
Now that you mention it, I am not sure what this comment really means - I assumed it wanted the calls to this object to be changed, but maybe it meant just the function name (as in `expectThrowsAnd`). Perhaps it should be removed either way for the ambiguity.

WojciechMazur: You're right, I've misunderstood the context of this comment. I belive original intent was to use `assertThrows` instead of `assertThrowsAnd`

nezosb:
I have updated dependencies in `build.sbt` as requested and left the comments in AssertThrows unchanged.

sjrd (approved):

ekrich:
Yes, I had a second to look at Scala.js yesterday and assertThrows is the
only util helper function. When I ported to JUnit we didn’t have this
upgrade yet so it was a note to change those. The JUnit change was huge so
so didn’t want to change more and I didn’t know exactly what the JUnit
upgrade was going to look like.

PR #2334, opened 2021-07-20
Fix #2321: Fix Clang discovery on Windows

### LINKED ISSUE #2127: Enhance clang discovery, compile, and linking options

### LINKED ISSUE #2321: Fix Clang discovery on Windows
On my Windows machine, the current implementation of ```private[scalanative] def discover(binaryName: String, envPath: String): Path``` in "\tools\src\main\scala\scala\scalanative\build\Discover.scala" fails to retrieve the proper location of clang.exe.

On my system this fix works (addition of ```if (binaryFile.isFile) return binaryFile.toPath```):
```
  /** Discover the binary path using environment
   *  variables or the command from the path.
   */
  private[scalanative] def discover(binaryName: String,
                                    envPath: String): Path = {
    val binaryNameOrPath = sys.env.getOrElse(envPath, binaryName)
    val binaryFile = new File(binaryNameOrPath)     // added by me
    if (binaryFile.isFile) return binaryFile.toPath // added by me
    
    val locateCmd        = if (Platform.isWindows) "where" else "which"
    val path = Process(Seq(locateCmd, binaryNameOrPath))
      .lineStream_!(silentLogger())
      .map { p => Paths.get(p) }
      .headOption
      .getOrElse {
        throw new BuildException(
          s"""No '$binaryNameOrPath' found in PATH or via '$envPath' environment variable.
             |Please refer to ($docSetup)""".stripMargin)
      }
    path
  }
```

Sorry but I'm too lazy/busy to make a PR for these two lines of code.
I hope this will still help.

ekrich:
Also closes #2127.

Documentation added.

For release notes:

If using the undocumented `CLANG_PATH` or `CLANGPP_PATH` to point to the executables `clang` and `clang++` respectively for CI or some other purpose, this has been replaced by `LLVM_BIN` which points to the `LLVM\bin` directory that contains the executables instead.


docs/user/setup.rst
```diff
@@ -110,6 +110,53 @@ installation of FreeBSD.
     $ wget https://raw.githubusercontent.com/scala-native/scala-native/master/scripts/scala-native.nix
     $ nix-shell scala-native.nix -A clangEnv
 
+**Windows**
+
+There are perhaps many different possible ways to setup an environment on
+Windows but this way works and is not too difficult. Powershell was used
+for installing and running Scala Native.
+
+1. Download and install Visual Studio Community 2019
```
ScalaWilliam, line 119: Hi, this could be replaced with chocolatey or scoop, eg: https://community.chocolatey.org/packages/visualstudio2019community


Similarly with LLVM:

https://community.chocolatey.org/packages/llvm

david-bouyssie: > Hi, this could be replaced with chocolatey or scoop, eg: https://community.chocolatey.org/packages/visualstudio2019community
> 
> Similarly with LLVM:
> 
> https://community.chocolatey.org/packages/llvm

I agree. Alternatively a full command line with appropriate arguments could be also useful.
Example:
```
".\vs_community__<version>.exe" --add Microsoft.VisualStudio.Component.PortableLibrary 
    --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 
    --add Microsoft.VisualStudio.Component.Windows10SDK.18362 
    --quiet --wait --norestart --noUpdateInstaller
```

ekrich: I will clean this up and mention chocolatey and scoop as alternatives. I needed to install the way I did because of corporate constraints which I will also mention. The last command line example is more complicated so I don't really want to add that although it seems more direct. BTW, I am a Windows novice so I'm sure many people have much fancier ways to go about this. We can always update the instructions after this PR based on feedback. Right now I am trying to fix the problem and wanted to add some docs that worked for me in a corporate environment.

I was actually surprised it would be this easy and it would just work. :smile:

WojciechMazur (approved):

PR #2338, opened 2021-08-04
Add a default signal handler to `TestMain`

### LINKED ISSUE #1262: Make segfaults in tests debuggable
As discussed in #1056, it's currently difficult to debug segfaults in unit tests. We could greatly improve the situation if we provide a default catch-all segfault handler that prints detailed segfault information before aborting the program. Then it would be clear where exactly did the program crashes, unlike current opaque behavior.

nezosb:
Now, similarly to Bash programs, fatal signals in `TestMain` will be reported as exit code 128 + n, where n is the signal value. This way, `ProcessRunner` will be able to distinguish between planned exits and fatal signals and report accordingly. Reporting through exit codes was the only possible choice here - IO operations in handlers lead to unspecified behavior and we cannot return from the handler to the main program to report there, as returning from certain signals leads to the handler being called again.

Signal handler is being set only for signals that, by default, terminate the process and two that stop it for unexpected reasons (SIGTTIN and SIGTTOUT). Otherwise, the testing could stop when, for example, someone simply resizes the terminal window.

This feature was tested locally by running external C code in `unit-tests` as I was unable to acquire any errors of this nature in a Scala code. This way, SIGSEGV, SIGBUS and SIGFPE were all tested. I did not include any tests in this PR, as I did not want to bloat `scripted-tests` (although I can add those if necessary).

This PR should resolve #1262 and #1828.
Special thanks to LeeTibbert for a documented significant research on this subject.


nezosb:
I have changed SignalConfig implementation from C to Scala and changed which signals are being caught. This time, they were chosen strictly by hand from those supported in posixlib and clib. Hopefully, this way we can avoid any errors/warnings from trying to set a handler to signals unsupported by the OS distribution, as they all should be supported.

test-interface/src/main/scala/scala/scalanative/testinterface/signalhandling/SignalConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // Only these select "signals" can work on Windows
+    setHandler(SIGABRT)
+    setHandler(SIGFPE)
+    setHandler(SIGILL)
+    setHandler(SIGSEGV)
+    setHandler(SIGTERM)
+
+    if (isLinux() || isMac()) {
```
WojciechMazur, line 40: It would not work for other UNIX-based systems, eg. BSD family. 
```suggestion
    if (!isWindows) {
```

test-interface/src/main/scala/scala/scalanative/testinterface/signalhandling/SignalConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+     * can only communicate through exitcode values. In bash programs,
+     * exitcode > 128 signifies a fatal signal n, where n = exitcode - 128.
+     * This is the convention used here.
+     */
+    val defaultHandler = CFuncPtr1.fromScalaFunction { (sig: CInt) =>
+      exit(128 + sig)
+    }
+
+    def YELLOW = "\u001b[0;33m"
+    def RESET = "\u001b[0;0m"
```
WojciechMazur, line 24: This can be a val
```suggestion
    val YELLOW = "\u001b[0;33m"
    val RESET = "\u001b[0;0m"
```

test-interface/src/main/scala/scala/scalanative/testinterface/signalhandling/SignalConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    def setHandler(sig: CInt): Unit = {
+      if (signal(sig, defaultHandler) == SIG_ERR)
+        println(
```
WojciechMazur, line 28: We can try to print it to standard error
```suggestion
        Console.err.println(
```

WojciechMazur (commented):
Got few minor changes to fix, it would be good if we could experiment with getting more information about the origin of the signal through stack trace.

Also here's a quick snippet resulting in Segmentation Fault using Scala code, which might be useful for testing: 
```
    import scalanative.unsafe._
    import scalanative.runtime.Intrinsics.{stackalloc => _, _}
    import scalanative.runtime._
    val ref = stackalloc[Int]
    println(loadInt(loadRawPtr(toRawPtr(ref))))
```

test-interface/src/main/scala/scala/scalanative/testinterface/signalhandling/SignalConfig.scala
nezosb: Oh right! I'm not sure how I missed that...

test-interface/src/main/scala/scala/scalanative/testinterface/signalhandling/SignalConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+    /* Default handler for signals like SIGSEGV etc.
+     * Since it needs to be able to handle segmentation faults, it has to
+     * exit the program on call, otherwise it will keep being called indefinetely.
+     * This and IO operations in handlers leading to unspecified behavior mean we
+     * can only communicate through exitcode values. In bash programs,
+     * exitcode > 128 signifies a fatal signal n, where n = exitcode - 128.
+     * This is the convention used here.
+     */
+    val defaultHandler = CFuncPtr1.fromScalaFunction { (sig: CInt) =>
+      exit(128 + sig)
```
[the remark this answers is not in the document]
nezosb: It turns out I misread a presentation - I thought that write() and read(), despite being async-signal-safe, could still lead to errors when used in signal handler, leading to my original comment about IO operations. That is of course not the case. Moreover, I discovered that all of the libunwind methods used in StackTrace_PrintStackTrace() in StackTrace.c are also async-signal-safe according to http://www.nongnu.org/libunwind/man/libunwind(3).html#section_5. This means that we should be able to print the stack trace, after rewriting aforementioned method to support write() instead of printf().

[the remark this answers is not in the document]
ekrich: Since I have one case where the stack unwind is the crash, it might be good to try and print the signal caught first and the stacktrace and then exit.

nezosb:
As requested, I have added stack trace printing before exiting the signal handler, preceded by a message with signal caught. Since StackTraceElement.fromSymbol had to be slightly reimplemented to be async-signal-safe, a base implementation was created in nativelib, used in both `javalib` and in `TestMain`.

WojciechMazur (approved):

PR #2341, opened 2021-08-06
Fix #2339/2340 Add a publish-local script

### LINKED ISSUE #2339: Can't compile on macOS on jdk 1.8
Environment:
```
openjdk version "1.8.0_292"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_292-b10)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.292-b10, mixed mode)
```

Error:
```
[error] ## Exception when compiling 59 sources to /Users/catap/src/scala-native/tools/target/scala-2.11/classes
[error] java.lang.StackOverflowError
[error] scala.tools.nsc.transform.patmat.Solving$Solver$class.findTseitinModelFor(Solving.scala:474)
[error] scala.tools.nsc.transform.patmat.PatternMatching$OptimizingMatchTranslator.findTseitinModelFor(PatternMatching.scala:89)
[error] scala.tools.nsc.transform.patmat.Solving$Solver$class.findTseitinModelFor(Solving.scala:484)
[error] scala.tools.nsc.transform.patmat.PatternMatching$OptimizingMatchTranslator.findTseitinModelFor(PatternMatching.scala:89)
[error] scala.tools.nsc.transform.patmat.Solving$Solver$class.findTseitinModelFor(Solving.scala:484)
[error] scala.tools.nsc.transform.patmat.PatternMatching$OptimizingMatchTranslator.findTseitinModelFor(PatternMatching.scala:89)
...
```

ekrich:
The `publish-local` and `release` scripts now call a `publish-impl` script.

Should publish use the oldest versions in the series or the latest? We can fix this depending on the answer and I added a comment for future maintainers.

I was not able to test release as it calls `publishSigned` which errors. If I call `publish` I get errors that I need keys and such which I expect. Not sure what is going on there.

I didn't add a publish for windows yet. Not sure where it goes in the list yet.

### DIFF
--- a/scripts/publish-impl
+++ b/scripts/publish-impl
@@ -1,0 +1,31 @@
+#!/bin/bash
+# publishSigned or publishLocal
+publish=$1
+
+set -ex
+
+sbt clean
+
+# lowest versions supported
+for v in 2.11.12 2.12.13 2.13.4; do
+  sbt -Dsbt.supershell=false \
+    ++$v \
+    nscplugin/$publish \
+    junitPlugin/$publish \
+    nativelib/$publish \
+    clib/$publish \
+    posixlib/$publish \
+    javalib/$publish \
+    auxlib/$publish \
+    scalalib/$publish \
+    testInterfaceSbtDefs/$publish \
+    testInterface/$publish \
+    junitRuntime/$publish
+done
+
+sbt -Dsbt.supershell=false \
+  util/$publish \
+  nir/$publish \
+  tools/$publish \
+  testRunner/$publish \
+  sbtScalaNative/$publish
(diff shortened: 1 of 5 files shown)


scripts/publish-impl
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# publishSigned or publishLocal
+publish=$1
+
+set -ex
+
+sbt clean
+
+# lowest versions supported
+for v in 2.11.12 2.12.13 2.13.4; do
```
sjrd, line 10: We should use the latest version instead, in this case.

ekrich: Thank-you, I will fix that. Do you know why `publishSigned` is basically not found? I haven't looked to see if there is a change with something in sbt.

sjrd: Because sbt-pgp is not in `project/build.sbt`. It's in the global `plugins.sbt` of the machine running the publishing.

ekrich: I updated the docs so it should be good to go once CI completes.

WojciechMazur (approved):

PR #2344, opened 2021-08-09
Ignore no longer existing objects when inheriting Interflow state

WojciechMazur:
This PR fixes issue leading to failures of `release` builds when using linktime resolved properties. 
When resolving link-time property call to synthesised method is replaced with `Op.Copy` returning a resolved value. 
When inlining such a call, the owner in which given property is defined (eg. `scalanative.LinktimeInfo$`) is never used and might be removed from the given block state. However, we don't check if a given instance is defined in both current and inherited state. This might lead to unhandled NullPointerException and failure of the build.

This PR does also contain fixes for other issues leading to build failure. `Reach.addMissing` now creates `Path` using the dedicated method for URIs, instead of passing String from `URI.getPath`, on Windows such usage will lead to an Invalid Argument exception.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/interflow/State.scala
+++ b/tools/src/main/scala/scala/scalanative/interflow/State.scala
@@ -137,8 +137,12 @@ final class State(block: Local) {
   def inherit(other: State, roots: Seq[Val]): Unit = {
     val closure = heapClosure(roots) ++ other.heapClosure(roots)
 
-    closure.foreach { addr =>
-      val clone = other.heap(addr).clone()
+    for {
+      addr <- closure
+      // Ignore keys no longer existing in other state
+      obj <- other.heap.get(addr)
+      clone = obj.clone
+    } {
       clone match {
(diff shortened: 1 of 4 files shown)


tools/src/main/scala/scala/scalanative/interflow/State.scala
```diff
@@ -137,8 +137,12 @@ final class State(block: Local) {
...
     val closure = heapClosure(roots) ++ other.heapClosure(roots)
 
-    closure.foreach { addr =>
-      val clone = other.heap(addr).clone()
+    for {
+      addr <- closure
+      // Ignore keys no longer existing in other state
+      obj <- other.heap.get(addr)
+      clone = obj.clone
+    } {
```
sjrd, line 145: `clone` could still be define as a regular `val` in the body of the for:
```suggestion
    } {
      val clone = obj.clone()
```

sjrd (approved):

PR #2349, opened 2021-08-13
if Windows, append .exe to binaryName in discover(), to support mingw64 and cygwin compilers

philwalk:
This change enables discovery of `clang.exe` and `clang++.exe`,  in these environments:
```
CYGWIN_NT-10.0 d5 3.2.0(0.340/5/3) 2021-03-29 08:42 x86_64 Cygwin
MSYS_NT-10.0-19042 d5 3.2.0-340.x86_64 2021-07-02 08:36 UTC x86_64 Msys
```
Also verified that discovery still works when first `clang` in the PATH is here:
```
C:/Program Files (x86)/Microsoft Visual Studio/2019/Preview/VC/Tools/Llvm/x64
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-# vim
-*.swp
 scalalib/src/
 scala-partest/fetchedSources/
 target/
@@ -34,3 +32,6 @@ bin/
 
 # vscode
 /.vscode/
+
+# vim
+*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+# vim
+*.swp
 scalalib/src/
 scala-partest/fetchedSources/
 target/
--- a/tools/src/main/scala/scala/scalanative/build/Discover.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Discover.scala
@@ -150,7 +150,7 @@ object Discover {
 
     val command: Seq[String] = {
       if (Platform.isWindows) {
-        val binName = s"$binaryName.exe"
+        val binName = s"${binaryName}.exe"
         val arg = binPath.fold(binName)(p => s"$p:$binName")
         Seq("where", arg)
       } else {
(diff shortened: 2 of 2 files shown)


tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -150,7 +150,8 @@ object Discover {
 
     val command: Seq[String] = {
       if (Platform.isWindows) {
-        val arg = binPath.fold(binaryName)(p => s"$p:$binaryName")
+        val binName = s"$binaryName.exe"
```
ekrich, line 153: `val binName = s"${binaryName}.exe"`

ekrich, line 153:
```suggestion
        val binName = s"${binaryName}.exe"
```

.gitignore
```diff
@@ -1,3 +1,5 @@
+# vim
+*.swp
```
ekrich, line 2: Do you mind moving this down to the bottom with the organized IDE/Editor ignores? The top part is pretty ugly.

ekrich (changes requested):
Looks great, thanks for contributing. I guess I should have tested in other than `cmd` or Powershell 😄

.gitignore
philwalk: done

philwalk:
FYI, In order to fix problems with discovery of `clang++` in `cygwin`, I had to replace the installed symlink.

Originally it was a Windows `<JUNCTION>` file (not visible to java), and even after replacing that with a '<SYMLINK>', java was not finding it.   I experimented with `java.nio.file` symlink detection and de-referencing, but it turned out to be much simpler to just rename the symlink from `clang++` to `clang++.exe`.

WojciechMazur (approved):

ekrich:
@philwalk This already is merged but if you want to do another PR and add a small paragraph under the Windows section for Cygwin and MSys with your workaround and any other install tips etc. for those environments that would be great. https://github.com/scala-native/scala-native/blob/master/docs/user/setup.rst

You can build the docs by running `make html` in the docs directory. You need Python 3 and pip to install sphinx, and recommonmark. We should have a section in the contributors section on what to run to get docs working but alas we don't.

philwalk:
@ekrich - thanks, hopefully I can work on this again soon.

PR #2351, opened 2021-08-21
Allow to embed Rust in native sources

WojciechMazur:
After sugestion of @ekrich and some discussion with Scala tooling team in VirtusLab we've come up with implementation allowing to embed Rust sources in native sources of SN durring our last hacking session. 
Current solution supports both compilation of simple programs using `rustc` as well as more complicated projects using external dependencies using crates.io. It would allow for bootsrapping frameworks calling Rust from Scala Native. Some additional abstractions might be needed to introduce in order to fullfill Rusts memory management contract - especially  read-only references, as well as to provide bindings for rust allowing to allocate memory inside Scala Native's GC, most of that might be possible to introduce as third-party library. 
  
It's worth to note that compilation of Rust sources currently always creates static libraries instead of object files, this solution has it's drawbacks - especially larger size of created artifacts, however it allows us to use Rust's std library, which might be problematic to provide on it's own. Due to this decision we needed to introduce some additional types, that would keep track either given source is object file or static library.


armanbilge:
We were discussing this PR in Discord today. @ekrich mentioned that it was in good shape but there wasn't much interest. So I wanted to make sure to express my interest 😁 

The appeal of this would be to help integrate/bind with [Tokio](https://tokio.rs/) which is the modern answer to libuv (e.g., Node.js uses libuv but [Deno](https://deno.land/) uses Tokio). Specifically, the goal would be to use Tokio to provide a Native implementation of the [cross-platform fs2 I/O](https://fs2.io/#/io) APIs, which already support both JVM and Node.js.

Thanks for pursuing these initiatives!

ahjohannessen:
@WojciechMazur This PR is exciting :) What is the status on this?

unit-tests/native/src/test/resources/scala-native/rust-testlib.rs
```diff
@@ -0,0 +1,10 @@
+use std::ffi::CStr;
+
+#[no_mangle]
+pub extern "C" fn check_test_string(cstring:  *const i8) -> bool {
+  unsafe {
+  let c_str: &CStr = CStr::from_ptr(cstring);
+  let str: &str = c_str.to_str().unwrap();
+  return str == "Hello Rust";
+  }
+}
```
ekrich, line 10: Missing a return here.

unit-tests/native/src/test/resources/scala-native/rustlib/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[no_mangle]
+pub extern "C" fn isEven(x: i32) -> bool {
+  x % 2 == 0
+}
```
ekrich, line 4: Missing return here too.

armanbilge:
I'm working on refreshing this PR in https://github.com/scala-native/scala-native/pull/2566.

PR #2356, opened 2021-08-27
Setup MiMa checks

WojciechMazur:
This PR adds checks proper setup for MiMa binary incompatibilities checks, based on Scala.js. 

Needs to be rebased onto #2207 due to a bug in the binary version resolving mechanism. 

 * Binary compatibility checks were enabled in all published projects.
 * Added filters for known incompatibilities inside package-private classes and classes treated as internal
 * Fixed some binary incompatibilities in public API

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -73,13 +73,17 @@ lazy val docsSettings: Seq[Setting[_]] = {
 
 // The previous releases of Scala Native with which this version is binary compatible.
 val binCompatVersions = Set("0.4.0")
+lazy val neverPublishedProjects = Set(windowslib).map(_.id)
 
 lazy val mimaSettings = Seq(
+  mimaFailOnNoPrevious := false,
   mimaBinaryIssueFilters ++= BinaryIncompatibilities.moduleFilters(name.value),
-  mimaPreviousArtifacts ++= binCompatVersions.map { version =>
-    ModuleID(organization.value, moduleName.value, version)
-      .cross(crossVersion.value)
-  }
+  mimaPreviousArtifacts ++= binCompatVersions
+    .filter(_ => !neverPublishedProjects.contains(thisProject.value.id))
+    .map { version =>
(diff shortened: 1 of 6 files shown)


project/BinaryIncompatibilities.scala
```diff
@@ -4,11 +4,65 @@ import com.typesafe.tools.mima.core._
...
+    exclude[Problem]("scala.scalanative.checker.*"),
+    exclude[Problem]("scala.scalanative.interflow.*"),
+    exclude[Problem]("scala.scalanative.linker.*"),
+    exclude[Problem]("scala.scalanative.build.NativeLib.*"),
+    exclude[Problem]("scala.scalanative.build.LLVM.*"),
+    exclude[Problem]("scala.scalanative.build.NativeConfig*Impl*"),
+    exclude[Problem]("scala.scalanative.build.GC.this")
+  )
+
+  final val CLib, PosixLib, WindowsLib: Filters = Seq(
```
sjrd, line 38: How come all three of those artifacts have the same MiMa filters?

build.sbt
```diff
@@ -398,7 +377,6 @@ lazy val nscplugin =
   project
     .in(file("nscplugin"))
     .settings(mavenPublishSettings)
-    .settings(noMimaSettings)
```
sjrd: There is no binary guarantee whatsoever for the nscplugin. We should not check MiMa on this project.

build.sbt
```diff
@@ -977,7 +955,6 @@ lazy val junitPlugin =
   project
     .in(file("junit-plugin"))
     .settings(mavenPublishSettings)
-    .settings(noMimaSettings)
```
sjrd: Same for junit-plugin. Compiler plugins don't have any binary compat guarantee.

sjrd (changes requested):

sjrd (approved):

PR #2358, opened 2021-09-02
Implement `java.nio` on Windows

WojciechMazur:
This PR adds Windows implementation for all `java.nio` methods supported on Unix. 

Scope of changes: 
* Implement WindowsFileSysytem, its provider and Views
* Implement WindowsPath and it's parser
* Moved Unix FileSystem implementation to its own package
* Conditionally select FS providers using linktime condition
* Adjust some of the java.nio test to be JVM compliment. 
* Added windowslib bindings used by java.nio

### DIFF
--- a/javalib/src/main/scala/scala/scalanative/nio/fs/FileHelpers.scala
+++ b/javalib/src/main/scala/scala/scalanative/nio/fs/FileHelpers.scala
@@ -88,7 +88,7 @@ object FileHelpers {
     }
 
     def listWindows() = Zone { implicit z =>
-      val searchPath = path + raw"\*"
+      val searchPath = raw"$path\*"
       if (searchPath.length.toUInt > FileApiExt.MAX_PATH)
         throw new IOException("File name to long")
 
--- a/javalib/src/main/scala/scala/scalanative/nio/fs/FileHelpers.scala
+++ b/javalib/src/main/scala/scala/scalanative/nio/fs/FileHelpers.scala
@@ -11,12 +11,17 @@ import scala.collection.mutable.UnrolledBuffer
 import scala.reflect.ClassTag
 import java.io.{File, IOException}
 import java.nio.charset.StandardCharsets
-import scala.scalanative.windows.ErrorHandlingApi._
+import scala.scalanative.windows._
+import scala.scalanative.windows.HandleApiExt.INVALID_HANDLE_VALUE
 import scala.scalanative.windows.FileApi._
 import scala.scalanative.windows.FileApiExt._
-import scala.scalanative.windows.HandleApiExt._
+import scala.scalanative.windows.FileApiOps._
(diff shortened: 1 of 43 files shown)


javalib/src/main/scala/scala/scalanative/nio/fs/FileHelpers.scala
```diff
@@ -82,7 +87,36 @@ object FileHelpers {
       }
     }
 
-    def listWindows() = ???
+    def listWindows() = Zone { implicit z =>
+      val searchPath = path + raw"\*"
```
ekrich, line 91: Can you use triple quotes for this?

ekrich (commented):
Is there any preference/guidance or alignment with Scala.js when using Java classes?
It seems there are the package alias `ju.Map, jl.Boolean` and the class alias `JMap, JBoolean` strategies.

WojciechMazur:
> Is there any preference/guidance or alignment with Scala.js when using Java classes?
> It seems there are the package alias `ju.Map, jl.Boolean` and the class alias `JMap, JBoolean` strategies.

@ekrich 
I don't think there is any recommended approach. Even in Scala.js you can find both types of aliases.

javalib/src/main/scala/scala/scalanative/nio/fs/FileHelpers.scala
WojciechMazur: Sure, done

unit-tests/shared/src/test/scala/javalib/nio/file/UnixPathTest.scala
```diff
@@ -2,12 +2,22 @@ package javalib.nio.file
...
+import org.junit.Assume._
 
 import scalanative.junit.utils.AssertThrows.assertThrows
+import org.scalanative.testsuite.utils.Platform.isWindows
 
-class PathTest {
+class UnixPathTest {
+
+  @Before
+  def checkIsUnix() {
```
sjrd, line 15: Same here: `assumeIsUnix`.

Also, perhaps this can be an `@BeforeClass` in the companion object instead.

unit-tests/jvm/src/test/resources/BlacklistedTests.txt
```diff
@@ -43,7 +43,7 @@ scala/scala/PrimitiveTest.scala
 scala/scala/bugcompat/LongFloatPrimitiveTest.scala
 
 # Tests that work on java 8, but fail on java 16
-scala/javalib/nio/file/PathTest.scala
+#scala/javalib/nio/file/UnixPathTest.scala
```
sjrd, line 46: Why the `#`? Was that intentional?

sjrd (changes requested):

sjrd (approved):

PR #2366, opened 2021-09-16
Remove non standard value for Windows TokenInformationClass bindings

WojciechMazur:
One of defined TokenInformationClass bindings was non standard, leading to failures in the CI. 
It looks like for Windows Runtime library in version 10.0.22000.0 used in CI `TokenOriginatingProcessTrustLevel` enum is no longer defined. However, it looks like this value was defined in older version 10.0.19041.0, used locally on my machine. 
The decision about the removal of this value can be also justified by the fact that mentioned enum value is not defined in [official Microsoft documentation](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ne-winnt-token_information_class) 

Additionally based on the logs from the CI deprecation warnings in the None GC were removed by usage of special Windows macro definition. This warning was probably not present before (they're already defined in `Parsing.h` file using `sscanf` and `getenv` functions treated as deprecated) and it might suggest that main `Windows.h` file header is now being included earlier and though ignoring previous definition of `_CRT_SECURE_NO_WARNINGS`

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/none/gc.c
+++ b/nativelib/src/main/resources/scala-native/gc/none/gc.c
@@ -1,7 +1,7 @@
 // sscanf and getEnv is deprecated in WinCRT, disable warnings
 // These functions are not used directly, but are included in "Parsing.h".
-// The definition used to disable warnings needs to be placed before the first 
-// include of Windows.h, depending on the version of Windows runtime 
+// The definition used to disable warnings needs to be placed before the first
+// include of Windows.h, depending on the version of Windows runtime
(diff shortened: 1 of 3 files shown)


sjrd (approved):

david-bouyssie:
I don't know why but on my system I also have an issue with TokenIsSandboxed, while it seems to be officially supported.

I have to comment this C function to make it work:
https://github.com/scala-native/scala-native/blob/90f9c504a71cbb867fd18130b27304db9e00aab4/windowslib/src/main/resources/scala-native/windows/winnt/tokenInformationClass.c#L132

PR #2370, opened 2021-09-17
Allow to run all unit tests on Windows

WojciechMazur:
This PR fixes issues and incompatibilities when trying to run unit tests on Windows. 
It does disable some of the unit tests when execution on Windows: 
*  `CVarArgListTest` - due to no compatible implementation for passing var args
* `DateTest` - due to problems with proper printing of timezones (eg. instead of `CEST` we would get long `Central European Summer Time` name)

It also does remove a separate compilation of native sources which is no longer needed in the CI - all native sources are compiled and tests in unit tests.


sjrd (approved):
:tada:

unit-tests/native/src/test/scala/scala/scalanative/regex/ExecTest.scala
```diff
@@ -106,6 +107,10 @@ class ExecTest {
     var line: String = null
     while ({ line = r.readLine(); line != null }) breakable {
       lineno += 1
+      // Compat for Windows
+      if (isWindows) {
+        line = line.replaceAll("\r", "")
```
ekrich, line 112: Should there be a function for this since it is repeated?

unit-tests/shared/src/test/scala/javalib/util/PropertiesTest.scala
```diff
@@ -196,7 +196,10 @@ class PropertiesTest {
       val ps = new PrintStream(out)
       props.list(ps)
       ps.close()
-      assertEquals(out.toString.trim, result.trim)
+      assertEquals(
+        out.toString.replaceAll(System.lineSeparator(), "\n").trim,
```
ekrich, line 200: Is this just an artifact of the test or is it possible the parser is incorrect? I am thinking that it is consistent if it stays on one platform and can read another but testing line endings if done on one platform and tested on another can make them inconsistent.

ekrich (commented):
Overall looks good, so just questions and comments.

unit-tests/shared/src/test/scala/javalib/util/PropertiesTest.scala
WojciechMazur: If we would remove adjustment of line separator the test would fail, however, this behaviour is compliant with the JVM, both end us with the same result: 
On JVM:
 ```
[error] name=alice> but was:<...isting properties --[]
[error] name=alice>, took 0.011 sec
[error]     at javalib.util.PropertiesTest.assertResult$1(PropertiesTest.scala:201)
 ``` 
 On SN: 
```
[error] name=alice> but was:<...isting properties --[]
[error] name=alice>, took 0.001 sec
[error]     at java.lang.StackTrace$.currentStackTrace(Unknown Source)
```

ekrich: Ok, thank-you.

PR #2371, opened 2021-09-17
Allow to execute scripted tests on Windows

WojciechMazur:
This PR fixes scripted tests to allow execution of them on Windows. 
Due to the limitations of Windows, some tests needed to be made OS-specific, mostly in `java-io-{1,2}` scripted tests. 
Also to allow proper communication via sockets in `java-net-socket` scripted tests file encoding used on the JVM endpoint was changed to UTF-8 (default on Windows is Cp1025, however, we don't support it)
`link-order` test was also adjusted to use Windows compliant naming convention for compiled sources. 

Execution of scripted tests was added to CI - running as separate task only for Scala 2.12 latest version (similarly as it is done linux tests)

### DIFF
--- a/scripted-tests/run/java-io-file-1/Utils.scala
+++ b/scripted-tests/run/java-io-file-1/Utils.scala
@@ -1,0 +1,18 @@
+import java.util.Locale
+
+object Utils {
+  val isWindows = System
+    .getProperty("os.name", "unknown")
+    .toLowerCase()
+    .startsWith("windows")
+
+  def assertOsSpecific(
+      pred: Boolean,
+      msg: => String
+  )(onUnix: Boolean, onWindows: Boolean) = {
+    val expected =
+      if (isWindows) onWindows
+      else onUnix
+    assert(pred == expected, msg)
+  }
+}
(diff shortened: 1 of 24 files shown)


scripted-tests/run/java-io-file-1/Utils.scala
```diff
@@ -0,0 +1,18 @@
+import java.util.Locale
+
+object Utils {
+  val isWindows = System
+    .getProperty("os.name", "unknown")
+    .toLowerCase()
```
sjrd, line 6:
```suggestion
    .toLowerCase(Locale.ROOT)
```
with the appropriate `import`.

scripted-tests/run/java-io-file-2/Utils.scala
```diff
@@ -0,0 +1,18 @@
+import java.util.Locale
+
+object Utils {
+  val isWindows = System
+    .getProperty("os.name", "unknown")
+    .toLowerCase()
```
sjrd, line 6:
```suggestion
    .toLowerCase(Locale.ROOT)
```

sjrd (changes requested):

sjrd (approved):

PR #2373, opened 2021-09-19
Reimplement `Thread.sleep` to be OS-specific and multithreading safe

WojciechMazur:
This PR changes the implementation of `Thread.sleep` function, to allow usage of it from Windows  - currently used functions `sleep` and `usleep` are POSIX-specific. Instead, it provides separate implementation depended on Windows-native `Sleep` function. 

Additionally, Unix implementation was improved to prepare for multithreaded execution - current implementation would be invoked from sleep in case of interruption by signal - this scenario is common eg. when using Boehm GC in multithreaded mode, leading to premature leaving the `Thread.sleep` function. To make it easier and more accurate, the new implementation uses the` unistd.nanosleep` function returning remaining time of requested sleep.  

Newly created objects in package `java.lang.impl` would be extended in the future, as they will contain OS-specific implementation for interacting with Threads (starting os threads, parking mechanism etc) in the future. This change was cherry picked from my [multithreading develop branch](https://github.com/WojciechMazur/scala-native/tree/multithreading/develop/javalib/src/main/scala/java/lang/impl).

### DIFF
--- a/javalib/src/main/scala/java/lang/Thread.scala
+++ b/javalib/src/main/scala/java/lang/Thread.scala
@@ -1,8 +1,8 @@
 package java.lang
 
-import scalanative.unsigned._
 import scalanative.annotation.stub
-import scalanative.libc.errno
+import scalanative.meta.LinktimeInfo.isWindows
+import impl._
 
 class Thread private (runnable: Runnable) extends Runnable {
   if (runnable ne Thread.MainRunnable) ???
@@ -73,15 +73,6 @@ object Thread {
   }
 
   def sleep(millis: scala.Long, nanos: scala.Int): Unit = {
-    import scala.scalanative.posix.errno.EINTR
-    import scala.scalanative.unsafe._
-    import scala.scalanative.posix.unistd
-
-    def checkErrno() =
-      if (errno.errno == EINTR) {
-        throw new InterruptedException("Sleep was interrupted")
-      }
-
     if (millis < 0) {
       throw new IllegalArgumentException("millis must be >= 0")
     }
@@ -89,10 +80,8 @@ object Thread {
       throw new IllegalArgumentException("nanos value out of range")
     }
 
-    val secs = millis / 1000
-    val usecs = (millis % 1000) * 1000 + nanos / 1000
-    if (secs > 0 && unistd.sleep(secs.toUInt) != 0.toUInt) checkErrno()
-    if (usecs > 0 && unistd.usleep(usecs.toUInt) != 0) checkErrno()
+    if (isWindows) WindowsThread.sleep(millis, nanos)
+    else PosixThread.sleep(millis, nanos)
   }
 
   def sleep(millis: scala.Long): Unit = sleep(millis, 0)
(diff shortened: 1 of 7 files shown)


javalib/src/main/scala/java/lang/Thread.scala
```diff
@@ -1,8 +1,8 @@
 package java.lang
 
-import scalanative.unsigned._
 import scalanative.annotation.stub
-import scalanative.libc.errno
+import scalanative.meta.LinktimeInfo.isWindows
+import impl._
```
sjrd, line 5: Consider always using fully-qualified imports at the top, especially for generic names like that, which could come from other imports by accident:
```suggestion
import java.lang.impl._
```

sjrd (approved):

posixlib/src/main/scala/scala/scalanative/posix/unistd.scala
```diff
@@ -33,6 +33,10 @@ object unistd {
   def truncate(path: CString, length: off_t): CInt = extern
   def unlink(path: CString): CInt = extern
   def usleep(usecs: CUnsignedInt): CInt = extern
+  def nanosleep(
+      requested: Ptr[time.timespec],
+      remaining: Ptr[time.timespec]
+  ): CInt = extern
```
ekrich, line 39: This should be in the time header as far as I can tell. https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/time.h.html

posixlib/src/main/scala/scala/scalanative/posix/unistd.scala
```diff
@@ -33,6 +33,10 @@ object unistd {
   def truncate(path: CString, length: off_t): CInt = extern
   def unlink(path: CString): CInt = extern
   def usleep(usecs: CUnsignedInt): CInt = extern
```
ekrich, line 35: I see `sleep` that take unsigned and returns unsigned - https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/unistd.h.html#tag_13_80

ekrich: Yeah, not sure how that `usleep` got there - https://newbedev.com/is-there-an-alternative-sleep-function-in-c-to-milliseconds

WojciechMazur: I think I found out why it is no longer in the specification, as described in [this reference](https://man7.org/linux/man-pages/man3/usleep.3.html)
`     POSIX.1-2001 declares this function
       obsolete; use nanosleep(2) instead.  POSIX.1-2008 removes the
       specification of usleep().`
When it comes to incorrect sleep return type I would suggest creating an issue, so we don't forget about that. We should change it in the next major release (0.5.x)

posixlib/src/main/scala/scala/scalanative/posix/unistd.scala
WojciechMazur: Thank you for catching this up, I don't know why I've put it in unistd at this point.

posixlib/src/main/scala/scala/scalanative/posix/unistd.scala
ekrich: Done. https://github.com/scala-native/scala-native/issues/2395

PR #2377, opened 2021-09-20
Improve implementation of `NativeConfig.linktimeProperties`

WojciechMazur:
This PR introduces several improvements to the mechanism of defining link-time properties. 
The main motivation of this change was to make it easier to define properties that should depend on the NativeConfig instance, as well as to allow maintainers of Scala Native plugins for build tools other than sbt easier. 

* Added static method returning predefined link-time properties managed in Scala Native project based on passed instance of NativeConfig. 
* Moved checking of validity for supported values for LinktimeProperties to static method
* Using `withLinktimeProperties` now only affects custom properties. 
* Custom properties might override predefined ones.
* Removed support for discovering linktime properties in `Discovery` since we don't plan to use env variable to set them up.

* Removed no longer needed overrides for linktime properties in `partests` and compiler tests in `tools`


tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -237,4 +222,40 @@ object NativeConfig {
...
+      value != null && hasSupportedType
+    }
+
+    val invalid = properties.collect {
+      case (key, value) if !isNumberOrString(value) => key
+    }
+    if (invalid.nonEmpty) {
+      System.err.println(
+        s"Invalid link-time properties: \n ${invalid.mkString(" - ", "\n", "")}"
+      )
```
sjrd, line 254: This should be part of the exception message, instead of being printed to the console.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -237,4 +222,40 @@ object NativeConfig {
...
+  def checkLinktimeProperties(properties: LinktimeProperites): Unit = {
+    def isNumberOrString(value: Any) = {
+      def hasSupportedType = value match {
+        case _: Boolean | _: Byte | _: Char | _: Short | _: Int | _: Long |
+            _: Float | _: Double | _: String =>
+          true
+        case _ => false
+      }
+
+      value != null && hasSupportedType
```
sjrd, line 245: `value != null` is redundant. If `value` is null, the `case _: T` will not match for any `T`, and therefore `case _ => false` will be used.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -176,31 +179,13 @@ object NativeConfig {
...
-      }
+    def linktimeProperties: Map[String, Any] = {
+      predefinedLinktimeProperties(this) ++ customLinktimeProperties
+    }
 
-      copy(linktimeProperties = v)
+    override def withLinktimeProperties(v: Map[String, Any]): NativeConfig = {
+      checkLinktimeProperties(v)
+      copy(customLinktimeProperties = v)
     }
```
sjrd, line 189: I don't like that there is `linktimeProperties` that does not always return exactly `v` after `withLinktimeProperties(v)` has been called.

There are several possible solutions to this issue:
* Rename `withLinktimeProperties` to `withCustomLinktimeProperties`, and add a getter `customLinktimeProperties` that returns exactly the custom ones, or
* Change `linktimeProperties` so that it returns exactly what was passed in `withLinktimeProperties`, and add the predefined properties later in the pipeline (not in `NativeConfig`).

I think the second option is better, as it removes any logic from `NativeConfig`. It is a data bag and nothing else.

sjrd (changes requested):

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
WojciechMazur: Ok, that's sound fine. Currently, there is only one place where we use link-time properties (`trait LinktimeValueResolver`). 
I'm wondering what would be the best place to define predefined values in. If we would leave them in `object NativeConfig` we would be able to show resolved values in sbt using `show nativeConfig` which might be useful for debugging. On the other hand, since those would be injected anyway maybe we should not show them at all?

sjrd: I think it's fine not to show them at all at the sbt level.

sjrd (approved):

PR #2380, opened 2021-09-22
Delay discovery for clang and clang++

ekrich:
This change starts moving towards late discovery. Previously, discovery occurred early and threw a build exception which was reported to not allow setting `clang` and `clang++` via env vars passed to `NativeConfig` methods. This more closely matches what happens when calling the build tools API directly. If we do this with all the `Discover` calls then much less would need to be called by other build tools or programs.

The other early discovery methods do not throw errors so they may not be a problem. You should be able to do the following (code not tested):

```scala
nativeConfig ~= {
  _.withLTO(LTO.thin)
  .withClang(System.getenv("CC"))
  .withClangPP(System.getenv("CXX"))
}
```

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -42,6 +42,7 @@ object ScalaNativePluginInternal {
   lazy val scalaNativeBaseSettings: Seq[Setting[_]] = Seq(
     crossVersion := ScalaNativeCrossVersion.binary,
     platformDepsCrossVersion := ScalaNativeCrossVersion.binary,
+    // deprecated loose settings in NativeConfig since 0.4.0
     nativeClang := nativeConfig.value.clang.toFile,
     nativeClangPP := nativeConfig.value.clangPP.toFile,
     nativeCompileOptions := nativeConfig.value.compileOptions,
@@ -55,15 +56,15 @@ object ScalaNativePluginInternal {
   )
 
   lazy val scalaNativeGlobalSettings: Seq[Setting[_]] = Seq(
-    nativeConfig := build.NativeConfig.empty
-      .withClang(interceptBuildException(Discover.clang()))
-      .withClangPP(interceptBuildException(Discover.clangpp()))
-      .withCompileOptions(Discover.compileOptions())
-      .withLinkingOptions(Discover.linkingOptions())
-      .withLTO(Discover.LTO())
-      .withGC(Discover.GC())
-      .withMode(Discover.mode())
-      .withOptimize(Discover.optimize()),
+    nativeConfig := {
+      build.NativeConfig.empty
+        .withCompileOptions(Discover.compileOptions())
+        .withLinkingOptions(Discover.linkingOptions())
+        .withLTO(Discover.LTO())
+        .withGC(Discover.GC())
+        .withMode(Discover.mode())
+        .withOptimize(Discover.optimize())
+    },
     nativeWarnOldJVM := {
       val logger = streams.value.log
       Try(Class.forName("java.util.function.Function")).toOption match {
@@ -97,6 +98,8 @@ object ScalaNativePluginInternal {
       workdir
     },
     nativeConfig := {
+      // convert deprecated loose settings to NativeConfig
+      // change to empty when loose settings are removed
       nativeConfig.value
         .withClang(nativeClang.value.toPath)
         .withClangPP(nativeClangPP.value.toPath)
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -58,12 +58,6 @@ object ScalaNativePluginInternal {
   lazy val scalaNativeGlobalSettings: Seq[Setting[_]] = Seq(
     nativeConfig := {
       build.NativeConfig.empty
-        .withCompileOptions(Discover.compileOptions())
-        .withLinkingOptions(Discover.linkingOptions())
-        .withLTO(Discover.LTO())
-        .withGC(Discover.GC())
-        .withMode(Discover.mode())
-        .withOptimize(Discover.optimize())
     },
     nativeWarnOldJVM := {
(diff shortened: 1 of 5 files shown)


ekrich:
Not sure the CI failures have anything to do do with this PR.

WojciechMazur:
Actually, this change has broken `link-order` scripted test, as it's using `nativeClang` setting on of its tasks. It would need to be adjusted. For testing locally you can use `sbtScalaNative/scripted run/link-order` to test only this single test

ekrich:
@WojciechMazur Thanks, the scripted test `link-order` does not use tools at all so I added the discovery. I think overall this PR is a good idea as it can make it easier for tool integration and removes code from the `sbt` plugin.

ekrich:
@WojciechMazur Do you think failure is spurious? I am pretty much done but need to do some testing and maybe change the docs a bit. Initial review might be nice for suggestions etc.

ekrich:
@WojciechMazur This can be reviewed now. I would like to do some more testing and some documentation if we decide to proceed. The plus of this is from the tools integration point of view, only `clang/clang++` have to be on the path. The default `NativeConfig` will work. Once we get rid of the deprecated bare settings then the `sbt` integration get even cleaner.

WojciechMazur:
To be honest I'm not a real fan of introducing the `complete` method and need explicit invocation of it to get the actual state of the configuration. What if we would try to change the implementation of NativeConfig instead to some more implicit. `NativeConfig.Impl` class is a private implementation detail, what if we would move responsibility for to its implementation. What I have in my mind is as an example  
```
case class Impl(clangOpt: Option[String]) extends NativeConfig {
 // following implements NativeConfig methods
 lazy val clang: String = clangOpt.getOrElse(Discover.clang())
 def withClang(clang: String): NativeConfig = copy(clangOpt = Some(clang))
}
```
This way we would never need to think about explicit completion of NativeConfig and at the same time its evaluation would be delayed until the first usage of a given setting.
Unfortunately, this would collide with another model defining what NativeConfig is as describe in sjrd comment in my PR today: 
https://github.com/scala-native/scala-native/pull/2377#discussion_r716631548
What do you about that?

ekrich:
I agree that this change is not in line with @sjrd 's comment and what you would think a configuration object should represent. Does Scala.js have a concept of Discovery that is similar because I feel that discovery and exceptions are not very aligned either?

I think that `complete` is like `validate` and to me that could be very helpful. Right now we only check the version when we discover but discovering individual items means there is no possibility of looking for sanity and perhaps warning if the user has 2 different compiler versions or checking if the Rust version is compatible with LTO or numerous other things that could be checked. Currently we don't save the compiler versions either but that could be helpful. If we know better we could save a lot of headache by checking various things. Rust example: https://doc.rust-lang.org/rustc/linker-plugin-lto.html So I don't think calling a validate is a bad thing because then at the last minute we check everything before proceeding.

I think the idea of localizing the individual settings with lazy vals is perhaps better in some ways as it encapsulates that logic in the `NativeConfig` impl and then the data from Discover can be augmented by the user if desired. I can go down this path but I am unsure at this point. Would we consider `checkClangVersion` of user supplied `clang` and `clang++` for minimum version or just let that go? Would it be possible to validate at a future date or leave an empty stub or perhaps put validate on the config object?

ekrich:
@WojciechMazur I started a new PR and see a few problems in the system as a whole. 

1. The raw sbt settings probably need to be dropped as they inadvertently cause detection in the plugin.
2. Trying to do the `lazy val` approach seems like it would work but each time a `NativeConfig` is cloned via modification and the values are asked for again it causes a new detection - this doesn't cause a problem as it is idempotent but it causes extra computation which also make reasoning more difficult if debugging.

I think this is a very important PR in theory but these problems and the detection for the sake of the environment variables is not desirable which was mentioned by @sjrd to not add new ones. If there is a way to create `NativeConfig` in CI using env vars like the example in the first comment above we could validate the paths and versions and such and avoid most of the discovery.

Any help appreciated - I can also push my new PR if that would be helpful.

WojciechMazur:
Thank you @ekrich. I agree about the removal of deprecated settings, would aim with that change to happen in the 0.5.x release. 
I also agree about the issue with `lazy vals`, even though in most cases settings should be resolved only once - in most cases, the usage happens after setting up all setting keys. I'm curious about the PR you spoke about, I'll be glad to see it (it can also be just a link to the repo to get the idea what it contains)
I also had recently thought about yet another approach - what if we would encapsulate setting keys in some kind of (immutable) wrapper (over the mutable underlying value) that would be responsible for caching and validation? This way we would be able to safely copy it's instances. I would think about replacing it with something like sbt settings. I don't have yet any clear implementation in my mind, but it can contain some interface like on the snippet below. What do you think?

```
trait SettingKey[T]{  {
 private var defined: Option[T] = None
 private lazy val discovered: Option[T] = None
 def discover(): Option[T] = {
  discovered = ???
 }
 def validate(v: T): Boolean = ???

 def get(): T = defined
   .orElse(discovered)
   .filter(validate)
   .getOrElse(throw BuildException("Canot resolve setting key"))
   
  def set(v: T) = if(validate(t)) {
    defined = Some(v)
  } else throw new ValidationException()
}
```
And then we would define NativeConfig something like
```
class Impl(clangSetting: SettingKey[String]) extends NativeConfig {
  def clang(): String: clangSetting.get()
  def withClang(v: String): String = clangSetting.set(v)
}
```

In perfect case NativeConfig.Impl would no longer be a box of values, but rather a box of receipts for values. I'm not sure if it's clear enough

ekrich:
Sure, let me work on the new PR a bit more and then I can push it. I do think I may be able to get around some of the weirdness. I also agree the immutable setup we have now is a bit awkward but I am not ready to give up on it yet. I also have and idea about separating discovery from validation which may be ok - the super late discovery in this PR is not great and trying to do the validation in `complete` is not good either.

ekrich:
I tried this approach and it didn't got that well. Maybe I didn't use the case class approach and that caused the issues I was having on the clone that happens. The new attempt is certainly not a box a values but I think it may at least illustrated my thinking. https://github.com/scala-native/scala-native/pull/3110

ekrich:
Late discovery is not super great because then the global instance of `NativeConfig` is empty so showing it is pretty much useless. I tried various approaches and none are as clean and simple as the case class approach to `NativeConfig`.
The global config does discovery on `sbt` project load so we get early errors if something is not correct. Removing all the logic out of `sbt` and into the build will really help build tool implementers. This was merged via PR # https://github.com/scala-native/scala-native/pull/3125

PR #2389, opened 2021-09-28
Skip linking on subsequent nativeLink calls in SN sbt plugin

jchyb:
Added file information caching based on the last modified date of a file. If a file was changed since the last `nativeLink`, the project is rebuild and new file information about the classpath and output is cached. This allows the plugin to skip the building process if no files were modified, resulting in quicker subsequent runs.


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -109,28 +110,75 @@ object ScalaNativePluginInternal {
...
-
-        val logger = streams.value.log.toLogger
-        build.Config.empty
-          .withLogger(logger)
-          .withMainClass(maincls)
-          .withClassPath(classpath)
-          .withWorkdir(cwd)
-          .withCompilerConfig(nativeConfig.value)
+
+        FileInfo.hash(outpath)
```
WojciechMazur, line 138: I think it would be great if we could move hashing the output path outside this method to follow the single-responsibility principle. Also since we have the method `buildIfChanged` maybe let's rename it to simply `build`?  
I'm also wondering wouldn't last modified info be enough for our use case, instead of hashing its binary content.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -109,28 +110,75 @@ object ScalaNativePluginInternal {
...
+                Tracked
+                  .lastOutput[Seq[HashFileInfo], HashFileInfo](
+                    cacheFactory.make("outputFileInfo")
+                  ) {
+                    case (_, None) => makeBuild()
+                    case (_, Some(outHashInfo)) =>
+                      if (changed || outHashInfo != FileInfo.hash(outpath))
+                        makeBuild()
+                      else outHashInfo
+                  }
```
WojciechMazur, line 161: This probably can be also squished to only two branches, which can be helpful if we would move hashing out of makeBuild() body
```suggestion
                  ) { (_, prev) => 
                   val outputHashInfo = FileInfo.hash(outpath) 
                   if(changed || !prev.contains(outputHashInfo)) {
                     makeBuild()
                   }
                   outputHashInfo
                  }
```

WojciechMazur (commented):
I've tried it locally and works really nice, especially when trying to run tests multiple times in a row, when trying to catch a spurious bug, thank you! 
I only have some minor comments, overall looks good.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
jchyb: Initially last modified was used, but when trying to fix some other (unrelated) issue, I found this: https://www.boost.org/doc/libs/1_32_0/libs/filesystem/doc/operations.htm#last_write_time , specifically the warning under last_write_time. Apparently the last modified value can have quite a low resolution depending on the os and filesystem, although I have not found any more info than this. I simply thought of hashing as a safer choice.

WojciechMazur (approved):
In failures in Windows CI should be fixed in merged #2391. Can you please rebase the branch on current upstream to make sure that it passes the tests

ekrich:
I know that the sbt stuff to support this is a big deal but it would be great if this was in tools so that Mill doesn't have to do something similar. Maybe create an issue for some future person?

PR #2390, opened 2021-09-29
Fix invalid nullability resolution for boxed primitives

WojciechMazur:
This PR fixes recently observed failures in the CI. The previous resolution of nullability of boxed primitives was wrong and probably was the aftermath of the bug fixed in #2133 
Since Boxed pointers were treated as NOT nullable, the optimizer removed the `redundant` check to make `Ptr[_] == null` comparison. We can only treat numeral primitives as not nullable, where pointers should always to treated as potentially containing null


sjrd (approved):
LGTM if it fixes the CI.

PR #2391, opened 2021-09-30
Enforce usage of LLVM 11 in Windows CI

WojciechMazur:
Since 29th September some Windows CI runners can be found to have preinstalled LLVM 12.0.0. To make our tests consistent and to follow testing with a minimum supported version we need to enforce usage of the older versions. Otherwise, `choco` used to install our dependencies would exit with an error.


sjrd (approved):

ekrich:
Do we want to add the minimum value for Windows into our Discovery?

PR #2393, opened 2021-09-30
Add missing FileChannel functionality in javalib

jchyb:
Add missing FileChannel functionality
This includes:
* Now FileInputStream, FileOutputStream and RandomAccessFile can return a FileChannel, with file pointers being consistent between the two (like it happens in JVM)
* With that came a minor refactorization of the four classes. Before, a FileChannel was composed of a Random Access File, which could be composed of a FileInputStream and FileOutputStream. Now, RAF can contain all three of the other, with FileInputStream and FileOutputStream depending on FIleChannel as well. To achieve this, some methods were moved from the three classes to FileChannel.
* File locking via lock() and tryLock() methods, with majority of the tests ported from Apache Harmony.
* MappedByteBuffer support via the map() method, with finalization done through WeakReferenceRegistry implemented in a previous PR. This is important, as Java Api does not provide any other way to unmap a file. Some of the tests were ported from apache harmony.

The finalization ended up being somewhat messy. My hope was to use partially applied functions, however tracing references in GC revealed that `this` is always being passed to them, so the data and method was moved to another class, which also registers the function to WeakReferenceRegistry. Perhaps we should redesign WeakReferenceRegistry to simply contain Finalization objects instead of functions, to remove ambiguity about the feature (I’m thinking about the potential future contributors here).

### DIFF
--- a/javalib/src/main/scala/java/nio/ByteArrayBits.scala
+++ b/javalib/src/main/scala/java/nio/ByteArrayBits.scala
@@ -1,9 +1,11 @@
 package java.nio
 
+import scala.scalanative.unsafe.Ptr
+
 // Ported from Scala.js
 private[nio] object ByteArrayBits {
   def apply(
-      array: Array[Byte],
+      array: Ptr[Byte],
       arrayOffset: Int,
       isBigEndian: Boolean,
       indexMultiplier: Int = 1
@@ -13,7 +15,7 @@ private[nio] object ByteArrayBits {
 
 @inline
 private[nio] final class ByteArrayBits(
-    array: Array[Byte],
+    array: Ptr[Byte],
     arrayOffset: Int,
     isBigEndian: Boolean,
     indexMultiplier: Int
--- a/javalib/src/main/scala/java/nio/CharBuffer.scala
+++ b/javalib/src/main/scala/java/nio/CharBuffer.scala
@@ -152,7 +152,7 @@ abstract class CharBuffer private[nio] (
   override def toString(): String = {
     if (_array != null) {
       // even if read-only
-      new String(_array.toArray, position() + _arrayOffset, remaining())
+      new String(_array.toArray(), position() + _arrayOffset, remaining())
     } else {
       val chars = new Array[Char](remaining())
       val savedPos = position()
--- a/javalib/src/main/scala/java/nio/CharBuffer.scala
+++ b/javalib/src/main/scala/java/nio/CharBuffer.scala
@@ -21,7 +21,7 @@ object CharBuffer {
 
 abstract class CharBuffer private[nio] (
     _capacity: Int,
-    private[nio] val _array: Array[Char],
+    private[nio] val _array: GenArray[Char],
     private[nio] val _arrayOffset: Int
 ) extends Buffer(_capacity)
     with Comparable[CharBuffer]
@@ -40,7 +40,7 @@ abstract class CharBuffer private[nio] (
     if (n == 0) -1
     else if (_array != null) {
       // even if read-only
-      target.put(_array, _arrayOffset, n)
+      GenBuffer(target).generic_put(_array, _arrayOffset, n)
       n
     } else {
       val savedPos = position()
@@ -66,7 +66,7 @@ abstract class CharBuffer private[nio] (
 
   @noinline
   def get(dst: Array[Char], offset: Int, length: Int): CharBuffer =
-    GenBuffer(this).generic_get(dst, offset, length)
+    GenBuffer(this).generic_get(ScalaArray(dst), offset, length)
 
   def get(dst: Array[Char]): CharBuffer =
     get(dst, 0, dst.length)
@@ -77,7 +77,7 @@ abstract class CharBuffer private[nio] (
 
   @noinline
   def put(src: Array[Char], offset: Int, length: Int): CharBuffer =
-    GenBuffer(this).generic_put(src, offset, length)
+    GenBuffer(this).generic_put(ScalaArray(src), offset, length)
 
   final def put(src: Array[Char]): CharBuffer =
     put(src, 0, src.length)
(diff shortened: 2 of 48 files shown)


ekrich:
Jan, are you on Discord? https://discord.gg/scala

jchyb:
Yes, with the same nickname (jchyb). Sorry for the delayed response, tomorrow I should be more easily available.

javalib/src/main/scala/java/nio/CharBuffer.scala
```diff
@@ -40,7 +40,7 @@ abstract class CharBuffer private[nio] (
     if (n == 0) -1
     else if (_array != null) {
       // even if read-only
-      target.put(_array, _arrayOffset, n)
+      GenBuffer(target).generic_put(_array, _arrayOffset, n)
```
WojciechMazur, line 43: Isn't that a spurious change? In other files you have not changed used methods

javalib/src/main/scala/java/nio/GenHeapBuffer.scala
```diff
@@ -100,7 +100,13 @@ private[nio] final class GenHeapBuffer[B <: Buffer](val self: B)
 
     val len = remaining()
     System
-      .arraycopy(_array, _arrayOffset + position(), _array, _arrayOffset, len)
+      .arraycopy(
+        _array.toArray(),
```
WojciechMazur, line 104: We probably should make _array.toArray() a val, otherwise we risk creation and copying of two arrys

javalib/src/main/scala/java/nio/GenHeapBufferView.scala
```diff
@@ -105,9 +105,9 @@ private[nio] final class GenHeapBufferView[B <: Buffer](val self: B)
     val len = remaining()
     val bytesPerElem = newHeapBufferView.bytesPerElem
     System.arraycopy(
-      _byteArray,
+      _byteArray.toArray(),
```
WojciechMazur, line 108: Also here we should make it a val

javalib/src/main/scala/java/nio/ByteArrayBits.scala
```diff
@@ -1,9 +1,11 @@
 package java.nio
 
+import scala.scalanative.unsafe.Ptr
+
 // Ported from Scala.js
 private[nio] object ByteArrayBits {
   def apply(
-      array: Array[Byte],
+      array: Ptr[Byte],
```
WojciechMazur, line 8: Probalby, in the place of each usage we already have Array[Byte] so this change probably can be reverted

javalib/src/main/scala/java/nio/GenArray.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.scalanative.unsafe._
+import scala.scalanative.runtime.ByteArray
+
+private[nio] sealed trait GenArray[T] {
+  def update(index: Int, value: T): Unit
+  def apply(index: Int): T
+  def toArray(): Array[T]
+  def toPtr(): Ptr[Byte]
+  val length: Int
+}
```
WojciechMazur, line 11: Missing blank line

javalib/src/main/scala/java/nio/MappedByteBufferImpl.scala
```diff
@@ -0,0 +1,162 @@
...
+            fd.handle,
+            null,
+            flProtect,
+            0.toUInt,
+            0.toUInt,
+            null
+          )
+        if (mappingHandle == null) throwException()
+
+        val dwFileOffsetHigh = (position >>> 32).toInt.toUInt
```
WojciechMazur, line 124: We should be albe to make toUInt directly

javalib/src/main/scala/java/nio/MappedByteBufferImpl.scala
```diff
@@ -0,0 +1,162 @@
...
+        if (isWindows) {
+          channel.position(prevSize)
+          for (i <- prevSize until minSize)
+            channel.write(ByteBuffer.wrap(Array[Byte](0.toByte)))
+          channel.position(prevPosition)
+        }
+      }
+    }
+
+    val (ptr: Ptr[Byte], windowsMappingHandler: Option[Handle]) =
```
WojciechMazur, line 106: This funciton is quite long, maybe let's split it into two os-specic methods at this point each returning Impl. `throwException` can be moved to object as private, and probably with changed name, eg. `failMapping`

WojciechMazur (commented):
Found a few places that might need a fixes

javalib/src/main/scala/java/nio/channels/FileLock.scala
```diff
@@ -34,7 +36,7 @@ abstract class FileLock private (
     shared
 
   final def overlaps(pos: Long, siz: Long): Boolean =
-    (pos + siz) >= position || (position + size) >= pos
+    (pos + siz) > position && (position + size) > pos
```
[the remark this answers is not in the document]
jchyb: Yes, it failed the harmony tests and conceptually does not make sense either. (f.e. (pos=0, siz=1), (position=10, size=1) would clear the inequality, and yet those do not overlap)

javalib/src/main/scala/java/nio/CharBuffer.scala
jchyb: It's necessary, other files always use the GenBuffer abstraction to avoid changing public api, so there was no need to change anything there. target.put() takes an Array, while GenArray().generic_put takes a GenArray, which is what we have now.

javalib/src/main/scala/java/nio/ByteArrayBits.scala
jchyb: In HeapByteBuffer we use it to write. When we change ScalaArray to Ptr[Byte], changes to Ptr[Byte] are still saved to ScalaArray, but changing PtrArray to Array, the changes are not saved to the Ptr[Byte], as the data is copied from there, not cast. So we have to pass Ptr[Byte] there. As for concerns about the GC, ByteArrayBits is always used in companion of other classes, which hold the reference to the real Array, so the Ptr[] data will not be cleaned preemptively.

PR #2394, opened 2021-09-30
Sbt nativeLink command should fail in case of clang failure

### LINKED ISSUE #2363: Sbt should fails when scala native can't link anything
When I run `nativeLink` and it fails, the `sbt` says it success and haven't return non zero code:
```
[error] ld: symbol(s) not found for architecture x86_64
[error] clang: error: linker command failed with exit code 1 (use -v to see invocation)
[info] Linking native code (immix gc, thin lto) (17631 ms)
[info] Total (48158 ms)
[success] Total time: 81 s (01:21), completed Sep 9, 2021 1:24:18 PM
```

WojciechMazur:
This PR fixes #2363, now in case of clang failure when compiling or linking it would throw appropriate BuildException.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -65,7 +65,7 @@ private[scalanative] object LLVM {
         val result = Process(compilec, config.workdir.toFile) ! Logger
           .toProcessLogger(config.logger)
         if (result != 0) {
-          sys.error(s"Failed to compile ${inpath}")
+          throw new BuildException(s"Failed to compile ${inpath}")
         }
       }
       objPath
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -131,9 +131,11 @@ private[scalanative] object LLVM {
       s"Linking native code (${config.gc.name} gc, ${config.LTO.name} lto)"
     ) {
       config.logger.running(compile)
-      Process(compile, config.workdir.toFile) ! Logger.toProcessLogger(
-        config.logger
-      )
+      val result = Process(compile, config.workdir.toFile) ! Logger
+        .toProcessLogger(config.logger)
```
sjrd, line 135: This reads very weirdly. Can you convince scalafmt to format it as
```suggestion
      val result = Process(compile, config.workdir.toFile) !
        Logger.toProcessLogger(config.logger)
```
?

sjrd (approved):

tools/src/main/scala/scala/scalanative/build/LLVM.scala
WojciechMazur: Yes, that's passible. Fixed

PR #2397, opened 2021-10-01
Fix #2396: URI normalize doesn't seem to work

### LINKED ISSUE #2396: URI normalize doesn't seem to work
```scala
val uri = new URI("/a/b")
uri.normalize()
```

In JVM this returns "/a/b", in Native we get:
```
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
    at java.lang.String.substring(Unknown Source)
    at java.net.URI.normalize(Unknown Source)
```

jchyb:
Before, it used to always throw StringIndexOutOfBoundsException. Now, related instructions have been swapped, so this should no longer be the case. Additional tests were also added.

### DIFF
--- a/unit-tests/shared/src/test/scala/javalib/net/URITest.scala
+++ b/unit-tests/shared/src/test/scala/javalib/net/URITest.scala
@@ -237,4 +237,33 @@ class URITest {
     }
   }
 
+  @Test def normalize(): Unit = {
+    def testNormalize(relative: Boolean): Unit = {
+      val first = if (relative) "" else "/"
+      assertEquals(new URI(s"${first}a/b"), new URI(s"${first}a/b").normalize())
+      assertEquals(
+        new URI(s"${first}a/b"),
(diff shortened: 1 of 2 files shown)


unit-tests/shared/src/test/scala/javalib/net/URITest.scala
```diff
@@ -237,4 +237,33 @@ class URITest {
     }
   }
 
+  @Test def normalize(): Unit = {
+    def testNormalize(relative: Boolean): Unit = {
+      val first = if (relative) "" else "/"
+      assertEquals(new URI(s"${first}a/b").normalize(), new URI(s"${first}a/b"))
```
WojciechMazur, line 243: Please make the expected result be present on the left side of `assertEquals`, and the actual value on the right side. That's an assumption taken by JUnit, otherwise it would lead to confusing messages in the logs in case of the failed tests.

WojciechMazur (approved):

PR #2405, opened 2021-10-05
Explain how to create universal binary on macOS

catap:


docs/user/sbt.rst
```diff
@@ -215,6 +215,26 @@ Use the following approach in `sbt` to set the target triple:
...
+.. code-block:: scala
+
+    lazy val sandbox64 = project.in(file("sandbox"))
+        .settings(nativeConfig ~= { _.withTargetTriple("arm64-apple-darwin20.6.0") })
+
+    lazy val sandboxM1 = project.in(file("sandbox"))
+        .settings(nativeConfig ~= { _.withTargetTriple("x86_64-apple-darwin20.6.0") })
+
+This `build.sbt` produces two different binary, one for `x86_64` and another one for `arm64`.
+You may easy to combine it to one so called fat binary or universal binary via
```
WojciechMazur, line 230:
```suggestion
You may easily combine them to one so called fat binary or universal binary via
```

docs/user/sbt.rst
```diff
@@ -215,6 +215,26 @@ Use the following approach in `sbt` to set the target triple:
...
+
+.. code-block:: sh
+
+     lipo -create sandbox64/target/scala-2.12/sandbox64-out sandboxM1/target/scala-2.12/sandboxM1-out -output sandbox-out
+
+which produces `sandbox-out` which can be used at any platform.
```
WojciechMazur, line 236:
```suggestion
which produces `sandbox-out` that can be used at any platform.
```
Or even better we can include it the sentence before the code block

docs/user/sbt.rst
```diff
@@ -215,6 +215,26 @@ Use the following approach in `sbt` to set the target triple:
...
+
+    lazy val sandboxM1 = project.in(file("sandbox"))
+        .settings(nativeConfig ~= { _.withTargetTriple("x86_64-apple-darwin20.6.0") })
+
+This `build.sbt` produces two different binary, one for `x86_64` and another one for `arm64`.
+You may easy to combine it to one so called fat binary or universal binary via
+
+.. code-block:: sh
+
+     lipo -create sandbox64/target/scala-2.12/sandbox64-out sandboxM1/target/scala-2.12/sandboxM1-out -output sandbox-out
```
WojciechMazur, line 234: Some links to the lipo project/reference might be helpful

docs/user/sbt.rst
```diff
@@ -215,6 +215,26 @@ Use the following approach in `sbt` to set the target triple:
 
     nativeConfig ~= { _.withTargetTriple("x86_64-apple-macosx10.14.0") }
 
+you may create a few dedicated projects with different target triple.
+For example if I have two `sandbox` project like:
```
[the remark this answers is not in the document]
WojciechMazur: Since lipo is macOS specific we might include this information here
```suggestion
For example, if I have multiple project definitions for different macOS architectures, eg:
```

WojciechMazur (changes requested):
Thank you for adding this guide to the docs, there are some typos that need to be fixed. I think lipo is macOS specific, do you know if there is some similar project for other OSs?

catap:
@WojciechMazur yes, `lipo` is apple specified things and it can be used for `macOS` and `iOS` only.

I have remember a project with name `FatELF` or something like that, which included some linux kernel patches and gcc patches.

Anyway, it was more than 10 years ago when I tried to do it on Linux and I have no idea regarding status of this project.

catap:
Indeed, it is called `FatELF` and it is still requires a lot of patches: https://github.com/icculus/fatelf/tree/main/patches

PR #2427, opened 2021-11-02
Support Scala 2.13.7

WojciechMazur:
This PR adds Scala 2.13.7 to the sbt build. 

* Fixed error with diverging implicits (implicit conversions needs to be explicitly imported)
* Made Scala 2.13.7 default version for Scala 2.13 builds
* Updated scalalib override patches 
* Updated publishing scripts to contain Scala 2.13.7 + small refactor
* Updated CI to use Scala 2.13.7 as default
* Updated MiMa checks to use also 0.4.1 artifact


unit-tests/native/src/test/scala/scala/scalanative/unsafe/CArrayOpsTest.scala
```diff
@@ -5,6 +5,8 @@ import org.junit.Test
 import org.junit.Assert._
 
 import scalanative.unsafe.Nat._
+// Scala 2.13.7 needs explicit import for implciit conversions
```
sjrd, line 8:
```suggestion
// Scala 2.13.7 needs explicit import for implicit conversions
```
Same elsewhere.

sjrd (approved):

PR #2431, opened 2021-11-06
Fix #2418: Building docs should be a part of the CI.

### LINKED ISSUE #2418: Building docs should be a part of the CI

ankusharya:
In this commit:
- Add action `build-docs`
- Modify `docs/user/sbt.rst` to remove the warning

Explanation about `build-docs` action
- Build on top of [sphinx-action](https://github.com/marketplace/actions/sphinx-build)
- Update Sphinx to `4.2.0` 
- Warning in building docs bubble as GitHub warning
- Action `build-docs` is tested locally by [nektos/act](https://github.com/nektos/act)

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -3,13 +3,13 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - main
 jobs:
   build-docs:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: ammaraskar/sphinx-action@0.4
         with:
-          pre-build-command: "pip install Sphinx==4.2.0 recommonmark==0.7.1"
           docs-folder: "docs/"
+          build-command: 'sphinx-build -b html . _build/html -W --keep-going -n'
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -11,5 +11,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ammaraskar/sphinx-action@0.4
         with:
+          pre-build-command: "pip install Sphinx==7.0.1 recommonmark==0.7.1"
           docs-folder: "docs/"
-          build-command: 'sphinx-build -b html . _build/html -W --keep-going -n -Dexclude_patterns=["**/requirements.txt"]'
+          build-command: 'sphinx-build -b html . _build/html -W --keep-going -n'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+  push:
+    branches:
+      - master
+jobs:
+  build-docs:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ammaraskar/sphinx-action@master
```
WojciechMazur, line 12: We should use a fixed version, current lastest is 0.4
```suggestion
      - uses: ammaraskar/sphinx-action@0.4
```

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,15 @@
...
+        with:
+          pre-build-command: "pip install Sphinx==4.2.0 recommonmark==0.7.1"
+          docs-folder: "docs/"
```
WojciechMazur, line 15: The current setup would not treat warnings as errors, most of the problems with our docs building in the past were in most cases informed as warnings, but they were ignored, eg. missing entry in the index.  We should add "-W" option to enable fatal warnings, `--keep-going` to not stop after the first warning and optionally `-n` which would try to check for missing references
```suggestion
          build-command: 'make html SPHINXOPTS="-W --keep-going -n"'
          docs-folder: "docs/"
```

WojciechMazur (commented):
Thank you for your contribution. I have two comments that should be addressed before merging this change.

.github/workflows/build-docs.yml
ankusharya: Hi, 
Before raising pull request, I thoroughly checked for all the options to make warning as error. 
I have checked on adding `-W` but it does not work.
To solve it I had already raised issue.
- [Add -W support](https://github.com/ammaraskar/sphinx-action/issues/34)
- [Add number of warning as step output](https://github.com/ammaraskar/sphinx-action/issues/35)

Even with one issue resolved, we can enable warning as error.

I also looked into catching `warning` as they bubble as Github warning, but i dont see any reference of any such `workflow commands`
If you find any resources/approach, please share. I will implement it.
Any thoughts?

WojciechMazur: @ankusharya I think that maybe problems with log files might be also a problem with running with act. We should check if it behaves the same in GH Actions environments in the same way. I don't have a large experience with this tool, but even then I came to corner cases where build was failing locally with act and worked well in the CI

PR #2434, opened 2021-11-12
Fix #2433: spaces not being allowed in Paths on Windows

### LINKED ISSUE #2433: Spaces in paths not supported in javalib on Windows
Example:
```scala
import java.nio.file.Paths
object Main {
    val path = Paths.get("test file")
  }
}
```
Native bahavior:
```
java.nio.file.InvalidPathException: java.nio.file.InvalidPathException: Illegal character ` ` in path at index 4: test file
        at scala.collection.immutable.WrappedString$$$Lambda$1.apply(Unknown Source)
        at scala.scalanative.nio.fs.windows.WindowsPathParser$.apply(Unknown Source)
        at Main$.main(Unknown Source)
        at <none>.__scrt_common_main_seh:(d:\a01\_work\2\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288)(Unknown Source)
        at <none>.BaseThreadInitThunk(Unknown Source)
        at <none>.RtlUserThreadStart(Unknown Source)
```
Expected (jvm) behavior:
Runs fine, we can of course also do any actual operations on that file.

jchyb:
According to the Windows documentation, merely characters with codes 0-31 are considered to be the control characters forbidden for paths. Since spaces (of code 32) are allowed, they were removed from the check.

A test for the issue was also added.

### DIFF
--- a/javalib/src/main/scala/scala/scalanative/nio/fs/windows/WindowsPathParser.scala
+++ b/javalib/src/main/scala/scala/scalanative/nio/fs/windows/WindowsPathParser.scala
@@ -121,7 +121,8 @@ object WindowsPathParser {
       allowedChars: String = ""
   ): String = {
     def isControlOrReservedChar(c: Char) = {
-      c < '\u0020' || pathReservedChars.contains(c)
+      // '\u0001F' - last control character (no. 31)
+      c <= '\u001F' || pathReservedChars.contains(c)
     }
     def checkValidSegment(
         segment: String,
--- a/javalib/src/main/scala/scala/scalanative/nio/fs/windows/WindowsPathParser.scala
+++ b/javalib/src/main/scala/scala/scalanative/nio/fs/windows/WindowsPathParser.scala
@@ -120,8 +120,8 @@ object WindowsPathParser {
       to: Int,
       allowedChars: String = ""
   ): String = {
-    def isSpaceOrReservedChar(c: Char) = {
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/scala/scalanative/nio/fs/windows/WindowsPathParser.scala
```diff
@@ -120,8 +120,8 @@ object WindowsPathParser {
       to: Int,
       allowedChars: String = ""
   ): String = {
-    def isSpaceOrReservedChar(c: Char) = {
-      c.isSpaceChar || pathReservedChars.contains(c)
+    def isControlOrReservedChar(c: Char) = {
+      c < '\u0020' || pathReservedChars.contains(c)
```
ekrich, line 124: I would consider either `val space = '\u0020' or leaving a comment at this location to make this a touch more readable. Other than that this looks great.

ekrich (changes requested):

javalib/src/main/scala/scala/scalanative/nio/fs/windows/WindowsPathParser.scala
jchyb: Sure, good idea

ekrich (approved):
LGTM, a better change than my idea.

WojciechMazur (approved):

PR #2441, opened 2021-12-01
Refactor build to use MultiScalaProjects

WojciechMazur:
This PR does change our build definition to allow for safe cross-compilation with the upcoming Scala 3 support, however it does not include yet any Scala 3 specific changes. 

* All projects (except `sbtScalaNative`) are now defined as `MultiScalaProject` known from Scala.js - it means each binary version of each project has it's own individual projects, eg. `scalalib` has now subprojects: `scalalib2_11`, `scalalib2_12`, `scalalib2_13`. This change was needed for the incoming need of selective dependency on Scala 3 or Scala 2.13 dependencies. 
Also, this would allow us as again to run scripted-tests with Scala version other then Scala 2.12
* Content of existing `build.sbt` has been moved and splitter to `project/Build.scala`, `project/Settings.scala`, project/Deps.scala` and `project/Commands.scala` 
* Command aliases, eg. `test-runtime` are now replaced with full commands taking binary scala version as an argument or using currently used scalaBinaryVersion used in the root project.
* Github workflow definitions were adjusted to use the new syntax for commands or running single tasks withing projects. Additionally, concurrency settings were added to the workflows allowing to cancel current CI run in case if new commit would be pushed to the same branch.
* Publishing scripts were adjusted to handle new projects layout. 

The largest drawback of the new build is increased time of loading time when starting sbt, since we define 3x more projects which needs to be resolved


ekrich:
Quick question. If at some time we want a native `tools` project then would that become a separate project or is there another way to do cross projects?

WojciechMazur:
> Quick question. If at some time we want a native `tools` project then would that become a separate project or is there another way to do cross projects?

Seems a bit unlikely to happen, since we would need still JVM and scalac to compile Scala, but yes, probably we would create a new project, similarly to `tests` and `testsJVM` as it would be easier then to ensure compatibility with external tools like sbt-cross-project

ekrich:
Another question, do you plan to add the ability to get the Java version so we can add tests for `require-jdk11` etc? We really need to have the ability to test new Java features to support JDK11 and JDK17. As seen here - https://github.com/scala-js/scala-js/blob/main/project/Build.scala#L1799-L1814

PR #2442, opened 2021-12-02
Clean up reading unix env

ekrich:


WojciechMazur (approved):
Good to know that the workaround is no longer needed, good job!

PR #2443, opened 2021-12-05
Fix #735: improve throughput of zone allocation

### LINKED ISSUE #735: Optimize zone allocation for speed
https://github.com/scala-native/scala-native/pull/734 introduces initial support for zone allocation. 

1. Unlike [scala-offheap](http://github.com/densh/scala-offheap) we don't do anything remotely smart to improve the performance of the zone allocator yet. Performance of both `alloc` and `close` can be substantially improved by using a memory-pool of pages with bump allocation. The core memory pool management can be ported from [scala-offheap](http://github.com/densh/scala-offheap) with minimal changes.

2. We should optimize away closure allocation that demarcates the start and end of the zone.

densh:
This change replaces naive implementation of the zone allocator we had previously with one that performs bump allocation for small allocations (<2K).

Allocation throughput after this change:

```
2 bytes : 21115 ns / 1000 allocations
4 bytes : 20467 ns / 1000 allocations
8 bytes : 20535 ns / 1000 allocations
16 bytes : 20005 ns / 1000 allocations
32 bytes : 19696 ns / 1000 allocations
64 bytes : 19915 ns / 1000 allocations
128 bytes : 20309 ns / 1000 allocations
256 bytes : 27632 ns / 1000 allocations
512 bytes : 41469 ns / 1000 allocations
1024 bytes : 55459 ns / 1000 allocations
2048 bytes : 83120 ns / 1000 allocations
4096 bytes : 773972 ns / 1000 allocations
8192 bytes : 1755135 ns / 1000 allocations
16384 bytes : 4041602 ns / 1000 allocations
```

Allocation throughput before this change:

```
2 bytes : 81216 ns / 1000 allocations
4 bytes : 81147 ns / 1000 allocations
8 bytes : 80590 ns / 1000 allocations
16 bytes : 82284 ns / 1000 allocations
32 bytes : 85628 ns / 1000 allocations
64 bytes : 80658 ns / 1000 allocations
128 bytes : 90329 ns / 1000 allocations
256 bytes : 91666 ns / 1000 allocations
512 bytes : 93411 ns / 1000 allocations
1024 bytes : 122811 ns / 1000 allocations
2048 bytes : 419922 ns / 1000 allocations
4096 bytes : 1087628 ns / 1000 allocations
8192 bytes : 2239114 ns / 1000 allocations
16384 bytes : 4660193 ns / 1000 allocations
```

Additionally, this change dramatically reduces memory overhead of small zone allocations. Previously each allocation had a corresponding malloc invocation, that was persisted in a linked list of boxed pointers. Now, the linked list only stores large contiguous memory areas and zone allocator bump allocates within them, without any additional book keeping per allocation.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/MemoryPool.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/MemoryPool.scala
@@ -1,0 +1,185 @@
+package scala.scalanative.runtime
+
+import scala.scalanative.unsafe._
+import scala.scalanative.unsigned._
+
+/** Efficient pool of fixed-size memory pages. Allocations from underlying
+ *  allocator are performed in big chunks of memory that are sliced into pages
+ *  of requested size.
+ *
+ *  Pages and chunks are organized in an intrusive linked list way to minimise
+ *  memory overhead and re-use the same nodes for the whole lifetime of the
+ *  pool.
+ *
+ *  Memory is reclaimed back to underlying allocator once the pool is finalized.
+ */
+final class MemoryPool private {
+  private[this] var chunkPageCount: ULong = MemoryPool.MIN_PAGE_COUNT
+  private[this] var chunk: MemoryPool.Chunk = null
+  private[this] var page: MemoryPool.Page = null
+  allocateChunk()
+
+  /** Allocate a chunk of memory from system allocator. */
+  private def allocateChunk(): Unit = {
+    if (chunkPageCount < MemoryPool.MAX_PAGE_COUNT) {
+      chunkPageCount *= 2.toULong
+    }
+    val chunkSize = MemoryPool.PAGE_SIZE * chunkPageCount
+    val start = libc.malloc(chunkSize)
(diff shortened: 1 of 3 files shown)


densh:
The benchmark that I used for throughput estimates:
```
object Test {
  def run(count: Int): Long = {
    val n = 1000
    val k = 1000
    val results = new Array[Long](n)
    var i = 0
    while (i < n) {
      val start = nanoTime()
      Zone { implicit z =>
        var j = 0
        while (j < k) {
          alloc[Byte](count)
          j += 1
        }
      }
      val end = nanoTime()
      results(i) = end - start
      i += 1
    }
    val sorted = results.sorted
    sorted((n * 0.5).toInt)
  }
    for (i <- 1 to 14) {
      val size = scala.math.pow(2, i).toInt

      // warmup runs
      run(size)
      run(size)
      run(size)
      run(size)
      run(size)

      val time = run(size)

      println(s"${size} bytes : ${time} ns / 1000 allocations")
    }
  }
}
```

WojciechMazur (approved):

PR #2448, opened 2021-12-07
Fix ARM64 MacOS varargs

keynmol:
This is very rough, just to demonstrate the issue and the solution.

Want to gather feedback on the following:

1. Is it time to refactor and extract common parts from x86 and arm64 versions?
2. Should we handle all platforms in this PR?
3. Is memory padding from the swift code necessary? https://github.com/apple/swift/blob/main/stdlib/public/core/VarArgs.swift#L604-L631
    - I couldn't quite get through the `ifdef`s in swift code, just did this and it worked

TL;DR: varargs are broken on MacOS ARM64

Reason: on most other platforms is a struct (like this one https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#definition-of-va-list) but on this particular one (M1) it's a direct pointer to arguments serialised in memory.

### To reproduce the problem

Use any function with variadic arguments on Apple M1 (SN 0.4.1). I've put this example in a sandbox:

```scala
import scala.scalanative.libc.stdio

object Test {
    Zone { implicit z =>
      val myStr = c"2567 1234"
      val struct = stackalloc[CStruct2[Int, Int]]
      stdio.sscanf(myStr, c"%d %d", struct.at1, struct.at2)

      stdio.printf(c"Original string: %s, parsed.. First: %d, second: %d", myStr, struct._1, struct._2)
    }
  }
}
```

Without the fix in this PR it will segfault upon enountering `sscanf`

With the fix, it will finish successfully:

```
[info] Linking (410 ms)
[info] Checking intermediate code (31 ms)
[info] Dumping intermediate code (linked) (143 ms)
[info] Discovered 638 classes and 4313 methods
[info] Optimizing (debug mode) (370 ms)
[info] Checking intermediate code (20 ms)
[info] Dumping intermediate code (optimized) (159 ms)
[info] Dumping intermediate code (lowered) (272 ms)
[info] Generating intermediate code (537 ms)
[info] Produced 8 files
[info] Compiling to native code (942 ms)
[info] Linking native code (immix gc, none lto) (81 ms)
[info] Total (2713 ms)
Original string: 2567 1234, parsed.. First: 2567, second: 1234[success] Total time: 5 s, completed 7 Dec 2021, 20:32:08
```

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -8,9 +8,15 @@ import scala.scalanative.unsafe._
 object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isWindows")
   def isWindows: Boolean = resolved
+  
+  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isMac")
+  def isMac: Boolean = resolved
 
   @resolvedAtLinktime(
     "scala.scalanative.meta.linktimeinfo.isWeakReferenceSupported"
   )
   def isWeakReferenceSupported: Boolean = resolved
+  
+  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isArm64")
+  def isArm64: Boolean = resolved
 }
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -8,7 +8,7 @@ import scala.scalanative.unsafe._
 object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isWindows")
   def isWindows: Boolean = resolved
-  
+
   @resolvedAtLinktime(
     "scala.scalanative.meta.linktimeinfo.isWeakReferenceSupported"
   )
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -9,14 +9,8 @@ object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isWindows")
   def isWindows: Boolean = resolved
   
-  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isMac")
-  def isMac: Boolean = resolved
-
   @resolvedAtLinktime(
     "scala.scalanative.meta.linktimeinfo.isWeakReferenceSupported"
   )
   def isWeakReferenceSupported: Boolean = resolved
-  
-  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isArm64")
-  def isArm64: Boolean = resolved
 }
--- a/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
@@ -1,10 +1,8 @@
 package scala.scalanative.runtime
 
-import java.util.Locale
-
 object PlatformExt {
   private lazy val osArch =
-    System.getProperty("os.arch").toLowerCase(Locale.ROOT)
+    System.getProperty("os.arch").toLowerCase()
   
   lazy val isArm64 = osArch == "aarch64"
 }
--- a/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
@@ -1,0 +1,10 @@
+package scala.scalanative.runtime
+
+import java.util.Locale
+
+object PlatformExt {
+  private lazy val osArch =
+    System.getProperty("os.arch").toLowerCase(Locale.ROOT)
+  
+  lazy val isArm64 = osArch == "aarch64"
+}
--- a/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
@@ -4,5 +4,7 @@ object PlatformExt {
   private lazy val osArch =
     System.getProperty("os.arch").toLowerCase()
 
-  lazy val isArm64 = osArch == "aarch64"
+  // On M1 macs it's arm64, on windows we set it to aarch64
+  // see ./nativelib/src/main/resources/scala-native/platform.c for details
+  lazy val isArm64 = osArch == "arm64" || osArch == "aarch64"
 }
--- a/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
@@ -3,6 +3,6 @@ package scala.scalanative.runtime
 object PlatformExt {
   private lazy val osArch =
     System.getProperty("os.arch").toLowerCase()
-  
+
   lazy val isArm64 = osArch == "aarch64"
 }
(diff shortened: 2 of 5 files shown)


nativelib/src/main/scala/scala/scalanative/unsafe/CVarArgList.scala
```diff
@@ -75,6 +80,58 @@ object CVarArgList {
...
+      }
+    }
+    val resultStorage =
+      z.alloc(sizeof[Long] * storage.size.toULong).asInstanceOf[Ptr[Long]]
+    val storageStart = storage.asInstanceOf[LongArray].at(0)
+    libc.memcpy(
+      toRawPtr(resultStorage),
+      toRawPtr(storageStart),
+      wordsUsed.toULong * sizeof[Long]
+    )
```
keynmol, line 128: Copied from the x86_64 unix version

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
```diff
@@ -8,9 +8,15 @@ import scala.scalanative.unsafe._
 object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isWindows")
   def isWindows: Boolean = resolved
+  
+  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isMac")
+  def isMac: Boolean = resolved
```
WojciechMazur, line 13: Linktime resolved values/conditions were designed to dead code eliminate untaken branches at link time, and the main goal was to get rid of usages of native functions that are not available in Windows/Unix. MacOS does not define any os-specific functions that would not be available in other UNIX based operating systems. 
We want to limit the amount of link-time resolved values to a minimum, and use runtime resolved values whenever possible. 
Usage of this method can be replaced with `scala.scalanative.runtime.Platform.isMac`

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
```diff
@@ -8,9 +8,15 @@ import scala.scalanative.unsafe._
...
 
   @resolvedAtLinktime(
     "scala.scalanative.meta.linktimeinfo.isWeakReferenceSupported"
   )
   def isWeakReferenceSupported: Boolean = resolved
+  
+  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isArm64")
+  def isArm64: Boolean = resolved
```
WojciechMazur, line 21: Now, after merging https://github.com/scala-native/scala-native/pull/2452 we can detect CPU architecture at runtime in a convenient way in the same way as on the JVM. As long as we don't use any Arm64 specific functions we should use runtime resolved value instead. 
I would suggest moving this method to `scala.scalanative.runtime.PlatformExt` (`Platform` is extern module and it can only contain extern methods, so we can define runtime resolved `object PlatformExt`) where we would use the same logic as on the JVM to check if we are using Arm64 based machine.

nativelib/src/main/scala/scala/scalanative/unsafe/CVarArgList.scala
```diff
@@ -75,6 +80,58 @@ object CVarArgList {
...
+
+    // On MacOS Arm64 the variadic arguments list is stored in memory
+    // without a header structure (such as one used on x86)
+    new CVarArgList(toRawPtr(storageStart))
```
WojciechMazur, line 132: Since the only change is the absence of the header, let's keep it simple with only a single implementation for now, in case if in the future we would locate more inconsistencies we might refactor it to make addressing them easier. For now, I believe we should just rename previous method to `toCVarArgList_Unix` and use the runtime condition to check if the header is used.

WojciechMazur (changes requested):

WojciechMazur:
Thank you for your contribution! I've left some comments in the review for you

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
keynmol: Gotcha, thanks

keynmol:
Oops, still not right, and local compilation is acting weirdly for me.
Give me a bit more time.

nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
```diff
@@ -0,0 +1,10 @@
+package scala.scalanative.runtime
+
+import java.util.Locale
+
+object PlatformExt {
+  private lazy val osArch =
+    System.getProperty("os.arch").toLowerCase(Locale.ROOT)
```
WojciechMazur, line 7: Scala Native does not define `java.util.Locale`, but it can be provided via external libraries. For usage inside Scala Native us just `.toLowerCase` it should be compliant with usage of `Locale.ROOT`

nativelib/src/main/scala/scala/scalanative/unsafe/CVarArgList.scala
```diff
@@ -121,13 +129,16 @@ object CVarArgList {
...
       wordsUsed.toULong * sizeof[Long]
     )
-
-    val resultHeader = z.alloc(sizeof[Header]).asInstanceOf[Ptr[Header]]
-    resultHeader.gpOffset = 0.toUInt
-    resultHeader.fpOffset = (countGPRegisters.toULong * sizeof[Long]).toUInt
-    resultHeader.regSaveArea = resultStorage
-    resultHeader.overflowArgArea = resultStorage + registerSaveWords
-    new CVarArgList(toRawPtr(resultHeader))
+    if (PlatformExt.isArm64 && Platform.isMac())
```
ekrich, line 132: Can we leave change `isMac()` to `isMac`?

WojciechMazur (approved):

keynmol:
@WojciechMazur Please don't merge this yet (unless you have a way to test it), as I need to re-test it on my machine which last night refused to cooperate on this PR.

WojciechMazur:
> @WojciechMazur Please don't merge this yet (unless you have a way to test it), as I need to re-test it on my machine which last night refused to cooperate on this PR.

Ok, acknowledged. Thank you for notifying me. We definitely need to find a way to tests ARM in the CI

nativelib/src/main/scala/scala/scalanative/runtime/PlatformExt.scala
```diff
@@ -4,5 +4,7 @@ object PlatformExt {
   private lazy val osArch =
     System.getProperty("os.arch").toLowerCase()
 
-  lazy val isArm64 = osArch == "aarch64"
+  // On M1 macs it's arm64, on windows we set it to aarch64
+  // see ./nativelib/src/main/resources/scala-native/platform.c for details
+  lazy val isArm64 = osArch == "arm64" || osArch == "aarch64"
```
keynmol, line 9: I thought I'd confirm it with C:

```
❯ bat test.c
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: test.c
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ #include <sys/utsname.h>
   2   │  #include <stdio.h>
   3   │  int main(int argc, char *argv[]) {
   4   │     struct utsname buffer;
   5   │     if (uname(&buffer) >= 0) {
   6   │       printf("%s", buffer.machine);
   7   │     }
   8   │
   9   │ }
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

❯ clang --version
Apple clang version 13.0.0 (clang-1300.0.29.3)
Target: arm64-apple-darwin21.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

❯ clang test.c

❯ ./a.out
arm64%
```

WojciechMazur: What is value of os.arch system property on the JVM? Maybe there is more JVM compliance changes that I initially thought

keynmol: Hehe

```scala
❯ scala-cli repl
scala> System.getProperty("os.arch")
val res0: String = aarch64

```

WojciechMazur: So it looks like clang and all other decent tools are depending on the `uname`, and we do the same. I've not managed to fully find out how does JDK set up `os.arch` on Unix. On Windows, there are like 4 variants using C precompiler definition, but for Unix, it looks more like manually set values in the build. Guess we need to stick to the `uname` for now.

PR #2450, opened 2021-12-08
Start adding JDK 11 support

ekrich:
Trying to run tests under JDK 11 fails miserably so if some brave sole wants to help out it would be much appreciated. The build support code should will work for Native but not JVM as this is definitely a WIP.

This is not so much about supporting new Java methods but just making Scala Native run and test using Java 11 and then Java 17, the next Long Term Support (LTS) versions after Java 8. Some code and methods such as the `java.lang.Character` code added so far is needed.

New methods and code beyond what is need so it works is welcome as well.


ekrich:
`sandbox/run` does work but `tests/testOnly *CharacterTest*` fails.

build.sbt
```diff
@@ -25,6 +25,23 @@ lazy val disabledDocsSettings: Seq[Setting[_]] = Def.settings(
   Compile / doc / sources := Nil
 )
 
+val javaVersion = settingKey[Int](
+  "The major Java SDK version that should be assumed for compatibility. " +
+    "Defaults to what sbt is running with."
+)
+
+// JDK version we are running with
+javaVersion in Global := {
```
xuwei-k, line 34:
```suggestion
Global / javaVersion := {
```

build.sbt
```diff
@@ -882,6 +899,20 @@ lazy val tests =
...
+        val testDir = (Test / sourceDirectory).value
+        val rootTestDir =
+          testDir.getParentFile.getParentFile.getParentFile
+
+        val javaV = javaVersion.value
+        javaV match {
+          case v if (v >= 11) =>
+            Seq(rootTestDir / "require-jdk11" / "src" / "test")
+          case _ => Nil
+        }
```
WojciechMazur, line 915: We can make it more generic 
```suggestion
        (9 to (Global / javaVersion).value).map { v =>
          rootTestDir / s"require-jdk$v" / "src" / "test"
        }
```

build.sbt
```diff
@@ -882,6 +899,20 @@ lazy val tests =
             Seq((Test / sourceDirectory).value / "scala-2.12+")
           case _ => Nil
         }
+      },
+      // temp hack
```
WojciechMazur, line 903: I guess we don't need this comment, it probably wouldn't be jest temporal ;)

WojciechMazur (commented):
Overall looks good, just a to comments

build.sbt
```diff
@@ -882,6 +899,20 @@ lazy val tests =
...
+      Test / unmanagedSourceDirectories ++= {
+        // unit-tests/native/src/test
+        val testDir = (Test / sourceDirectory).value
+        val rootTestDir =
+          testDir.getParentFile.getParentFile.getParentFile
```
WojciechMazur, line 908: We can also simplify this one
```suggestion
        // unit-tests/native
        val testDir = (Test / baseDirectory).value
        val rootTestDir = testDir.getParentFile
```

build.sbt
```diff
@@ -882,6 +899,20 @@ lazy val tests =
             Seq((Test / sourceDirectory).value / "scala-2.12+")
           case _ => Nil
         }
+      },
+      // temp hack
+      Test / unmanagedSourceDirectories ++= {
```
WojciechMazur, line 904: The same settings should be applied to the JVM project to make sure that our tests our compliant with the JVM runtime. It would be the easiest to move this logic to the `sharedTestSource` method. In fact, we can also move `require-jdk*` directories inside the `shared` directory.

build.sbt
ekrich: I really like this suggestion. I haven't changed it yet as I am working on the source location.

ekrich:
On this recent push I move the code to `shared/src/test/require-jdk11` for example. I have done this before with `shared/src/main/<something special>` with good results. See picture:

image


Here in `test` it seems to automatically find the source. 

Here is a Java 8 example which does not add the above source directory.

```
[info] Detected JDK version 8
[info] testDir: /Users/eric/workspace/scala-native/unit-tests/jvm
[info] sharedTestDir: /Users/eric/workspace/scala-native/unit-tests/shared/src/test
[info] Res: List()
[info] testDir: /Users/eric/workspace/scala-native/unit-tests/native
[info] sharedTestDir: /Users/eric/workspace/scala-native/unit-tests/shared/src/test
[info] Res: List()
```
This shows no additional paths added but we see the following:

```
[info] compiling 230 Scala sources to /Users/eric/workspace/scala-native/unit-tests/native/target/scala-2.12/test-classes ...
[error] /Users/eric/workspace/scala-native/unit-tests/shared/src/test/require-jdk11/org/scalanative/testsuite/javalib/lang/CharacterTestOnJDK11.scala:19:53: type mismatch;
[error]  found   : Int(65536)
[error]  required: Char
[error]     assertEquals("\uD800\uDC00", Character.toString(0x10000))
```

build.sbt
```diff
@@ -864,6 +881,25 @@ lazy val testsCommonSettings = Def.settings(
   )
 )
 
+lazy val javaVersionBasedTestSources = Def.settings(
+  Test / unmanagedSourceDirectories ++= {
+    // unit-tests/native/src/test
+    val testDir = (Test / baseDirectory).value
```
WojciechMazur, line 887: It is not correct, baseDirectory would point to `unit-tests/native` or `unit-tests/jvm`

WojciechMazur (commented):
It would be also great to add a single JDK 11 workflow in the CI, however, I can do it later. I sometimes had issues with `Buffer.position` method on some JDKs newer then 8

build.sbt
```diff
@@ -864,6 +881,25 @@ lazy val testsCommonSettings = Def.settings(
...
+    sLog.value.info(s"testDir: $testDir")
+    val sharedTestDir =
+      testDir.getParentFile / "shared/src/test"
```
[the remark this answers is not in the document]
WojciechMazur: https://github.com/scala-native/scala-native/blob/c3f13c43599b92da102b8883405a5311fca5b6d0/build.sbt#L834 code which does use glob filters to include all Scala files from `shared/src/test`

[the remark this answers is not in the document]
ekrich: Because of that `allScalaFromDir` then this works differently because Scala.js doesn't have that. Unless there is an easy way to change, I think we need to go with a non-standard setup.

build.sbt
ekrich: Sorry, left the comment from before.

build.sbt
ekrich: Got this one - thanks.

build.sbt
ekrich: Thanks, resolved.

build.sbt
ekrich: Yep, was temporary, removed.

build.sbt
ekrich: Used the baseDirectory as suggested.

build.sbt
ekrich: Yes, I created a setting and added it to both native and jvm tests.
Now the following works:
```scala
sbt:scala-native> testsJVM/testOnly *CharacterTest*
[info] Test run started
[info] Test org.scalanative.testsuite.javalib.lang.CharacterTestOnJDK11.toStringCodePoint started
[info] Test run finished: 0 failed, 0 ignored, 1 total, 0.039s
```

build.sbt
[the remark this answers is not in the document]
ekrich: This is what I was missing, why it behaved not as expected.

ekrich:
Next step is to fix the `Buffer.position` and others so native tests will pass.

ekrich:
@WojciechMazur I guess this PR is fine as is but I would sure like to get the tests to run under JDK 11 at least. It is really hard to tell if the problem is a compiler problem or a library problem or a combo. I can run the JDK11 test on the JVM with Java 11 so that works ok.

WojciechMazur:
> @WojciechMazur I guess this PR is fine as is but I would sure like to get the tests to run under JDK 11 at least. It is really hard to tell if the problem is a compiler problem or a library problem or a combo. I can run the JDK11 test on the JVM with Java 11 so that works ok.

Hey, after overriding some of `MappedByteBuffer` methods to return `MappedByteBuffer` instead of `ByteBuffer` I was able to run it locally. If you had some errors while compiling our linker/tools please let me know.

WojciechMazur (approved):

ekrich:
@WojciechMazur I have checked on JDK 11 and 17 and everything is good. Appreciate you finishing in the JDK 9+ PR - I was really tired. I do have a couple of additional changes to the NIO classes I found that I will put in a separate PR.

Can you add one compile/test on JDK 11 and 17 in CI just for Linux?

PR #2454, opened 2021-12-09
Adapt `unsafe` package for upcoming Scala 3 support

WojciechMazur:
This PR introduces changes that are needed to proper implementation of `unsafe` package methods. 

* `scalanative.unsafe.extern` annotation was moved to `scalanative.unsafe` package  object - Scala 3 does not allow to use classes and methods with the same names defined in the separate files, usage of `@extern` would be treated as trying to use `extern` method 
* No parens methods `alloc[T]` and `stackalloc[T]` were deprecated and empty parens variants of this methods were added - In Scala 3 `alloc[T](10)` can be interpreted as `alloc[T].apply(10)` which means instead of allocation memory of 10x size of T we would allocate 1x size of T and would try to access it's 10th element, it would lead to multiple silent yet deadly bugs 
* All usages of no parens `alloc` and `stackalloc` methods were replaced with empty parens variant
* All usages of `alloc(n)` and `stackalloc(n)` methods without explicit type, were given explicit type to make sure that we would not try access nth element of allocated memory as described above

Even though changing `alloc[T]` to `alloc[T]()` is less user friendly this syntax is matches standard denotation for non-pure methods, which both alloc and stackalloc definitely are.

### DIFF
--- a/javalib/src/main/scala/java/io/File.scala
+++ b/javalib/src/main/scala/java/io/File.scala
@@ -289,7 +289,7 @@ class File(_path: String) extends Serializable with Comparable[File] {
    */
   private def simplifyExistingPath(path: String)(implicit z: Zone): String = {
     if (isWindows) {
-      val resolvedName = alloc[WChar](FileApiExt.MAX_PATH)
+      val resolvedName: Ptr[WChar] = alloc[WChar](FileApiExt.MAX_PATH)
       GetFullPathNameW(
         toCWideStringUTF16LE(properPath),
         FileApiExt.MAX_PATH,
@@ -298,7 +298,7 @@ class File(_path: String) extends Serializable with Comparable[File] {
       )
       fromCWideString(resolvedName, StandardCharsets.UTF_16LE)
     } else {
-      val resolvedName = alloc[Byte](limits.PATH_MAX)
+      val resolvedName: Ptr[Byte] = alloc[Byte](limits.PATH_MAX)
       if (realpath(toCString(path), resolvedName) == null) {
         throw new IOException(
           s"realpath can't resolve: ${fromCString(resolvedName)}"
@@ -677,7 +677,7 @@ class File(_path: String) extends Serializable with Comparable[File] {
           val privilegeSetLength = stackalloc[windows.DWord]
           !privilegeSetLength = emptyPriviligesSize.toUInt
 
-          val privilegeSet = stackalloc[Byte](!privilegeSetLength)
+          val privilegeSet: Ptr[Byte] = stackalloc[Byte](!privilegeSetLength)
           memset(privilegeSet, 0, !privilegeSetLength)
 
           val grantedAcccess = stackalloc[windows.DWord]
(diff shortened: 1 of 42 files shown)


nativelib/src/main/scala/scala/scalanative/unsafe/package.scala
```diff
@@ -99,6 +99,10 @@ package object unsafe {
   /** C-style alignment operator. */
   @alwaysinline def alignmentof[T](implicit tag: Tag[T]): CSize = tag.alignment
 
+  // Scala 3 does not allow to use extern method and extern annotation class
+  // defined in separate files in the same package
+  type extern = scala.scalanative.annotation.extern
```
sjrd, line 104: Why not declare `final class extern` right here? That would satisfy Scala 3's constraint without changing the source API at all. It might even preserve binary compat? (not sure)

WojciechMazur: For some reason I thought that's not possible, probably I had some issues with that in 3.0.2 Scala version. Anyway, now it works just fine and we can define them both in a single file

javalib/src/main/scala/java/io/File.scala
```diff
@@ -298,7 +298,7 @@ class File(_path: String) extends Serializable with Comparable[File] {
       )
       fromCWideString(resolvedName, StandardCharsets.UTF_16LE)
     } else {
-      val resolvedName = alloc[Byte](limits.PATH_MAX)
+      val resolvedName: Ptr[Byte] = alloc[Byte](limits.PATH_MAX)
```
ekrich, line 301: @WojciechMazur Why is the type required here?

javalib/src/main/scala/scala/scalanative/nio/fs/unix/UnixFileSystemProvider.scala
```diff
@@ -24,7 +24,7 @@ class UnixFileSystemProvider extends GenericFileSystemProvider {
   }
 
   private def getUserDir(): String = {
-    val buff = stackalloc[CChar](4096.toUInt)
+    val buff: Ptr[CChar] = stackalloc[CChar](4096.toUInt)
```
ekrich, line 27: And sometimes you use it for `stackalloc` and sometimes not.

WojciechMazur: I've use it for all usages of `stackalloc[T](n)`, for `stackalloc[T]()` it's not needed. If it's not used somewhere then its probably my mistake and explicit type should be added there

javalib/src/main/scala/java/io/File.scala
WojciechMazur: When we would remove `alloc[T]` method this workaround would no longer be needed. Alternatively we can not include this method only in Scala3 unsafe package, but then breaking both binary and source compatibility does not sound very nice, especially when someone is using a lot of `stackalloc[T]` or `alloc[T]` calls

ekrich: Thank-you so much for explaining - sorry I didn't review prior. Hard to keep track given your velocity.

PR #2457, opened 2021-12-11
Ensure `javalib` source compatibility for Scala 3 build

WojciechMazur:
This PR introduces a series of fixes allowing us to compile `javalib` project with Scala 3 compiler

* Replaced usages do-while loops with `while-do` 
* Ensured correct types (coercion)
* Ensure methods defined in Java stdlib are not defined as `val`s 
* Add explicit result types 
* Added required parenthesis for lambda parameters
* Fix no longer compliant usages of `@swtich` annotations
* Fixes to `java.nio` Buffers due to incorrectly resolved types of `GenBuffer[T](this)` method calls (needs explicit type)
* Update scalafmt config to handle Scala 3 sources
* Duplicate classes registered by Scala 3 compiler, renamed them in Scala 3 sources
* Duplicate enum definitions for Scala 2 / 3 projects
* Split `java.util.Formatter` into `Impl` class and standard class to allow for the definition of inner enums
* Duplicate `java.util.ProcessBuilder`  to allow for the definition of inner enums

# Additional explanations: 
## Renamed classes
 Some classes defined in `javalib` needs to be renamed, otherwise, Scala 3 compiler would complain about their usage "in the incorrect period" - this happens if for some reason in the compiler phase registered some symbol, for eg. using `getRequiredClass` method, but also finds the same symbol somewhere else when compiling sources. It leads to fatal error. This is why we need to rewrite some classes, eg. `java.lang.Serializable` to `java.lang._Serializable` (due to the same for `j.l.Object` in Scala 2). The names would be adjusted when encoding names in NIR. 

## Duplication of enums
Scala 3 does not allow us to define enums in the same way as we did in Scala 2, instead, we need to use new syntax using `enum X extends Enum[X]` (extending is needed to ensure they're Java-compatible). Because of that, we need to provide a separate implementation for defined enums for both Scala 2 / 3 in the separate directories

## Duplication of classes containing inner enums 
There were more problematic cases where enum was defined within the body of some classes. In such a case we had no other option but to duplicate the whole class definition. In the case of `java.util.Formatter` (1.6k LOC) it was not trivial, especially since we didn't want to sync such a big files. In that case common `FormatterImpl` class was extracted implementing the logic of formatted and created two `Formatter` implementations containing constructors and definitions of enums, one for each Scala epoch version. 
In the case of ProcessBuilder extracting common logic was not trivial, but due to the relatively small size, I've decided to duplicate the whole class body.

## Fixes to `java.nio` Buffers
It looks like in Scala 3 compiler has sometimes problems with deriving correct type for given class/method invocation. This is why we needed to help it and replace calls to `GenBuffer(this)` with calls containing explicit type eg `GenBuffer[ByteBuffer](this)`. To limit the amount of code changes in the future the inlined calls were replaced with factored out method, eg `private def genBuffer = GenBuffer[ByteBuffer](this)`. 
We do the same for other `Gen*` type classes

## Correct signatures
Scala 3 compiler is better at telling us what should be signature for given method defined in JDK. This is we had a lot of compiler errors when some method was defined as `val foo` instead of `def foo()`. This also means that we needed to factor out  rhs of such definitions, eg. create additional value that would store lazily computed value.

### DIFF
--- a/javalib/src/main/scala-2/java/lang/Enum.scala
+++ b/javalib/src/main/scala-2/java/lang/Enum.scala
@@ -1,3 +1,6 @@
+// Classes in this file needs special handling in Scala 3, we need to make sure
+// that they would not be compiled with Scala 3 compiler
+
 package java.lang
 
 abstract class Enum[E <: Enum[E]] protected (_name: String, _ordinal: Int)
--- a/javalib/src/main/scala-2/java/lang/Enum.scala
+++ b/javalib/src/main/scala-2/java/lang/Enum.scala
@@ -1,4 +1,4 @@
-// Classes in this file needs special handling in Scala 3, we need to make sure
+// Classes in this file need special handling in Scala 3, we need to make sure
 // that they would not be compiled with Scala 3 compiler
 
 package java.lang
--- a/javalib/src/main/scala-2/java/lang/annotation/RetentionPolicy.scala
+++ b/javalib/src/main/scala-2/java/lang/annotation/RetentionPolicy.scala
@@ -1,3 +1,5 @@
+// Enums are not source-compatbile, make sure to sync this file Scala 3 implementation
+
 package java.lang.annotation
 
 final class RetentionPolicy private (name: String, ordinal: Int)
--- a/javalib/src/main/scala-2/java/lang/annotation/RetentionPolicy.scala
+++ b/javalib/src/main/scala-2/java/lang/annotation/RetentionPolicy.scala
@@ -1,4 +1,4 @@
-// Enums are not source-compatbile, make sure to sync this file Scala 3 implementation
+// Enums are not source-compatible, make sure to sync this file's Scala 3 implementation
 
 package java.lang.annotation
 
--- a/javalib/src/main/scala-2/java/lang/annotation/RetentionPolicy.scala
+++ b/javalib/src/main/scala-2/java/lang/annotation/RetentionPolicy.scala
@@ -1,4 +1,4 @@
-// Enums are not source-compatible, make sure to sync this file's Scala 3 implementation
+// Enums are not source-compatible, make sure to sync this file with Scala 3 implementation
 
 package java.lang.annotation
 
--- a/javalib/src/main/scala-3/java/lang/ProcessBuilder.scala
+++ b/javalib/src/main/scala-3/java/lang/ProcessBuilder.scala
@@ -1,3 +1,7 @@
+// Any changes to this files needs to be synced with it's Scala2 counter part.
+// Due to enum restrictions it is impossible to create common implementation
+// for both Scala versions
+
 package java.lang
 
 import java.util.{ArrayList, List}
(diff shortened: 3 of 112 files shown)


javalib/src/main/scala-2/java/lang/Enum.scala
```diff
@@ -1,3 +1,6 @@
+// Classes in this file needs special handling in Scala 3, we need to make sure
```
jchyb, line 1:
```suggestion
// Classes in this file need special handling in Scala 3, we need to make sure
```
Typo, here and in other scala-2 files

javalib/src/main/scala-3/java/util/Formatter.scala
```diff
@@ -0,0 +1,93 @@
+package java.util
+// Ported from Scala.js, commit: 0383e9f, dated: 2021-03-07
+
```
jchyb, line 3: (I assume) missing explanation

javalib/src/main/scala-3/java/util/concurrent/TimeUnit.scala
```diff
@@ -0,0 +1,110 @@
+package java.util.concurrent
+
```
jchyb, line 2: Missing explanation (I think)

javalib/src/main/scala-3/scala/scalanative/compat/ScalaStream.scala
```diff
@@ -0,0 +1,21 @@
+package scala.scalanative.compat
+
+import java.util.stream.WrappedScalaStream
```
jchyb, line 3: Missing explanation also here I think

javalib/src/main/scala-2/java/lang/annotation/RetentionPolicy.scala
```diff
@@ -1,3 +1,5 @@
+// Enums are not source-compatbile, make sure to sync this file Scala 3 implementation
```
jchyb, line 1:
```suggestion
// Enums are not source-compatible, make sure to sync this file's Scala 3 implementation
```
Typo, here and in other scala-2 files

javalib/src/main/scala-3/java/lang/ProcessBuilder.scala
```diff
@@ -0,0 +1,194 @@
+// Any changes to this files needs to be synced with it's Scala2 counter part.
```
jchyb, line 1:
```suggestion
// Any changes to this file need to be synced with it's Scala2 counterpart.
```
Typo, here and in other scala-3 files

javalib/src/main/scala-3/java/lang/annotation/Retention.scala
```diff
@@ -1,3 +1,3 @@
 package java.lang.annotation
```
jchyb, line 1: Missing explanation? Perhaps the file in particular is self explanatory, I am more sure about this in other files

jchyb (commented):
Looks very cool, I really like the idea of explanations in the compat files. I found some NIT typos

PR #2458, opened 2021-12-13
Use encoded names for methods and fields in NIR

WojciechMazur:
This PR changes the way names of methods and fields are stored in NIR format. Currently, names are always stored in decoded form, eg. `def foo_=(v: Int)` would have the name `foo_=` in NIR 
This behaviour however might lead to causing problems in Scala 3. In some cases, it can produce names containing partially encoded names, leading to missing definitions while linking. 
To make sure this would not happen we now store names primarily in encoded form, eg. `foo_$eq` with exceptions for extern members which needs to be always decoded (eg. if the native name contains special characters like dots in some of the LLVM intrinsic methods).

This change is binarily incompatible in NIR, so its version was bumped up.

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
@@ -7,6 +7,7 @@ import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import scala.collection.mutable
 import scala.scalanative.nir.serialization.{Tags => T}
+import scala.reflect.NameTransformer
 
 final class BinaryDeserializer(buffer: ByteBuffer, bufferName: String) {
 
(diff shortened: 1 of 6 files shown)


sjrd:
Instead of breaking compatibility of NIR, you could have a deserialization hack that encodes all names upon deserialization of an older NIR format. So that would mean only bump the reversion of NIR to 9, and when reading NIR <= 5.8, encode all non-extern names.

We do this kind of stuff all the time in the Scala.js IR deserializer to preserve backward compatibility. Here is a PR from this year where we did this: https://github.com/scala-js/scala-js/pull/4439

nir/src/main/scala/scala/scalanative/nir/serialization/BinaryDeserializer.scala
```diff
@@ -362,4 +372,73 @@ final class BinaryDeserializer(buffer: ByteBuffer, bufferName: String) {
 
     readPosition()
   }
+
+  // Copy pasted from Scala 3.1.0 compiler
+  // Used to encode names from NIR versions <= 5.8 produced with decoded names
+  private object NameTransformer {
```
sjrd, line 378: There is already `scala.reflect.NameTransformer` in the standard library, which provides this functionality.

PR #2463, opened 2021-12-14
Add Scala `scala.runtime.JProcedure` Scala implementations to auxlib

WojciechMazur:
This PR adds Scala implementation for Scala 3 `JProcedure` classes, which might be used when defining Lambdas returning `Unit`.  
JProcedure were originally defined as Java sources


ekrich:
@shadaj Wasn't this in your PR for 32bit as well?

PR #2464, opened 2021-12-14
Ensure Scala 3 source-compatibility for tests

WojciechMazur:
This PR ensures that our unit tests and source compatible with Scala 3. 

* Fix compile-time errors for unit tests
* Unit tests that are Scala 2 specific (`SymbolTest`, `ReflectiveProxyTest`) were moved to scala-2 directory
* Adapt sharedSources setting to handle scala version specific sources


unit-tests/shared/src/test/scala/javalib/util/DefaultFormatterTest.scala
```diff
@@ -709,7 +709,7 @@ class DefaultFormatterTest {
   }
 
   @Test def formatForGeneralConversionType_bB(): Unit = {
-    val triple: Array[Array[Object]] = Array(
+    val triple: Array[Array[Any]] = Array[Array[Any]](
```
ekrich, line 712: Both sides and a cast are required?

WojciechMazur: It definitely shouldn't be needed. I'll fix it

ekrich: Looks much better!

PR #2465, opened 2021-12-16
Adapt LazyVals to work in Scala Native

WojciechMazur:
This PR does add a new plugin phase which is responsible for rewriting calls to `scala.runtime.LazyVals` which cannot be supported in Scala Native because we do not support static class constructors and we don't provide access to JDK `unsafe` methods, like `getFieldOffset`. 

* Added override for `scala.runtime.LazyVals`
* Added `scala.scalanative.runtime.LazyVals` which are used instead of Scala implementation - they're using the same algorithm, however, are using a pointer to resolved field instead of calculating it using offset (we we're not able to get offsets correctly) 
* Added `AdaptLazyVals` phase which does: 
  * collect references to bitmap fields used by LazyVals within compilation unit
  * rewrites usages of internal `OFFSET` fields with intrinsic `classFieldRawPtr` function call based on trees collected in previous step 
  * rewrites all not-compliant calls to Scala LazyVals methods with their Scala Native counter parts
* Reverted mitigations for failing tests due to LazyVals issues
* Added building and running sandbox in the CI

### DIFF
--- a/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/AdaptLazyVals.scala
+++ b/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/AdaptLazyVals.scala
@@ -1,0 +1,132 @@
+package scala.scalanative.nscplugin
+
+import dotty.tools._
+import dotc._
+import dotc.transform.{LazyVals, MoveStatics}
+import dotc.ast.tpd._
+import plugins._
+import core.Flags._
+import core.Contexts._
+import core.Names._
+import core.Symbols._
+import core.StdNames._
+import scala.annotation.{threadUnsafe => tu}
+
+// This phase is responsible for rewriting calls to scala.runtime.LazyVals with
+// its scala native specific counter-part. This is needed, becouse LazyVals are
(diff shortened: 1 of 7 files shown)


nscplugin/src/main/scala-3/scala/scalanative/nscplugin/AdaptLazyVals.scala
```diff
@@ -0,0 +1,132 @@
...
+          val Apply(_, List(cls: Literal, fieldname: Literal)) = vd.rhs
+          vd.symbol -> fieldname
+      }.toMap
+    }
+
+    ctx
+  }
+
+  // Replace all usages of all unsuportted LazyVals methods with their
+  // Scala Native specific implementation (taking Ptr intead of object + offset)
```
jchyb, line 59:
```suggestion
  // Replace all usages of all unsupported LazyVals methods with their
  // Scala Native specific implementation (taking Ptr instead of object + offset)
```

jchyb (commented):
Very impressive work (as always)! I've found some very minor NIT typos

PR #2468, opened 2021-12-18
Generate static method forwarders inside companion module

WojciechMazur:
This PR fixes issues when trying to use `@main` annotation - created `main` method would be defined as a static method inside the class, while we need it to be defined as a method of the module. This also contains fixes to accessing static members, eg. enum fields

* Generate static method forwarders inside companion module
* Ignore moving methods annotated with `@scala.static` from companion module to companion class. 
* Added method for explicitly calling  forwarder of static methods
* Added tests for checking if we can access fields and methods of static members


sjrd:
I think this should go in the entirely opposite direction instead. IIUC, this PR, for an actual `static def foo` in a `class A`, will go out of its way to make it accessible as a `def foo` in `object A`. Both Scala/JVM and Scala.js work the other way around: for a `def foo` in an `object A`, make it available as a `static def foo` in `class A`. This is both easier to do (I believe) and more efficient.

Again IIUC, the reason to do this is because the linker looks for main methods as `def foo` in `object A`. IMO this is what we need to fix instead. The linker should look for main methods as `static def foo` in `class A`. And the NirGen should emit static forwarders for methods of the companion object.

For backward compatibility, the linker should fall back on using a `def foo` in `object A` if it cannot find a static method.

sjrd:
Oh shoot I was a few seconds too slow to react :(

WojciechMazur:
There are some good points here, I'll try to prepare a PR that would revert this change and would generate static forwarders in class instead. I believe it should also to back-ported then to Scala 2 compiler plugin as well

PR #2469, opened 2021-12-20
Make reachablity phase more user friendly

WojciechMazur:
This PR includes some improvements to `Reach`, by providing better logging of missing definitions and fixes frequently observed issues. 

* Fix frequent class cast exception (Unavailable -> ScopeInfo) 
* Fix `IllegalArgumentException` when trying to report missing definition using `nir.Position.NoPostion` 
* Throw verbose exception when required definition (ex. Scala hashCode) is missing
* Include unavailable definitions in while reporting missing defns

### DIFF
--- a/tools/src/main/scala/scala/scalanative/linker/Reach.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/Reach.scala
@@ -117,14 +117,17 @@ class Reach(
     }
     def fallback = global match {
       case Global.Member(owner, sig) =>
-        infos(owner) match {
-          case scope: ScopeInfo =>
-            scope.linearized
-              .find(_.responds.contains(sig))
-              .map(_.responds(sig))
-              .flatMap(lookup)
-          case _ => None
-        }
+        infos
+          .get(owner)
+          .collect {
+            case scope: ScopeInfo =>
+              scope.linearized
+                .find(_.responds.contains(sig))
+                .map(_.responds(sig))
+                .flatMap(lookup)
+          }
+          .flatten
+
       case _ => None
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/linker/Reach.scala
```diff
@@ -849,12 +849,14 @@ class Reach(
...
 
   protected def addMissing(global: Global, pos: Position): Unit = {
-    val prev = missing.getOrElse(global, Set.empty)
-    val position = NonReachablePosition(
-      path = Paths.get(pos.source),
-      line = pos.line + 1
-    )
-    missing(global) = prev + position
+    val prev = missing.getOrElseUpdate(global, Set.empty)
+    if (pos != nir.Position.NoPosition) {
```
sjrd, line 865: Isn't this going to *ignore* missing stuff references from an empty position altogether, potentially letting the linker phase to complete without error where there should be one? If that's the case, we should instead at least put something in `missing(global)`, even if it is with a `path: None` or something like that.

WojciechMazur: It would not, `missing(global)` is set using `missing.getOrElseUpdate(global, Set.empty)`. It means we know it is missing but don't know where, because it might happen that for some reason the position is empty, eg. when modified in the scalac phase without setting a new position

sjrd: Ah OK. Looks good, then.

sjrd (approved):

PR #2470, opened 2021-12-20
Allow to run scripted tests with Scala 3

WojciechMazur:
This PR fixes some of scripted tests to allow running them with Scala 3. 

* One of the tests was using `scala.App` which is not supported in Scala 3, due to removal of `DelayedInit` feature. Test source was moved to `scala-2` directory, new test using `@main` annotation was added inside `scala-3` directory. 
* Replaced usage of procedural syntax
* Fixed `java-io-N` tests to allow sharing sources between sbt and application sources


scripted-tests/run/java-io-file-1/Platform.scala
```diff
@@ -1,5 +1,5 @@
 // This file is used only when compiling sources using Scala Native
 
 object Platform {
-  val isWindows = scalanative.runtime.Platform.isWindows
+  val isWindows = scalanative.runtime.Platform.isWindows()
```
lolgab, line 4: Shouldn't `isWindows` be defined as parenthesis-less in `Platform` too since it not a side-effectful `def` ? Also this could be a good time to do it since we are in the 0.5 breaking window

Issue #567, opened 2017-03-19
Proposal: embed resources into executable


regb:
Currently, the resource folder seems to be ignored by the packaging system. On the JVM, a JAR file contains resources as well, and the JDK exposes some API to access these files, in particular:

    def getResource(name: String): URL
    def getResourceAsStream(name: String): InputStream

It would be quite nice if we could pack the content of the resource folder into the executable, with these functions providing a way to extract the data as files.

ekrich:
Excellent proposal: This may be related to #155 where we would like DLLs or shared object files as there are cross platform and build changes needed. Libraries may need resources as well.

This may be a bit more high level as it involves user application packaging. 

The next level up to package for a distribution seems to be covered by https://github.com/sbt/sbt-native-packager

I looked around to found this first for just resource files in C.
http://stackoverflow.com/questions/1997172/is-there-a-linux-equivalent-of-windows-resource-files

Points to the following:
http://www.linuxjournal.com/content/embedding-file-executable-aka-hello-world-version-5967

And Mac uses resource bundles.
https://developer.apple.com/library/content/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW1

Windows:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa380599(v=vs.85).aspx

densh:
Thanks for the proposal @regb, we'll definitely review for inclusion in one of post-0.2 releases !

jonas:
If we disregard support for dynamic linking, resources could be bundled by packaging the executable as a zip file with the binary as the prelude and open `argv[0]` to read resources from the zip

benwaffle:
GTK+ apps use [gresources](https://developer.gnome.org/gio/stable/glib-compile-resources.html) for bundling files into the executable. It produces a .c file containing a byte array encoded as a [GVariant](https://developer.gnome.org/glib/stable/glib-GVariant.html) db with all the data and registers it on startup. On Linux it puts it in its own ELF section.

Andrei-Pozolotin:
+1 https://gitter.im/scala-native/scala-native?at=5ab570c1e4ff28713a68da2a

ahjohannessen:
+1

felipe-vieira0:
+1 for this!

WojciechMazur:
Base64 approach was implemented in PR #2475

PR #2475, opened 2021-12-23
Add getResourceAsStream() via embedding resources

jchyb:
This PR adds resource file embedding into an output binary file, as inspired by #567. Files embedded this way can be accessed by using getClass().getResourceAsStream(). getResources() and getResource() are not available for now as the way they would have to be would vastly differentiate Scala Native from how it works on the JVM. An all new URI format for embeddings would have to be specified.

Resources are taken from a classpath, with the paths being embedded in the order of appearance. In practice, this means that resources of dependencies with the same path relative to classpath root will be shadowed by those of the main compiled project, and will not be embedded, like in JVM. Then the files are read and both found relative paths (with classpath elements as roots) and contents of the files are embedded in the codegen phase of the compiler. Later, during runtime a map between the relative paths and contents of the file is being created. 

Both unit tests and scripted tests were added. Unit-tests test general functionality of getResourceAsStream() and are run against the JVM. Scripted tests however focus more on the embedding part, testing if and which files are being embedded.

~~To avoid unnecessarily embedding the source files, a filtering mechanism based on file extension was added. Otherwise, every „.h” and „.c” file (which are especially prominent in our resource folders), would be embedded into the code, heavily increasing the size of the resulting binary.~~ 

EDIT: Now only few source file extensions are being filtered, including ".scala" and ".class". If any file is a part of the resources/scala-native it will also independently be filtered. This includes the aforementioned ".c" and ".h" files and also any readme's etc. 

A documentation entry was also added specifying the various intricacies of embeddal, including the filtering mechanism described here.


jchyb:
I will try rebase when Scala 3 testing infrastructure will be completely merged. I know I am quite late with this, so it is completely fine by me if we focus on Scala 3 now and skip this PR until the next release I think.

docs/lib/javalib.rst
```diff
@@ -619,4 +619,39 @@ Some notes on the implementation:
...
+Embedding Resources
+-------------------
+
+In Scala Native, resources are implemented via embedding a resource in a resulting
+binary file. Only `getClass().getResourceAsInputStream()` is implemented.
+For that to work, you have to specify an additional NativeConfig option:
+
+```scala
+nativeConfig ~= {
+  _withEmbedResources(true)
```
ekrich, line 631:
```suggestion
  _.withEmbedResources(true)
```

nativelib/src/main/scala/java/lang/resource/EmbeddedResourceHelper.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  lazy val resourceFileIdMap =
+    EmbeddedResourceHelper.getAllFilePaths().zipWithIndex.toMap
+
+  // Decodes, constructs and returns all embedded resource file paths.
+  private def getAllFilePaths(): Array[String] = {
+    val filePathAmount = EmbeddedResourceReader.getEmbeddedSize()
+    val res = Array.ofDim[Array[Byte]](filePathAmount)
+
+    for (id <- 0 until filePathAmount) {
```
ekrich, line 15: Do you want to use a generator vs a simple while loop?

ekrich (changes requested):

ekrich:
Do you think you could just ignore the `resources/scala-native` completely. People may want to put other files in there like READMEs or makefiles etc. and it designed specifically for just C interopt code.

docs/lib/javalib.rst
jchyb: Good catch! Thank you

jchyb:
This sounds like a good idea - maybe it could even get rid of the current filtering mechanism. I will try doing that later.

nativelib/src/main/scala/java/lang/resource/EmbeddedResourceHelper.scala
ekrich: Yes, I think you should be able to use `while` loops with counters or `java.util.ScalaOps` to avoid using Scala collections but feel like you are using them. I brought `ScalaOps` over from Scala.js recently and it is pretty much genius.

tools/src/main/scala/scala/scalanative/codegen/ResourceEmbedder.scala
```diff
@@ -0,0 +1,199 @@
...
+        Attrs.None,
+        extern(name),
+        Type.Ptr,
+        Val.Const(
+          arrayValue
+        )
+      )
+    }
+
+    def generateExtern2DArray(name: String, content: IndexedSeq[Val.Const]) = {
```
WojciechMazur, line 123: ArrayValue does not need indexedSeq, regular `Seq` is enough. Changing that would save us from the need to convert all sequences to indexed ones

WojciechMazur (commented):
Good work! Overall PR looks good in my opinion. 
I've left some comments on the things that were bothering me and left some tips/suggestions according to simplifying the code

docs/lib/javalib.rst
```diff
@@ -619,4 +619,41 @@ Some notes on the implementation:
...
+This will include the resource files found on the classpath in the resulting
+binary file. Please note that files with following extensions cannot be embedded
+and used as a resource:
+
+```
+".class", ".nir", ".scala", ".java", ".jar",
+```
+
+This is to avoid unnecesarily embedding source files. If necessary, please
+consider using a different file extension for embeddal. Files found in the
```
ekrich, line 644: This line has "embeddal."

docs/lib/javalib.rst
```diff
@@ -619,4 +619,41 @@ Some notes on the implementation:
...
+resources/scala-native directory will not be embedded as well. It is recommended
+to add the ".c" nad ".h" files there.
+
+Reasoning for the lack of getResource() and getResources():
+
+In Scala Native, the outputted file that can be run is a binary, unlike JVM's
+classfiles and jars. For that reason, were getResources() URI methods implemented,
+a new URI format using a seperate FileSystem would have to be added (f.e. instead
```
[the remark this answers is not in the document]
ekrich: So do you mean "for example" -> "e.g." vs "f.e"?

nativelib/src/main/scala/java/lang/Class.scala
```diff
@@ -144,8 +149,32 @@ final class _Class[A] {
   ): java.lang.reflect.Method = ???
   @stub
   def getMethods(): Array[Method] = ???
-  @stub
-  def getResourceAsStream(name: java.lang.String): java.io.InputStream = ???
+
+  def getResourceAsStream(name: java.lang.String): java.io.InputStream = {
+    val absoluteName =
+      if (name(0) == '/') {
```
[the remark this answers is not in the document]
jchyb: I tried to see what happens on JVM and it turns out that something interesting does. All file names from the specified directory in a jar (so here relative ".") are put to an InputStream and returned. I'll try to implement that for the embedded files. For our use case I fear that there wouldn't be much use for encoding source paths like this, since we cannot access them either way

docs/lib/javalib.rst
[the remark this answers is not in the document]
WojciechMazur: I meant like `were getResources() URI methods implemented,` I cannot understand that, I guess it should be like `when `getResources()` URI methods would be implemented`.

nativelib/src/main/scala/java/lang/Class.scala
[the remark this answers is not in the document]
jchyb: Turns out this doesn't happen on JVM when compiled into a JAR, so I ended up not adding that feature

WojciechMazur (approved):

PR #2476, opened 2021-12-23
Generate static method forwarders in Scala 2

WojciechMazur:
This PR does port generation of static method forwarders from Scala.js and existing implementation for Scala 3 compiler plugin. 

* Ported static forwarders generation to Scala 2 compiler plugin
* Added ScalaNativeOptions for Scala 2 compiler plugin
* Made static forwarders tests shared between Scala 2/3 tests
* Applied fixes to allow to invoke static methods and access static members

### DIFF
--- a/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenExpr.scala
+++ b/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenExpr.scala
@@ -654,8 +654,6 @@ trait NirGenExpr[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
           } else {
             val ty = genType(sym.tpe)
             val qual = genExpr(qualp)
-            // if (sym.isStaticMember) buf.genModule(sym.owner)
-            // else genExpr(qualp)
             buf.fieldstore(ty, qual, name, rhs, unwind)
           }
 
--- a/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenExpr.scala
+++ b/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenExpr.scala
@@ -633,8 +633,11 @@ trait NirGenExpr[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
     ): Val = {
       if (sym == BoxedUnit_UNIT) {
         Val.Unit
-      } else {
+      } else if (!isImplClass(sym.owner)) {
         genApplyStaticMethod(sym, receiver, Seq())
+      } else {
+        val module = genModule(sym.owner)
+        genApplyMethod(sym, statically = true, module, Seq())
       }
     }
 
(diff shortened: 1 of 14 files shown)


nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -617,24 +617,24 @@ trait NirGenExpr[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
...
+    def genStaticMember(receiver: Tree, sym: Symbol)(implicit
+        pos: nir.Position
+    ): Val = {
       if (sym == BoxedUnit_UNIT) {
         Val.Unit
       } else {
-        val ty = genType(sym.tpe)
-        val module = genModule(sym.owner)
-        genApplyMethod(sym, statically = true, module, Seq())
+        genApplyStaticMethod(sym, receiver, Seq())
```
sjrd, line 637: This change is in fact not backward binary compatible. It can break the scenario where there exists a third-party library implementing some part of the JDK (such as scala-java-time) and compiled with 0.4.1. That library would not have static forwarders for the members. But now the compiler will emit calls to those non-existent static forwarders, which will break.

I am not actually aware of any third-party JDK implementation for Scala Native. So if none exists, we're good.

If there is, we can work around this problem in two possible ways:
* Either we introduce some linking magic for compatibility: when trying to reach a static method from a class compiled with 0.4.0 or 0.4.1, if it does not exist, we try calling a method of the same name in the companion object instead. This is completely doable (and we've done worse things in the history of Scala.js IR), but non-trivial.
* Or we still capitalize on the not-so-mature aspect of Scala Native to say that it is a reasonable breakage: there can only be 1-3 such libraries in the wild, and they would be able to recompile and republish newer versions with Scala Native 0.4.2. Those newer versions would not break their own backward binary compatibility, so we don't create a fracture in the binary ecosystem. The only downside is that users of those libraries would have to upgrade those libraries at the same time as they upgrade Scala Native. But this is probably a reasonable price to pay given the state of the ecosystem.

WojciechMazur: I've added a step in `Reach` that does rewrite calls to static methods. I've tested it locally by rewriting all static calls (by disabling one of the conditions) - there were 2 corner cases that I've addressed: 
 * ImplClasses was containing only static methods, I've decided to not produce static methods at all for all ImplClasses members 
 * `lifterTree` (Scala intrinsic method?) was defined only as static in class - I've added additional case that would try to use methods defined in class instead of companion module

sjrd: > * `lifterTree` (Scala intrinsic method?) was defined only as static in class - I've added additional case that would try to use methods defined in class instead of companion module

I think this was an artifact of how you stress-tested the rewrites. I don't think this can happen in practice, because `lifterTree` should only be called from the same compilation unit, I think (so separate compilation and binary compat are not a concern).

Perhaps a more accurate stress-test would be to disable the generation of all *static forwarders*, but otherwise keep everything the same (and in particular, emit already-static methods normally, as static methods).

WojciechMazur: After disabling static forwarders everything seems to be working fine, all unit tests in Scala 2.11 2.12 and 2.13 are able to link and are passing the tests.
In seed `liftedTree` methods are defined as `private static` so they should never be a problem for us.

sjrd (approved):

PR #2480, opened 2021-12-23
Enforce usage of empty parens variant of `stackalloc[T]` and `alloc[T]` in Scala 3

WojciechMazur:
This PR does remove empty-parens variants of methods `alloc[T]` and `stackalloc[T]` from Scala 3 unsafe package.
Even though this change is not source compatible with Scala 2 it does prevent users from getting runtime errors when `stackalloc[T](n)` would be interpreted as `stackalloc[T].apply(n)` - instead of allocation memory of size `n*sizeOf(T)` it would allocate `1*sizeOf(T)` and would try to access received pointer at offset `n` leading to undefined behaviour.  

Additionally, since deprecation annotations are ignored in methods defined as macros, a deprecation warning is now a part of the `stackalloc[T]` and `alloc[T]` definition.  

* Removed no-parens `stackalloc[T]` and `alloc[T]` methods from Scala 3 unsafe package
* Added warning in no-parens `stackalloc[T]` and `alloc[T]` methods in Scala 2 
* Fixed all remaining usages of no parens variants of mentioned method in Scala Native

### DIFF
--- a/nativelib/src/main/scala-2/scala/scalanative/unsafe/UnsafePackageCompat.scala
+++ b/nativelib/src/main/scala-2/scala/scalanative/unsafe/UnsafePackageCompat.scala
@@ -79,6 +79,19 @@ private object MacroImpl {
   import scala.reflect.macros.blackbox.Context
 
   def alloc1[T: c.WeakTypeTag](c: Context)(tag: c.Tree, z: c.Tree): c.Tree = {
+    c.warning(
+      c.enclosingPosition,
+      s"Empty param method `alloc[T]` is deprecated, " +
+        "in Scala 3 `alloc[T](n)` can be interpretted as " +
+        "`alloc[T].apply(n)` leading to runtime erros, " +
+        "use `alloc[T]()` instead "
+    )
+    alloc1Impl(c)(tag, z)
+  }
+
+  private def alloc1Impl[T: c.WeakTypeTag](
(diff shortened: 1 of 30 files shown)


nativelib/src/main/scala-2/scala/scalanative/unsafe/UnsafePackageCompat.scala
```diff
@@ -79,6 +79,19 @@ private object MacroImpl {
   import scala.reflect.macros.blackbox.Context
 
   def alloc1[T: c.WeakTypeTag](c: Context)(tag: c.Tree, z: c.Tree): c.Tree = {
+    c.warning(
```
sjrd, line 82: Is there by any chance a `c.deprecationWarning` that we could use? If yes, that would take into account the `-deprecation` flag, which would be awesome.

sjrd (approved):

nativelib/src/main/scala-2/scala/scalanative/unsafe/UnsafePackageCompat.scala
WojciechMazur: I've not found any mentions of such a method in either black-box or white-box macros, there are only 3 methods that can b used for logging: `echo`, `warning`, `abort`

PR #2496, opened 2022-01-03
Allow to define top-level extern definitons in Scala 3

WojciechMazur:
This PR adds a new feature to Scala 3 compiler plugin allowing to define use use `extern` definitions defined in the top-level of a given file. This allows for an easier definition of externs in Scala 3 and makes it more similar to the ones defined in C. 

* Added runtime tests checking for correct behaviour of top-level defined extern variables and functions
* Added compilation tests checking for cases of non-explicit return type for top-level externs
* Added transformation of top-level externs in `PrepNativeInterop` phase - it would add `@extern` annotation to methods and fields defined with `extern` at rhs, but not having explicit `@extern` annotation
* Adapted detection of extern symbols to handle top-level definitions
* Replaced caching mechanism in `PrepNativeInterop` with `ContextCached` wrapper used in other phases. 
* Added fixes to class constructors to allow definition of extern fields.

### DIFF
--- a/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenExpr.scala
+++ b/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenExpr.scala
@@ -122,12 +122,20 @@ trait NirGenExpr(using Context) {
 
       desugarTree(lhsp) match {
         case sel @ Select(qualp, _) =>
-          val rhs = genExpr(rhsp)
+          def rhs = genExpr(rhsp)
           val sym = sel.symbol
           val name = genFieldName(sym)
-          if (sym.owner.isExternModule) {
-            val externTy = genExternType(sel.tpe)
-            genStoreExtern(externTy, sym, rhs)
+          if (sym.isExtern) {
+            // Ignore intrinsic call to extern in class constructor
+            // It would always present and would lead to undefined behaviour otherwise
+            val shouldIgnoreAssign =
+              curMethodSym.get.isClassConstructor &&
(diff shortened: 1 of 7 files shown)


nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -122,12 +122,20 @@ trait NirGenExpr(using Context) {
 
       desugarTree(lhsp) match {
         case sel @ Select(qualp, _) =>
-          val rhs = genExpr(rhsp)
+          def rhs = genExpr(rhsp)
```
ekrich, line 125: Are you sure `def` is required here? It seems the `rhsp` is evaluated so the two uses of `rhs` don't seem like they would change unless `rhsp` is also a function.

WojciechMazur: Yes, it is needed, otherwise, we might generate a call to `extern` (evaluated as `throw UndefinedError`), which result would not be assigned a an argument for other calls. Also since it's throwing an exception it would not be dead code eliminated - or rather everything after throwing an exception would be treated as dead-code. That's why we need to delay generating `rhsp`. Because `rhs` is always used only once the usage of `def` is better then adding `lazy val` due to overhead, and we don't need to duplicate `genExpr` in case if we would like to stick with `val`

ekrich: Got it. You can see how when reading the code, that using `def` there would generate a question as to why.

PR #1571, opened 2019-04-26
Add support for 32-bit platforms through a linker-level switch

shadaj:
This PR introduces compile/runtime logic for targeting 32-bit platforms. Most notably, this introduces the `Size`/`USize` types, which can be used to safely interact with platform-dependent `ssize_t`/`size_t` types without having to recompile Scala for each platform. Uses of this type are lowered at link time to the appropriate native type.

PR #2502, opened 2022-01-04
Update libunwind to LLVM 12.0.1

shadaj:
Split from #1571


scripts/update_libunwind.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+shopt -s globstar
+
+LIBUNWIND_FOLDER=$1
```
WojciechMazur, line 5: Where can we find upstream sources for libunwind? If it's a git project, we might just clone it to some tmp directory in this script, this way updating in the future might be way easier. Otherwise, some links might be usefull.

WojciechMazur (commented):
Looks good to me, I only have a question according to the update libunwind script.

scripts/update_libunwind.sh
shadaj: Yeah, it's just from the LLVM repo. I'll update the script to clone the sources into a temporary repo.

shadaj: Should be fixed now!

WojciechMazur (approved):

PR #2506, opened 2022-01-05
Port URLDecoder from Scala.js

kpodsiad:


javalib/src/main/scala/java/net/URLDecoder.scala
```diff
@@ -0,0 +1,84 @@
...
+      if (!Charset.isSupported(enc))
+        throw new UnsupportedEncodingException(enc)
+      else
+        Charset.forName(enc)
+    }
+
+    decodeImpl(s, charset)
+  }
+
+  private def decodeImpl(s: String, charset: => Charset): String = {
```
plokhotnyuk, line 28: This decoder is too slow. I've already provided a fast path patch for similar http4s decoder that saves more than 5% CPU cycles on our services: https://github.com/http4s/http4s/pull/5556

kpodsiad: Thanks @plokhotnyuk, I'll swap implementations

unit-tests/shared/src/test/scala/javalib/net/URLDecoderTest.scala
```diff
@@ -0,0 +1,89 @@
...
+class URLDecoderTest {
+
+  private final val utf8 = "utf-8"
+  private final val ReplacementChar = '\uFFFD'
+
+  @Test
+  def decodeTest(): Unit = {
+    def test(encoded: String, expected: String, enc: String = utf8): Unit ={
+      println(s">$expected<")
+      println(s">${URLDecoder.decode(encoded, enc)}<")
```
kpodsiad, line 22: @WojciechMazur I left these on purpose, because there is some strange behavior on native tests, it's the output
```shell
>a b c<
>a b c<
>a b c<
>a b c<
>aßc<
>aßc<
>a  c<
>a  c<
>a b c<
>a b c<
>a £ c<
>a £ c<
>a b c<
>a b c<
>a b c<
>a b c<
>a b c<
>a b c<
>a b c<
>a b c<
[info] Test run started
[info] Test javalib.net.URLDecoderTest.decodeTest started
[error] Test javalib.net.URLDecoderTest.decodeTest failed: org.junit.ComparisonFailure: expected:<a[ b ]c> but was:<a[ b ]c>, took 0.016 sec
[error]     at java.lang.StackTrace$.currentStackTrace(Unknown Source)
[error]     at java.lang.Throwable.fillInStackTrace(Unknown Source)
```
seems like strings aren't compared by value

kpodsiad: I don't know why yet, but problem concerns `utf-16`

kpodsiad: `Charset.decode` returns invalid bytes:
```scala
val decoded = URLDecoder.decode(encoded, enc)
println(expected.toList.map(_.toByte))
println(decoded.toList.map(_.toByte))
```
Native
```shell
List(97, 32, 98, 32, 99)
List(97, -1, 32, -1, 98, -1, 32, -1, 99)
```
vs JVM
```shell
List(97, 32, 98, 32, 99)
List(97, 32, 98, 32, 99)
```

WojciechMazur: Actually, the algorithm ported from http4s is incorrect when it comes to UTF-16, if we would try to run it on the JVM we would get the same results as on the native. There are multiple cases needed attention when trying to add to it support for UTF-16 (auto-endian): 
 *   `out.put(charset.encode(String.valueOf(c)))` would always produce array starting with 0xFFFD header (or 0xFDFF) header  
 * in the first `%` check we always add resolved value from 2 next chars as a single byte, which is incorrect in UTF-16, the char should be properly encoded because at the very end we decode whole byte array. That's why Scala.js was using default CharBuffer and accessory ByteBuffer for handling escaped values and their proper encoding. 
 * Since the algorithm is optimized for the UTF-8 it's probably going to work faster in most cases where utf-8 content is expected, in our case implementation needs to be maximally flexible and compliant with the JVM behaviour 
  * there are some unused valued used eg `mark` ;) 

I've seen that http4s is using iterator instead of accessing elements by index,  but this definitely does not lead to any improvement, since the access time to the next element in each case is constant.  Here's a quick comparison of results taken on my machine with JVM. (each loop was doing 1 million decodings of the same string, so the actual difference is 6-7ns per URL string decoding in favour of using iterator). If we would like to use Scala.js implementation with iterators I'm leaving it at the end of this comment.
 ```
8 iter:  290358340ns
8 index: 284355863ns
9 iter:  280812061ns
9 index: 287956120ns
 ```

Also, a quick explanation - when we run the `testsJVM` in Scala Native build we would **not** use the JDK implementation from `javalib`, but from the JDK distribution, we're currently using in the sbt. They're used to check if our implementation and tests are compliant with official distributions.  

```
>List(-2, -1, 0, 97, 0, 32, 0, 98, 0, 32, 0, 99)<
>List(-2, -1, 0, 97, -2, -1, 0, 32, -2, -1, 0, 98, -2, -1, 0, 32, -2, -1, 0, 99)<
[error] (run-main-2) java.lang.AssertionError: assertion failed: (utf-16,a b c,a%fe%ff%00%20b%fe%ff%00%20c)
[error] java.lang.AssertionError: assertion failed: (utf-16,a b c,a%fe%ff%00%20b%fe%ff%00%20c)
[error]         at scala.runtime.Scala3RunTime$.assertFailed(Scala3RunTime.scala:8)
[error]         at Test$.test(Test.scala:54)
[error]         at Test$.main(Test.scala:45)
[error]         at Test.main(Test.scala)
[error]         at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]         at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error]         at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
````

```
private def decodeImpl(s: String, charset: => Charset): String = {
    val len = s.length
    lazy val charsetDecoder = charset.newDecoder()

    lazy val byteBuffer = ByteBuffer.allocate(len / 3)
    val charBuffer = CharBuffer.allocate(len)

    def throwIllegalHex() = {
      throw new IllegalArgumentException(
        "URLDecoder: Illegal hex characters in escape (%) pattern"
      )
    }

    val in = CharBuffer.wrap(s)
    while (in.hasRemaining()) {
      in.get() match {
        case '+' => charBuffer.append(' ')

        case '%' if in.remaining() < 2 =>
          throwIllegalHex()

        case '%' =>
          val decoder = charsetDecoder
          val buffer = byteBuffer
          buffer.clear()
          decoder.reset()
          var inEscape = true
          while (in.remaining() >= 2 && inEscape) {
            val c1 = Character.digit(in.get(), 16)
            val c2 = Character.digit(in.get(), 16)
            if (c1 < 0 || c2 < 0)
              throwIllegalHex()

            buffer.put(((c1 << 4) + c2).toByte)
            // Peak next char to check if it's also an escape. 
            // If so drop next char to allign with state before loop
            inEscape = in.hasRemaining() && in.get(in.position()) == '%'
            if (inEscape) in.get()
          }

          buffer.flip()
          val decodeResult = decoder.decode(buffer, charBuffer, true)
          val flushResult = decoder.flush(charBuffer)

          if (decodeResult.isError() || flushResult.isError())
            throwIllegalHex()

        case c =>
          charBuffer.append(c)
      }
    }

    charBuffer.flip()
    charBuffer.toString
  }
```

kpodsiad: @WojciechMazur  I force-pushed implementation form your comment

WojciechMazur (approved):

PR #2508, opened 2022-01-06
Implement time zone offset

ekrich:
This PR allows us to get time zone offset which is needed for `java.util.Date` and downstream for `java.time`. We probably would also need to get Time Zone itself as that is needed for `Date` as well but since `Date` is mostly deprecated there is no much need right now. Time Zone could also be used for a lightweight `Locale` type implementation. A full blown `Locale` implementation is already support as a 3rd party package.

There are a few rationale for this addition. 

- Scala.js can provide `java.util.Date` because it has JS `Date`.
- Downstream `java.time` can use this to provide `LocalDate` and `LocalDateTime`
- Scala Native has the cross platform CI that allows for testing on Windows and Unix type platforms.

Currently, there is not a good place for this so it is in runtime so technically private and subject to change. We could in the future provide a `util` package for a few fundamental functions such as this. Windows is supported and the Unix C functions are well supported GNU extensions.

### DIFF
--- a/nativelib/src/main/resources/scala-native/time_zone_offset.c
+++ b/nativelib/src/main/resources/scala-native/time_zone_offset.c
@@ -1,11 +1,11 @@
 #if defined(_WIN32)
 #define WIN32_LEAN_AND_MEAN
-    #include <windows.h>
-    #include <wchar.h>
+#include <windows.h>
+#include <wchar.h>
 #else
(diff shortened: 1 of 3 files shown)


WojciechMazur:
When it comes to 3rd parties I belive resolution of timezone offset could be implemented outside of Scala Native. If we're going to implement `Date` it seems to be good enough to include it in the repo

WojciechMazur (approved):

PR #2511, opened 2022-01-06
Fix compilation errors when using lazy vals in quoted macros

WojciechMazur:
This PR resolves an issue with the compilation of Scala 3 macros using lazy vals, based on a bug found in shapeless3 https://github.com/typelevel/shapeless-3/pull/61#discussion_r779376350 

The problem was caused by trying to inline and evaluate code that was modified in the `AdaptLazyVals` phase. When Scala 3 `Inliner` phase would try to invoke the modified code it would try to invoke Scala Native Intrinsic method (eg. `Intrinsic.classFieldRawPtr`) on the JVM, which would lead to throwing `UndefinedBehaviourException` and resulting in failure of the compiler. 

To mitigate this issue global rewrites performed in the `AdaptLazyVals` (that would be propagated to produce ByteCode or Tasty) phase were replaced with local transformations performed when generating NIR. This way products of compilation other than NIR files would not be affected by our specific rewrites.

* Added test case for macro leading to compilation error 
* Rewritten `AdaptLazyVals` phase to be defined as a helper class, instead of compiler plugin phase
* Updated NIR generation to locally transform AST using `AdaptLAzyVals` instance

### DIFF
--- a/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/AdaptLazyVals.scala
+++ b/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/AdaptLazyVals.scala
@@ -16,7 +16,7 @@ import scala.annotation.{threadUnsafe => tu}
 // its scala native specific counter-part. This is needed, because LazyVals are
 // using JVM unsafe API and static class constructors which are not supported
 // in Scala Native.
-// In theory it could be defined as seperate compilation phase (it was in the past), but
+// In theory it could be defined as separate compilation phase (it was in the past), but
 // it would lead to the problems when using macros - rewritten lazy fields method would
 // be inlined and evaluated at compile time, however modified AST contains Scala Native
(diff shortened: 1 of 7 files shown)


nscplugin/src/main/scala-3/scala/scalanative/nscplugin/AdaptLazyVals.scala
```diff
@@ -12,22 +12,16 @@ import core.Symbols._
...
-  val name = "scalanative-adaptLazyVals"
-}
-
-class AdaptLazyVals extends PluginPhase {
-  val phaseName = AdaptLazyVals.name
-
-  override val runsAfter = Set(LazyVals.name, MoveStatics.name)
-  override val runsBefore = Set(GenNIR.name)
-
+// In theory it could be defined as seperate compilation phase (it was in the past), but
```
ekrich, line 19:
```suggestion
// In theory it could be defined as separate compilation phase (it was in the past), but
```

PR #2518, opened 2022-01-09
ProcessMonitor should not busy wait if there are no active processess

### LINKED ISSUE #2501: Runtime.exec spawns infinitely busylooping ProcessMonitor background thread
```scala
    sys.runtime.exec(Array("/bin/echo"))
    Thread.sleep(60_000)
  }
```

This took me a while to track down, but using Runtime.exec gets you a free busylooping ProcessMonitor thread consuming 100% CPU for the remaining runtime of your program, even when all child processes have exited and none are being waited on. (See process_monitor.cpp)

It looks like this was done to support the timeout variant of waitFor the Java Process API offers. That does indeed appear to be a bit tricky to implement without OS-specific APIs, but this is arguably worse than not implementing the timeout variants at all and just implementing the non-timeout cases using vanilla waitpid / waitpid WNOHANG. Something like this maybe:
```scala
  private def checkResult(pid: Int): CInt = {
    val status: Ptr[CInt] = stackalloc[CInt]()

    val ret = waitpid(pid, status, WNOHANG)

    if (ret != 0)
      !status
    else
      -1
  }

  private def waitForPid(pid: Int): CInt = {
    val status: Ptr[CInt] = stackalloc[CInt]()

    waitpid(pid, status, 0)

    !status
  }
```

I suspect the busylooping solution is fundamentally unsalvagable. According to this stackoverflow post there might be some pipe tricks that could fit the non-busylooping and cross-platform waiting with timeout bill, and various OS-specific solutions, but that's a little outside of my experience:
https://stackoverflow.com/questions/282176/waitpid-equivalent-with-timeout

WojciechMazur:
This PR fixes #2501 by limiting the amount of busy waiting in the background thread. Previously, spawning any processes would lead to creating a thread that would constantly busy-wait in the background - this happens because in case if there are no active subprocesses, call to `waitpid(-1, $status, 0)` would return immediately. 
To fix this issue a condition and a counter of active processes were introduced. Each freshly created subprocess starts it's execution by notifying ProcessMonitor incrementing the counter and signalling the condition. When there would be no active processes the background ProcessMonitor thread would wait until the condition would be signalled.

### DIFF
--- a/javalib/src/main/scala/java/lang/process/UnixProcess.scala
+++ b/javalib/src/main/scala/java/lang/process/UnixProcess.scala
@@ -142,6 +142,8 @@ object UnixProcess {
   @link("pthread")
   @extern
   private[this] object ProcessMonitor {
+    @name("scalanative_process_monitor_notify")
+    def notifyMonitor(): Unit = extern
     @name("scalanative_process_monitor_check_result")
     def checkResult(pid: Int): CInt = extern
     @name("scalanative_process_monitor_init")
@@ -198,6 +200,7 @@ object UnixProcess {
          * to run on the child before execve inside of a method.
          */
         def invokeChildProcess(): Process = {
+          ProcessMonitor.notifyMonitor()
           if (dir != null) unistd.chdir(toCString(dir.toString))
           setupChildFDS(!infds, builder.redirectInput(), unistd.STDIN_FILENO)
           setupChildFDS(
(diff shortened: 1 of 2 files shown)


shado23:
That's definitely a welcome improvement. In the long run, the busy looping may still be problematic when spawning longer running background processes.

PR #2527, opened 2022-01-11
Upgrade scalafmt to 3.3.2

ekrich:
We are currently waiting on an update to allow more than 2 `///` starting a comment. This is covered in https://github.com/scalameta/scalafmt/pull/3051

Change summary:

1. Add space after inline comment - this is fine.
```scala
-        .hasNext //used to materialize underlying LazyList (since 2.13)
+        .hasNext // used to materialize underlying LazyList (since 2.13)
```
2. Do not indent next `@` tag in scaladoc - this is fine.
```scala
    *  @param source
-   *    @param count
+   *  @param count
```
3. ~~This change I am not sure about~~ I think this change is fine as well after auditing the other code. It removes the indent on infix continuation in case statements - https://github.com/scalameta/scalafmt/pull/3018
It does not make a ton of changes but the infix continuation is not exactly consistent with other infix continuation wrapping with indent in the code base.
```scala
+++ b/Users/eric/workspace/scala-native/javalib/src/main/scala/java/lang/StackTraceElement.scala
@@ -39,5 +39,5 @@
         getClassName == that.getClassName &&
-          getMethodName == that.getMethodName &&
-          getFileName == that.getFileName &&
-          getLineNumber == that.getLineNumber
+        getMethodName == that.getMethodName &&
+        getFileName == that.getFileName &&
+        getLineNumber == that.getLineNumber
       case _ =>
```

[diff.txt](https://github.com/scala-native/scala-native/files/7849047/diff.txt)

### DIFF
--- a/javalib-ext-dummies/src/main/scala/java/util/Locale.scala
+++ b/javalib-ext-dummies/src/main/scala/java/util/Locale.scala
@@ -40,9 +40,9 @@ final class Locale(
     that match {
       case that: Locale =>
         getLanguage() == that.getLanguage() &&
-          getCountry() == that.getCountry() &&
-          getVariant() == that.getVariant() &&
-          extensions == that.extensions
+        getCountry() == that.getCountry() &&
+        getVariant() == that.getVariant() &&
+        extensions == that.extensions
       case _ =>
         false
     }
--- a/javalib-ext-dummies/src/main/scala/java/util/Locale.scala
+++ b/javalib-ext-dummies/src/main/scala/java/util/Locale.scala
@@ -40,9 +40,9 @@ final class Locale(
     that match {
       case that: Locale =>
         getLanguage() == that.getLanguage() &&
-        getCountry() == that.getCountry() &&
-        getVariant() == that.getVariant() &&
-        extensions == that.extensions
+          getCountry() == that.getCountry() &&
+          getVariant() == that.getVariant() &&
+          extensions == that.extensions
       case _ =>
         false
     }
--- a/unit-tests/shared/src/test/scala/javalib/util/DefaultFormatterTest.scala
+++ b/unit-tests/shared/src/test/scala/javalib/util/DefaultFormatterTest.scala
@@ -196,11 +196,12 @@ class DefaultFormatterTest {
         classOf[UnsupportedEncodingException],
         new Formatter(notExist, "ISO 1111-1")
       )
-    } finally if (notExist.exists()) {
-      // Fail on RI on Windows, because output stream is created and
-      // not closed when exception thrown
-      assertTrue(notExist.delete())
-    }
+    } finally
+      if (notExist.exists()) {
+        // Fail on RI on Windows, because output stream is created and
+        // not closed when exception thrown
+        assertTrue(notExist.delete())
+      }
   }
 
   @Test def constructorPrintStream(): Unit = {
--- a/unit-tests/shared/src/test/scala/javalib/util/zip/InflaterTest.scala
+++ b/unit-tests/shared/src/test/scala/javalib/util/zip/InflaterTest.scala
@@ -116,9 +116,11 @@ class InflaterTest {
   def setUp(): Unit = {
     outPutBuff1 = Array[Byte](120, 94, 99, 100, 102, 97, -25, 72, 45, 42, -87,
       52, 5, 0, 6, -12, 2, 17) ++ Array.fill[Byte](483)(0)
-    outPutDiction = Array[Byte](120, 63, 13, 10, 107, 2, 20, 99, 97, 101, 102,
-      74, 76, 98, 99, -25, -32, 100, 40, 54, 46, 47, 2, 13, 112, 2,
-      127) ++ Array.fill[Byte](474)(0)
+    outPutDiction =
+      Array[Byte](120, 63, 13, 10, 107, 2, 20, 99, 97, 101, 102, 74, 76, 98, 99,
+        -25, -32, 100, 40, 54, 46, 47, 2, 13, 112, 2, 127) ++ Array.fill[Byte](
+        474
+      )(0)
   }
 
   @Test def finished(): Unit = {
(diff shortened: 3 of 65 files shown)


ekrich:
Current counts
```sh
% scripts/scalafmt --test 2> diff.txt         
error: --test failed
% wc -l diff.txt                              
     771 diff.txt

```

WojciechMazur:
Thank you Eric for your summary, the issue you mentioned was fixed and merged so I think we can update scalafmt to the next release containing the fix

ekrich:
The undesired change that makes `///`  go to `// /` is fixed. The diff is now 614 lines.
[diff.txt](https://github.com/scala-native/scala-native/files/7915417/diff.txt)

WojciechMazur:
Sure, go ahead, let's merge it after the release. Thank you!

unit-tests/shared/src/test/scala/javalib/util/zip/InflaterTest.scala
```diff
@@ -116,9 +116,11 @@ class InflaterTest {
   def setUp(): Unit = {
     outPutBuff1 = Array[Byte](120, 94, 99, 100, 102, 97, -25, 72, 45, 42, -87,
       52, 5, 0, 6, -12, 2, 17) ++ Array.fill[Byte](483)(0)
-    outPutDiction = Array[Byte](120, 63, 13, 10, 107, 2, 20, 99, 97, 101, 102,
-      74, 76, 98, 99, -25, -32, 100, 40, 54, 46, 47, 2, 13, 112, 2,
-      127) ++ Array.fill[Byte](474)(0)
+    outPutDiction =
+      Array[Byte](120, 63, 13, 10, 107, 2, 20, 99, 97, 101, 102, 74, 76, 98, 99,
+        -25, -32, 100, 40, 54, 46, 47, 2, 13, 112, 2, 127) ++ Array.fill[Byte](
```
WojciechMazur, line 121: Can we move the Array.fill to the next line?

unit-tests/shared/src/test/scala/javalib/util/DefaultFormatterTest.scala
```diff
@@ -196,11 +196,12 @@ class DefaultFormatterTest {
         classOf[UnsupportedEncodingException],
         new Formatter(notExist, "ISO 1111-1")
       )
-    } finally if (notExist.exists()) {
-      // Fail on RI on Windows, because output stream is created and
-      // not closed when exception thrown
-      assertTrue(notExist.delete())
-    }
+    } finally
```
WojciechMazur, line 199: I wish the previous style could stay, it was nice and really readable

javalib-ext-dummies/src/main/scala/java/util/Locale.scala
```diff
@@ -40,9 +40,9 @@ final class Locale(
     that match {
       case that: Locale =>
         getLanguage() == that.getLanguage() &&
-          getCountry() == that.getCountry() &&
-          getVariant() == that.getVariant() &&
-          extensions == that.extensions
+        getCountry() == that.getCountry() &&
+        getVariant() == that.getVariant() &&
```
WojciechMazur, line 44: Do you maybe know if there is a setting that would allow us to stick with the previous layout? It actually made a lot of sense to indent the continued condition to make it more visible and to not mismatch it with some other expression

WojciechMazur (commented):
Overall looks nice, I've left some comments but it's fine if we would not be able to revert some of the changes

ekrich:
@WojciechMazur I will look into these and get back. Thanks for the review.

unit-tests/shared/src/test/scala/javalib/util/DefaultFormatterTest.scala
ekrich: I think this one is designed that way because it works for braceless syntax. I can try and figure something out but since this is pretty much a one off it may change more with an option.

unit-tests/shared/src/test/scala/javalib/util/zip/InflaterTest.scala
ekrich: Ok, that worked and looks nice

javalib-ext-dummies/src/main/scala/java/util/Locale.scala
ekrich: I talked to the powers at be but apparently this is top-level in a case and the format is different - https://github.com/scalameta/scalafmt/pull/3018 I don't really understand why but ...

kitbellew: it's top level but you can exclude `&&`

ekrich: I see that you.

ekrich: @kitbellew found the rule - https://github.com/scala-js/scala-js/blob/master/CODINGSTYLE.md#long-expressions-with-binary-operators

ekrich:
We know are now targeting the yet to be released 3.3.4.

Based on the new changes in `scalafmt` due to some new `indentOperator.exemptScope = aloneArgOrBody`special rules, we have a decision to make. Here is the reference: https://scalameta.org/scalafmt/docs/configuration.html#indentoperator

Option 1: If we leave the default which is not recommended (according to the docs) there are no changes to the current pushed changes.

Option 2: `indentOperator.exemptScope = aloneEnclosed` indents more continuations.
[diff.txt](https://github.com/scala-native/scala-native/files/7967302/diff.txt)
300 lines in diff.

Option 3: `indentOperator.exemptScope = aloneArgOrBody` added for Scala.js and undos much of the current indenting - 433 lines in diff.
[diff.txt](https://github.com/scala-native/scala-native/files/7967309/diff.txt)

Option 4: `indentOperator.exemptScope = all`
545 lines in diff.
[diff.txt](https://github.com/scala-native/scala-native/files/7967317/diff.txt)

The default (Option 1) removed some of the indenting - these changes are pushed. Option 2 seems to add more indentation which might be more readable. Option 3 is a more complicated rule. Option 4 removes the indents.

Myself, I am favoring Option 2 as Option 1 is not recommended even though we have to add an additional setting.

WojciechMazur:
@ekrich I'm also in favor of Option 2, I think we can use it

WojciechMazur (approved):

PR #2532, opened 2022-01-20
Add missing java.io.channels.Channels methods

jchyb:
Missing Channels methods were added, along with some newly prepared tests. `newInputStream(ch: AsynchronousByteChannel)` and `newOutputStream(ch: AsynchronousByteChannel)` were omitted for now, as `AsynchronousByteChannel`s are not implemented as well - perhaps it is better to wait for multithreading with both.

All tests worked with java 17 jvm, but two failed on java 8 despite them following the java documentation, so they were assumed-out. 

Inspired by reports about missing methods used in os-lib on scala-native discord.


WojciechMazur (approved):
Thank you for this PR, I can't see any issues worth discussion. Also, I appreciate testing it against newer version of the JDK

PR #2558, opened 2022-02-10
Don't include `.tasty` files in embeded resources

WojciechMazur:


jchyb (commented):
I think we also list excluded extensions in the javalib docs (also sorry about that, I must have forgot to test resources locally on scala 3)

WojciechMazur:
@jchyb Thanks for remainder, fixed

PR #2565, opened 2022-02-19
Port `java.util.BitSet`

armanbilge:
All credit to @er1c for the effort in https://github.com/scala-js/scala-js/pull/4217.

The changes I made were to use 64-bit words and related adjustments.

### DIFF
--- a/javalib/src/main/scala/java/util/BitSet.scala
+++ b/javalib/src/main/scala/java/util/BitSet.scala
@@ -1,6 +1,6 @@
-package java.util
+// Ported from Scala.js commit: c0be6b6 dated: 2021-12-22
 
-// Ported from Scala.js
+package java.util
 
 import java.io.Serializable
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/util/BitSet.scala
```diff
@@ -0,0 +1,678 @@
+package java.util
+
+// Ported from Scala.js
```
ekrich, line 3: Do you mind putting the SHA and the date of the port? There are some other files in the javalib you can use as an example, maybe HashMap? Also for the test file.

armanbilge: Yes, absolutely! Done.

WojciechMazur (approved):
Thank you for porting this! Looks good. 
You should now only run `./scripts/scalafmt` in your project and push the changes to fix the check-lint step

PR #2566, opened 2022-02-20
Refresh #2351 "Allow to embed Rust in native sources"

armanbilge:
Picking up from https://github.com/scala-native/scala-native/pull/2351:
- [x] merge in `main` and resolve conflicts
- [x] fix bincompat issues on Scala 3
- [ ] fix LTO tests. So far I'm trying to update ubuntu/llvm in CI, is this acceptable?
    ```
    Error:  /usr/bin/ld: error: LLVM gold plugin has failed to create LTO module: Unknown attribute kind (70) (Producer: 'LLVM13.0.0-rust-1.58.1-stable' Reader: 'LLVM 6.0.0')
    Error:  clang: error: linker command failed with exit code 1 (use -v to see invocation)
    Error:  Failed to link /home/scala-native/scala-native/unit-tests/native/.2.13/target/scala-2.13/tests-test-out
    Error:  (tests2_13 / Test / nativeLink) Failed to link /home/scala-native/scala-native/unit-tests/native/.2.13/target/scala-2.13/tests-test-out
    ```
- [ ] what else?


unit-tests/native/src/test/scala/scala/scalanative/unsafe/RustIntegrationTest.scala
```diff
@@ -0,0 +1,52 @@
...
+object RustIntegrationTest {
+  @BeforeClass
+  def assumeSupportsRust() = {
+    // LTO w/ rust requires a newer LLVM than the minimum LLVM version for SN
+    assumeFalse(
+      "RustIntegrationTest not available in the current build",
+      sys.env.get("SCALANATIVE_LTO").exists(_ != "none")
+    )
+  }
+}
```
armanbilge, line 29: Hmm, this doesn't work because it's a _linking_ problem, not a runtime problem.

WojciechMazur:
Thank you for restoring the original PR, good work! 

In a meantime, since it was first published I was thinking that maybe we could introduce some plugin style integration for different sources - I think it would be the best solution, but embedded support might be good as well for now. However, we could try to create an abstraction for Rust stuff in some 3rd party library, e.g. containing abstraction over references (read-only access to pointer)

armanbilge:
I liked the idea about a plugin style integration, so I did an experiment to see how it might work in https://github.com/scala-native/scala-native/pull/2572.

ekrich:
Overall, this is very cool but I don't think the want to get into the Rust library building business. Rust libraries can either come with `extern C` ability or we can add glue code for Rust as we do for C. Building a plugin outside Scala Native core is doable with the code here if someone wants that capability.

I think it would be appropriate to just allow `.rs` files in the `scala-native` directory and just call `rustc`. In my initial experiments this worked fine. I did notice that if a Rust source includes another, it gets compiled first but then Rust just passes it over if it is already compiled. Since our build works like building with `make` we include each file.

We could go the plugin route as well but it might be overkill at this. I think adding the `rs` extension along with the proper arguments for `rustc` would be fine along with some documentation for a PR and then we would at least have some Rust support for `0.5.x`.

He-Pin:
Maybe zig is simpler？

ekrich:
We can add rust. I will tackle it after the current build tools PR is merged.

PR #2571, opened 2022-02-21
Refactor multi-arch CI setup

WojciechMazur:
This PR contains changes allowing for faster testing of Scala Nativ in multi-arch CI setups and simplifies existing workflows. 

The current setup was problematic in usage - it was using multiarch container (emulated using qemu) in which we installed dedicated JDK distribution and compiled (almost) natively for target architecture all our codebase. This approach, due to emulation was really slow - just compiling scala sources and building tests native binary could take from 1h up to 3 hours. 

The new approach uses (native) amd64 host for fast compilation of our sources and producing executable binary for a given target, while the emulator is being only used to run the binary file. It is done by using crossdock based docker image, containing the toolchain and libraries needed for cross-compilation.  The only drawback of new approach is the need for manual compilation of needed C dependencies (zlib, boehm gc) because crossdock does has an easy setup for installing libraries, however due to limited amount of dependencies is not a big problem for us. 

To make testing possible we introduce an internal workaround inside `ProcessRunner` - we can now use JVM system properties to pass additional emulator arguments which would run our cross-compiled test runner binary. In the future we might think about introducing such an option in a public API, however, it might be problematic in usage. 

Minimal arm64 CI tests workflow has been established, however, it is currently limited to only starting cross-compiled test-runner. Actual tests would be added after fixing arm64 runtime issues in the follow up PRs.

* Default amd64 Linux tests are now run being run in Github Actions without an additional testing container. 
* Removed redundant Linux tests by simplifying env matrix 
* Replaced multiarch based docker image with crossdock based setup - cross-compile on amd64, test on emulated target arch
* Added workaround for passing emulator config to ProcessRunner via JVM system properties
* Added new workflow testing new cross-compilation workflow
* Added additional flags used when compiling sources needed for exceptions handling on arm64

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -55,8 +55,9 @@ private[scalanative] object LLVM {
           if (config.targetsWindows) Seq("-g")
           else Nil
         }
+        val expectionsHandling = List("-fexceptions", "-fcxx-exceptions", "-funwind-tables")
         val flags = opt(config) +: "-fvisibility=hidden" +:
-          stdflag ++: platformFlags ++: config.compileOptions
+          stdflag ++: platformFlags ++: expectionsHandling ++: config.compileOptions
         val compilec =
           Seq(compiler) ++ flto(config) ++ flags ++ target(config) ++
             Seq("-c", inpath, "-o", outpath)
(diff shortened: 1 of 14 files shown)


tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -55,8 +55,10 @@ private[scalanative] object LLVM {
           if (config.targetsWindows) Seq("-g")
           else Nil
         }
+        val expectionsHandling =
```
ekrich, line 58: Little late but this should have been `exceptionsHandling`.

armanbilge: I fixed this in https://github.com/scala-native/scala-native/pull/2572 :)

PR #2572, opened 2022-02-22
Extract `NativeSourcePlugin`

armanbilge:
This PR branches off of https://github.com/scala-native/scala-native/pull/2566 to try extracting out a plugin for native sources as suggested in https://github.com/scala-native/scala-native/pull/2566#issuecomment-1046936972.


tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
```diff
@@ -0,0 +1,256 @@
...
+    // C++14 or newer standard is needed to compile code using Windows API
+    // shipped with Windows 10 / Server 2016+ (we do not plan supporting older versions)
+    if (config.targetsWindows) Seq("-std=c++14")
+    else Seq("-std=c++11")
+
+  protected override def compiler(config: Config): String = config.clangPP.abs
+
+}
+
+case object RustSourcePlugin extends NativeSourcePlugin {
```
armanbilge, line 115: I kept this here for now, but I guess the idea is this can move into another repo/library? :)

tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
```diff
@@ -0,0 +1,256 @@
...
+
+import NativeSourcePlugin._
+
+trait NativeSourcePlugin {
+
+  def extensions: Seq[String]
+
+  def compile(config: Config, path: Path): Option[CompilationContext]
+
+}
```
armanbilge, line 17: This is the new interface for compiling native sources.

tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
```diff
@@ -0,0 +1,256 @@
...
+object NativeSourcePlugin {
+
+  case class CompilationContext(
+      command: Seq[String],
+      result: CompilationResult
+  )
+
+}
+
+sealed abstract class ClangSourcePlugin extends NativeSourcePlugin {
```
armanbilge, line 28: I hope this is not over-engineered 😅

tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
```diff
@@ -0,0 +1,256 @@
...
+case object CSourcePlugin extends ClangSourcePlugin {
+
+  /** C file extension: ".c" */
+  val cExt = ".c"
+
+  /** Assembler file extension: ".S" */
+  val assemblerExt = ".S"
+
+  /** List of source patterns used: ".c, .cpp, .S" */
+  val extensions = Seq(cExt, assemblerExt)
```
armanbilge, line 92: I wasn't sure where `.S` should go, so I snuck it in here.

Also need to fix that comment :)

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -56,6 +56,8 @@ sealed trait NativeConfig {
    */
   def embedResources: Boolean
 
+  def nativeSourcePlugins: Seq[NativeSourcePlugin]
```
armanbilge, line 59: Is the `NativeConfig` the right place to keep these?

tools/src/main/scala/scala/scalanative/build/Discover.scala
```diff
@@ -42,6 +42,9 @@ object Discover {
     path
   }
 
+  def rustc: Path = discover("rustc", "RUSTC")
+  def cargo: Path = discover("cargo", "CARGO_HOME")
```
armanbilge, line 46: If the rust integration is extracted, I guess these should be moved/removed as well.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -128,7 +135,9 @@ object NativeConfig {
       linkStubs = false,
       optimize = true,
       linktimeProperties = Map.empty,
-      embedResources = false
+      embedResources = false,
+      nativeSourcePlugins =
+        Seq(CSourcePlugin, CppSourcePlugin, RustSourcePlugin)
```
armanbilge, line 140: I put them all here for now, for backwards-compatibility. Maybe in future it should be empty, and default C/CPP plugins are populated elsewhere?

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -31,44 +24,36 @@ private[scalanative] object LLVM {
...
    *    The paths of the `.o` files.
    */
-  def compile(config: Config, paths: Seq[Path]): Seq[Path] = {
+  def compile(
+      config: Config,
+      paths: Seq[Path]
+  ): Seq[CompilationResult] = {
+    import NativeSourcePlugin._
+
+    val plugins = LlSourcePlugin +: config.compilerConfig.nativeSourcePlugins
```
armanbilge, line 33: IIUC the `.ll` files are part of the core functionality—Scala Native will not work without them. So it seems that it should not be optional.

WojciechMazur, line 33: According to the previous comment - if we would only define additional plugins in NativeConfig we should here define also CLang and Cpp plugins. Maybe, if in the future someone would like to introduce better Cpp integration we could add settings to nativeConfig allowing for disabling our internal implementations.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
WojciechMazur: I think that could be the only way we could use to pass extensions without the usage of reflection. 
I can imagine the following scenario: 
We can define an sbt plugin, which could automatically populate some fields of nativeConfig,  

```
object ScalaNativeRustIntegrationPlugin extends AutoPlugin {
  override def requires: Plugins = plugins.ScalaNativePlugin
  
  object autoImport {
   val rustIntegrationConfig = taskKey[RustPLuginConfig](???) 
   val rustIntegrationPlugin = taskKey[RustIntegarionPlugin](???)
   
   case class RustPluginConfig() 
   class RustIntegarionPlugin(config: RustPluginConfig) extends NativeSourcePlugin  
 }
}
```

and then in users code 

```
val foo = project.in(file("foo")))
  .enablePlugins(ScalaNativeRustIntegrationPlugin)
  .settings(
    rustIntegrationConfig := ???,
    nativeConfig ~= _.withNativeSourceCompilerPlugins(rustIntegrationPlugin.value)
  )
```

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -31,44 +24,36 @@ private[scalanative] object LLVM {
...
-        val flags = opt(config) +: "-fvisibility=hidden" +:
-          stdflag ++: platformFlags ++: config.compileOptions
-        val compilec =
-          Seq(compiler) ++ flto(config) ++ flags ++ target(config) ++
-            Seq("-c", inpath, "-o", outpath)
+    paths.par.flatMap { path =>
+      val inpath = path.toAbsolutePath()
+
+      val compiler = plugins
+        .find(p => p.extensions.exists(ext => inpath.toString.endsWith(ext)))
```
WojciechMazur, line 39: In the future, we might think about improvement for that, eg.  building a Map (extension -> impl)

WojciechMazur, line 39: Also, we should check if we don't have clashes between plugins when two or more of them would use the same set of extensions

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -166,33 +174,55 @@ private[scalanative] object NativeLib {
...
 
+  private def srcExtensions(plugins: Seq[NativeSourcePlugin]): Seq[String] =
+    plugins.flatMap(_.extensions)
+
   /** Used to find native source files in jar files */
-  private val jarSrcRegex: String = {
-    val regexExtensions = srcExtensions.mkString("""(\""", """|\""", ")")
+  private def jarSrcRegex(plugins: Seq[NativeSourcePlugin]): String = {
+    val regexExtensions =
+      srcExtensions(plugins).mkString("""(\""", """|\""", ")")
```
WojciechMazur, line 183: There are a lot of plugins passing here, maybe we move all of this method to a single `Impl` taking and plugins list. This might allow for some optimizations also. In most cases, we would probably create at most 2 instances per run, (1 instance per findNativeLibs and findNativePaths call)

WojciechMazur (commented):
Great job on this topic, I really like the separation of compilation logic depending on the source. 
I don't have today more time to spend on it, but it looks promising. 

I wonder if in the final implementation we can merge some ideas from #2325 eg. filtering sources.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -106,6 +108,11 @@ sealed trait NativeConfig {
   def withEmbedResources(
       value: Boolean
   ): NativeConfig
+
+  def withNativeSourcePlugins(plugins: Seq[NativeSourcePlugin]): NativeConfig
+
+  def withNativeSourcePlugin(plugin: NativeSourcePlugin): NativeConfig =
+    withNativeSourcePlugins(nativeSourcePlugins :+ plugin)
```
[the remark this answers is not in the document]
armanbilge: Actually, I wrote these with different purpose.

`withNativeSourcePlugin` _appends_ a single plugin to your list of plugins.

`withNativeSourcePlugins` _replaces_ the entire list of plugins. Otherwise there is no way to remove and/or change the order.

I have some more thoughts on this topic, in the other comments.

tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
```diff
@@ -0,0 +1,256 @@
...
+package build
+
+import java.io.File
+import java.nio.file.{Files, Path, Paths}
+import scalanative.build.IO.RichPath
+import scalanative.build.Mode._
+
+import NativeSourcePlugin._
+
+trait NativeSourcePlugin {
```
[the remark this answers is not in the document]
armanbilge: Yes, that seems good 👍

tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
```diff
@@ -0,0 +1,256 @@
...
+        command = Seq(compiler(config)) ++
+          LLVM.flto(config) ++
+          flags ++
+          LLVM.target(config) ++
+          Seq("-c", inpath.abs, "-o", outpath.abs),
+        result = ObjectFile(outpath)
+      )
+    }
+    else if (alreadyExists) Some(CompilationContext(Nil, ObjectFile(outpath)))
+    else None
```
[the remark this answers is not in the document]
armanbilge: Thanks, good catch :)

tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
```diff
@@ -0,0 +1,256 @@
...
+
+}
+
+sealed abstract class ClangSourcePlugin extends NativeSourcePlugin {
+
+  protected def stdflag(config: Config): Seq[String]
+
+  protected def compiler(config: Config): String = config.clang.abs
+
+  protected def forceRebuild: Boolean = false
```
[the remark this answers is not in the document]
armanbilge: Makes sense 👍

tools/src/main/scala/scala/scalanative/build/NativeSourcePlugin.scala
armanbilge: Yes, I agree :) It can be complicated to modify the CI in the primary repo to accomodate rust's special requirements. I think all the rust integration can live very naturally in another repo: a compiler plugin, some std library wrappers like you suggested, etc.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
armanbilge: You made another comment:
> Maybe, if in the future someone would like to introduce better Cpp integration we could add settings to nativeConfig allowing for disabling our internal implementations.

I think instead of adding more settings, we should keep the C and Cpp plugins as part of the configurable list instead of force-enabling them. They can be included by default but then they will be easy to remove/replace. They should definitely remain part of the core repo always :)

I think the Rust makes sense as another repo.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -232,6 +248,7 @@ object NativeConfig {
         | - optimize:           $optimize
         | - linktimeProperties: $listLinktimeProperties
         | - embedResources:     $embedResources
+        | - nativeSourcePlugins:${nativeSourcePlugins.mkString(",")}
```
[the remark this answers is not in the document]
armanbilge: Yes, this should be improved 👍

tools/src/main/scala/scala/scalanative/build/LLVM.scala
armanbilge: See https://github.com/scala-native/scala-native/pull/2572#discussion_r812598777.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
armanbilge: Yes, actually we can make it even easier for the user :)

```scala
object ScalaNativeRustIntegrationPlugin extends AutoPlugin {
  override def requires: Plugins = plugins.ScalaNativePlugin
  
  object autoImport {
   val rustIntegrationConfig = settingKey[RustPluginConfig](???) 
   
   case class RustPluginConfig()
 }

  override def projectSettings = Seq(
    nativeConfig := {
      val plugin = new RustSourceCompilerPlugin(rustIntegrationConfig.value)
      nativeConfig.value.withNativeSourceCompilerPlugin(plugin)
    }
  )
}
```

Then user code is just:
```scala
val foo = project.in(file("foo"))
  .enablePlugins(ScalaNativeRustIntegrationPlugin)
  .settings(
    rustIntegrationConfig := ???,
  )
```

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
armanbilge: Oh yeah, that will be much better 😆

tools/src/main/scala/scala/scalanative/build/LLVM.scala
armanbilge: Probably a good idea :) I was thinking there can be some priority/precedence based on their ordering, but that has got limited use-case and potential for confusion probably :)

armanbilge:
> I wonder if in the final implementation we can merge some ideas from https://github.com/scala-native/scala-native/pull/2325 eg. filtering sources.

@ekrich do you know what the connection of this PR is to https://github.com/scala-native/scala-native/pull/2325? It wasn't obvious to me.

ekrich:
I haven't had enough time yet to look at this PR. I don't think they are super related. Some of the parts of the build plugin concept had to do with filtering for windows that is now done in a different way and also conditional linking solved some of the things that were needed. I will put a comment on #2325 as an update.

tools/src/main/scala/scala/scalanative/build/NativeSourcesCompilerPlugin.scala
```diff
@@ -0,0 +1,282 @@
...
+  protected def forceRebuild(input: Path): Boolean = false
+
+  private def opt(config: Config): String =
+    config.mode match {
+      case Mode.Debug       => "-O0"
+      case Mode.ReleaseFast => "-O2"
+      case Mode.ReleaseFull => "-O3"
+    }
+
+  def compile(config: Config, inpath: Path): Option[CompilationContext] = {
```
ekrich, line 54: Could you explain why this returns an option because in the C/C++ plugin it seems to be not needed?

armanbilge: That's a good question. I'm actually not 100% sure, this was the existing method signature. The rust compiler does return `None` in some instances. It seems to indicate that there is nothing to be done. But I'm not sure what differentiates it from returning a `CompilationContext` with an empty command. @WojciechMazur will probably know :)

https://github.com/scala-native/scala-native/blob/653dfabe2ebbfde9495e861f4166ae3e98be8c9d/tools/src/main/scala/scala/scalanative/build/NativeSourcesCompilerPlugin.scala#L183

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -216,22 +239,25 @@ object NativeConfig {
...
+        | - GC:                  $gc
+        | - mode:                $mode
+        | - LTO:                 $lto
+        | - linkStubs:           $linkStubs
+        | - check:               $check
+        | - checkFatalWarnings:  $checkFatalWarnings
+        | - dump:                $dump
+        | - optimize:            $optimize
+        | - linktimeProperties:  $listLinktimeProperties
+        | - embedResources:      $embedResources
```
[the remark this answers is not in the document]
armanbilge: I forgot I renamed the config in question. Actually we need more spacers 😆

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -126,7 +127,12 @@ private[scalanative] object LLVM {
         } else Seq("-rdynamic")
       flto(config) ++ platformFlags ++ Seq("-o", outpath.abs) ++ target(config)
     }
-    val paths = objectsPaths.map(_.abs)
+    // Make sure that libraries are linked as the last ones
+    val paths = {
+      compilationOutput.filter(_.isInstanceOf[ObjectFile]) ++
+        compilationOutput.filter(_.isInstanceOf[Library])
```
ekrich, line 133: Can you use `partition` here to avoid 2 transits?

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -175,10 +192,13 @@ private[scalanative] object NativeLib {
 
   private val jarPattern = Pattern.compile(jarSrcRegex)
 
-  private def isNativeFile(name: String): Boolean =
+  private def isNativeFile(name: String): Boolean = {
     jarPattern.matcher(name).matches()
+  }
```
ekrich, line 197: Spurious change?

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -175,10 +192,13 @@ private[scalanative] object NativeLib {
...
 
-  private def readDir(path: Path): Option[Path] =
+  private def readDir(
+      path: Path
+  ): Option[Path] =
```
ekrich, line 201: Here too?

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -192,7 +212,11 @@ private[scalanative] object NativeLib {
 
   /** Used to find native source files in directories */
   private def srcPatterns(path: Path): String =
-    srcExtensions.mkString(s"glob:${srcPathPattern(path)}**{", ",", "}")
+    srcExtensions.mkString(
+      s"glob:${srcPathPattern(path)}**{",
+      ",",
+      "}"
+    )
```
ekrich, line 219: Here?

tools/src/main/scala/scala/scalanative/build/NativeLib.scala
```diff
@@ -218,7 +242,10 @@ private[scalanative] object NativeLib {
    *  @return
    *    The source pattern
    */
-  private def destSrcPattern(workdir: Path, destPath: Path): String = {
+  private def destSrcPattern(
+      workdir: Path,
+      destPath: Path
+  ): String = {
```
ekrich, line 248: And here?

ekrich (commented):
Just a few small things - cosmetic. Will look more later.

PR #2574, opened 2022-02-24
Synchronise generation of gyb templates and remove line directive

WojciechMazur:
This PR does cherry-pick changes made to the generation of gyb templates to lower the amount of diffs when introducing small changes. This change was originally a part of #1571. Original script was update to make it more user friendly.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -7,7 +7,9 @@ version = "3.4.0"
 docstrings.style = AsteriskSpace
 project.git = true
 project.excludePaths = [
-  "glob:**/scalalib/**"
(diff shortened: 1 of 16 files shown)


ekrich:
Do we already use

```
// format: off
...
// format: on
```
or do you format after generation?

WojciechMazur:
@ekrich Thanks for the suggestion, I've disabled the formatting of these files. it should make them a bit more readable (!sic)  
There is 1 file which is to big to disable formatting using `// format: off` directive, it was added to scalafmt.conf exlucded paths. For all the others a single header ``// format: off`` is enough, there is no need for end directive

PR #2579, opened 2022-02-27
Fix runtime problems on MacOS using M1 chips

### LINKED ISSUE #2494: CVarArgList not working correctly on macOS M1
Multiple tests fail on Scala 3 and Scala 2.13.

Summary:

- Floats and Doubles don't work.
- All tests with greater than 6 args don't work.
- Mixed args don't work.

Here are some samples.
```scala
[error] Test scala.scalanative.unsafe.CVarArgListTest.mixArgs7 failed: java.lang.AssertionError: 1 2 3 4 5 6 -1717986918 0.0 2.2 0.0 3.3 0.0 4.4 0.0 != 1 2 3 4 5 6 7 1.1 2.2 3.3 4.4 5.5 6.6 7.7, took 0.000 sec
[error] Test scala.scalanative.unsafe.CVarArgListTest.doubleArgs3 failed: java.lang.AssertionError: 0.0 0.0 0.0 != 1.1 2.2 3.3, took 0.000 sec
[error] Test scala.scalanative.unsafe.CVarArgListTest.floatArgs9 failed: java.lang.AssertionError: 0.0 0.0 0.0 0.0 0.0 0.0 1.1 0.0 2.2 != 1.1 2.2 3.3 4.4 5.5 6.6 7.7 8.8 9.9, took 0.000 sec
[error] Test scala.scalanative.unsafe.CVarArgListTest.ushortArgs7 failed: java.lang.AssertionError: 1 2 3 4 5 6 0 != 1 2 3 4 5 6 7, took 0.000 sec
[info] Test run finished: 51 failed, 0 ignored, 127 total, 0.083s
```

WojciechMazur:
This PR is aimed to fix #2494  and other runtime issues when using macOS with M1 (arm64) chips. Because currently, we can't have any self-hosted Mac/M1 based machines that could be used in the CI all of the introduced changes are being tested only locally on the M1 host.  

- [x]  Fix CVarArgs - it should use the same logic as proposed by @shadaj in the 32bit branch, but with different padding 
- [x] Fix ProcessMonitor deadlocks - currently used `dispatch_semapthore` was leading to deadlocks, we replace it with named semaphores that would be used in both Mac and Linux. 
- [x] Use `TMPDIR` env to determinate `java.io.tmpdir` system property - on MacOs 12 wiht M1 files created in `/tmp` directory have wrong permissions. We should create them in the path described by `TMPDIR` 
- [x] Adapt TimeTest - TimeZone is not being abbreviated on Arm64 - same problems as in the FreeBSD
- [x] Fix failing java.io.FilesTest - 2 up to 4 tests are frequently failing due to bad descriptor or bad permissions
- [x] Fix failing java.io.SocketTest - Connecting to non-existing socket with timeout is blocking (for 1 min) and return incorrect exception
- [x] Fix failing scalanative.unsafe.ExternTest - incorrect results, probably due to incorrect Scala definition of snprintf method taking varargs (`...)`, where positional args are being passed.

### DIFF
--- a/javalib/src/main/scala/java/nio/file/Files.scala
+++ b/javalib/src/main/scala/java/nio/file/Files.scala
@@ -64,9 +64,10 @@ object Files {
       else throw new UnsupportedOperationException()
 
     val targetFile = target.toFile()
+    val targetExists = targetFile.exists()
 
     val out =
-      if (!targetFile.exists() || (targetFile.isFile() && replaceExisting)) {
+      if (!targetExists || (targetFile.isFile() && replaceExisting)) {
         new FileOutputStream(targetFile, append = false)
       } else if (targetFile.isDirectory() &&
           targetFile.list().isEmpty &&
@@ -81,8 +82,15 @@ object Files {
         throw new FileAlreadyExistsException(targetFile.getAbsolutePath())
       }
 
-    try copy(in, out)
-    finally out.close()
+    try {
+      val copyResult = copy(in, out)
+      // Make sure that created fi
+      if (!targetExists) {
+        targetFile.setReadable(true, ownerOnly = false)
+        targetFile.setWritable(true, ownerOnly = true)
+      }
+      copyResult
+    } finally out.close()
   }
 
   def copy(source: Path, out: OutputStream): Long = {
--- a/javalib/src/main/scala/java/nio/file/Files.scala
+++ b/javalib/src/main/scala/java/nio/file/Files.scala
@@ -84,7 +84,7 @@ object Files {
 
     try {
       val copyResult = copy(in, out)
-      // Make sure that created fi
+      // Make sure that created file has correct permissions
       if (!targetExists) {
         targetFile.setReadable(true, ownerOnly = false)
         targetFile.setWritable(true, ownerOnly = true)
--- a/nativelib/src/main/resources/scala-native/platform/posix/process_monitor.cpp
+++ b/nativelib/src/main/resources/scala-native/platform/posix/process_monitor.cpp
@@ -7,13 +7,17 @@
 #include <sys/wait.h>
 #include <sys/time.h>
 #include <unordered_map>
-#ifdef __APPLE__
-// Semaphores on OSX are deprectated
-#include <dispatch/dispatch.h>
-#else
 #include <semaphore.h>
 #include <errno.h>
 #include <sys/mman.h>
+#include <unistd.h>
+
+// The +1 accounts for the null char at the end of the name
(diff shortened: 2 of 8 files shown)


ekrich:
Ok, I ran this and got the following. Went from 137 test failures to 7.
[M1.txt](https://github.com/scala-native/scala-native/files/8155205/M1.txt)

javalib/src/main/scala/java/nio/file/Files.scala
```diff
@@ -81,8 +82,15 @@ object Files {
         throw new FileAlreadyExistsException(targetFile.getAbsolutePath())
       }
 
-    try copy(in, out)
-    finally out.close()
+    try {
+      val copyResult = copy(in, out)
+      // Make sure that created file has correct permissions
```
WojciechMazur, line 87: On MacOS M1 the files created by copying were given empty permissions, this should match the permission of files created by the JVM

nativelib/src/main/resources/scala-native/platform/posix/process_monitor.cpp
```diff
@@ -7,13 +7,21 @@
...
+#include <stdlib.h>
+#include <fcntl.h>
+
+// The +1 accounts for the null char at the end of the name
+#ifdef __APPLE__
+#include <sys/posix_sem.h>
+#define SEM_MAX_LENGTH PSEMNAMLEN + 1
+#else
+#include <limits.h>
+#define SEM_MAX_LENGTH _POSIX_PATH_MAX + 1
```
WojciechMazur, line 24: This is and creation of semaphore is copy-pasted from Phase.c in Commix impl. In both cases this is needed due to problems on MacOs - usage of sem_create would return -1 and errno would mention not implemented error. 
Also because previous `dispatch_semaphore_t` was leading to deadlocks on Arm64 we do replace both linux and mac implementation with shared named semaphore

nativelib/src/main/scala/scala/scalanative/unsafe/CVarArgList.scala
```diff
@@ -216,4 +218,48 @@ object CVarArgList {
...
+          value.toULong: CVarArg
+        case value: UShort =>
+          value.toULong: CVarArg
+        case value: UInt =>
+          value.toULong: CVarArg
+        case value: Float =>
+          value.toDouble: CVarArg
+        case o => arg
+      }
+    }
```
WojciechMazur, line 242: This implementation is exactly the same as for the x86 (in 32bit support branch), the only difference comes from mapping to Int64 instead of Int32

PR #2584, opened 2022-03-04
Allow to link as dynamic library (take 2)

adampauls:
A rebase and fixup of https://github.com/scala-native/scala-native/pull/2145. I'm not sure what stopped that one from getting through. Perhaps more tests?


WojciechMazur:
Thank you for refreshing that stale PR. I would fix the failures seen in the CI and add some missing parts for Scala 3 over the weekend

WojciechMazur:
Over the weekend I've understood that the previous implementation was incorrect, and in some cases, it would be quite difficult to maintain. Currently, we adapt all the exported functions to be safe to call from outside of SN, however, it might be problematic to call an exported function from Scala Native itself. Instead, we can de something different - we can generate additional definitions of exported methods that would delegate to the regular method after boxing the arguments. I would be a lot easier to maintain and debug. I've already prepared new implementation for Scala 3 and would backport it to Scala 2 soon. 
We can also get rid of new `isExported` attribute, which also have messed up our binary compatibility (due to `Attr` being implemented as a case class)

armanbilge:
@WojciechMazur I made a comment in https://github.com/scala-native/scala-native/pull/2145#issuecomment-1057563801 about exporting functions with C ABI for static linking (instead of dynamic). Will it be possible to support that as well?

adampauls:
@WojciechMazur would you mind putting up your new implementation for Scala 3? We might end up using Scala 3 anyways, so happy to work with whatever you have.

WojciechMazur:
@adampauls I had a problem with pushing changes and ended up erasing your changes in this branch reverting it to the state before rebasing :<. Sorry about that. To simplify stuff a bit I've force pushed the state from before my destructive changes to the original PR including your changes. Let's continue the discussion there and let's close this PR. It now contains improved Scala 3 implementation which should be ready to try. Also, I've added a generation of accessors for fields. 

@armanbilge Yes, adding a static library support should be possible, and actually should be trivial to add on top of that changes.

PR #2587, opened 2022-03-11
add tgmath for real #104

i10416:


WojciechMazur:
Thank you for this contribution as well, we should also define some minimal tests as well, at least to use the methods and make sure they're reachable when linking.

ekrich:
This looks good. I usually try to put the functions in spec order to make it easier to compare. What order did you use for the functions?
https://en.cppreference.com/w/c/numeric/tgmath

i10416:
> This looks good. I usually try to put the functions in spec order to make it easier to compare. What order did you use for the functions?

Oh, I didn't care about that. It sounds neat to put them in spec order. Thanks. I will tweak them.

ekrich:
There is one for floats. https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html#assertEquals(float,%20float,%20float)

i10416:
Thank you for reviewing. 
Btw, how precisely should we test these numeric values? Currently, tests randomly fail due to numerical errors.
In addition, can I leave rounding functions' tests for now as they depend on fenv, which has not been implemented yet.

i10416:
https://github.com/scala-native/scala-native/issues/2590

i10416:
Sorry for messing up 🙇

WojciechMazur (approved):

PR #2588, opened 2022-03-11
add locale.h and limits.h #104

i10416:


WojciechMazur (commented):
Thank you for this contribution, some minimal unit tests would be useful, especially since we're dealing with raw ptr offsets and it might be quite fragile to any changes. 

Also, we can probably can also try to workaround CStruct 21 limit by the usage of inner struct. 
Eg. we can wrap 
```
     frac_digits: CChar,
      p_cs_precedes: CChar,
      n_cs_precedes: CChar,
      p_sep_by_space: CChar,
      n_sep_by_space: CChar,
      p_sign_posn: CChar,
      n_sign_posn: CChar,
```
into `type InternalStruct=CStruct6[CChar,...]` or `CArray[CChar, Nat._6]` 
We should be then able to define CStructN[CString, ..., InternalStruct]. (note that we don't declare it a Ptr[InternalStruct], which would be fine in this case). We can then refer to it using `struct.at(idx)` which would yield Ptr[InternalStruct] which is safe to use

i10416:
> Also, we can probably can also try to workaround CStruct 21 limit by the usage of inner struct.

I think using array for struct is a bit clumsy approach, but it aligns with header definition.

i10416:
Test fails on Windows 🤔

i10416:
Ah, I wrote wrong tests for international monetary numeric formatting parameters :(

i10416:
Removed some fields that are not always guaranteed to be exist.

i10416:
Btw, do you have any idea why test is failing on M1 mac? Unfortunately, it is hard for me to debug M1 because I don't have M1 mac :(

WojciechMazur:
> Btw, do you have any idea why test is failing on M1 mac?

It's actually failing on Arm64 Linux (via qemu emulator), I'll try to investigate this

i10416:
> I'll try to investigate this

Thank you!

unit-tests/native/src/test/scala/scala/scalanative/libc/LocaleTest.scala
```diff
@@ -0,0 +1,91 @@
...
+        !toCString(""),
+        !clocale.positive_sign
+      )
+      assertEquals(
+        !toCString(""),
+        !clocale.negative_sign
+      )
+      // CHAR_MAX
+      assertEquals(
+        127,
```
WojciechMazur, line 69: CHAR_MAX can be either a signed or unsiged value, here's a snipped of limits.h header file.  I believe that's the root of problems on Arm64 in the CI. I suggest adding CHAR_MAX to limits.scala or even better create a new one in clib project (current one is defined in posixlib)

```
/* Minimum and maximum values a `signed char' can hold.  */
#  define SCHAR_MIN     (-128)
#  define SCHAR_MAX     127

/* Maximum value an `unsigned char' can hold.  (Minimum is 0.)  */
#  define UCHAR_MAX     255

/* Minimum and maximum values a `char' can hold.  */
#  ifdef __CHAR_UNSIGNED__
#   define CHAR_MIN     0
#   define CHAR_MAX     UCHAR_MAX
#  else
#   define CHAR_MIN     SCHAR_MIN
#   define CHAR_MAX     SCHAR_MAX
#  endif
```

clib/src/main/scala/scala/scalanative/libc/locale.scala
```diff
@@ -0,0 +1,177 @@
...
+    ptr
+  }
+
+  implicit class LConvOps(val p: Ptr[LConv]) extends AnyVal {
+    def decimal_point: CString = !p.at(0).asInstanceOf[Ptr[CString]]
+    def decimal_point_=(value: CString): Unit =
+      !p.at(0).asInstanceOf[Ptr[CString]] = value
+    def thousands_sep: CString = !p.at(8).asInstanceOf[Ptr[CString]]
+    def thousands_sep_=(value: CString): Unit =
+      !p.at(8).asInstanceOf[Ptr[CString]] = value
```
WojciechMazur, line 102: Even though this is correct on x64, at the moment we would introduce 32bit arch support the offsets would be incorrect. I suggest at least creating a constant value with the size of the pointer (8) and use it to calculate field offest. This way it would be easier to adapt it when adding x86 support.  Scala compiler  or Scala Native optimizer should be able to inline (sizeOfPtr * offset) calls

unit-tests/native/src/test/scala/scala/scalanative/libc/LocaleTest.scala
i10416: Thanks! can I directly add limits.h in this PR as limits.h contains only constants macros? Or should I make another PR?

clib/src/main/scala/scala/scalanative/libc/locale.scala
i10416: replaced 8 with `unsafe.sizeof`

unit-tests/native/src/test/scala/scala/scalanative/libc/LocaleTest.scala
i10416: Added limits.h in this PR.

PR #2594, opened 2022-03-15
add some int-related bindings

i10416:
rel #104

### DIFF
--- a/clib/src/main/resources/scala-native/inttypes.c
+++ b/clib/src/main/resources/scala-native/inttypes.c
@@ -1,0 +1,5 @@
+#include <inttypes.h>
+
+void scalanative_inttypes_imaxdiv(intmax_t q, intmax_t d, imaxdiv_t *result) {
+    *result = imaxdiv(q, d);
+}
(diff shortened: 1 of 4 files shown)


i10416:
running tests over 5h++???🤔

https://github.com/scala-native/scala-native/runs/5562641402?check_suite_focus=true

i10416:
> It might be nice to put sources in the code because some times it is hard to find the reference on which file the functions may be in - if not for the reviewer - https://en.cppreference.com/w/c/string/byte/strtoimax
> 
> Sometimes it is not intuitive either like these in stdlib. https://en.cppreference.com/w/c/numeric/math/div

Yeah, I agree. I will do that later.

ekrich:
I know the Scaladoc documentation is not great. I have used this link: https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html

You could maybe put the following:
```
@see [[https://en.cppreference.com/w/cpp/numeric/math/abs]]
```
There is also this basic doc that I found: https://docs.scala-lang.org/style/scaladoc.html

The Scaladoc does get published but nobody has been writing docs for C code because I guess we figured they could look at the C docs somewhere online. Your approach is of course much more professional but maybe you are raising the minimum standards too 😄 

Listed from the main README.md - https://javadoc.io/doc/org.scala-native/clib_native0.4_2.13/latest/scala/scalanative/libc/index.html

WojciechMazur (approved):

PR #2596, opened 2022-03-19
Port java.util.Optional from Scala.js

ekrich:


ekrich:
@sjrd You said, _"You can call j.u.Objects.equals instead. And inside that one, cast both operands to AnyRef."_

I looked at `Objects.equals` in Scala Native and found one side casted. https://github.com/scala-native/scala-native/blob/main/javalib/src/main/scala/java/util/Objects.scala#L13-L18

Maybe that is good enough or should I change that too? I added an `testEquals` test case for sanity.

In Scala.js I don't see any casting to `AnyRef`. https://github.com/scala-js/scala-js/blob/main/javalib/src/main/scala/java/util/Objects.scala#L19-L24

I also don't see `-Xfatal-warnings` in the Scala.js build.

sjrd:
> I looked at `Objects.equals` in Scala Native and found one side casted. https://github.com/scala-native/scala-native/blob/main/javalib/src/main/scala/java/util/Objects.scala#L13-L18
> 
> Maybe that is good enough or should I change that too?

That's good enough, indeed.

> I also don't see `-Xfatal-warnings` in the Scala.js build.

It's here:
https://github.com/scala-js/scala-js/blob/058532aa8c504b76431b40e3e1b51b2cdef87643/project/Build.scala#L539-L540

> In Scala.js I don't see any casting to `AnyRef`.

Perhaps because we don't cross-compile for Scala 3?

sjrd (approved):

ekrich:
Thank-you for the review. I needed to change the test I added as I didn't actually check `equals` on `Optional`.

I reviewed the Scala.js code but I don't get a warning/error when compiling using Java 11 and Scala 2.13 so I still can't explain the difference compared to Scala Native.

```Scala
sbt:Scala.js> javalib2_13 / compile
[info] compiling 235 Scala sources to /Users/eric/workspace/scala-js/javalib/.2.13/target/scala-2.13/classes ...
[success] Total time: 7 s, completed Mar 22, 2022, 7:16:48 AM
```

Issue #2438, opened 2021-11-24
Working directory in Config cannot be a relative path


jchyb:
When setting `build.Config` to use relative path for a working directory, linking will fail with a non user-friendly message.

For example, if we set in sbt-scala-native plugin:
```scala
val cwd = Paths.get("native") // relative path
Files.createDirectory(cwd)

val logger = streams.value.log.toLogger
build.Config.empty
  .withLogger(logger)
  .withMainClass(maincls)
  .withClassPath(classpath)
  .withWorkdir(cwd) // set here
  .withCompilerConfig(nativeConfig.value)
```

and try to link a hello world program, the following exception will be thrown:
```scala
[error] java.nio.file.NoSuchFileException: native/native/4.ll
[error]         at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
[error]         at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
[error]         at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
[error]         at sun.nio.fs.UnixFileSystemProvider.newFileChannel(UnixFileSystemProvider.java:177)
[error]         at java.nio.channels.FileChannel.open(FileChannel.java:287)
[error]         at java.nio.channels.FileChannel.open(FileChannel.java:335)
[error]         at scala.scalanative.io.VirtualDirectory$NioDirectory.merge(VirtualDirectory.scala:120)
[error]         at scala.scalanative.io.VirtualDirectory$NioDirectory.merge$(VirtualDirectory.scala:115)
[error]         at scala.scalanative.io.VirtualDirectory$LocalDirectory.merge(VirtualDirectory.scala:137)
[error]         at scala.scalanative.codegen.AbstractCodeGen.gen(AbstractCodeGen.scala:43)
[error]         at scala.scalanative.codegen.CodeGen$.$anonfun$emit$4(CodeGen.scala:59)
[error]         at scala.collection.parallel.AugmentedIterableIterator.map2combiner(RemainsIterator.scala:116)
[error]         at scala.collection.parallel.AugmentedIterableIterator.map2combiner$(RemainsIterator.scala:113)
[error]         at scala.collection.parallel.immutable.ParHashMap$ParHashMapIterator.map2combiner(ParHashMap.scala:80)
[error]         at scala.collection.parallel.ParIterableLike$Map.leaf(ParIterableLike.scala:1056)
[...]

PR #2605, opened 2022-04-20
Remove logic from outside build - 0.5.0 only

ekrich:
The idea here is to not have any build or naming logic in the build file. We don't want all the build tools (sbt, Mill, Gradle, Maven, etc.) re-implementing the same logic.

This removes `nativeLink / artifactPath` which is called for every single project, scala version, and for normal and test configuration so that makes the sbt load much slower. Note: add a println of the value in the build and see for yourself.

The idea is also to internalize any caching or other tricks needed to speed up compilation. Even though build tools typically handle recompilation caching some of this is already in the build tools so I think this should be handled within. This is not planned for this PR however.

I would hope that we can finalize this and then attack the Library and Rust PRs after because they can build on this idea.

Note: we have this PR which may need considering. I don't quite understand the specific use in the Portsfile. https://github.com/scala-native/scala-native/pull/2362


ekrich:
Now you can run the second task as well to show the entire build config.
```
show sandbox2_13 / nativeConfig
show sandbox2_13 / nativeBuildConfig
```

WojciechMazur (commented):
I really like this change, but let's schedule it for 0.5.0, it would be way easier for maintainers of the mill, scala-cli or other build tools to handle that.

ekrich:
I have heard rumblings about wanting C and C++ options separate and perhaps we can separate include and link paths from the options. Perhaps we should have the conformance option separate as well like the standard e.g. `-std=c++11`. I also plan to make discovery return the items that will be used for compiling and linking so if you want to augment/inspect/replace them then all these things are possible. This way we can remove settings from the build that can't be changed. In other words, you can pretty much set nothing or have complete control. I am also thinking about/formulating plans to allow each project to present options for that project read from `scala-native.properties`. This can allow the link filtering `-D_MYDEF` setting and other things to make the library developer better. We could also separate to GC and move some code around to the proper project.

I also plan to cache the discovery - it only needs to happen once (reloaded when the build is reloaded). Not sure how that works for different platfroms yet. With the properties file, we could also cache those for each sub-project to avoid extra computation. Also, checking time stamps on the native files could avoid recompilation similar to `make`.

ekrich:
Check if this is supported for workdir - https://github.com/scala-native/scala-native/issues/2438

ekrich:
Thanks for your contribution. We try and shared code with Scala.js when we can but this might be a bit much at this time.
https://github.com/scala-js/scala-js/blob/main/javalib/src/main/scala/java/lang/Character.scala

It would be good if you ported some tests from Scala.js. At least the methods you added. https://github.com/scala-js/scala-js/blob/main/test-suite/shared/src/test/scala/org/scalajs/testsuite/javalib/lang/CharacterTest.scala

ekrich:
@WojciechMazur This is ready for review. I would prefer to do several smaller PRs on the way to better tools rather than a big bang that is hard to review.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -14,8 +14,13 @@ object ScalaNativePlugin extends AutoPlugin {
 
     val nativeVersion = nir.Versions.current
 
+    val nativeBuildConfig =
+      taskKey[build.Config]("Full build configuration, Config")
+
```
WojciechMazur, line 19: `nativeBuildConfig` should probably be hidden from the users not to allow for its modification. MainClass and classPath shall always be up to date with the values coming from `selectMainClass` or other classes. Similarly to classpath which is managed by the sbt.
Workdir contains mostly internal data - *.hnir, *.ll, and *.o files. These should be kept away from the users and shall be managed only the build tool.
The only benefit of changing the workdir is to change the output directory for the result program file. I think that we can add a dedicated setting in the `NativeConfig` for the output directory where the final executable file would be moved upon successful linkage.

To make this task hidden we should move it to the `ScalaNativePluginInternal`

ekrich: Sure, I thought it was nice to be able to view it, but it does open it up for users so we probably don't want that. I think a way to copy the executable somewhere else might be a good feature if we get demand and some good use cases.

ekrich: Shouldn't the `tasks` in the plugin theoretically be `settings`.

ekrich: I think I have addressed the review comments now.

WojciechMazur (approved):

PR #2618, opened 2022-05-30
Fix CRC32 update with empty array

armanbilge:


ekrich:
This looks good but I am having a hard time finding out about CRC for empty Strings, Arrays, etc. Do you have some context on how you found this problem and the proper solution?

unit-tests/shared/src/test/scala/javalib/util/zip/CRC32Test.scala
```diff
@@ -70,4 +70,10 @@ class CRC32Test {
     )
   }
 
+  @Test def updateEmptyArray(): Unit = {
+    val crc = new CRC32()
+    crc.update(Array.emptyByteArray)
+    assertEquals(crc.getValue(), 0)
+  }
```
armanbilge, line 77: @ekrich if you checkout 2523b8d6c846e454fb951bbd8ca109ed6fec99f0 which adds this test, you can run it in the JVM and see it passes while running in Native fails. This is how I discovered the problem (JVM/native implementations were not matching) and this is how I found the "proper" solution (make the implementations match behavior).

ekrich: Thank-you for the explanation.

javalib/src/main/scala/java/util/zip/CRC32.scala
```diff
@@ -25,8 +25,9 @@ class CRC32 extends Checksum {
     update(buf, 0, buf.length)
 
   def update(buf: Array[Byte], off: Int, nbytes: Int): Unit = {
-    // avoid int overflow, check null buf
-    if (off <= buf.length && nbytes >= 0 && off >= 0 && buf.length - off >= nbytes) {
+    if (nbytes == 0) {
+      ()
+    } else if (off <= buf.length && nbytes >= 0 && off >= 0 && buf.length - off >= nbytes) { // avoid int overflow, check null buf
```
ekrich, line 30: Minor nit - if you specifically check for `nbytes == 0` then should the `else if` say `&& nbytes > 0`? Also, is the comment still correct? Seems like this line would wrap with scalafmt especially with the trailing comment.

armanbilge: > Also, is the comment still correct?

Not sure, not entirely sure what it was referring to.

> Seems like this line would wrap with scalafmt especially with the trailing comment.

Scalafmt in CI seems happy with the current formatting.

javalib/src/main/scala/java/util/zip/CRC32.scala
```diff
@@ -27,7 +27,7 @@ class CRC32 extends Checksum {
   def update(buf: Array[Byte], off: Int, nbytes: Int): Unit = {
     if (nbytes == 0) {
       ()
-    } else if (off <= buf.length && nbytes >= 0 && off >= 0 && buf.length - off >= nbytes) { // avoid int overflow, check null buf
+    } else if (off <= buf.length && nbytes > 0 && off >= 0 && buf.length - off >= nbytes) { // avoid int overflow, check null buf
```
ekrich, line 30: If you agree, I think we should just remove the comment - I don't think it is helping to make the code more readable.

armanbilge:
```suggestion
    } else if (off <= buf.length && nbytes > 0 && off >= 0 && buf.length - off >= nbytes) {
```

ekrich (approved):
LGTM. @WojciechMazur can take a final peak before merging.

WojciechMazur (approved):

PR #2646, opened 2022-06-12
Clarify release and LTO modes

armanbilge:


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -49,7 +49,7 @@ object ScalaNativePlugin extends AutoPlugin {
 
     val nativeLTO =
       taskKey[String](
-        "LTO variant used for release mode (either \"none\", \"thin\" or \"full\")."
+        "LTO variant used for release mode, either \"none\", \"thin\", or \"full\" (legacy)."
```
armanbilge, line 52: Maybe we shouldn't even mention `full` if it is legacy.

WojciechMazur (approved):

PR #2674, opened 2022-06-22
Port `ThreadLocalRandom`

armanbilge:
And a couple other things.


javalib/src/main/scala/java/lang/Math.scala
```diff
@@ -209,7 +209,7 @@ object Math {
       copySign(0.0d, a)
     } else if (Double.isInfinite(a) && babs < aabs) {
       copySign(Double.MAX_VALUE, a)
-    } else if (aabs == Double.MAX_VALUE && babs > aabs) {
+    } else if (aabs == Double.MAX_VALUE && babs > aabs && copySign(a, b) == a) {
```
armanbilge, line 212: > If `start` is equal to ± [`Double.MAX_VALUE`](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#MAX_VALUE) and `direction` has a value such that the result should have a larger magnitude, an infinity with same sign as `start` is returned.

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#nextAfter-double-double-

"`direction` has a value such that the result should have a larger magnitude" is the important bit. It's important we check that the signs match.

PR #2676, opened 2022-06-22
Port `ConcurrentHashMap`

armanbilge:


javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -0,0 +1,242 @@
...
+          throw new IllegalStateException(
+            "next must be called at least once before remove"
+          )
+        removeNode(last)
+        lastNode = null
+      }
+    }
+  }
+
+  abstract class CollectionView[K, V, E] extends Collection[E]
```
armanbilge, line 177: I had to add this one to satisfy the linker. Scala.js doesn't need it, and it doesn't appear to be part of the JDK's public API as far as I understand.

PR #2691, opened 2022-06-25
Port Piped{Input,Output}Stream from Apache Harmony

### LINKED ISSUE #2631: Missing implementation for piped streams
Scala Native Version: 0.4.4

The current javalib does not contain the `PipedInputStream` and `PipedOutputStream` classes which are useful for testing (e.g passing data to an `InputStream` to be read by an io function to test).

WojciechMazur:
Fixes #2631 
Direct port of Apache Harmony implementation of: 
* PipedInputStream
* PipedOutputStream
* PipedReader
* PipedWriter

Since Piped classes are dedicated for usage from multiple threads and might lead to deadlocks when used in single thread (as pointed in the Java API) it should be merged after implementing multithreading support in Scala Native 

TODO: 
- [x] Port remaining tests from Apache Harmony (converted from Java to Scala and commented out)

### DIFF
--- a/javalib/src/main/scala/java/io/OutputStream.scala
+++ b/javalib/src/main/scala/java/io/OutputStream.scala
@@ -8,7 +8,7 @@ abstract class OutputStream extends Object with Closeable with Flushable {
     write(b, 0, b.length)
 
   def write(b: Array[Byte], off: Int, len: Int): Unit = {
-    if (off < 0 || len < 0 || len > b.length - off)
+    if (off > b.length || off < 0 || len < 0 || len > b.length - off)
       throw new IndexOutOfBoundsException()
 
     var n = off
--- a/javalib/src/main/scala/java/io/PipedInputStream.scala
+++ b/javalib/src/main/scala/java/io/PipedInputStream.scala
@@ -1,0 +1,284 @@
+// Ported from Apache Harmony
+package java.io
+
+object PipedInputStream {
+  final val PipeSize = 1024
+}
+
+/** Constructs a new unconnected {@code PipedInputStream}. The resulting stream
+ *  must be connected to a {@link PipedOutputStream} before data may be read
+ *  from it.
+ */
+class PipedInputStream(private[io] val pipeSize: Int) extends InputStream {
+  private var lastReader: Thread = _
+  private var lastWriter: Thread = _
+  private var isClosed = false
+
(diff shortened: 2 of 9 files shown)


tanishiking (commented):
I guess porting from Apache Harmony applies the "redistribution" https://www.apache.org/licenses/LICENSE-2.0 (even though we have a modification from Java to Scala)

In that case,

> a. You must give any other recipients of the Work or Derivative Works a copy of this License; and

We already have one https://github.com/scala-native/scala-native/blob/main/LICENSE.md#license-notice-for-apache-harmony :+1:

---

> b. You must cause any modified files to carry prominent notices stating that You changed the files; and

I guess we have to keep these part in the source code https://github.com/apache/harmony/blob/02970cb7227a335edd2c8457ebdde0195a735733/classlib/modules/luni/src/main/java/java/io/PipedInputStream.java#L1-L16 am I correct?

---

> d. If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; ...

Looks like we need to copy the [NOTICE file](https://github.com/apache/harmony/blob/trunk/NOTICE) into our repository `NOTICE` file?

WojciechMazur:
> I guess we have to keep these part in the source code [https://github.com/apache/harmony/blob/02970cb7227a335edd2c8](https://github.com/apache/harmony/blob/02970cb7227a335edd2c8457ebdde0195a735733/classlib/modules/luni/src/main/java/java/io/PipedInputStream.java#L1-L16) 

Probably, we point in the License notices that we either include information that given code is ported from Apache Harmony or we include the whole original header. Typically the short notice is good enough, I think Scala.js does the same. 

> Looks like we need to copy the [NOTICE file](https://github.com/apache/harmony/blob/trunk/NOTICE) into our repository NOTICE file? 

Good point, we should add such file. Probably we can do that in similar format as Scala.js does: https://github.com/scala-js/scala-js/blob/037fa855a636a521721f6b126f915744136cd47e/NOTICE

ekrich:
I think we remove all the javadoc/scaladoc from the code as we publish that for reference but we want the Oracle javadoc to be the reference.

PR #2723, opened 2022-07-13
Port asynchronous test adapter code from Scala.js

lolgab:
The current test infrastructure doesn't work with asynchronous code.
Tests frameworks work around this problem by forcing the execution of the `scala.scalanative.runtime.loop()` before awaiting the futures. This is not needed in Scala.js since all the test infrastructure code is asynchronous.
This PR ports that code to Scala Native as well.
I tested with utest and you can now have asynchronous tests that behave exactly as with Scala.js.
For example, returning a future that doesn't complete hangs as with Scala.js instead of failing instantaneously.
```scala
import utest._
import scala.concurrent._
object MyTest extends TestSuite {
  val tests = Tests {
    test("doesn't complete") {
      val p = Promise[Int]()
      p.future
    }
  }
}
```

### DIFF
--- a/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
+++ b/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
@@ -12,7 +12,7 @@ package object async {
     if (isMultithreadingEnabled)
       Await.ready(future, Duration.Inf)
     else
-      scala.scalanative.runtime.loop()
+      while (!future.isCompleted) scala.scalanative.runtime.loop()
     future.map(_ => Success(()))
   }
 }
--- a/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
+++ b/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
@@ -1,17 +1,18 @@
 package scala.scalanative.junit
 
+import scala.util.{Try, Success}
 import scala.concurrent.{Await, Future}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.Duration
 import scala.scalanative.meta.LinktimeInfo.isMultithreadingEnabled
 
 package object async {
-  type AsyncResult = Unit
+  type AsyncResult = Future[Try[Unit]]
   def await(future: Future[_]): AsyncResult = {
     if (isMultithreadingEnabled)
-      Await.result(future, Duration.Inf)
-    else {
-      future.map(_ => Success(()))
-    }
+      Await.ready(future, Duration.Inf)
+    else
+      scala.scalanative.runtime.loop()
+    future.map(_ => Success(()))
   }
 }
--- a/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
+++ b/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
@@ -1,6 +1,7 @@
 package scala.scalanative.junit
 
 import scala.concurrent.{Await, Future}
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.Duration
 import scala.scalanative.meta.LinktimeInfo.isMultithreadingEnabled
 
@@ -10,8 +11,7 @@ package object async {
     if (isMultithreadingEnabled)
       Await.result(future, Duration.Inf)
     else {
-      scala.scalanative.runtime.loop()
-      future.value.get.get
+      future.map(_ => Success(()))
     }
   }
 }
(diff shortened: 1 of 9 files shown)


ekrich:
Very cool. If you could put the Date and SHA like this that would be super - <https://github.com/scala-native/scala-native/blob/main/javalib/src/main/scala/java/util/AbstractMap.scala#L1-L11>

lolgab:
@ekrich This wasn't a complete copy-paste but I adapted the code to work with the existing code. Does the Scala.js port commit still apply?
By the way, I'm making this PR ready for review even if some comments need to be updated, because I don't know exactly what to write.

armanbilge:
@lolgab this is awesome!! So IIUC, we should be able to remove the `loop()` from the JUnit integration?

https://github.com/scala-native/scala-native/blob/5924b310ae516baf7ca7326d0f2329e4726799be/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala#L1-L11

junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala
```diff
@@ -1,17 +1,18 @@
...
 package object async {
-  type AsyncResult = Unit
+  type AsyncResult = Future[Try[Unit]]
   def await(future: Future[_]): AsyncResult = {
     if (isMultithreadingEnabled)
-      Await.result(future, Duration.Inf)
-    else {
+      Await.ready(future, Duration.Inf)
+    else
       scala.scalanative.runtime.loop()
```
armanbilge, line 15: If the `loop()` is still necessary, I'm not sure how this PR improves the status quo?

WojciechMazur: Tl;dr It's necessary otherwise testing infrastructure would be broken. 

if we take a look at Scala.js ExecutionContext all the asynchronous operations are scheduled by the JS and their tasks queue. We don't have a native tasks scheduler like Scala.js does so there is no way to run it asynchronously. To mitigate this problem we use `loop()` which basically works like an aggressive version of `Thread.yield()`. Without that, we would just loop infinitely in await with lots of not-completed jobs I think that's what lolgab observed, however, if the future would be completed at some point it would also hang. 

The only improvement comes from parallelization of tests: as you can see I needed to modify 1 of the tests which modified a global shared state. Due to the changes made by this PR different test suites accessing this shared state were actually run in parallel, though it was necessary to synchronize the execution order.

armanbilge: > Tl;dr It's necessary otherwise testing infrastructure would be broken.

Yes, thanks, I realized this a while ago as well :)

---

> The only improvement comes from parallelization of tests

Do you mean parallelization of individual tests, or parallelization of test suites?

I am concerned about this breaking change:
https://github.com/scala-native/scala-native/blob/eddb084af7e559f16854d749fa2c092b9af6d501/test-interface-sbt-defs/src/main/scala/sbt/testing/Task.scala#L35-L48

In theory it allows for async tests, but in practice the test framework still has to call `loop()` anyway. So it seems to me that this is completely pointless.

WojciechMazur: So by default, we had parallelization of test suites already provided by sbt by spawning multiple runners, unless disabled using `set every parallelExecution := false`. On top of that, we now get another parallelization of test suites within the runner instance, and probably also of individual tests (have not checked that but the current algorithm should allow for that). 
Now as I'm thinking about it I'm not sure if it's compliant with how tests are executed on the JVM. I'm not sure but probably when running tests on the JVM the additional synchronization was not required in the mentioned test, but I'm not sure

Yes, it has to run loop anyway, but when running with enabled multithreading it is skipped and we use `Await.ready` similarly to the JVM.

armanbilge: Thanks!

> On top of that, we now get another parallelization of test suites within the runner instance

Is this really useful for single-threaded runtime? I understand the value for multi-threading of course, but as you say when multi-threading we can use `Await.ready` anyway.

Sorry for being a broken record, but I don't see why this breaking change is necessary. To be clear, I'm referring specifically to the change to the `Task` interface.

- It's not needed for multi-threading, because we know the existing interface can work exactly like the JVM. 
- For single-threading, we know it's not necessary, because we've been doing async tests without it anyway. It's needed on Scala.js because as you point out, the execution queue is controlled by the JS runtime, but because in Native we have to call `loop()` anyway all async tests become effectively synchronous, from the perspective of the test interface.

WojciechMazur:
As decided in the review, the PR does not bring any reasonable improvements, since it still require to manually run scalanativ e loop. Closing.

PR #2739, opened 2022-07-19
Add support for Scala 3.2.0

WojciechMazur:
* Fix AdaptLazyVals in nscplugin to handle usages of `getOffsetStatic` instead of `getOffset`
* Address compiler warnings for narrowing and refutable binding/patterns 
* Disable fatal-warnings to allow usage of non-local returns 
* Update scalalib patches
* Add runtime tests to the CI in the limited scope


ekrich:
Is the upcoming multi-threading going to need some differences for lazy vals? C atomics are added in C11.

PR title spelling should be `support`.

WojciechMazur:
@ekrich Yes, it would, the implementation of [LazyVals](https://github.com/lampepfl/dotty/blob/main/library/src/scala/runtime/LazyVals.scala) uses eg. a number of monitors to control safe initialization in multiple threads, however in Scala Native this part is completely removed. We would need to restore and this part in the future. There are still some more advanced pending changes for Lazy Vals which would be probably shipped in 3.3.x, but for now we can ignore this fact

PR #2744, opened 2022-07-22
Fix #2730: Package private at the top level of build breaks easy import

### LINKED ISSUE #2730: Package private at the top level of build breaks easy import
In the documents, we advertise the following:
```scala

nativeConfig ~= {
  _.withLTO(LTO.thin)
    .withMode(Mode.releaseFast)
    .withGC(GC.commix)
}
```

Now we get the following due to changes:
```
[error] /Users/eric/workspace/scala-native/project/Build.scala:547:13: object IO in package build cannot be accessed in package scala.scalanative.build
[error]             IO.createDirectory(trgDir)
```

One import is highly useful so perhaps we should move these to an `internal` package.

ekrich:
Move package private objects in `core` package.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -3,6 +3,9 @@ package build
 
 import java.nio.file.{Path, Paths}
 import scala.scalanative.util.Scope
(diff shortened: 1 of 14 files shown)


ekrich:
Test failures seem unrelated.

Issue #2733, opened 2022-07-17
`nir_2.12-0.4.5` contains `object scala.scalanative.nir.Attr#Volatile` ?


armanbilge:
```
$ wget https://repo1.maven.org/maven2/org/scala-native/nir_2.12/0.4.5/nir_2.12-0.4.5.jar
$ less nir_2.12-0.4.5.jar | head -n 30
Archive:  nir_2.12-0.4.5.jar
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
     313  Defl:N      134  57% 2010-01-01 00:00 7a5fd22e  META-INF/MANIFEST.MF
       0  Stored        0   0% 2010-01-01 00:00 00000000  scala/
       0  Stored        0   0% 2010-01-01 00:00 00000000  scala/scalanative/
       0  Stored        0   0% 2010-01-01 00:00 00000000  scala/scalanative/nir/
       0  Stored        0   0% 2010-01-01 00:00 00000000  scala/scalanative/nir/serialization/
     417  Defl:N      290  31% 2010-01-01 00:00 f7899198  scala/scalanative/nir/Attr$.class
    1820  Defl:N      864  53% 2010-01-01 00:00 dfc78456  scala/scalanative/nir/Attr$Abstract$.class
    1892  Defl:N      886  53% 2010-01-01 00:00 59db779c  scala/scalanative/nir/Attr$AlwaysInline$.class
    1788  Defl:N      793  56% 2010-01-01 00:00 726d452d  scala/scalanative/nir/Attr$BailOpt$.class
    2527  Defl:N     1129  55% 2010-01-01 00:00 b43f92d8  scala/scalanative/nir/Attr$BailOpt.class
    1862  Defl:N      882  53% 2010-01-01 00:00 133414b7  scala/scalanative/nir/Attr$DidOpt$.class
    1800  Defl:N      858  52% 2010-01-01 00:00 aec48f49  scala/scalanative/nir/Attr$Dyn$.class
    1812  Defl:N      867  52% 2010-01-01 00:00 6c34053d  scala/scalanative/nir/Attr$Extern$.class
     394  Defl:N      249  37% 2010-01-01 00:00 64b578ea  scala/scalanative/nir/Attr$Inline.class
    1884  Defl:N      885  53% 2010-01-01 00:00 8468c10b  scala/scalanative/nir/Attr$InlineHint$.class
[...]

PR #2745, opened 2022-07-23
Drop empty elements in `Paths.get`

armanbilge:
> If `more` specifies one or more elements then each non-empty string, including `first`, is considered to be a sequence of name elements (see [`Path`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html)) and is joined to form a path string.

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Paths.html#get-java.lang.String-java.lang.String...-

Targeted to 0.4.x.

### DIFF
--- a/javalib/src/main/scala/scala/scalanative/nio/fs/unix/UnixFileSystem.scala
+++ b/javalib/src/main/scala/scala/scalanative/nio/fs/unix/UnixFileSystem.scala
@@ -2,6 +2,7 @@ package scala.scalanative.nio.fs.unix
 
 import java.io.IOException
 import java.lang.Iterable
+import java.lang.StringBuilder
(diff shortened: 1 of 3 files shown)


sideeffffect:
Fails on weird bin check issues :confused: 
https://github.com/scala-native/scala-native/runs/7478968848?check_suite_focus=true#step%3A4%3A101=
https://github.com/scala-native/scala-native/runs/7478970859?check_suite_focus=true#step%3A4%3A837=
https://github.com/scala-native/scala-native/runs/7478970882?check_suite_focus=true#step%3A4%3A837=

armanbilge:
@sideeffffect see https://github.com/scala-native/scala-native/issues/2733.

PR #2747, opened 2022-07-23
Add `j.u.s.Stream#forEach`

armanbilge:
Targeting 0.4.x.

### DIFF
--- a/unit-tests/shared/src/test/scala/javalib/util/stream/StreamTest.scala
+++ b/unit-tests/shared/src/test/scala/javalib/util/stream/StreamTest.scala
@@ -1,7 +1,7 @@
 package javalib.util.stream
 
 import java.util.stream._
-import java.util.function.Function
+import java.util.function.{Consumer, Function}
 
 import org.junit.Test
 import org.junit.Assert._
@@ -49,7 +49,10 @@ class StreamTest {
   @Test def streamForeachWorks(): Unit = {
     val s = Stream.of(1, 2, 3)
     var sum = 0
-    s.forEach(sum += _)
+    val consumer = new Consumer[Int] {
+      def accept(i: Int): Unit = sum += i
+    }
+    s.forEach(consumer)
     assertEquals(6, sum)
   }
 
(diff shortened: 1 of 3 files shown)


unit-tests/shared/src/test/scala/javalib/util/stream/StreamTest.scala
```diff
@@ -46,6 +46,13 @@ class StreamTest {
     assertFalse(it.hasNext())
   }
 
+  @Test def streamForeachWorks(): Unit = {
+    val s = Stream.of(1, 2, 3)
+    var sum = 0
+    s.forEach(sum += _)
```
sideeffffect, line 52: This line seems to cause issues for Scala 2.11
```scala
[error] D:\a\scala-native\scala-native\unit-tests\shared\src\test\scala\javalib\util\stream\StreamTest.scala:52:22: missing parameter type for expanded function ((x$1) => sum.<$plus$eq: error>(x$1))
[error]     s.forEach(sum += _)
[error]                      ^
[error] one error found
[error] (testsJVM2_11 / Test / compileIncremental) Compilation failed
```
https://github.com/scala-native/scala-native/runs/7479078735?check_suite_focus=true#step%3A5%3A69=

sideeffffect (changes requested):

unit-tests/shared/src/test/scala/javalib/util/stream/StreamTest.scala
armanbilge: Thanks :)

ekrich:
Is there anyway this can be done without delegating to Scala stdlib? We have been trying to decouple like Scala.js. https://github.com/scala-js/scala-js/pull/4677 ScalaOps can help your sanity in the process.

armanbilge:
That's a great point. Unfortunately not in this PR, since it seems that the Java Stream is implemented in terms of the Scala one.

ekrich:
If you are still working on this could you add a comment at the top as a reminder?

armanbilge:
Seems more appropriate as an issue? Since it applies to the entire project, no? :)

ekrich:
Yes, I think that is more appropriate. I searched and could not find one. We can keep track of items as we find them.

PR #2748, opened 2022-07-23
Port `InflaterOutputStream` from Android Luni

armanbilge:
Targeting 0.4.x.

https://android.googlesource.com/platform/libcore/+/2e317a02b5a8f9b319488ab9311521e8b4f87a0a/luni/src/main/java/java/util/zip/InflaterOutputStream.java

### DIFF
--- a/javalib/src/main/scala/java/util/zip/InflaterOutputStream.scala
+++ b/javalib/src/main/scala/java/util/zip/InflaterOutputStream.scala
@@ -8,22 +8,15 @@ import java.io.OutputStream
 import java.util.Arrays
 
 class InflaterOutputStream private (
+    out: OutputStream,
     protected val inf: Inflater,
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/util/zip/InflaterOutputStream.scala
```diff
@@ -0,0 +1,103 @@
+// ported from android luni 2e317a02b5a8f9b319488ab9311521e8b4f87a0a
```
WojciechMazur, line 1: I believe we should add a license info mentioning the usage of Android Luni source code into `License.md` file (similarly to usage RE2 and others, at the end of the file)

armanbilge: @WojciechMazur good point, done.

PR #2759, opened 2022-07-28
Implement `UUID.randomUUID()` using `java.security.SecureRandom`

armanbilge:
Fixes https://github.com/scala-native/scala-native/issues/2600 by porting the changes in https://github.com/scala-js/scala-js/pull/4659. Targeting 0.4.x.

### DIFF
--- a/javalib-ext-dummies/src/main/scala/java/security/SecureRandom.scala
+++ b/javalib-ext-dummies/src/main/scala/java/security/SecureRandom.scala
@@ -1,0 +1,7 @@
+package java.security
+
+/** Fake implementation of `SecureRandom` that is not actually secure at all.
+ *
+ *  It directly delegates to `java.util.Random`.
+ */
+class SecureRandom extends java.util.Random
--- a/javalib/src/main/scala/java/util/UUID.scala
+++ b/javalib/src/main/scala/java/util/UUID.scala
@@ -1,6 +1,7 @@
 package java.util
 
 import java.lang.{Long => JLong}
+import java.security.SecureRandom
 
 final class UUID private (
     private val i1: Int,
@@ -134,14 +135,35 @@ object UUID {
   private final val NameBased = 3
   private final val Random = 4
 
-  private lazy val rng = new Random() // TODO Use java.security.SecureRandom
+  private lazy val rng = new SecureRandom()
 
   def randomUUID(): UUID = {
-    val i1 = rng.nextInt()
-    val i2 = (rng.nextInt() & ~0x0000f000) | 0x00004000
-    val i3 = (rng.nextInt() & ~0xc0000000) | 0x80000000
-    val i4 = rng.nextInt()
-    new UUID(i1, i2, i3, i4, null, null)
+    // ported from Apache Harmony
+
+    val data = new Array[Byte](16)
+    rng.nextBytes(data)
+
+    var msb = (data(0) & 0xffL) << 56
+    msb |= (data(1) & 0xffL) << 48
+    msb |= (data(2) & 0xffL) << 40
+    msb |= (data(3) & 0xffL) << 32
+    msb |= (data(4) & 0xffL) << 24
+    msb |= (data(5) & 0xffL) << 16
+    msb |= (data(6) & 0x0fL) << 8
+    msb |= (0x4L << 12) // set the version to 4
+    msb |= (data(7) & 0xffL)
+
+    var lsb = (data(8) & 0x3fL) << 56
+    lsb |= (0x2L << 62) // set the variant to bits 01
+    lsb |= (data(9) & 0xffL) << 48
+    lsb |= (data(10) & 0xffL) << 40
+    lsb |= (data(11) & 0xffL) << 32
+    lsb |= (data(12) & 0xffL) << 24
+    lsb |= (data(13) & 0xffL) << 16
+    lsb |= (data(14) & 0xffL) << 8
+    lsb |= (data(15) & 0xffL)
+
+    new UUID(msb, lsb)
   }
 
   // Not implemented (requires messing with MD5 or SHA-1):
--- a/unit-tests-ext/shared/src/test/scala/javalib/util/UUIDTestEx.scala
+++ b/unit-tests-ext/shared/src/test/scala/javalib/util/UUIDTestEx.scala
@@ -1,0 +1,27 @@
+// Ported from Scala.js, commit: dff0db4, dated: 2022-04-01
+
+package javalib.util
+
+import java.util.UUID
+
+import org.junit.Test
+import org.junit.Assert._
+
+/** Additional tests for `java.util.UUID` that require
+ *  `java.security.SecureRandom`.
+ */
+class UUIDTestEx {
+
+  @Test def randomUUID(): Unit = {
+    val uuid1 = UUID.randomUUID()
+    assertEquals(2, uuid1.variant())
+    assertEquals(4, uuid1.version())
+
+    val uuid2 = UUID.randomUUID()
+    assertEquals(2, uuid2.variant())
+    assertEquals(4, uuid2.version())
+
+    assertNotEquals(uuid1, uuid2)
+  }
+
+}


LeeTibbert:
@arman.  Thank you for the PR.  Is this your most up-to-date thinking?
We had some discussion in the Issue earlier today. I do not want to 
re-hash that.

If not current, then sorry to bother you.

If yes, this is HEAD, to my thinking  introducing a `secureRandom()` which  document itself as
insecure is creating a devil worse than the one it is trying to drive out and,
by my reading, leaves the well placed concern of your original Issue in place.
Am I missing something?

javalib-ext-dummies/src/main/scala/java/security/SecureRandom.scala
```diff
@@ -0,0 +1,7 @@
+package java.security
+
+/** Fake implementation of `SecureRandom` that is not actually secure at all.
+ *
+ *  It directly delegates to `java.util.Random`.
+ */
+class SecureRandom extends java.util.Random
```
armanbilge, line 7: @LeeTibbert I assume you are referring to this? This is not published. It is an internal dummy used for testing only.

javalib/src/main/scala/java/util/UUID.scala
```diff
@@ -134,14 +135,35 @@ object UUID {
...
-  private lazy val rng = new Random() // TODO Use java.security.SecureRandom
+  private lazy val rng = new SecureRandom()
 
   def randomUUID(): UUID = {
-    val i1 = rng.nextInt()
-    val i2 = (rng.nextInt() & ~0x0000f000) | 0x00004000
-    val i3 = (rng.nextInt() & ~0xc0000000) | 0x80000000
-    val i4 = rng.nextInt()
-    new UUID(i1, i2, i3, i4, null, null)
+    // ported from Apache Harmony
```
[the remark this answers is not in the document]
armanbilge: There is no 32 bit on the 0.4.x branch. If simple code like this has problems in 32-bit then that's a huge can of worms :)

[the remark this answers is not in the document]
ekrich: Well it is just an `int` so that should be fine - just a naive question I suppose.

javalib-ext-dummies/src/main/scala/java/security/SecureRandom.scala
LeeTibbert: Yes, exactly.  I recommend posting 4 Marine's on duty and having a platoon as relief.

What I would do if I were telling the story, and I am not, is not even go near `secureRandom()` unless
it is indeed believably & defensibly secure, not even in testing or dummy code.  
You are handling smallpox virus outside a containment facility.

That said, I suggest  doing at least three things:

1)   Most importantly, I would have the file say, in better works, "For testing only"

2)   I would move the code to the unit-test itself and make it private to that class to make it even
      more evident  that it is for testing purposes and harder to take the blade guards off.

      The name `javalib-ext-dummies` requires Institutional Knowledge that even I do not have,
      not that I am all knowing and I have missed a lot in the past year. If were using 'find` to find
      a `SecureRandom()` implementation for a project and did not know the security implications
      you imply but do not describe for `java.util.Random` I would grab the `javalib-ext-dummies` file
      in a flash and complain loudly when it was not Secure.  I am not trying to be harsh here or
      to reveal my tendency to leap before I look.  

3) I would probably have `SecureRandom` directly call 
`MASSIVELY_INSECURE_RANDOM_USE_AT_YOUR_OWN_RISK_AND_DO_NOT_COMPLAIN_WHEN_NOT_IF" + "_YOU_HURT_YOURSELF()"

armanbilge: I can definitely improve the comment. I don't think it should be moved to the unit test, that would not follow the existing pattern of `javalib-ext-dummies` (which of course can be challenged in a separate issue).

Note that this same pattern is used in https://github.com/scala-js/scala-js/pull/4659.

LeeTibbert: I appreciate your considering my concerns and addressing them as is consistent with your
intent.

It is always an open issue as to when to follow precedent and when to take another path.  
The Scala.js folks have already said that they would not provide a  known insecure variant if they
had to do things a second time.  You have heard and considered my thoughts about `javalib-ext-dummies`
so I will not belabor them.

armanbilge: > The Scala.js folks have already said that they would not provide a known insecure variant if they
had to do things a second time.

To clarify, they were talking about _publishing_ a known insecure variant.
https://github.com/scala-js/scala-js-fake-insecure-java-securerandom/

That is different from the internal dummy used for testing.

LeeTibbert: re: `scala-js-fake-insecure-java-securerandom`  I like the name!   

Thank you for the chuckle, I could use one or a few.

sjrd: > > The Scala.js folks have already said that they would not provide a known insecure variant if they
> > had to do things a second time.
> 
> To clarify, they were talking about _publishing_ a known insecure variant. https://github.com/scala-js/scala-js-fake-insecure-java-securerandom/
> 
> That is different from the internal dummy used for testing.

Indeed. And to be even clearer, we would definitely do redo the same thing for testing. This is the best pattern to test "partial" implementations (`randomUUID`), which require some externally provided supplementary class (`SecureRandom`) to actually link.

LeeTibbert:
I have been thinking about this issue whilst I have been offline for the heat of the day.

This PR is far better than the situation it is correcting.  I believe it also provides the basis
for a dialectic.

I believe there is a Scala Native solution which would remove the requirement
for a user/developer to provide their own SecureRandom implementation.

Arman, in the spirit of process improvement, not antagonism, with your go ahead
I can submit a WIP PR for consideration to make ideas easier to review and
compare,  If I do such a submission, may I use/submit the Test from this PR, giving 
credit to this collaboration?

The implementation I have in mind would fulfill the description of `randomUUID` which
has had us all concerned.  I always hate to say "It should not be that bad/hard' before
I have at least sandbox code.  (Avoiding reading `/dev/random` or kin is a _big_ simplification).
This kind of code is where SN shines.

The reason for spending more time on this issue is that the
code in question is in `java.util`. I know of no plans for that
package to go away in 0.5.0. Whatever we accept as
a solution is likely to be around for a long time. A fix which
avoids imposing a 'roll your own' requirement is less work
than documenting this requirement (which, BTW, I think, this
PR is missing).

Linux has had the `getentropy()` call for 4 or more years.  The BSDs
have had it since before dirt was created.  Windows has a plethora
of similar calls (I believe BmumbleCryptmumble is the current
recommendation).

State actors, financial institutions, a few Computer Science
grad students, and my niece in second grade will all be
using a hardware RNG.  The routines I mentioned above
are documented, or call routines documented as being
cryptography secure. On unix, they are closely related
to `/dev/random` & `dev/urandom', but easier to access.

By my understanding of the UUID algorithm, it reduces
some randomness.

Note that I am _not_ advocating routines to get the
UUID itself.  On unix `libuuid` exists on some but 
not all systems.  I think Windows has a built in
UUID generator but would have to check.
In short, I think there is a tangle here which
we can easily avoid.

Note also that I am not suggesting implementing `SecureRandom`,
although such would be possible with work & the routines I mentioned.

armanbilge:
> Arman, in the spirit of process improvement, not antagonism, with your go ahead
I can submit a WIP PR for consideration to make ideas easier to review and
compare, If I do such a submission, may I use/submit the Test from this PR, giving
credit to this collaboration?

Yes, absolutely, I cribbed the test from Scala.js anyway 😜 looking forward to seeing your idea.

ekrich:
>I believe there is a Scala Native solution which would remove the requirement
for a user/developer to provide their own SecureRandom implementation.

@LeeTibbert I don't think the intent is to really roll your own. I think the intent is to have it provided outside of the core of Scala Native. If you want to work on something it won't be wasted work and can be included somewhere that can be included if needed by the users. Even a micro library is a possibility or a full `java.security` package.

For example, `java.util.Date` has an added `toInstant` method and yet we don't provide a `java.time` API. The problem is that Java doesn't have `implicit`s to allow a `toInstant` method on `java.util.Date` that could be provided by a `java.time` API. This is one way Scala really shines by allowing new functionality to be added and not couple the package dependencies together.

javalib/src/main/scala/java/util/UUID.scala
```diff
@@ -134,14 +135,35 @@ object UUID {
   private final val NameBased = 3
   private final val Random = 4
 
-  private lazy val rng = new Random() // TODO Use java.security.SecureRandom
+  private lazy val rng = new SecureRandom()
 
   def randomUUID(): UUID = {
```
ekrich, line 140: Even though we do not normally document out Java library implementation, they do get published. I recommend adding a really good java/scaladoc outlining that this uses `java.security.SecureRandom` as a source of random numbers and that a third party library is required to use this method.

WojciechMazur (approved):

PR #2771, opened 2022-08-04
upgrading 'setup-ammonite' github action

yilinwei:
Upgrades the action to fix issue mentioned [here](https://github.com/yilinwei/setup-ammonite/issues/3).


yilinwei:
I have also released `0.1.2` which supports `win32` and `macos` VM's as well.

WojciechMazur (approved):

PR #2772, opened 2022-08-04
add onType function on CollectLocalValDeps and Lower.Impl

yuly16:
`CollectLocalValDeps` and `Lower.Impl` extend `transform`. However, the function `onType` in transform does nothing. Therefore we can add `override def onType(ty: Type): Type = ty` in the two classes to avoid unnecessary calculations.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/interflow/UseDef.scala
+++ b/tools/src/main/scala/scala/scalanative/interflow/UseDef.scala
@@ -51,7 +51,7 @@ object UseDef {
     }
 
     override def onType(ty: Type): Type = ty
-    
+
   }
 
   private def collect(inst: Inst): Seq[Local] = {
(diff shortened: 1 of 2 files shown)


tools/src/main/scala/scala/scalanative/interflow/UseDef.scala
```diff
@@ -49,6 +49,9 @@ object UseDef {
       }
       super.onNext(next)
     }
+
+    override def onType(ty: Type): Type = ty
```
WojciechMazur, line 53: Can you please add a small comment why are we're overriding `onType` here. It should be easier to follow up the code changes in the future.

WojciechMazur (changes requested):
Thanks for your contributions, it looks like some of the files are not formatted correctly, can you please run script `./scripts/scalafmt` and publish the changes again?

tools/src/main/scala/scala/scalanative/interflow/UseDef.scala
yuly16: No problem

PR #2777, opened 2022-08-06
Incremental Compilation

yuly16:
Add the incremental compilation support for scala native. 

We set `package` as the granularity of the incremental compilation. In other words, in the edit-compile-run loop, we only generate the llvm code of the changed package. The llvm code of unchanged parts(e.g. most of the libraries and unchanged packages) won't be generated again.

We use the [autotest for scala-native incremental compilation](https://github.com/yuly16/incremental-compilation-autotest) to test both the correctness and performance of incremental compilation. The benchmark refers to the [scala native benchmark](https://github.com/scala-native/scala-native-benchmarks). The time cost of `nativeLink` is presented below. All binaries are tested to be correct. 

1. Compiling in the first compilation(sbt clean + sbt nativeLink), the time cost is:
image

2. Compiling in later compilations(change a line in scala source code + sbt nativeLink), the time cost is:
image

Even if the incremental compilation adds some overhead in the first compilation, it brings significant performance improvement in later compilations. The overhead is caused by generating too many `*.ll` files. In the original version, the number of `*.ll` files equals to the number of cpu cores; when incremental compilation is applied, the number of `*.ll` files equals to the number of packages.


ekrich:
It looks like there are some Java 11 methods being used like `Path.of(String)` that are not implemented. Also, maybe it would be a good idea to add a new option to `NativeConfig` for incremental with a default of true. What do you think @WojciechMazur ?

ekrich:
It looks like there are some Java 11 methods being used like `Path.of(String)` that are not implemented. Also, maybe it would be a good idea to add a new option to `NativeConfig` for `incremental` with a default of `true`. What do you think @WojciechMazur ?

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -70,12 +72,21 @@ object Build {
         linked
       }
 
+      implicit var incCompilationContext: IncCompilationContext = null
```
WojciechMazur, line 75: Having an implicit null is never a good idea. Consider creating empty IncCompilationContext always, but run side-effecting methods like `collectFromPrev` explicitly on it's instance instead of making it a part of the constructor

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -31,44 +32,56 @@ private[scalanative] object LLVM {
...
       val inpath = path.abs
       val outpath = inpath + oExt
       val isCpp = inpath.endsWith(cppExt)
       val isLl = inpath.endsWith(llExt)
       val objPath = Paths.get(outpath)
+      val packageName = (config.workdir relativize path).toString
+        .replace(File.separator, ".")
+        .split("\\.")
+        .dropRight(1)
+        .mkString(".")
```
WojciechMazur, line 49: I think we can make it more stable by usage of `Path.subPath`, it might be more safe then parsing string path, here's example, which could be simpler if we were allowed to use collection conventers here (it's not possible due to cross compilation with Scala 2.12) 
```suggestion
      val packageName = { 
       val relPath = (config.workdir relativize path)
       val it = relPath.subPath(0, relPath.getNameCount() - 1).iterator
       // conversion from java to scala iterator
       new Iterator[Path]{def hasNext() = it.hasNext(); def next() = it.next()}
         .mkString(".")
```

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -31,44 +32,56 @@ private[scalanative] object LLVM {
...
-        }
-        val expectionsHandling =
-          List("-fexceptions", "-fcxx-exceptions", "-funwind-tables")
-        val flags = opt(config) +: "-fvisibility=hidden" +:
-          stdflag ++: platformFlags ++: expectionsHandling ++: config.compileOptions
-        val compilec =
-          Seq(compiler) ++ flto(config) ++ flags ++
-            asan(config) ++ target(config) ++
-            Seq("-c", inpath, "-o", outpath)
+        if (incCompilationContext == null || incCompilationContext.isChanged(
```
WojciechMazur, line 53: this condition can be a part of the previous one, which would eliminate additional nesting level.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -130,7 +143,21 @@ private[scalanative] object LLVM {
         asan(config) ++ target(config)
     }
     val paths = objectsPaths.map(_.abs)
-    val compile = config.clangPP.abs +: (flags ++ paths ++ linkopts)
+    val llvmLinkInfo = flags ++ paths ++ linkopts
+    def dump(strings: Seq[String], fileName: String): Unit = {
```
WojciechMazur, line 147: I remember it's a fix for passing too many file paths to the clang compiler. It would be useful to  add a comment why are we creating this file. 
As there is just a single usage we can inline it and eventually wrap into braces of `locally{...}` block to not polute namespace.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -14,7 +14,9 @@ import scala.scalanative.compat.CompatParColls.Converters._
 object CodeGen {
 
   /** Lower and generate code for given assembly. */
-  def apply(config: build.Config, linked: linker.Result): Seq[Path] = {
+  def apply(config: build.Config, linked: linker.Result)(implicit
+      incCompilationContext: IncCompilationContext = null
```
WojciechMazur, line 18: Again, null implicits are bad. Remove the default value of null please

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -44,7 +46,8 @@ object CodeGen {
 
   /** Generate code for given assembly. */
   private def emit(config: build.Config, assembly: Seq[Defn])(implicit
-      meta: Metadata
+      meta: Metadata,
+      incCompilationContext: IncCompilationContext = null
```
WojciechMazur, line 50: Also this null

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -75,7 +117,7 @@ object CodeGen {
       import build.Mode._
       (config.mode, config.LTO) match {
         case (Debug, _)                                  => separate()
-        case (ReleaseFast | ReleaseFull, build.LTO.None) => single()
+        case (ReleaseFast | ReleaseFull, build.LTO.None) => separate()
```
WojciechMazur, line 120: release + LTO.none should still be generated to a single file (for performance reasons in non lto builds) This configuration should also be checked at earlier stages so we don't introduce any additional and not utilize overhead. In such a case a warning should also be logged for the users, to make them aware of this limitation.

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -54,14 +57,53 @@ object CodeGen {
       // of available processesors. This prevents LLVM from optimizing
       // across IR module boundary unless LTO is turned on.
       def separate(): Seq[Path] =
-        partitionBy(assembly, procs)(_.name.top.mangle).par
+        assembly
+          .groupBy { defn =>
+            val packageName =
+              defn.name.top.id.split("\\.").dropRight(1).mkString(".")
```
WojciechMazur, line 63: We can use single char split to make code simpler to read and not use regexes
```suggestion
              defn.name.top.id.split('.').init.mkString(".")
```

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -54,14 +57,53 @@ object CodeGen {
...
+              }
+              if (incCompilationContext == null ||
+                  incCompilationContext.isChanged(pack)) {
+                val sorted = defns.sortBy(_.name.show)
+                val packagePrefix = config.workdir resolve
+                  Paths.get(
+                    pack
+                      .split(s"\\.")
+                      .dropRight(1)
+                      .mkString(File.separatorChar.toString)
```
WojciechMazur, line 80: File also have `separator` which would allow to skip char -> string conversion
```suggestion
                    packageName.split('.').init.mkString(File.seperator)
```

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -54,14 +57,53 @@ object CodeGen {
...
-              Impl(config, env, sorted).gen(id.toString, workdir)
+            case (pack, defns) =>
+              if (incCompilationContext != null) {
+                incCompilationContext.collectFromCurr(pack, defns)
+              }
+              if (incCompilationContext == null ||
+                  incCompilationContext.isChanged(pack)) {
+                val sorted = defns.sortBy(_.name.show)
+                val packagePrefix = config.workdir resolve
+                  Paths.get(
```
WojciechMazur, line 76: resolve also accepts String as a parameter, so we can skip pass it directory. In case of multiline method calls please use dots and braces

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -54,14 +57,53 @@ object CodeGen {
...
+                      .split(s"\\.")
+                      .dropRight(1)
+                      .mkString(File.separatorChar.toString)
+                  )
+                  .toFile
+                  .toPath
+                assert(packagePrefix.toFile.exists())
+                val packagePath = pack
+                  .split(s"\\.")
+                  .mkString(File.separatorChar.toString)
```
WojciechMazur, line 101: We can use simple replace here
```suggestion
                val packagePath = pack.replace(".", java.io.File.separator)
```

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -54,14 +57,53 @@ object CodeGen {
...
+                Impl(config, env, sorted).gen(packagePath, workdir)
+              } else {
+                val packagePrefix = config.workdir resolve Paths.get(
+                    pack
+                      .split(s"\\.")
+                      .dropRight(1)
+                      .mkString(File.separatorChar.toString)
+                  )
+                  .toFile
+                  .toPath
```
WojciechMazur, line 97: We can use path resolve + normalize to achieve the same effect
```suggestion
                val packagePrefix = config.workdir 
                 .resolve(pack.replace(".", File.separator))
                 .resolve("..")
                 .normalize
```

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -50,6 +50,9 @@ sealed trait NativeConfig {
   /** Shall we optimize the resulting NIR code? */
   def optimize: Boolean
 
+  /** Shall we use the incremental compilation? */
+  def incrementalCompilation: Boolean
```
WojciechMazur, line 54: The name for boolean variable should be a verb, instead of a noun. Probably the safest choice in this case would be `{use,enable}IncrementalCompilation` or `compileIncrementally`

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -63,6 +66,41 @@ object CodeGen {
           .toSeq
           .seq
 
+      // Incremental compilation code generation
+      def separateInc(): Seq[Path] =
```
WojciechMazur, line 70: Let's use a full name, `seperateIncrementally`

tools/src/test/scala/scala/scalanative/IncCompilationTest.scala
```diff
@@ -0,0 +1,134 @@
...
+                      |  }
+                      |  def main(args: Array[String]): Unit = {
+                      |    print(1)
+                      |    print("hello world?")
+                      |  }
+                      |}""".stripMargin
+      val entry = "A"
+      val changedTop = Set[String]("A", "A$")
+      val outDir = Files.createTempDirectory("native-test-out")
+      println(outDir)
```
WojciechMazur, line 30: There should be no `println` in the tests

tools/src/test/scala/scala/scalanative/IncCompilationTest.scala
```diff
@@ -0,0 +1,134 @@
...
+        .withLinkingOptions(Discover.linkingOptions())
+        .withLTO(Discover.LTO())
+        .withGC(Discover.GC())
+        .withMode(Discover.mode())
+        .withOptimize(Discover.optimize())
+      val config = makeConfig(outDir, entry, nativeConfig)
+      val ext = if (Platform.isWindows) ".exe" else ""
+      val artifact = outDir.resolve("result" + ext)
+
+      Build.build(config, artifact)
```
WojciechMazur, line 49: This part seems to be repeated in different tests. We can factor the common part to a helper method

tools/src/main/scala/scala/scalanative/build/IncCompilationContext.scala
```diff
@@ -0,0 +1,68 @@
+package scala.scalanative.build
+import java.io.{ByteArrayOutputStream, File, ObjectOutputStream, PrintWriter}
+import java.nio.ByteBuffer
+import java.nio.file.{Path, Paths}
+import scala.collection.concurrent.TrieMap
+import scala.io.Source
+import scala.language.implicitConversions
+import scala.scalanative.nir.Defn
+
+class IncCompilationContext {
```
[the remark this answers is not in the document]
yuly16: package `scala.scalanative.codegen` also uses `IncCompilationContext`, so we can't make it private in package `scala.scalanative.build`

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -56,9 +58,8 @@ private[scalanative] object LLVM {
       // If pack2hashPrev is not empty, we don't recompile native library.
       // Even if native library changes(This is very rare case). If native library
       // changes, we should clean the project first.
-      if ((isLl || !Files.exists(objPath)) &&
-          (incCompilationContext.pack2hashPrev.isEmpty ||
-            incCompilationContext.isChanged(packageName))) {
+      if ((isLl || !Files.exists(objPath)) && incCompilationContext
+            .shouldCompile(packageName)) {
```
WojciechMazur, line 62: Can we try to move this condition to separate line to make it more readable?
```suggestion
      if ((isLl || !Files.exists(objPath)) && 
          incCompilationContext.shouldCompile(packageName)) {
```

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -136,7 +136,7 @@ sealed trait NativeConfig {
   def withOptimize(value: Boolean): NativeConfig
 
   /** Create a new config with given incrementalCompilation value */
-  def withIncrementalCompilation(value: Boolean): NativeConfig
+  def withUseIncrementalCompilation(value: Boolean): NativeConfig
```
WojciechMazur, line 139: I guess we can remove `use` in the setter to make it more straightforward. The getter `useIncrementalCompilation` is fine, and should not be changed
```suggestion
  def withIncrementalCompilation(value: Boolean): NativeConfig
```

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -114,12 +118,11 @@ object CodeGen {
...
         config.LTO,
-        config.compilerConfig.incrementalCompilation
+        config.compilerConfig.useIncrementalCompilation
       ) match {
-        case (_, _, true)                                   => separateInc()
-        case (Debug, _, _)                                  => separate()
         case (ReleaseFast | ReleaseFull, build.LTO.None, _) => single()
-        case (ReleaseFast | ReleaseFull, _, _)              => separate()
+        case (_, _, true) => seperateIncrementally()
+        case _            => separate()
```
WojciechMazur, line 125: We have only a single check for flag here. We can simplify the whole pattern matching expression by excluding boolean and checking it in the case block
```suggestion
        case (ReleaseFast | ReleaseFull, build.LTO.None) => single()
        case _ => 
           if (config.compilerConfig.useIncrementalCompilation) seperateIncrementally()
           else separate()
```

.gitignore
```diff
@@ -38,3 +38,5 @@ bin/
 
 # Virtual env generated dependecies, for generating docs
 .venv
+
+log
```
WojciechMazur, line 42: Spurious change in the .gitignore

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -70,12 +72,22 @@ object Build {
...
+
       // optimize and generate ll
       val generated = {
         val optimized = ScalaNative.optimize(fconfig, linked)
         ScalaNative.codegen(fconfig, optimized)
       }
 
+      if (config.compilerConfig.useIncrementalCompilation) {
+        incCompilationContext.dump()
+      }
```
WojciechMazur, line 89: At this point, there is the last usage of incOmpilationContext. We should be able to clear the underlying maps at this point to allow for their GC. Maybe we can introduce method `clear` to the context which would be called now

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -31,16 +32,34 @@ private[scalanative] object LLVM {
    *  @return
    *    The paths of the `.o` files.
    */
-  def compile(config: Config, paths: Seq[Path]): Seq[Path] = {
+  def compile(config: Config, paths: Seq[Path])(implicit
+      incCompilationContext: IncCompilationContext = new IncCompilationContext(
+        config.workdir
+      )
```
WojciechMazur, line 38: There's a total of 5 instantiations of `IncCompilationContext` as the default argument. I don't think it should be needed. We probably only should have a single in the `Build::build` method. If for some reason implicit argument would not available we risk losing the state loaded previously.

WojciechMazur (commented):
Looks good, I only have some last comments

tools/src/main/scala/scala/scalanative/codegen/CodeGen.scala
```diff
@@ -121,8 +117,10 @@ object CodeGen {
         config.compilerConfig.useIncrementalCompilation
```
WojciechMazur, line 117: This parameter is never checked, can be removed

lolgab:
This is great! Thank you very much!
I have a question about how the split in `.ll` files happens.
Why does the split happen per `package`?
Have you considered an approach similar to what Scala.js does with `SmallModulesFor(List("my-package"))` where every other package not in the list is merged together (probably the same way as today) and you generate a file per class for `my-package`?
Is it because of something specific to how Scala Native works?

WojciechMazur (approved):

PR #2819, opened 2022-08-31
optimization on interflow 

yuly16:
1. add a framework test. With this framework test, it is easy to debug scala native nativeLink.
2. In the interflow phase, create a unique set to avoid duplicate instances of the same `Global` in `todo`. In the previous version, for example, when using scala-native to compile `scalafmt` project, thousands of instances of the same `Global` appear in `todo` queue, which make the size of `todo` queue too large, while actually we only need to visit this `Global` once. By introducing a set `unique`, when `pushTodo` is called, we check if this `Global` has been in `unique`. If no, we visit the `Global` and push it to `unique`; else, discard this `Global`
3. In the function `advance` in `MergeProcessor`, we change the type of `todo` as SortedSet, which makes the code cleaner and avoids transforming `todo` to array and sort it when executing `advance` as the previous version.
4. Now we can set the max inline depth and max caller/callee size from `NativeConfig`. It is necessary because very large projects(e.g. scalafmt) spend too much time in inlining in releaseFull Mode. In the previous scala-native version, compiling these large projects takes several hours or even causes out of heap memory. Now the optimization phase can be done by limiting the max inline depth or max caller/callee size.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -19,11 +19,6 @@ object ScalaNativePlugin extends AutoPlugin {
         "User configuration for the native build, NativeConfig"
       )
 
-    val optimizerConfig =
-      taskKey[build.OptimizerConfig](
-        "Configuration of the Scala Native optimizer"
-      )
-
     val nativeClang =
       taskKey[File]("Location of the clang compiler.")
 
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -19,6 +19,11 @@ object ScalaNativePlugin extends AutoPlugin {
         "User configuration for the native build, NativeConfig"
       )
 
+    val optimizerConfig =
+      taskKey[build.OptimizerConfig](
+        "Configuration of the Scala Native optimizer"
+      )
+
     val nativeClang =
       taskKey[File]("Location of the clang compiler.")
 
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -86,7 +86,6 @@ object ScalaNativePluginInternal {
       .withGC(Discover.GC())
       .withMode(Discover.mode())
       .withOptimize(Discover.optimize()),
-    optimizerConfig := build.OptimizerConfig.empty,
     nativeWarnOldJVM := {
       val logger = streams.value.log
       Try(Class.forName("java.util.function.Function")).toOption match {
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -86,6 +86,7 @@ object ScalaNativePluginInternal {
       .withGC(Discover.GC())
       .withMode(Discover.mode())
       .withOptimize(Discover.optimize()),
+    optimizerConfig := build.OptimizerConfig.empty,
     nativeWarnOldJVM := {
       val logger = streams.value.log
       Try(Class.forName("java.util.function.Function")).toOption match {
--- a/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
+++ b/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
@@ -35,12 +35,6 @@ sealed trait NativeConfig {
   /** The LTO mode to use used during a release build. */
   def lto: LTO
 
-  /** The maximum inline depth during the optimization phase */
-  def maxInlineDepth: Option[Int]
-
-  /** The maximum caller and callee size during the optimization phase */
-  def maxCallerSize: Int
-
   /** Shall linker check that NIR is well-formed after every phase? */
   def check: Boolean
 
@@ -129,12 +123,6 @@ sealed trait NativeConfig {
   /** Create a new config with the given lto mode. */
   def withLTO(value: LTO): NativeConfig
 
-  /** Create a new config with the given max inline depth. */
-  def withMaxInlineDepth(value: Int): NativeConfig
-
-  /** Create a new config with the max caller size. */
-  def withMaxCallerSize(value: Int): NativeConfig
-
   /** Create a new config with given check value. */
   def withCheck(value: Boolean): NativeConfig
 
@@ -177,8 +165,6 @@ object NativeConfig {
       targetTriple = None,
       gc = GC.default,
       lto = LTO.default,
-      maxInlineDepth = 1000,
-      maxCallerSize = 8192,
       mode = Mode.default,
       check = false,
       checkFatalWarnings = false,
@@ -201,8 +187,6 @@ object NativeConfig {
       gc: GC,
       mode: Mode,
       lto: LTO,
-      maxInlineDepth: Int,
-      maxCallerSize: Int,
       linkStubs: Boolean,
       check: Boolean,
       checkFatalWarnings: Boolean,
@@ -245,12 +229,6 @@ object NativeConfig {
 
     def withLTO(value: LTO): NativeConfig =
       copy(lto = value)
-
-    def withMaxInlineDepth(value: Int): NativeConfig =
-      copy(maxInlineDepth = value)
-
-    def withMaxCallerSize(value: Int): NativeConfig =
-      copy(maxCallerSize = value)
 
     def withCheck(value: Boolean): NativeConfig =
       copy(check = value)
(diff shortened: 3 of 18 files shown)


ekrich:
A couple of questions to potential improve the code some.

1. What is the relationship between the `todo` in `Interflow` and `MergeProcessor`?
2. Is it possible to make the `todo` in the `MergeProcessor` a `mutable.TreeSet` to preserve the order?
3. Is it also possible to make the `todo` in `MergeProcessor` a `val` and then just use `remove` on the `head` item?

JD557:
Out of curiosity, can you share some benchmarks?

yuly16:
> Out of curiosity, can you share some benchmarks?

Hi, the benchmark I used is the branch `add-scala-native` in the link `https://github.com/tgodzik/scalafmt`

yuly16:
> A couple of questions to potentially improve the code some.
> 
> 1. What is the relationship between the `todo` in `Interflow` and `MergeProcessor`?
> 2. Is it possible to make the `todo` in the `MergeProcessor` a `mutable.TreeSet` to preserve the order?
> 3. Is it also possible to make the `todo` in `MergeProcessor` a `val` and then just use `remove` on the `head` item?

Hi Ekrich, 
Sorry, I just noticed that I didn't send the message several days ago. 
1. the `todo` in `Interflow` contains the `Method` preparing to be optimized, while `todo` in `MergeProcessor` contains `blocks` in a `Method`. The two `todo` have totally different meanings so they have no relationship.
2. I think it is okay
3. It is not very easy to do it. In the original version, `todo` in `MergeProcessor` is `var`, since we need to do `filter`/`filterNot` to `todo`, which is easier to write the code if `todo` is `var` instead of `val`.

tools/src/main/scala/scala/scalanative/interflow/Eval.scala
```diff
@@ -27,7 +32,7 @@ trait Eval { self: Interflow =>
           if (unwind ne Next.None) {
             throw BailOut("try-catch")
           }
-          val value = eval(op)
+          val value = eval(op, inlineDepth)
```
WojciechMazur, line 35: `inlineDepth` can probably can be a part a member of state

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -155,6 +167,8 @@ object NativeConfig {
       targetTriple = None,
       gc = GC.default,
       lto = LTO.default,
+      maxInlineDepth = 1000,
```
WojciechMazur, line 170: I think this default value is a bit to big. If we would like allow to infinite number of inlines, then probably on `Option[Int]` would be a better choice. `None` would then mean unlimited

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -117,6 +123,12 @@ sealed trait NativeConfig {
   /** Create a new config with the given lto mode. */
   def withLTO(value: LTO): NativeConfig
 
+  /** Create a new config with the given max inline depth. */
+  def withMaxInlineDepth(value: Int): NativeConfig
+
+  /** Create a new config with the max caller size. */
+  def withMaxCallerSize(value: Int): NativeConfig
```
WojciechMazur, line 130: Can we put them in a dedicated object, eg. `OptimizerConfig` 
I think we can then have 
```
trait NativeConfig {
...
def optimizerConfig: OptimizerConfig
def withOptimizerConfig(mapping: OptimizerConfig => OptimizerConfig): NativeConfig
...
```

tools/src/main/scala/scala/scalanative/interflow/Interflow.scala
```diff
@@ -147,7 +156,12 @@ class Interflow(val mode: build.Mode, val is32BitPlatform: Boolean)(implicit
 object Interflow {
   def apply(config: build.Config, linked: linker.Result): Seq[Defn] = {
     val interflow =
-      new Interflow(config.mode, config.compilerConfig.is32BitPlatform)(linked)
+      new Interflow(
+        config.mode,
+        config.compilerConfig.is32BitPlatform,
+        config.compilerConfig.maxInlineDepth,
+        config.compilerConfig.maxCallerSize
```
WojciechMazur, line 163: With so many arguments, maybe we can make a small refactor and pass config directly to the Interflow?

tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
```diff
@@ -18,7 +18,8 @@ final class MergeProcessor(
         local -> offset
     }.toMap
   val blocks = mutable.Map.empty[Local, MergeBlock]
-  var todo = mutable.Set.empty[Local]
+  val todoOrdering: Ordering[Local] = Ordering.by(offsets)
+  var todo = mutable.SortedSet.empty[Local](todoOrdering)
```
WojciechMazur, line 22: It probably should be immutable field with mutable collection. 
```suggestion
  val todo = mutable.SortedSet.empty[Local](todoOrdering)
```

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -117,6 +123,12 @@ sealed trait NativeConfig {
   /** Create a new config with the given lto mode. */
   def withLTO(value: LTO): NativeConfig
 
+  /** Create a new config with the given max inline depth. */
+  def withMaxInlineDepth(value: Int): NativeConfig
```
WojciechMazur, line 127: In `Inline` we define also a max size of inlined methods. Currently, it's hard coded to 8 instructions. Maybe we can also allow to tune this condition as well

tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
```diff
@@ -18,7 +18,8 @@ final class MergeProcessor(
         local -> offset
     }.toMap
   val blocks = mutable.Map.empty[Local, MergeBlock]
-  var todo = mutable.Set.empty[Local]
+  val todoOrdering: Ordering[Local] = Ordering.by(offsets)
```
WojciechMazur, line 21: If todoOrdering is not reused, maybe we can inline the Ordering defintion

WojciechMazur:
Hey, thank you for this  PR. There seems to be a merge conflict with your previous, already merged PR. Can you resolve them, please, as I'm not sure which version of test is more accurate?

yuly16:
Sure, I will fix it

ekrich:
@yuly16 This has been waiting for a month now, can you rebase?

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
ekrich: This looks like it got put on the `Config` not the `NativeConfig`.

He-Pin:
@yuly16 ping~~

yuly16:
Hi Keer and Eric, I will fix it within two days.

kerr ***@***.***> 于2022年10月27日周四 18:00写道：

ekrich:
@yuly16 I also think @WojciechMazur meant for the `OptimizerConfig` to be part of `NativeConfig` rather than a new `sbt` task and only accessible from `Config` via a programmer.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -19,6 +19,11 @@ object ScalaNativePlugin extends AutoPlugin {
         "User configuration for the native build, NativeConfig"
       )
 
+    val optimizerConfig =
```
ekrich, line 22: I don't think we need a new setting as this is being added into `NativeConfig`.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -26,6 +26,9 @@ object ScalaNativePluginInternal {
   val nativeWarnOldJVM =
     taskKey[Unit]("Warn if JVM 7 or older is used.")
 
+  val nativeWorkdir =
```
ekrich, line 29: This was removed in a previous PR so I think that is a rebase issue.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -86,6 +89,7 @@ object ScalaNativePluginInternal {
       .withGC(Discover.GC())
       .withMode(Discover.mode())
       .withOptimize(Discover.optimize()),
+    optimizerConfig := build.OptimizerConfig.empty,
```
ekrich, line 92: I don't think this line is needed at all since `OptimizerConfig.empty` is created when `NativeConfig.empty` is created.

tools/src/test/scala/scala/scalanative/IncCompilationTest.scala
```diff
@@ -87,34 +141,10 @@ class IncCompilationTest extends codegen.CodeGenSpec with Matchers {
...
     val classpath = makeClasspath(outDir)
     Config.empty
-      .withBasedir(outDir)
+      .withWorkdir(outDir)
       .withClassPath(classpath.toSeq)
       .withMainClass(entry)
       .withCompilerConfig(setupNativeConfig)
   }
 
-  private def nativeLink(
```
ekrich, line 96: This is more or less correct now. We only provide a `basedir` for the compiler to do everything and should not have a `workdir`. The `artifactPath` is now calculated by `NativeConfig` and not passed to `build` but returned by `build` or you can retrieve it from `NativeConfig` as well.

ekrich (changes requested):
Seems like just some rebase issues based on things changing quite a bit in the mean time.

ekrich:
@yuly16 Thanks for this improvement.

PR #2854, opened 2022-09-19
Publish local command for easier local development

keynmol:
Let's bikeshed on the name!

The purpose of this command is to be able to run

```
sbt> publish-local-dev 3.2.0
```

And get everything published for a project to use a local version of SN with this particular Scala version.
This includes the nscplugin, sbt plugin, and test stuff.

Did I miss any projects?

### DIFF
--- a/project/Commands.scala
+++ b/project/Commands.scala
@@ -13,7 +13,8 @@ object Commands {
     "test-tools" ::
       "test-mima" ::
       "test-runtime" ::
-      "test-scripted" :: _
+      "test-scripted" ::
+      "publish-local-dev" :: _
   }
 
   lazy val testRuntime = projectVersionCommand("test-runtime") {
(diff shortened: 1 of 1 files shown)


ekrich:
Well, maybe it could replace the `release` and the `publish-local` (`publish-impl`) scripts. So maybe publish-local without an arg could publish all versions?

WojciechMazur (approved):

ekrich:
@WojciechMazur It was pointed out that the local shell scripts don't work on Windows. Do you think it is appropriate for a followup here to remove/replace the scripts I mentioned above? You use the release script so would that be okay?
cc @keynmol

keynmol:
I think Eric's suggestion is legit - I will attempt to make a more general command that can fully replace the script.

Ideally, we'd have **everything** working through SBT, as that shifts the command portability to SBT, which has been solved a while back and is maintained more vigorously.

ekrich:
Important to put WIP if needed because the merge comes fast at times.

PR #2869, opened 2022-09-25
Generate LLVM metadata 

### LINKED ISSUE #29: LLDB integration
It's already possible to debug Scala native code with LLDB but we need to provide source code position mapping to have better end-user experience.
- [x] Add support for positions to NIR
- [x] Emit positions in the frontend
- [x] Preserve positions during transformations
- [x] Map positions to LLVM's debug information in the code generator

keynmol:
This is a WIP as they come, putting it up in case I suddenly run out of time to work on it.
For demonstration purposes I managed to find 1 (**one**) file that doesn't trip up the validation

image

#### What is working

1. Metadata, modelling, indexing, and rendering (most of it is modelled)
2. Writing out subprograms, locations, compile units, files, basic types and pointers

#### What isn't working:

- [x] Tons of these "warnings" for _some_ functions - they all have something in common but I'm not sure what - at first I assumed pointers. The issue is probably super simple and very stupid.

```llvm
[error] !dbg attachment points at wrong subprogram for function
[error] !30 = distinct !DISubprogram(linkageName: "_SM28scala.util.package$chaining$G4load", scope: !6, file: !7, line: 18, type: !31, scopeLine: 18, spFlags: DISPFlagDefinition, unit: !6)
```

A fascinating aspect of this is that the error message from clang uses its own identifiers - all those `!31` and `!6` are not from the file at all.

- [x] Tons of these:

```llvm
[error] inlinable function call in a function with debug info must have a !dbg location
[error]   %_30001 = call dereferenceable_or_null(24) i8* @"_SM34scala.collection.mutable.ArraySeq$G4load"()
```

This is basically missed invocations that I think come from `*Compat` files in codegen.

- [ ] Declaring local variables: https://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-declare
- [ ] Demangled `name` in DISubprogram
- [ ] (for future PR) - nested scoping to match Scala scopes

### Suggestions

- [ ] Create a single implicit `RenderingContext` instead of repeated implicit parameters
- [x] Think about refactoring GenIdx somehow to either remove it completely or at least embed it in DwarfBuilder

Closes #29 (last remaining TODO item)


keynmol:
@WojciechMazur @densh 

I've put this up as the start of the work, and in the hopes that either of you might have a brilliant idea on what could be causing the `!dbg attachment points at wrong subprogram for function` errors on _some_ functions.

WojciechMazur:
Thanks for bootstrapping work on this! So far I have no idea why these errors are present, but I'll try to check out your branch and investigate it. At first glance, it looks correct, but I don't know debug symbols metadata well

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -36,7 +36,7 @@ private[scalanative] object LLVM {
       incCompilationContext: IncCompilationContext
   ): Seq[Path] = {
     // generate .o files for all included source files in parallel
-    paths.par.map { path =>
```
keynmol, line 39: TODO: restore this, as I was trying to not have interleaved output from so many failing compilations.

A better approach is to just grab one of the failing `.ll` files and run it through either `llc` or `clang -c <file.ll>` - it's enough to reproduce the errors.

tools/src/main/scala/scala/scalanative/build/Logger.scala
```diff
@@ -51,7 +51,7 @@ object Logger {
     def debug(msg: String): Unit = err.println(s"[debug] $msg")
     def info(msg: String): Unit = out.println(s"[info] $msg")
     def warn(msg: String): Unit = out.println(s"[warn] $msg")
-    def error(msg: String): Unit = err.println(s"[error] $msg")
+    def error(msg: String): Unit = err.println(s"[error!] $msg")
```
keynmol, line 54: TODO: remove this.

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
```diff
@@ -84,6 +130,14 @@ private[codegen] abstract class AbstractCodeGen(
     defns.foreach { defn => if (defn.isInstanceOf[Defn.Var]) onDefn(defn) }
     defns.foreach { defn => if (defn.isInstanceOf[Defn.Declare]) onDefn(defn) }
     defns.foreach { defn => if (defn.isInstanceOf[Defn.Define]) onDefn(defn) }
+
+    // defns.foreach {
```
keynmol, line 134: While I guess the original block of code separates variables from declarations, from defines, I wonder if it'd be faster to do a single pass over this collection.

It's likely to be insignificant though.

keynmol:
Ok so in a function like this:

```llvm
define i8* @"_SM28scala.util.package$chaining$G4load"()  !dbg !37 {
_1.0:
  %_12 = bitcast i8* bitcast ([201 x i8*]* @"__modules" to i8*) to i8**
  %_13 = getelementptr i8*, i8** %_12, i32 84
  %_4 = bitcast i8** %_13 to i8*
  %_14 = bitcast i8* %_4 to i8**
  %_5 = load i8*, i8** %_14, !dereferenceable_or_null !{i64 8}
  %_6 = icmp ne i8* %_5, null
  br i1 %_6, label %_2.0, label %_3.0
_2.0:
  ret i8* %_5
_3.0:
  ;%_7 = call i8* @"scalanative_alloc_small"(i8* bitcast ({ { i8*, i32, i32, i8* }, i32, i32, { i8* }, [4 x i8*] }* @"_SM28scala.util.package$chaining$G4type" to i8*), i64 8), !dbg !34
  ;%_15 = bitcast i8* %_4 to i8**
  ;store i8* %_7, i8** %_15
  ;call nonnull dereferenceable(8) i8* @"_SM28scala.util.package$chaining$RE"(i8* dereferenceable_or_null(8) %_7)
  ret i8* %_4
}
...
!37 = distinct !DISubprogram(linkageName: "_SM28scala.util.package$chaining$G4load", isDefinition: true, scope: !32, unit: !32, file: !2, line: 18, type: !36, spFlags: DISPFlagDefinition)
!33 = distinct !DISubprogram(linkageName: "_SM28scala.util.package$chaining$D20scalaUtilChainingOpsL16java.lang.ObjectL16java.lang.ObjectEO", isDefinition: true, scope: !32, unit: !32, file: !2, line: 18, scopeLine: 18, type: !31, spFlags: DISPFlagDefinition)
!34 = !DILocation(line: 18, column: 9, scope: !33)
```

If one uncomments the %_7 assignment and the rest, the error starts appearing:

```llvm
warning: overriding the module target triple with arm64-apple-macosx12.0.0 [-Woverride-module]
1 warning generated.
!dbg attachment points at wrong subprogram for function
!30 = distinct !DISubprogram(linkageName: "_SM28scala.util.package$chaining$G4load", scope: !4, file: !5, line: 18, type: !15, spFlags: DISPFlagDefinition, unit: !4)
i8* ()* @"_SM28scala.util.package$chaining$G4load"
  %_7 = call i8* @scalanative_alloc_small(i8* bitcast ({ { i8*, i32, i32, i8* }, i32, i32, { i8* }, [4 x i8*] }* @"_SM28scala.util.package$chaining$G4type" to i8*), i64 8), !dbg !29
!29 = !DILocation(line: 18, column: 9, scope: !28)
!28 = distinct !DISubprogram(linkageName: "_SM28scala.util.package$chaining$D20scalaUtilChainingOpsL16java.lang.ObjectL16java.lang.ObjectEO", scope: !4, file: !5, line: 18, type: !18, scopeLine: 18, spFlags: DISPFlagDefinition, unit: !4)
!28 = distinct !DISubprogram(linkageName: "_SM28scala.util.package$chaining$D20scalaUtilChainingOpsL16java.lang.ObjectL16java.lang.ObjectEO", scope: !4, file: !5, line: 18, type: !18, scopeLine: 18, spFlags: DISPFlagDefinition, unit: !4)
inlinable function call in a function with debug info must have a !dbg location
  %0 = call nonnull dereferenceable(8) i8* @"_SM28scala.util.package$chaining$RE"(i8* dereferenceable_or_null(8) %_7)
warning: ignoring invalid debug info in /Users/velvetbaldmime/projects/scala-native/sandbox/.3/target/scala-3.1.3/native/scala/util.ll
```

So it's a case of indeed wrong metadata, and horrible error - basically `%_7` points at the wrong location (in the wrong function) - it should be pointing at `!37` in all function calls.

keynmol:
**UPDATE** (cross-posted from Discord)

Aight, I think it's another milestone - with the latest changes, the sandbox binary 

1. links without any LLVM complaints
2. I can run MacOS' dsymutil on it: /Library/Developer/CommandLineTools/usr/bin/dsymutil /Users/velvetbaldmime/projects/scala-native/sandbox/.3/target/scala-3.1.3/sandbox-out which seems to produce a separate folder with debugging symbols
3. the produced file responds to LLVM's dwarf dump:

> ~/projects/llvm-project/build/bin/llvm-dwarfdump sandbox-out.dSYM/Contents/Resources/DWARF/sandbox-out | grep DW_AT_MIPS_linkage_name | head
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD12scala_$eq$eqL16java.lang.ObjectzEO")
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD4TYPEL15java.lang.ClassEo")
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD6equalsL16java.lang.ObjectzEO")
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD7valueOfbL14java.lang.ByteEo")
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD8hashCodeiEO")
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD8toStringL16java.lang.StringEO")
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD9byteValuebEO")
                DW_AT_MIPS_linkage_name    ("_SM14java.lang.ByteD9compareToL14java.lang.ByteiEO")
                DW_AT_MIPS_linkage_name    ("_SM15java.lang.Byte$D7comparebbiEO")
                DW_AT_MIPS_linkage_name    ("_SM15java.lang.Byte$D7valueOfbL14java.lang.ByteEO")


And contains Scala Native symbols.
I don't know how else to verify that everything is okay 😄 All the .o files produced from Scala Native code also contain debugging symbols 🎉 
Pretty sure the the other 80% is using https://www.nongnu.org/libunwind/man/unw_get_proc_info(3).html to somehow try and extract that information during unwinding...
And the next 80% is cleaning up the PR.

JD557:
Just a nitpick: I see some packages/classes with referencing DWARF, but I'm not sure if this is correct.
My understanding is that we are adding debug information to the LLVM bytecode, which is then used to either generate DWARF or PDB information depending on the target, right?

ebruchez:
Found this while searching whether Scala Native supports producing debugging information ;) It sounds very promising. Has anyone given this a try? The last comment about unwinding seems to imply that this is not usable yet even to play with?

keynmol:
> Found this while searching whether Scala Native supports producing debugging information ;) It sounds very promising. Has anyone given this a try? The last comment about unwinding seems to imply that this is not usable yet even to play with?

Hi @ebruchez!

With lots and lots of caveats, this PR will embed _some_ metadata that LLVM will then convert to DWARF. 
This is still a far cry from "debugging" information in a sense where it's usable via, say, lldb - variables, values, expressions, etc.

At this point my main focus is source locations and function names, which should eventually lead (with lots of extra work) to exceptions that no longer say "Unknown source" in stack traces.

But in its current state, it's not super useful, even though you can try it out locally by doing `publish-local-dev 3.2.2` (if your project uses Scala 3.2.2) and then using 0.5.0-SNAPSHOT as the version of scala-native plugin.

Progress on this is stalled by my lack of time and the absolute horrible experience with LLVM metadata error messages.

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
```diff
@@ -840,9 +1029,36 @@ private[codegen] abstract class AbstractCodeGen(
...
+
+    def addDebug() =
+      if (config.debugMetadata)
+        dwarfTok.foreach { tok =>
+          /** There are situations where the position is empty, for example in
+           *  situations where a null check is generated (and the function call
+           *  is throwNullPointer)
+           *
+           *  in this case we can only use NoPosition
+           */
```
WojciechMazur, line 1056: These functions are generated in the Lower, and we probably should be able to mitigate some of these problems in the future. Of course, we still can also get NoPosition from the compiler plugin, but that should be rare.

tools/src/main/scala/scala/scalanative/codegen/LLVMDebugInformation.scala
```diff
@@ -0,0 +1,375 @@
...
+import DebugInformationSection._
+import LLVMDebugInformation._
+import Val.Str
+import Val.Num
+import scala.scalanative.nir.{Position, Global}
+import scala.reflect.ClassTag
+
+private[llvm] trait GenIdx {
+  def id(tok: DebugInformationSection.Token): Int
+  def gen: DebugInformationSection.Token
```
WojciechMazur, line 13: It's side-effecting so it should be called with parenthesis

tools/src/main/scala/scala/scalanative/codegen/LLVMDebugInformation.scala
```diff
@@ -0,0 +1,375 @@
...
+}
+private[llvm] object GenIdx {
+  def create = new GenIdx {
+    private val mp = collection.mutable.Map.empty[Token, Int]
+    override def id(tok: Token): Int = mp(tok)
+    override def gen: Token = {
+      val newVal = new Token
+      val newIdx = mp.size
+
+      mp.synchronized {
```
WojciechMazur, line 23: Each GenIdx instance is accessible by a single thread (AbstractCodeGen.Impl instance) so synchronization should not be needed

tools/src/main/scala/scala/scalanative/codegen/LLVMDebugInformation.scala
```diff
@@ -0,0 +1,375 @@
...
+      }
+
+    def cachedBy[K, T <: LLVMDebugInformation](
+        k: K,
+        in: => LLVMDebugInformation
+    )(implicit
+        ct: ClassTag[T]
+    ): Incr[T] =
+      keyCache.synchronized {
+        keyCache
```
WojciechMazur, line 333: Maybe we can use a concurrent hash map so we can skip locking? Eg. `scala.collection.concurrent.TrieMap`

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
```diff
@@ -70,12 +122,19 @@ private[codegen] abstract class AbstractCodeGen(
...
+  )(implicit
+      sb: ShowBuilder,
+      dwf: DebugInformationSection.Builder
+  ): Unit = {
     import sb._
     def onDefn(defn: Defn): Unit = {
       val mn = mangled(defn.name)
       if (!generated.contains(mn)) {
         newline()
+        if (defn.pos != Position.NoPosition)
```
WojciechMazur, line 136: You can use `defn.pos.isEmpty` or `defn.pos.isDefined` to make it shorter

WojciechMazur (commented):
Once again thank you for the enormous amount of work. I believe that's the correct direction to go! Sorry for the late review. So far I don't have some bigger concerns, I like how the `LLVMDebugInfo` was introduced and implemented in a bigger scope. We'd probably need to polish some things here, but let's wait for the cleanups and further work in this area. 

I believe the next step would be to try to force the usage of `libdwarf` to work with these changes. Based on that we probably should be able to get source positions, however unfortunetlly I've not managed to make any POC here. I've tried to spend a bit of time on that, but it's API was quite problematic to comprehend.

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
keynmol: Done in [f55c49b](https://github.com/scala-native/scala-native/pull/2869/commits/f55c49ba668729a24e5c252226f80f880e2f77f4)

tools/src/main/scala/scala/scalanative/codegen/LLVMDebugInformation.scala
keynmol: Done in [f55c49b](https://github.com/scala-native/scala-native/pull/2869/commits/f55c49ba668729a24e5c252226f80f880e2f77f4)

project/Build.scala
```diff
@@ -442,9 +442,10 @@ object Build {
     MultiScalaProject("sandbox", file("sandbox"))
       .enablePlugins(MyScalaNativePlugin)
       .settings(nativeConfig ~= { c =>
-        c.withCompileOptions(c.compileOptions :+ "-g")
-          .withLinkingOptions(c.linkingOptions :+ "-g")
+        c
           .withDebugMetadata(true)
```
ekrich, line 446: Seems like this wrap is not needed.

tanishiking (commented):
Hey @keymol, thanks a lot for working on this PR, debugging information on scala-native is something I really like to have in scala-native!

How can I help you to finish this PR?

keynmol:
Hey @tanishiking!

> How can I help you to finish this PR?

My general feeling is that _this_ PR needs relatively little work to be merged behind a config flag disabled by default. I'll defer to @WojciechMazur on what exactly he wants to change here.

The larger picture is shown on the diagram below. The parts linked on the diagram:

1. PR with decoding DWARF line locations: https://github.com/indoorvivants/macho-elf-coff-parser/pull/7
2. Project for decoding various binary formats (at this point only Mach-O, the mac one): https://github.com/indoorvivants/macho-elf-coff-parser

What's 100% missing currently is the glue code between the Scala Native exception runtime and the libunwind to collect the traces and the program counters. There's a whole bag of problems there as I need to learn about relocations and various other quirks of binaries.

The good news is that metadata on its own can be useful - someone™ can build on top of it and add symbols for local variables and other things, which should automatically make LLVM debugger more usable with Scala Native binaries (that's my theory)


d2(2)

I don't have a lot of free time, so if we can work on the shortest possible path to merging this (disabled by default), then we can prioritise bringing in the sources from macho-elf-coff-parser into SN runtime and see what's missing to make an actual exception work

tanishiking:
Thank you for the detailed explanation!
Though, I'm not familiar with the things related to libunwind (I need to study...) I would like to be work on things like local variables and Subprogram names

tanishiking:
It seems `sandbox2_13/nativeLink` fails in this branch, but rebase on `main` fixed the stuff https://github.com/scala-native/scala-native/commit/6ecd8445d7b93022e620e6cfbefb3edf9f2934ae

tanishiking:
It looks like, something fails when LTO is enabled, let me check :eyes:

keynmol:
Setting it as ready for review because @WojciechMazur somehow fixed the build (I am _positive_ that I added both of those changes before and it didn't work :D)

tanishiking:
By generating the `.debug_aranges` section, we can efficiently look up the line number and file name using a mechanism similar to [add2line](https://linux.die.net/man/1/addr2line), that depends on `.debug_aranges`. Maybe we can copy-and-paste the source code of `addr2line` into `scala-native`.
Ref: [22288 – eu-addr2line doesn't find a rust file:line](https://sourceware.org/bugzilla/show_bug.cgi?id=22288).

Additionally, I came across an interesting blog post that showcases the implementation of a stacktrace  with line numbers and filenames, using the `libbacktrace`: [C/C++: printing stacktrace containing file name, function name, and line numbers using libbacktrace - Jiyang Tang](https://tjysdsg.github.io/libbacktrace)
[According to the comment in libbacktrace](https://github.com/ianlancetaylor/libbacktrace/blob/cdb64b688dda93bbbacbc2b1ccf50ce9329d4748/dwarf.c#L2450-L2452)  it just reads through the entire ".debug_info" section, instead of looking up by the `.debug_aragens` section (which might be a bit time-consuming).

Since there's a discussion around emitting debug_aranges by default or not https://github.com/rust-lang/rust/issues/45246#issuecomment-963734305, having an option to enhance the stacktrace without `.debug_aranges` would be good.

---

That being said, it seems we have some options to enhance the stacktrace of scala-native:

- Read through `.debug_info` or look up using `.debug_aranges`
  - I personally think reading through `.debug_info` would be a good starting point.
- Implement in C or Scala?
  - While @keynmol's DWARF parser implementation seems really interesting, maybe using C libraries such as elfutils / addr2line / libbacktrace would be an easier option for now.

keynmol:
> Implement in C or Scala? 

IMO, putting _more_ C/C++ code into Scala Native doesn't seem right. Not taking into account various additional licensing and build pains, implementing the logic we need in Scala will ensure long term maintainability and much broader set of potential contributors.

In my project the DWARF part is ready - and it's file-agnostic, so if implemented once, it can be used on ELF/Mach-O/etc.

The code doesn't use aranges: https://github.com/indoorvivants/macho-elf-coff-parser/blob/main/main.scala#L33 (mainly because I didn't know about it :D)

I think it will be better to go full Scala, and gradually support more file formats, rather than go through the effort of integrating libbacktrace (which does exactly the same as my Scala code mostly) only to then rip it out.

tanishiking:
Hey, I'm  doing an experiment to add file and line information to backtraces using debug information. I'm basing my work on this PR in the following branch:https://github.com/tanishiking/scala-native/tree/fileline-stacktrace

I migrated the mach-o and DWARF parser code from this repository https://github.com/indoorvivants/macho-elf-coff-parser (Thank you very much @keynmol this is awesome!)

Here's the approach I'm taking:

- Search for the DIE with the `DW_TAG_subprogram` attribute using the address obtained from `libunwind`. The line number can be found using the `DW_AT_decl_line` attribute.
- Search for the CU DIE (with `DW_TAG_compile_unit`), which is parent of the found `DW_AT_subprogram` DIE. The last CU DIE before the subprogram DIE is the one we're interested in. 
  - The filename can be found using the `DW_AT_name` attribute.
  - Parsing the `.debug_line` section would be straightforward, but it seems rather complicated.

For example, when running the following code (in sandbox):

```scala
object Test {
    f()
  }

  def f() = g()

  def g() = error()

  def error() = throw new Error("test")
}
```

the resulting stacktrace looks like this: :tada:

```
java.lang.Error: test
        at java.lang.StackTrace$.$anonfun$currentStackTrace$1(Throwables.scala:54)
        at java.lang.StackTrace$$$Lambda$3.apply(Throwables.scala:54)
        at scala.scalanative.unsafe.Zone$.apply(Zone.scala:27)
        at java.lang.StackTrace$.currentStackTrace(Throwables.scala:50)
        at java.lang.Throwable.fillInStackTrace(Throwables.scala:125)
        at Test$.error(Test.scala:9)
        at Test$.g(Test.scala:7)
        at Test$.f(Test.scala:5)
        at Test$.main(Test.scala:1)
        at Test.main(Test.scala:1)
        at <none>.start(Unknown Source)
```

Still it's not ready for review (and this PR isn't yet merged) and there're things to do, we're in the right direction :)

TODO:
- [ ] Read a separate debug file.
- [ ] Make it work with PIE.
  - Currently, the DIE lookup fails if we link with `-pie` (which is usually the case by default these days).
  - When the executable is a PIE, the address retrieved from libunwind is randomized due to ASLR. To workaround this, we need to adjust the address by searching for something like `/proc/self/maps` and subtracting it (for macOS, we can use `vmmap`?).
- [ ] Support other executables; ELF and COFF (PE?)

tanishiking (approved):
I left some comments, but overall it looks good to me. Thanks again @keynmol for this incredible work! :tada:

As commented in the PR description, let's work on demangling the subroutine name (otherwise, we cannot set a break point if users' don't know the `linkageName`) and generating local variables 👍

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
```diff
@@ -709,9 +898,7 @@ private[codegen] abstract class AbstractCodeGen(
               }
               str(", !")
               str(deref)
-              str(" !{i")
-              str(platform.sizeOfPtrBits)
-              str(" ")
+              str(" !{i64 ")
```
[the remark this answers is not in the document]
WojciechMazur: It could be fixed in seperate PR, but that's not a big deal. The previous code was incorrect, becouse `derencable_or_null` always required i64.

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
```diff
@@ -41,12 +60,45 @@ private[codegen] abstract class AbstractCodeGen(
...
+      val debugInfo = dir.write(Paths.get(s"$id-debug.ll")) { writer =>
+        implicit val sb: ShowBuilder = new FileShowBuilder(writer)
+
+        db.anon(
+          LLVMDebugInformation.`llvm.module.flags`(
+            Seq(
+              db.anon(
+                LLVMDebugInformation
+                  .IntAttr(
+                    7,
```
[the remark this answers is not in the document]
WojciechMazur: Now there is an enum defined. Thank you for linking the source!

tools/src/main/scala/scala/scalanative/codegen/AbstractCodeGen.scala
```diff
@@ -860,13 +1074,14 @@ private[codegen] abstract class AbstractCodeGen(
...
-        if (unwind ne Next.None) {
+        if (unwind eq Next.None) addDebug()
+        else {
           str(" to label %")
           currentBlockSplit += 1
           genBlockSplitName()
           str(" unwind ")
-          genNext(unwind)
+          genNext(unwind, dwarfTok)
+          addDebug()
```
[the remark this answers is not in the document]
WojciechMazur: Should be fixed now, I've removed redundant generation of dbg postion in genNext

tanishiking (approved):
LGTM again from my side, thanks again @keynmol @WojciechMazur !

WojciechMazur (approved):

PR #2884, opened 2022-10-07
Fix `FileChannel#write` for read-only buffers

armanbilge:
Fixes a bug discovered in https://github.com/typelevel/fs2/pull/3008.


javalib/src/main/scala/java/nio/channels/FileChannelImpl.scala
```diff
@@ -338,7 +338,14 @@ private[java] final class FileChannelImpl(
...
     val srcLim: Int = buffer.limit()
     val lim = math.abs(srcLim - srcPos)
-    write(buffer.array(), 0, lim)
+    val bytes = if (buffer.hasArray()) {
+      buffer.array()
+    } else {
+      val bytes = new Array[Byte](lim)
+      buffer.get(bytes)
+      bytes
+    }
```
armanbilge, line 347: read-only buffers do not give direct access to the underlying `Array`

WojciechMazur (approved):

PR #2898, opened 2022-10-11
Port `j.u.ArrayDeque` from JSR 166

armanbilge:
This PR replaces the `java.util.ArrayDeque` implementation originally ported from Scala.js with a manual, line-by-line port of the public domain Java implementation distributed in [JSR 166](https://gee.cs.oswego.edu/dl/concurrency-interest/index.html). The new implementation passes all the existing tests, in addition to 60 new tests ported from the JSR 166 test suite.

This fixes problems inherited from the Scala.js implementation such as https://github.com/scala-js/scala-js/issues/4734 and hopefully generally improves the `ArrayDeque`'s performance and robustness.

This PR should be backported to 0.4.x.

### DIFF
--- a/javalib/src/main/scala/java/util/ArrayDeque.scala
+++ b/javalib/src/main/scala/java/util/ArrayDeque.scala
@@ -1320,14 +1320,10 @@ class ArrayDeque[E](
    *    a copy of this deque
    */
   override def clone(): ArrayDeque[E] = {
-    try {
-      val result = super.clone().asInstanceOf[ArrayDeque[E]]
-      result.elements = Arrays.copyOf(elements, elements.length)
-      return result
-    } catch {
-      case e: CloneNotSupportedException =>
-        throw new AssertionError()
-    }
+    val result = new ArrayDeque[E](Arrays.copyOf(elements, elements.length))
+    result.head = this.head
+    result.tail = this.tail
+    result
   }
 
   /** debugging */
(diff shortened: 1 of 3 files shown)


ekrich:
Please merge this after my javalib PR.

armanbilge:
@ekrich thanks for your comments.

1. I looked at your Javalib PR and I don't believe it is conflicting with this one. You did not make any modifications to `ArrayDeque` and to the best of my knowledge I did not use any Scala lib in this port. So I think we should be safe there :)

3. Note that the Scala.js implementation is based on `js.Array`, which is a growable array-like type. The original port of `ArrayDeque` from Scala.js simply replaced `js.Array` with `j.u.ArrayList`. This new implementation is completely in terms of `Array` which I hope will have better performance.

    Furthermore, I ran the JSR 166 test suite against the old `ArrayDeque` implementation (that is based on Scala.js). It fails 12 of the 60 tests. Unfortunately the PR you linked will not fix any of those, as it only fixes the performance bug, for which there is no test actually.

ekrich:
They are creating a new ring buffer version. The reason for merging mine first is that I change the location of `AssertThrows` but I also change `ScalaOps` but maybe you are not using it. I make a ton of changes, probably not that many that would need rebasing but mine is ready now anyway.

WojciechMazur (approved):

ekrich:
@armanbilge You sly dude went to see the master Doug Lea - nice going!

PR #2909, opened 2022-10-16
Optimize method `AbstractStringBuilder.append0(CharSequence, Int, Int)`

david-bouyssie:
This PR changes the implementation of the following method:
https://github.com/scala-native/scala-native/blob/cf6d1a4de1e1a9386fc437119158dc7384b9a58e/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala#L123

This PR also attributes Android Luni in the source code and updates the project attribution in `LICENSE.md`

The code modification is based on the "Android Luni libcore2" implementation (Apache 2.0 license) that can be found at:
https://android.googlesource.com/platform/libcore2/+/master/luni/src/main/java/java/lang/AbstractStringBuilder.java#151

A significant difference compared to the reference code is that value is not cloned if the backing Array is shared.
For this purpose the following code fragment has been removed:
```scala
if (shared) {
    value = value.clone();
    shared = false;
}
```

This may have side effects in combination of the other related PR https://github.com/scala-native/scala-native/pull/2908.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1793,7 +1793,15 @@ THE SOFTWARE.
 
 # License notice for Android Luni
 
-Implementation of `InflaterOutputStream` is taken based on the [Android Luni](https://android.googlesource.com/platform/libcore/) project.
+Scala Native's `javalib/` contains parts that are derived from either:
+* the latest "Apache 2.0 licensed" [libcore snapshot](https://android.googlesource.com/platform/libcore/+/2e317a02b5a8f9b319488ab9311521e8b4f87a0a/luni/) of the Android Luni project.
+* the "Apache 2.0 licensed" [libcore2 archive](https://android.googlesource.com/platform/libcore2/+/master/luni/) of the Android Luni project.
+
+Those parts are either marked with `// ported from Android Luni` or include the full copyright preamble in the source code file.
+
+For instance, the implementation of `InflaterOutputStream` is based on this source file:
+https://android.googlesource.com/platform/libcore/+/2e317a02b5a8f9b319488ab9311521e8b4f87a0a/luni/src/main/java/java/util/zip/InflaterInputStream.java
+
 The original license notice is included below:
 
 ```
--- a/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
+++ b/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
@@ -1,3 +1,4 @@
+// Contains parts ported from Android Luni
 package java.lang
 
 import java.io.InvalidObjectException
--- a/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
+++ b/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
@@ -139,7 +139,7 @@ abstract class AbstractStringBuilder private (unit: Unit) {
     if (newCount > value.length)
       enlargeBuffer(newCount)
 
-    chars match {
+    chars0 match {
       case str: String => str.getChars(start, end, value, count)
       case asb: AbstractStringBuilder =>
         System.arraycopy(asb.value, start, value, count, length)
@@ -147,7 +147,7 @@ abstract class AbstractStringBuilder private (unit: Unit) {
         var i = start
         var j = count // Destination index.
         while (i < end) {
-          value(j) = chars.charAt(i)
+          value(j) = chars0.charAt(i)
           j += 1
           i += 1
         }
--- a/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
+++ b/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
@@ -121,15 +121,38 @@ abstract class AbstractStringBuilder private (unit: Unit) {
   }
 
   final def append0(
-      chars: CharSequence,
-      start: scala.Int,
-      end: scala.Int
+    chars: CharSequence,
+    start: scala.Int,
+    end: scala.Int
   ): Unit = {
     val chars0 = if (chars != null) chars else "null"
-    if (start < 0 || end < 0 || start > end || end > chars0.length()) {
+
+    val nChars = chars0.length()
+    if (nChars == 0) return
+
+    if (start < 0 || end < 0 || start > end || end > nChars)
       throw new IndexOutOfBoundsException()
+    
+    val length = end - start
+    val newCount = count + length
+    if (newCount > value.length)
+      enlargeBuffer(newCount)
+
+    chars match {
+      case str: String => str.getChars(start, end, value, count)
+      case asb: AbstractStringBuilder =>
+        System.arraycopy(asb.value, start, value, count, length)
+      case _ =>
+        var i = start
+        var j = count // Destination index.
+        while (i < end) {
+          value(j) = chars.charAt(i)
+          j += 1
+          i += 1
+        }
     }
-    append0(chars0.subSequence(start, end).toString)
+    
+    this.count = newCount
   }
 
   def capacity(): scala.Int = value.length
--- a/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
+++ b/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
@@ -126,33 +126,10 @@ abstract class AbstractStringBuilder private (unit: Unit) {
     end: scala.Int
   ): Unit = {
     val chars0 = if (chars != null) chars else "null"
-
-    val nChars = chars0.length()
-    if (nChars == 0) return
-
-    if (start < 0 || end < 0 || start > end || end > nChars)
+    if (start < 0 || end < 0 || start > end || end > chars0.length()) {
       throw new IndexOutOfBoundsException()
-    
-    val length = end - start
-    val newCount = count + length
-    if (newCount > value.length)
-      enlargeBuffer(newCount)
-
-    chars match {
-      case str: String => str.getChars(start, end, value, count)
-      case asb: AbstractStringBuilder =>
-        System.arraycopy(asb.value, start, value, count, length)
-      case _ =>
-        var i = start
-        var j = count // Destination index.
-        while (i < end) {
-          value(j) = chars.charAt(i)
-          j += 1
-          i += 1
-        }
     }
-    
-    this.count = newCount
+    append0(chars0.subSequence(start, end).toString)
   }
 
   def capacity(): scala.Int = value.length
--- a/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
+++ b/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
@@ -122,9 +122,9 @@ abstract class AbstractStringBuilder private (unit: Unit) {
   }
 
   final def append0(
-    chars: CharSequence,
-    start: scala.Int,
-    end: scala.Int
+      chars: CharSequence,
+      start: scala.Int,
+      end: scala.Int
   ): Unit = {
     val chars0 = if (chars != null) chars else "null"
 
@@ -133,7 +133,7 @@ abstract class AbstractStringBuilder private (unit: Unit) {
 
     if (start < 0 || end < 0 || start > end || end > nChars)
       throw new IndexOutOfBoundsException()
-    
+
     val length = end - start
     val newCount = count + length
     if (newCount > value.length)
@@ -152,7 +152,7 @@ abstract class AbstractStringBuilder private (unit: Unit) {
           i += 1
         }
     }
-    
+
     this.count = newCount
   }
 
--- a/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
+++ b/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
@@ -126,10 +126,33 @@ abstract class AbstractStringBuilder private (unit: Unit) {
     end: scala.Int
   ): Unit = {
     val chars0 = if (chars != null) chars else "null"
-    if (start < 0 || end < 0 || start > end || end > chars0.length()) {
+
+    val nChars = chars0.length()
+    if (nChars == 0) return
+
+    if (start < 0 || end < 0 || start > end || end > nChars)
       throw new IndexOutOfBoundsException()
+    
+    val length = end - start
+    val newCount = count + length
+    if (newCount > value.length)
+      enlargeBuffer(newCount)
+
+    chars match {
+      case str: String => str.getChars(start, end, value, count)
+      case asb: AbstractStringBuilder =>
+        System.arraycopy(asb.value, start, value, count, length)
+      case _ =>
+        var i = start
+        var j = count // Destination index.
+        while (i < end) {
+          value(j) = chars.charAt(i)
+          j += 1
+          i += 1
+        }
     }
-    append0(chars0.subSequence(start, end).toString)
+    
+    this.count = newCount
   }
 
   def capacity(): scala.Int = value.length
--- a/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
+++ b/javalib/src/main/scala/java/lang/AbstractStringBuilder.scala
@@ -5,20 +5,21 @@ import java.io.InvalidObjectException
 import java.util.Arrays
 import scala.util.control.Breaks._
 
-abstract class AbstractStringBuilder private (unit: Unit) {
+protected abstract class AbstractStringBuilder private (unit: Unit) {
   import AbstractStringBuilder._
 
   protected var value: Array[Char] = _
   protected var count: scala.Int = _
   protected var shared: scala.Boolean = _
 
-  final def getValue(): Array[scala.Char] = value
+  private[lang] final def getValue(): Array[scala.Char] = value
+
   final def shareValue(): Array[scala.Char] = {
     shared = true
     value
   }
 
-  final def set(chars: Array[scala.Char], len: scala.Int): Unit = {
+  /*final def set(chars: Array[scala.Char], len: scala.Int): Unit = {
     val chars0 = if (chars != null) chars else new Array[scala.Char](0)
     if (chars0.length < len) {
       throw new InvalidObjectException("")
@@ -27,7 +28,7 @@ abstract class AbstractStringBuilder private (unit: Unit) {
     shared = false
     value = chars0
     count = len
-  }
+  }*/
 
   def this() = {
     this(())
@@ -55,7 +56,7 @@ abstract class AbstractStringBuilder private (unit: Unit) {
     shared = false
   }
 
-  final def appendNull(): Unit = {
+  protected final def appendNull(): Unit = {
     val newSize = count + 4
     if (newSize > value.length) {
       enlargeBuffer(newSize)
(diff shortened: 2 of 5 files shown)


LeeTibbert:
David,  Thank you for this PR and careful License attribution.

This Looks Good To Me (LGTM).  It definitely looks like an improvement.

LeeTibbert (approved):
Looks Good to Me (LGTM).  Thank you.

david-bouyssie:
I signed the CLA.

**Note to the reviewers relative to the CLA**
The modifications of this PR are based on old/archived version of Android Luni:
https://android.googlesource.com/platform/libcore2/+/master/luni/src/main/java/java/lang/AbstractStringBuilder.java

david-bouyssie:
@LeeTibbert 

re: removed code fragment

I'm still a bit anxious about the removal of this code fragment
```scala
if (shared) {
    value = value.clone();
    shared = false;
}
```

Normally the `enlargeBuffer()` should take care of the "sharing" logic and thus it should be safe.
But I'm still wondering why it is there in original Android Luni version.

What do you think?

LeeTibbert:
To put into words what I think we both intend.  The correctness & immutability of any String created by
a `sb.toString()` is paramount.  

_Any_ append (and other conditions we have discussed, such as shrinkage) to an SB which has had 
`toString()` called on 
it since the last allocation of the SB's underlying Array should cause a new SB array (in example above SB1)
to be allocated. 

If I go to a calm, quiet place and meditate for a while and look for central effect, 
I think what we are implementing we are describing is a 'lazy' Array[Char] allocation.

Please keep me honest here:
That is, currently a String is created and it always allocates an Array[Char]. 

With the proposed algorithm, a String is allocated and uses the Array[Char] from 
the parent StringBuilder. 

If that StringBuilder is single-use, we have saved one Array allocation.

If that StringBuilder is multi-use and does anything to touch its backing Array,
a new Array must be allocated. That allocation would have happened anyway
in the present scenario.  

The size of the new Array and its growth patterns are areas for discussion and
experimentation.   I think each of the append methods and `trimToSize` and `setLength()`
need to be considered separately. It may be possible to so something smart, given
known current size and intended new size.

I would describe this as workloads with a lot of single-use-StringBuilders get an unconditional
win.  Workloads with a lot of multiple-use StringBuilders get "no worse than today" behavior with
the potential for improvements.

How much improvement in the multi-use case depends on if we can make each of the "new-Array-allocation-forced"
cases capable doing something better than today.

I would have to study the existing code.   As a raw thought, imagine appending a 80 character String to a 
StringBuilder which has a shared 16 byte Array.  An `ensureCapacity` will, I think, allocate a new Array of at least
96 Chars.  When the hypothetical new "unshare-current" does it allocation, it could use knowledge that it is called
from an `append` and knows the size that will be added to directly allocate the 96 bytes (and perhaps a bit extra)
I can imaging "dumb" implementations in which "unsure-current" blindly allocates an new Array of size current,
and the append calls `ensureCapacity()` which allocates another array, hosed! 

The latter algorithm would be my first draft implementation and I hope that I would realize that it was dumb!

I have to go off and deal with Life.  I'll trace the propose (multiple) append, `trimToSize`, and `setLength()`
code paths to see 1) are we introducing any regressions 2) can we use knowledge of where we are to
reduce allocations or to make the large enough to avoid "un-share & then ensureCapacity" allocations.

david-bouyssie:
>To put into words what I think we both intend. The correctness & immutability of any String created by
a sb.toString() is paramount.

Agree

>Any append (and other conditions we have discussed, such as shrinkage) to an SB, which has had toString() called on it since the last allocation of the SB's underlying Array, should cause a new SB array (in example above SB1)
to be allocated.

I think it's not mandatory. It's only required if SB1 capacity is exceeded.
Otherwise, I don't see why it breaks the String immutability contract.

Here is a small picture that illustrates my current understanding:
image

LeeTibbert:
> I think it's not mandatory. It's only required if SB1 capacity is exceeded.

You are correct. On thinking about it and reading the _String code I think I
figured out that the `append` family need not care about `shared` and
can go about doing their thing.

LeeTibbert:
I spent some time looking at the `_String` class, constructor, and code. It seems to pretty consistently use
array, origin, length  to access into the data.  That means, as David has shown graphically, that any operation
which adds to the 'shared' array and the child string will not "see" it.  

I have not yet searched the source file to see if any 'array.length' usages have crept into the source, in
contravention of the array,offset,length idiom. The former would invalidate immutability.  Just being
extra careful with venerable code which has been touched by many people over the years.

LeeTibbert:
I seemed to remember that there are discussions and long studies of how compilers, such as gcc & llvm handle
shared strings. There are well identified things they can and can not do. We may not be the first people
addressing this issue.

LeeTibbert:
So, methods which can shrink the array or (if any) alter the  contents of the array before 
the largest "shared" size need to allocate a new array (and some of those may be doing so already).

I think I convinced myself that `shared` should change from boolean to a count of the size of the
array when it was last given to `toString()`.  The starting count could be -1. I am pretty sure that
would work in all cases. 0 might work, with some spurious allocations is rare cases, but I did
not run that to ground.

LeeTibbert:
To try to converge my thoughts.  I keep coming back to the "wastage" decision point.

I do not know the allocation size of the minimum Object. I think it is something like 16 bytes
on 64 bit systems.   

Current behavior gives us a wastage of "minimum size of an Object allocation (in this case Array itself, not contents)."
For the sake of discussion lets call that MSO (a.k.a 16).

In a new scheme, a fixed size "allowed wastage" of MSO would give us a large set of the benefits we are seeing
with today's explicit + MSO memory budget.   

What I like about this is that:
1) it can be implemented today (well, quickly) 

2) it is a single point where better algorithms can be tried in private

3) it maintains today's behavior of huge objects not becoming extra huge.
       
Like a ziggurat, better algorithms can be built on top.

David, I do not mean to be beating the same drum about "fixed size" wastage and
annoying you. I know that you have heard that approach before and declined it.

I think the key step is figuring out how to determine/measure "better"
some trade off between number of allocations (a proxy for execution speed) and
total amount of memory used.

I suspect that allowing a "wastage" of "MSO + 16" would give an execution speed up
for most workloads with unnoticeable  or "acceptable" increased required memory used.
This would have to be tested.

I have also considered wilder algorithms, based on an expected string size of 80 or 132
characters and the buffer growth progression David mapped out.  Those are outside of
the scope of this PR.

david-bouyssie:
> So, methods which can shrink the array or (if any) alter the contents of the array before the largest "shared" size need to allocate a new array (and some of those may be doing so already).
> 
> I think I convinced myself that `shared` should change from boolean to a count of the size of the array when it was last given to `toString()`. The starting count could be -1. I am pretty sure that would work in all cases. 0 might work, with some spurious allocations is rare cases, but I did not run that to ground.

To help further readers (including us) to follow the discussion, could you please move this comment to:
https://github.com/scala-native/scala-native/pull/2908

david-bouyssie:
> To try to converge my thoughts. I keep coming back to the "wastage" decision point.
> 
> I do not know the allocation size of the minimum Object. I think it is something like 16 bytes on 64 bit systems.
> 
> Current behavior gives us a wastage of "minimum size of an Object allocation (in this case Array itself, not contents)." For the sake of discussion lets call that MSO (a.k.a 16).
> 
> In a new scheme, a fixed size "allowed wastage" of MSO would give us a large set of the benefits we are seeing with today's explicit + MSO memory budget.
> 
> What I like about this is that:
> 
> 1. it can be implemented today (well, quickly)
> 2. it is a single point where better algorithms can be tried in private
> 3. it maintains today's behavior of huge objects not becoming extra huge.
> 
> Like a ziggurat, better algorithms can be built on top.
> 
> David, I do not mean to be beating the same drum about "fixed size" wastage and annoying you. I know that you have heard that approach before and declined it.
> 
> I think the key step is figuring out how to determine/measure "better" some trade off between number of allocations (a proxy for execution speed) and total amount of memory used.
> 
> I suspect that allowing a "wastage" of "MSO + 16" would give an execution speed up for most workloads with unnoticeable or "acceptable" increased required memory used. This would have to be tested.
> 
> I have also considered wilder algorithms, based on an expected string size of 80 or 132 characters and the buffer growth progression David mapped out. Those are outside of the scope of this PR.

I think the "wastage" decision part, should be treated in the .toString PR/issue.
So as above, could you please move this comment to:
https://github.com/scala-native/scala-native/pull/2908

david-bouyssie:
> re: data for "wastage" decision

Idem there, could you please move this comment to:
https://github.com/scala-native/scala-native/pull/2908

Thanks

LeeTibbert:
Entries moved as requested,  they do make more sense in #2908.  
Thank you for picking up on this and the politely suggested better approach.

Sorry for dropping them in the "wrong" place.
I deleted my entries. I think if you delete yours, the 'off-topic' discussion will 
disappear, so a future reader can go from entry to entry here and not wonder
what glitched.

david-bouyssie:
> Entries moved as requested, they do make more sense in #2908. Thank you for picking up on this and the politely suggested better approach.
> 
> Sorry for dropping them in the "wrong" place. I deleted my entries. I think if you delete yours, the 'off-topic' discussion will disappear, so a future reader can go from entry to entry here and not wonder what glitched.

Thanks for the move.

I deleted my suggestions, I think you can delete your last comment, then I'll delete this one and it will be totally clean ;-)

david-bouyssie:
> I spent some time looking at the `_String` class, constructor, and code. It seems to pretty consistently use array, origin, length to access into the data. That means, as David has shown graphically, that any operation which adds to the 'shared' array and the child string will not "see" it.
> 
> I have not yet searched the source file to see if any 'array.length' usages have crept into the source, in contravention of the array,offset,length idiom. The former would invalidate immutability. Just being extra careful with venerable code which has been touched by many people over the years.

The backing Array field is defined like this in the _String class:
```
 protected[_String] var value: Array[Char] = _
```

Although having a protected field in a final class, is not very logical to me, it let me think that risk of bad use of `value.length()` are rather limited.

It's also important to note that the public method `String.length()` is defined like this:
```scala
def length(): Int = count
```

Finally regarding the number of usages,  it looks like String.value.length is only used there:

```scala
  /**
   * Constructs a new string with the same sequence of characters as {@code
   * toCopy}. The returned string's <a href="#backing_array">backing array</a>
   * is no larger than necessary.
   */
  def this(string: String) {
    this()
    value = if (string.value.length == string.count) string.value
    else Arrays.copyOfRange(string.value, string.offset, string.offset + string.length)
    offset = 0
    count = value.length
  }
```

I think the `Arrays.copyOfRange()` line it could be replaced by else `Arrays.copyOfRange(string.value, string.offset, string.offset + string.count)` for better consistency/readability (according to the test just above).
Also count could be set as `count = value.count`.

david-bouyssie:
So back to the original question, code fragment removal.
To me it's seems safe on a theoretical point of view.

However how could we validate this? Can we rely on exist test cases? Should we design a new one? If yes, how?

LeeTibbert:
| I think these String class specific changes could be addressed separately.
I agree. I think the work in that comment is essential in convincing ourselves
and others that we are not introducing backdoor mutable strings.

I know that I feel better knowing that there is only one use of .length and
that we can fix it.

A discussion for the spin-off Topic. I think that  would always to the copy
to prevent second, and later, generations of 'at-risk' Strings. I think
your other suggestions about getting rid of .length are also prudent.

LeeTibbert:
Yes, back to the Present.  Two days ago I had convinced myself that the fragment should
be _retained_.  Let me start over and see if I can retrace and confirm the reasoning that
lead me there (and the maximum downside of both retaining and removing it).

Your diagrams might help me walk thru the possibilities.
```
if (shared) {
    value = value.clone();
    shared = false;
}

```

LeeTibbert:
After clearing my mind & heart, studying the code & your diagrams, and going for a short
walk to let the back brain work, I believe that my final opinion is that `clone` section
can and should be deleted.

Is that your current understanding and direction?

Let me sketch out what I think a test might look like and see what you think.
Something like
* allocate a StringBuilder

* add some text

* call sb.toString() and capture result1

* append some text to the SB.

* call String(result1) and capture result2

* are the contents of result1 and result2 the same?
   I do not remember if String has a .sameContents,
   if it does, I would probably be a chicken and use that.
   If not then .hashCode() should work  "result1 == result2" 
   _should_ work, (and probably uses .hashCode() underneath)
   but I have been hosed too many times by quirks in "==".

This uses the String(string) constructor I/we were concerned 
above to test if the append() of this PR changes the content
of a previously instantiated String from sb.

What do you think?

david-bouyssie:
> After clearing my mind & heart, studying the code & your diagrams, and going for a short walk to let the back brain work, I believe that my final opinion is that `clone` section can and should be deleted.
> 
> Is that your current understanding and direction?

Yes it is, just be anxious to adopt a solution different to Android Luni's one 😅

>Let me sketch out what I think a test might look like and see what you think.

I would say time to code 😄

More seriously, there is no `.sameContent()` method, but something very similar: `def contentEquals(cs: CharSequence)`.
It looks like a good test to me, thanks.

david-bouyssie:
re: code example "Constructs a new string with the same sequence of characters"

I did a mistake there. I copy pasted code from my current Sandbox, and this doesn't match current SN code, which is:
```
  def this(string: _String) = {
    this()
    value = string.value
    offset = string.offset
    count = string.length()
  }
```

We could either:
* replace `string.length()` by `string.count` to ease code understanding (correctness should be identical).
* adopt the code previously shown

It implies two different behaviors when doing something like `val s2 = new String(s1)` :
* in the first case, it means that s2 is another wrapper of `s1.value`. What are typical a use cases for that?
* in the second case, the backing Array is copied in a new one if wastage (unused chars) is observed in s1. So it leads to backing Array compaction at the cost of a new allocation (see the extended discussion about this topic in PR #2908).

Javadoc says:
_Initializes a newly created String object so that it represents the same sequence of characters as the argument; in other words, the newly created string is a **copy** of the argument string._

How a **copy** should be specified/satisfied?

LeeTibbert:
| How a copy should be specified/satisfied?

This constructor definition is somewhat strange.
It hints, I think, at JVM implementation details (i.e. "wasteage" under the hood).

I think we have to check the other String() constructors
which pass in an Array.  Can their 'count' ever be less than
their 'length"?   

The conservative approach is to _always_  copy the String argument.

Now all of the discussion below depends upon no other _String constructor
being able to create Strings where the count != length.  There must be
someplace, or else why does Java have this constructor.


IIUC, java {Double, Float} .toString, and soon various ways to call String{Builder, Buffer}
can generate __Strings with unused/unseen tails. Those are  candidates for physical compacting
but are already "logically" compact.

I concur that the "replace string.length() by string.count to ease code understanding" is
correct.  It could stand a comment about "logical" & physical compaction.

My suggestion is to do the "always copy" and create an Issue that there is a possible
opportunity for optimization there and possibly give proposed code.   That allows
the bugs (or lack thereof, hopefully) of this and closely related PRs to shine through.
Yes, the chicken approach. It also allows time to study the Java constructors and
and discover/appreciate the intent of the Java developers having this constructor.

I understand entirely and support your decision if you want to make  the first `string.count()`
alternative (with one line comment) you posted just prior.

david-bouyssie:
I checked the different String constructors. Here are the ones, I think, that require focus:
image

I added three symbols:
* the green tick means that the corresponding constructor is doing a copy of input date in a fresh/compact backing Array.
* the yellow exclamation mark means that the value is shared, but in a way that should not break String immutability
* the red cross means that the current implementation is breaking String immutability and should be changed (see issue https://github.com/scala-native/scala-native/issues/2925)

Note: I have not included the internal constructor used to wrap a shared backing Array without doing copy, things should be under control there.

Thus I'm not really convinced that the _always copy_ conservative way is required. Even if the issue #2908 was merged today,  I think it would still let the symbols above in the same state.
Consequently, I would prefer to:
* just do the `string.length` -> `string.count` change and fix the issue #2908.
* add specific tests in PR #2908 that would ensure consistency for all possible String construction kinds.

LeeTibbert:
Nice This is the kind of analysis & completeness needed, for now & future reference.

* "The conservative approach is to always copy the String argument."  Sorry I had meant
   that in the context of what, correctly, became #2925.  You have convinced me that
   "copy" is not need in the context of this PR.

* "that would ensure consistency for all possible String construction kinds."  A bit of work but
  exactly what is needed.

You  are probably _way_ ahead of me.
   
`StringBuilderTest` could have one or two append tests similar to what you did
 a day or so about appending Doubles/Floats.  These naturally live in the PR.

I think the "string constructor is immutable" tests` would live in StringTest.scala
(or such).   That is where I would look for them if I ever  had cause to wonder.
Depending on how you want to do things, "constructor immutability" could be another Issue & 
its eventual PR.   Yes, it would add confidence if that was available after this PR ensemble
settles but there is a strong reason for separation of concerns.

david-bouyssie:
>Nice This is the kind of analysis & completeness needed, for now & future reference.

Thanks.

>You are probably _way_ ahead of me.

Not that much, I'm just waking up some hours before, so get more time for ideas digestion generation/digestion 😄.

>StringBuilderTest could have one or two append tests similar to what you did a day or so about appending Doubles/Floats. These naturally live in the PR.

Will do so tomorrow, this is the way.

>I think the "string constructor is immutable" tests` would live in StringTest.scala
(or such). That is where I would look for them if I ever had cause to wonder.
Depending on how you want to do things, "constructor immutability" could be another Issue &
its eventual PR. Yes, it would add confidence if that was available after this PR ensemble
settles but there is a strong reason for separation of concerns.

Actually I think that "constructor immutability" tests have to be treated either before or within PR #2908.
I think the for fix for the issue #2925 should come first. In the PR fixing this one we could introduce additional paranoid unit test, that could serve of proofs of robust String immutability.
Once this get merged, I will have to rebase the PR #2908 on top of that, to be 100% sure that everything is fine.
It looks like the roadmap has been refined...

LeeTibbert:
| before or within PR https://github.com/scala-native/scala-native/pull/2908

PR #2908 is already a 400Kg gorilla, a nice 400Kg gorilla, but still 400Kg.

My suggestion would be that either you or I create a  "string immutability" Issue
and then the PR.  I do not mind doing that, if it allows you to maintain your focus.

The reason that I suggest this is that having some confidence that the String constructors
are indeed immutable means that anything you/we do that breaks that becomes easier
to detect.  Actually, what I/we want to do is be able to assert the opposite: i.e. that 
we did not break anything.

The PR might take a while to get merged.  If we can get away with it touching StringTest.scala
but not String.scala, it should not block any of the PRs in this stream.  We would have the confidence
of the tests and could use them privately whilst they moved their way thru the PR review queue.

It may make more sense for me to take a run at Issue #2925. A fix there will have to touch StringTest.
The issue itself is also closer to our present concerns.

I think you might apply for the title "Grand Marshall" if you/we manage to get this all sequenced, with
good character.

david-bouyssie:
Commit https://github.com/scala-native/scala-native/pull/2909/commits/a1ea82b236df39c4ab011b8e58516344b92a6236 contains my last wanted modifications for this PR.

I hope reviewer will agree 🤞

david-bouyssie:
Sorry for the title of the last commit, I just realized it could be better...

LeeTibbert:
@WojciechMazur  Sorry to disturb.  
If you can create a moment, could you please release this to CI.

This LGTM, modulo CI issues.  

Thank you.

LeeTibbert:
| Another option would be to wait for a rebase on PR https://github.com/scala-native/scala-native/pull/2929, once it get merged.

Given that we can run the tests of PR #2929 locally until it gets merged, and we know it is coming, I think that
not waiting & merging this one first is the better way to go.  That would get one whole `kit & caboodle` nailed
down.

WojciechMazur (approved):

PR #2939, opened 2022-10-24
Fix CI deprecation warnings and multiach builds

WojciechMazur:
* Updated actions/cache v2 -> v3
* Removed usages of deprecated `set-output` commands
* Replaced ammonite script with scala-cli runner due to deprecated node 12 setup action
* Fixed issues with building multiarch containers: refactored testing process, now using fast i386/ubuntu images for linux-x86 builds due to failures in building Boehm GC on that platform

There is still 1 remaining deprecation in Windows jobs - we would need to wait for new release of https://github.com/al-cheb/configure-pagefile-action or reimplement setting up pagefile size limit from scratch.


ekrich:
What was the general problem with the x86 Boehm compile issue?

WojciechMazur:
> What was the general problem with the x86 Boehm compile issue?

In summary makefile of Boehm GC was skipping linking of the final library. I'm not sure why it happend. It worked completely fine on arm64 platform and locally on amd64. It's strange because we have not introduced any changes to the build process. The only explanation might have been changed behavior in preinstall binaries in GitHub actions, however, it seems unlikely as well. 

Anyway, after spending way too much time on trying to fix that, I've decided to find an alternative that would not involve building Boehm GC. I've found out that multi-platform images for ubuntu work quite well. 
I wanted to move arm64 builds as well to arm64-v8/ubuntu images but they worked too slow when emulated. So finally i386 are build in i386 platform, while arm64 is cross-build on amd64 for fast compilation and tested in emulator of arm64.

PR #2949, opened 2022-10-27
Fix encoding of `scala.Nothing` and `scala.Null` in type signatures

### LINKED ISSUE #2858: Linking error in Scala 2: `Not found Top(scala.Nothing)`
Porting ZIO to Scala Native we observe this linking error in Scala 2 (Scala 3 links fine).
I couldn't reproduce without using [izumi-reflect](https://github.com/zio/izumi-reflect).

This reproduction fails to link on Scala.

```scala
//> using platform "scala-native"
//> using scala "2"
//> using lib "dev.zio::izumi-reflect::2.2.0"

import izumi.reflect._

object Main {
  def foo[T: Tag] = ???
  def main(args: Array[String]): Unit = foo
}
```

There is a workaround in izumi-reflect around a scalac bug with `Nothing`
https://github.com/zio/izumi-reflect/blob/45d34c1ab03f040d98ab38464c573ff45a97fef5/izumi-reflect/izumi-reflect/src/main/scala-2/izumi/reflect/TagMacro.scala#L42
Maybe it can be related?

### LINKED ISSUE #2920: Spurious "missing definitions while linking"
When trying to run ZIO tests, there are some issues when linking. They are very suspicious, most likely wrong:
```
[error] Found 1 missing definitions while linking
[error] Not found Top(scala.Nothing)
[error] 	at /home/runner/work/zio/zio/test-tests/shared/src/test/scala/zio/test/mock/PolyMockSpec.scala:15
```
```
[error] Found 1 missing definitions while linking
[error] Not found Top(scala.Nothing)
[error] 	at /home/runner/work/zio/zio/test/shared/src/main/scala/zio/test/mock/MockRandom.scala:33
```
https://github.com/zio/zio/actions/runs/3283699177/jobs/5408751779#step:7:687 in the PR https://github.com/zio/zio/pull/7031

/cc @lolgab

WojciechMazur:
fixes #2858 
fixes #2920 

`scala.Nothing` and `scala.Null` were not encoded as `scala.runtime.Nothing$` and `scala.runtime.Null$` leading to missing definitons when linking.

### DIFF
--- a/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenName.scala
+++ b/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenName.scala
@@ -31,6 +31,8 @@ trait NirGenName[G <: Global with Singleton] {
       if (fullName == "java.lang._String") "java.lang.String"
       else if (fullName == "java.lang._Object") "java.lang.Object"
       else if (fullName == "java.lang._Class") "java.lang.Class"
+      else if (fullName == "scala.Nothing") "scala.runtime.Nothing$"
+      else if (fullName == "scala.Null") "scala.runtime.Null$"
       else fullName
     }
     val name = sym match {
--- a/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenName.scala
+++ b/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenName.scala
@@ -163,7 +163,9 @@ object NirGenName {
     "java.lang._Object" -> "java.lang.Object",
     "java.lang._String" -> "java.lang.String",
     "java.lang.annotation._Retention" -> "java.lang.annotation.Retention",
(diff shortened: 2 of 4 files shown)


ekrich:
I am sure many people have wondered but I haven't seen anything documenting why `java.lang.Object` and `java.lang.Class` have underscore prefixes - maybe to de-conflict from the loaded Java classes?

WojciechMazur:
> I am sure many people have wondered but I haven't seen anything documenting why `java.lang.Object` and `java.lang.Class` have underscore prefixes - maybe to de-conflict from the loaded Java classes?

That's exactly the case. Otherwise, inside the scala compiler, we would either get incorrect symbols or we could crash the compiler (mostly in Scala 3, eg. with j.l.Serializable) if it depends on some symbols defined in the Java stdlib. In such cases we need to define a Scala class with a different name and transform it inside upon compilation.

PR #2974, opened 2022-11-08
Add JDK9 constructors to `j.m.BigInteger`

armanbilge:
@plokhotnyuk uses these in jsoniter-scala.

https://github.com/plokhotnyuk/jsoniter-scala

### DIFF
--- a/unit-tests/shared/src/test/scala/javalib/math/BigIntegerTest.scala
+++ b/unit-tests/shared/src/test/scala/javalib/math/BigIntegerTest.scala
@@ -1,6 +1,7 @@
 package javalib.math
 
 import java.math._
+import java.util.Arrays
 
 import org.junit.Test
 import org.junit.Assert._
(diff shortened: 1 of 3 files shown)


unit-tests/shared/src/test/scala/javalib/math/BigIntegerTest.scala
```diff
@@ -175,4 +176,149 @@ class BigIntegerTest {
 
     assertFalse(jbi1 == jbi2)
   }
+
+  // ported from Scala.js commit cbf86bb dated Oct 23 2020
```
armanbilge, line 180: @ekrich you mean this comment?

ekrich: I didn't see that down there. I guess we had always put them on the top based on the latest port. I had to put some comments in some of the code I ported as there was some Scala Native specific code so it wasn't a straight port. Added for JDK 9+ might be helpful there.

armanbilge: I added it to make it clear where the Scala.js ported test came from.

> Added for JDK 9+ might be helpful there.

I wrote those tests myself. Scala.js does not support these APIs and thus has no tests for them.

ekrich: Sounds good, thanks. It seems sometime it should be cut and dry but is messier in practice.

WojciechMazur (approved):

PR #2998, opened 2022-11-15
Cleanup incremental compilation

WojciechMazur:
* IncCompilationContext was used only in CodeGen - remove it's declarations in other scopes, move to scalanative.codegen package and rename to `IncrementalCodeGenContext` 
* Cleanup the inc codegen context code
* Remove spurious changes in generated files after Lowering phase - make sure `GenerateReflectiveProxies` produces deterministic output.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -74,20 +74,10 @@ object Build {
         linked
       }
 
-      implicit val incCompilationContext: IncCompilationContext =
-        new IncCompilationContext(fconfig.workdir)
-      if (fconfig.compilerConfig.useIncrementalCompilation) {
-        incCompilationContext.collectFromPreviousState()
-      }
-
       // optimize and generate ll
       val generated = {
         val optimized = ScalaNative.optimize(fconfig, linked)
         ScalaNative.codegen(fconfig, optimized)
-      }
-
-      if (fconfig.compilerConfig.useIncrementalCompilation) {
-        incCompilationContext.dump()
       }
 
       val objectPaths = fconfig.logger.time("Compiling to native code") {
@@ -112,9 +102,6 @@ object Build {
         }
 
         libObjectPaths ++ llObjectPaths
-      }
-      if (fconfig.compilerConfig.useIncrementalCompilation) {
-        incCompilationContext.clear()
       }
 
       // finally link
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -61,68 +61,57 @@ object Build {
       if (Files.notExists(config.workdir)) {
         Files.createDirectories(config.workdir)
       }
+      // validate classpath - use fconfig below
+      val fconfig = {
+        val fclasspath = NativeLib.filterClasspath(config.classPath)
+        config.withClassPath(fclasspath)
+      }
 
-      if (canSkipBuild(config)) {
-        config.logger.debug(
-          "Skipping Scala Native build - inputs and configuration have not changed."
-        )
-        config.artifactPath
-      } else {
-        // validate classpath - use fconfig below
-        val fconfig = {
-          val fclasspath = NativeLib.filterClasspath(config.classPath)
-          config.withClassPath(fclasspath)
+      // find and link
+      val linked = {
+        val entries = ScalaNative.entries(fconfig)
+        val linked = ScalaNative.link(fconfig, entries)
+        ScalaNative.logLinked(fconfig, linked)
+        linked
+      }
+
+      // optimize and generate ll
+      val generated = {
+        val optimized = ScalaNative.optimize(config, linked)
+        ScalaNative.codegen(config, optimized)
+      }
+
+      val objectPaths = fconfig.logger.time("Compiling to native code") {
+        // compile generated LLVM IR
+        val llObjectPaths = LLVM.compile(fconfig, generated)
+
+        /* Used to pass alternative paths of compiled native (lib) sources,
+         * eg: reused native sources used in partests.
+         */
+        val libObjectPaths = scala.util.Properties
+          .propOrNone("scalanative.build.paths.libobj") match {
+          case None =>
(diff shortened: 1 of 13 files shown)


tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -144,4 +140,35 @@ object Build {
         NativeLib.compileNativeLibrary(config, linkerResult, nativeLib)
       )
   }
+
+  /** Check if build can be skipped. This function returns true only when inputs
+   *  of the build has not changed since last compilation. It means that all
+   *  files on the classpath were not modified after building last artifact and
+   *  the cached checksum of config is equal to the checksum of current config.
+   */
+  private def canSkipBuild(config: Config): Boolean = {
```
ekrich, line 149: I think this could be a little bit aggressive and could miss if native files e.g. `.c` have changed. Nativelibs are found from the classpath and the checksums for classpath copied directory or jar extracted files are checked and if the checksum has changed then the files are overlayed. This makes sure that you can change, add, or remove native files in the scala_native directory. I think this check takes about 100ms when I was working on this recently.

WojciechMazur: I've made some local tests and there is no problem with the detection of changes to native sources. 
I think that the current checks does not conflict with current hash checks in `NativeLib.unpackNativeCode`. 

If the native file would change the fact of this event would lead to producing a new, modified jar/directory with an updated modification time attribute. It should be cheap to compare. If non of the files on the classpath were modified then hash their content cannot change as well (especially since we're working on a copy there).
Otherwise, if any of files on the classpath was modified, then the build would not be skipped. This would lead to the standard procedure of copying and comparing jars/dirs you described. 

The checksum of config would detect if a new file was added to the classpath, but does not check its content, all of this should be relatively cheap and is based on the fact that hashCode of string value is deterministic and would not change between JVM runs. 

When testing however i found a 1 corner case in finding nativelibs: let's assume we have following file structure: 
```
sandbox/src/main/
├── resources
│   └── scala-native
│       └── testlib.c
└── scala
    └── Test.scala
```
At first compilation it would work correctly. However, if we remove the testlib.c some unexpected thing would happend.
```
sandbox/.2.13/target/scala-2.13/native/
├── native-code-clib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-3
├── native-code-clib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-4
├── native-code-javalib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-1
├── native-code-javalib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-2
├── native-code-nativelib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-0
├── native-code-nativelib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-1
├── native-code-posixlib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-2
├── native-code-posixlib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-3
├── native-code-sandbox_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-0
├── native-code-windowslib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-4
├── native-code-windowslib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-5
```
Because `sandbox/resource/scala-native`  is now empty it was filtered out. However, the previous `native-code-sandbox` directory is still present and would be compiled, and though would still allow to link final application even though native code was removed.
 Also because we enumerate all the files based on their idx on the classpath we would duplicate all of the remaining native-code directories.

ekrich: So this does indeed seem to be a corner case that is pretty ugly.
> At first compilation it would work correctly. However, if we remove the testlib.c some unexpected thing would happened(sic).
The code looks for the `scala-native` dir and files inside but it also doesn't remove it if it is empty. I am not sure why the directories are getting re-expanded with new numbers but perhaps each should be assigned with a hash of the classpath entry vs a sequential number. The sequence number could be because perhaps different classpath items could have duplicate names coming from different places. A comment would have been nice, right?

So overall there are a couple of problems being worked on at the same time. I was working on how to fix all the issues we have in the backlog due to our Config discovery which doesn't work correctly. I came to the conclusion that we need to cache the config the first time so all the discovery is done only once. The `sbt` process loads our code and it stays in memory until the next `reload` or restart. No setting or the classpath can change until that happens. I have run into a few fits and starts with that code as the mutable settings setup I had didn't work well and there are quite a few things going on with discovery that are actually quite tricky and keeping the current immutable builder style seems like a good plan.

Caching between restarts is another angle which I haven't thought about. So far we have changed things so each phase of the build can pass files on the the next and then the next phase can decide what to do which is good because then we can use local reasoning in each phase and avoid work that doesn't need to be done. I did move around the timing stuff a bit in the build so that unrelated items were not getting timed together. It might be good to look at that so we can find where the time is being spent.

I think it would be better to just do the incremental compilation and put the caching off for a later change. I was going to look into why all the work was being done by the different phases when nothing has changed but hadn't gotten that far. I think we should be careful how far we go down this road as we are treading close to build tool features. It basically is one; but I think then I think we have to carefully consider the invalidation and clean up as you have shown above in the native lib area. For example, when I change the build code I `reload` the build and then `clean` to make sure I have a clean slate. This might not be too much to ask - almost everywhere that is needed.

I do think that we could do certain things like a classpath hash that can be used by the different phases in the build similar the checking if the artifactPath file exists or not.

ekrich: Looking at this again I think it is ok. If we decide to change the build to cache the config then maybe we won't have to hash the config unless there are things we want to cleanup and we could do that on a case by case basis like for classpath changes between runs or whatever else makes sense.

The classpath native code removal seems sane as well if we want to manage this so a clean does not need to be called. That may save us some figuring out when a person runs into it - the empty native dir case anyway. We could maybe even change the "something changed, refresh the whole native lib" too at some point.

PR #3001, opened 2022-11-16
Fix incorrect declarations of extern methods defined in `extern trait`

WojciechMazur:
Extern declarations produced by method declarations in extern traits were producing incorrect type signature containing boxed types. This PR fixes this behavior.
Fixes compilation errors present in #2996


ekrich:
Thank-you. I rebased and pushed the PR mentioned above.

PR #3009, opened 2022-11-23
Allow to calculate real sizeOf class based on `MemoryLayout`

WojciechMazur:
* Adds `runtime.Intrinsic.sizeOf[T]` methods evaluated to NIR `Ops.SizeOf(global)`
* Compile time against `sizeOf[T]` where T is a trait  - there is no memory layout for traits
* Adapts MemoryLayout to use FieldLayout of class as a reference to class size
* Make `is32BitPlatform` information an implicit context stored in `PlatformInfo` object

`class Foo()` would yield size of 8 - HeaderSize 8bytes, no fields
`class Foo(x: Int)` would yield size of 16 = HeaderSize + 4 bytes field x + 4 bytes padding
`class Foo(x: Int, y: Int)` would yield size of 16 = HeaderSize + 2 fields 4 bytes each
```
class Foo(x: Int, y: Int)  -> sizeOf = 16
class Bar(x: Foo) -> sizeOf = 16 (inner fields of reference types are still just a pointer, they always count as sizeOf[Ptr] 
```

Current solution might be used to replace `unsafe.sizeOf[T: Tag]`, however it has a different semantics in terms to reference kinds

It's worth mentioning that size of class might change between compilations - any unused field can be removed and though reduce the final size of class. This assumption was presented in the added `IntrinsicsTest` forcing usage of all class fields.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
@@ -26,7 +26,6 @@ object Lower {
     import meta._
 
     implicit val linked: Result = meta.linked
-    val is32BitPlatform = meta.is32BitPlatform
 
     val Object = linked.infos(Rt.Object.name).asInstanceOf[Class]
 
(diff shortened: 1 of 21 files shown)


tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -758,8 +757,8 @@ object Lower {
         pos: Position
     ): Unit = {
       val Op.Sizeof(ty) = op
-
-      val memorySize = MemoryLayout.sizeOf(ty, is32BitPlatform)
+      val memorySize = MemoryLayout.sizeOf(ty, Some(meta))
+      println(s"genSizeOf: $ty=${memorySize}")
```
yawen-guan, line 761: Seems forget to delete the println here.

PR #3035, opened 2022-12-13
Generate scaladoc for javalib in Scala 2

ekrich:
Interestingly I get the following error using `javalib2_12/doc` but not `javalib2_12/compile`.
```Scala
[error] /Users/eric/workspace/scala-native/javalib/src/main/scala/java/lang/process/UnixProcessGen2.scala:253:42: type mismatch;
[error]  found   : Int
[error]  required: scala.scalanative.posix.poll.pollEvent_t
[error]     (which expands to)  Short
[error]     (fds + 0).events = pollEvents.POLLIN | pollEvents.POLLRDNORM
```
The LHS is a `CShort` or `Short` which can be fixed by adding `( ... ).toShort` but the error does not occur using 2.13. Instead we get the following which may be the unicode issue?
```scala
[info] Main Scala API documentation to /Users/eric/workspace/scala-native/javalib/.2.13/target/scala-2.13/api...
[error] stack trace is suppressed; run last javalib2_13 / Compile / doc for the full output
[error] (javalib2_13 / Compile / doc) java.nio.charset.MalformedInputException: Input length = 1
```
Scaladoc 3 generation has the tasty issue but that could be because Scala 3.1.3 is broken?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Getting Started and full documentation can be found at [https://www.scala-native
 ## Online Scaladoc
 
 [![Scaladoc nativelib](https://javadoc.io/badge2/org.scala-native/nativelib_native0.4_3/javadoc.svg?label=nativelib)](https://javadoc.io/doc/org.scala-native/nativelib_native0.4_3)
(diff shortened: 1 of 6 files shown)


WojciechMazur:
I've opened an issue for Scala 3 problem with scaladoc: https://github.com/lampepfl/dotty/issues/16709

PR #3037, opened 2022-12-14
Support new Scala3 lazy vals enabled by -Ylightweight-lazy-vals flag

WojciechMazur:
This PR adds support for lightweight lazy vals and it's targeting dotty-experimental branch. 
This change can be backported to the main branch upon release of Scala 3.2.2 (main branch uses only stable releases, -Ylightweight-lazy-vals is available since 3.2.2-RC1). 
The new lazy vals showed the incorrectness in code generation of try-finally blocks which are using linktime if conditions - finally block could have been visited more then once.  

* Scala-next projects are now always using `-Ylightweight-lazy-vals` flag 
* Fix issue with multiple executions of finally block
* Provide a stub implementation for `java.util.concurrent.CountDownLatch`  
* Upgrade experimental dotty version to latest nightly

### DIFF
--- a/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
+++ b/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
@@ -580,6 +580,16 @@ class IssuesTest {
     free(null)
   }
 
+  @Test def `can initialize lazy vals using linktime if`() = {
+    object Foo {
+      val fooLiteral = "foo"
+      lazy val fooLazy =
+        if (scala.scalanative.meta.LinktimeInfo.isWindows) fooLiteral
+        else fooLiteral
+    }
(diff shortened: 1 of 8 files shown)


unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
```diff
@@ -580,6 +580,16 @@ class IssuesTest {
     free(null)
   }
 
+  @Test def `can initialize lazy vals using linktime if`() = {
```
ekrich, line 583: Could this just be toggle case? I ask because I spent a bunch of time on Scala.js normalizing all the tests. The only exception is `_Issue341` as a postfix. See https://github.com/scala-js/scala-js/pull/4266

PR #3039, opened 2022-12-16
Add support for marking objects using custom GC roots

WojciechMazur:
This PR adds API allowing to mark range of memory addresses as a roots of GC. This allows to mark objects referenced only from memory outside GC heap. 

Current API `GC.addRoots(addrLow: Ptr[_], addrHigh: Ptr[_])` and `GC.removeRoots(addrLow, addrHigh)` is based on `GC_add_roots` and `GC_remove_roots` present in the Boehm GC API


nativelib/src/main/resources/scala-native/gc/immix_commix/GCRoots.c
```diff
@@ -0,0 +1,42 @@
...
+    while (current != NULL) {
+        if (AddressRange_Contains(range, current->range)) {
+            bool detached = false;
+            if (prev == NULL) {
+                *head = current->next;
+                GC_Roots *expected = NULL;
+                detached = atomic_compare_exchange_strong(
+                    (_Atomic(GC_Roots *) *)head, &expected, current->next);
+            } else {
+                prev->next = current->next;
```
wks, line 30: There is a risk that if two threads try to concurrently remove items from the singly linked list, one of them will access freed memory on this line.

Suppose the singly linked list has the following nodes: head -> A -> B -> C -> D;

Thread 1 wants to remove C.  It calls `GC_Roots_RemoveByRange`. `current` is now pointing to C and `prev` is pointing to B. Thread 1 is just about to execute line 30, but context switching happened.

Then thread 2 starts executing. It wants to remove B.  It called `GC_Roots_RemoveByRange` and it quickly finished.  Now the linked list looks like: head -> A -> C -> D, while B is freed.

Then thread 1 wakes up, and continues to execute line 30.  Now `prev` already points to freed memory.  The atomic compare-exchange doesn't help because no matter whether it is successful or not, some freed memory have already been overwritten.

I guess the "custom GC roots" feature is for internal use for the ScalaNative runtime.  In this case, I think it will not be called very often.  It should be OK to use mutex for synchronisation.

WojciechMazur: Thank you for this feedback and for bringing awareness. I agree that it might be a real issue and it would require improvements. The main purpose of custom GC roots was CaptureCalculus SafeZones experiments added in https://github.com/scala-native/scala-native/pull/3120. In the current implementation, it would add/remove gc roots once per usage of the SafeZone block in the code. 

I'll convert this comment into the dedicated issue

PR #3054, opened 2022-12-27
docs: deprecate posix.getopt as it isn't posix 2018

i10416:
`getopt` is no longer part of POSIX 2018 and should be moved to unistd.scala.

references #3052

### DIFF
--- a/docs/lib/posixlib.rst
+++ b/docs/lib/posixlib.rst
@@ -186,7 +186,7 @@ C Header          Scala Native Module
 .. _scala.scalanative.posix.fenv: https://github.com/scala-native/scala-native/blob/main/clib/src/main/scala/scala/scalanative/libc/fenv.scala
 .. _scala.scalanative.posix.float: https://github.com/scala-native/scala-native/blob/main/clib/src/main/scala/scala/scalanative/libc/float.scala
 .. _scala.scalanative.posix.fnmatch: https://github.com/scala-native/scala-native/blob/main/clib/src/main/scala/scala/scalanative/libc/fnmatch.scala
-.. _scala.scalanative.posix.getopt: https://github.com/scala-native/scala-native/blob/main/posixlib/src/main/scala/scala/scalanative/posix/getopt.scala
+.. _scala.scalanative.posix.getopt: https://github.com/scala-native/scala-native/blob/main/posixlib/src/main/scala/scala/scalanative/posix/getopt.scala [#getopt_no_longer_posix_2018]_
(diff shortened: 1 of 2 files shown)


ekrich (commented):
Do you plan to follow this up with a PR to fix the `unistd` code as well? If not you could consider updating that code as well.

- Change `optarg` and friends to vars and remove `@name` annotations.
- Remove the corresponding functions from `unistd.c`.
- Not sure if the extern `var`s are being used for linking or not with the current `getopt` but if it isn't remove them.
- Add some release notes `0.5.0.md` about the deprecation and using `unistd`.

Thanks for working on this. 

cc/ @WojciechMazur I am pretty sure this is not binary compat so it would be great to get an opinion if this PR as is should target `0.4.x` and then the one I describe about should be targeted for `0.5.0`.

i10416:
Thank you for guidance.
I will follow up this PR in https://github.com/scala-native/scala-native/pull/3048

i10416:
I think deprecation notice should be announced in 0.4.next.md and removal should be announced in 0.5.0.md.

WojciechMazur (approved):

WojciechMazur:
I agree, the plan to deprecate in 4.x and remove in 5.x seems good to me

posixlib/src/main/scala/scala/scalanative/posix/getopt.scala
```diff
@@ -3,6 +3,9 @@ package posix
 
 import scalanative.unsafe._
 
+@deprecated(
+  "getopt is no longer part of POSIX 2018 and will be moved to unistd"
```
i10416, line 7:
```suggestion
  "getopt is no longer part of POSIX 2018 and will be removed. Use unistd instead"
```

Ah, I found getopt-related stuffs already in unistd. `will be moved to unistd` is not correct.

docs/lib/posixlib.rst
```diff
@@ -241,7 +241,8 @@ C Header          Scala Native Module
                 specification because Scala Native supports only
                 passing structures by reference.  See code for details
 		and usage.
-
+.. [#getopt_no_longer_posix_2018] getopt.scala, introduced in PR `Fix #202: Support assignment to extern variables #1348 <https://github.com/scala-native/scala-native/pull/1348/>`_ 
+                                  is no longer part of POSIX 2018 and will be moved to unistd.scala in the future.
```
i10416, line 245:
```suggestion
                                  is no longer part of POSIX 2018 and will be unified to unistd.scala in the future.
```

i10416:
> I will follow up this PR in https://github.com/scala-native/scala-native/pull/3048

Hmm, #3048  is not originally meant to remove deprecated getopt.scala. I will create another PR to drop getopt.scala.

PR #3060, opened 2022-12-29
chore: use assertEquals instead of assert

i10416:


i10416:
By the way, I wonder why `==` is used instead of ScalaTest's `===` operator in tests. Actually, I think this makes little difference. Perhaps, is there plan to drop ScalaTest in the future?

WojciechMazur:
> Perhaps, is there plan to drop ScalaTest in the future?

Actually there no such plans. I think that in most cases `==` was used just for convenience

WojciechMazur (approved):

PR #3061, opened 2022-12-29
Make initialization of `System.properties` a lazy operation

WojciechMazur:
Currently, System.properties are being initialized in the body of `System` constructor, even though typically they are not used. 
These changes delay this operation. 
Early initialization of System properties also caused multiple issues in initialization of cyclic dependencies  - mostly due to trasnsitive dependencies on other methods defined in `java.lang.System`


ekrich (approved):
LGTM.

PR #3062, opened 2022-12-29
Handle whitespaces in the files passed to the clang when linking

WojciechMazur:
Config file passed to the LLVM when linking was never escaping whitespace. It could have lead to error when linking. Now whenever any option passed to the llvm config file contains whitespace it would be escaped by using double-quotes

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -181,9 +181,11 @@ private[scalanative] object LLVM {
       val pw = new PrintWriter(configFile)
       try
         llvmLinkInfo.foreach {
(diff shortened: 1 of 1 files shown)


ekrich (commented):
LGTM :-)

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -308,4 +310,9 @@ private[scalanative] object LLVM {
     if (config.targetsWindows) Nil
     else Seq("-fPIC")
 
+  private def escapeWhitespaces(str: String): String = {
```
ekrich, line 313: I don't think the {}'s are needed here so you could make it Scala 3 style.

PR #3082, opened 2023-01-09
tidy: replace apply() with empty to avoid allocation

i10416:
extract the lowest-hanging fruits from  #3067; just small micro-optimization

I feel this kind of chores should be done by Scalafix🤔


i10416:
could someone re-run the test that failed due to 503  Service Unavailable?

WojciechMazur (approved):

PR #3097, opened 2023-01-17
Refactor javalib test in to proper packages

ekrich:
Other improvements:

- Test run twice 5900 total -> 5878 total now
- Regex tests now run on JVM, some assumeFalse, was 4892 total, skip 42 -> 4967 total, skip 64 now 

ArrayListTest is the only test not moved - maybe a candidate for Scala.js test port.


ekrich:
> Could you elaborate just a bit about the "5900 - 5878" part? It is not face evident, or at least not
to my face. Are those line numbers in a log file or something. The "Test run twice" seems clear.

The tests where in 2 places and they were merged.

> When you say "Regex tests" do you mean the javalib or the RE2 or both?
It makes sense to run the javalib tests and running as many as sensible
on JVM is an advancement.

The javalib ones are now in shared. They were in native so they couldn't be run via
```
sbt> testsJVM3/testOnly org.scalanative.testsuite.javalib.util.regex.PatternTest
sbt> testsJVM3/testOnly org.scalanative.testsuite.javalib.util.regex.MatcherTest
```
They are also run in the standard `tests3/test` as well.

ekrich:
I reviewed again so it should be ready to go.

WojciechMazur (approved):

PR #3102, opened 2023-01-23
Port `j.u.TreeMap` and friends

armanbilge:
Syncs us up with some recent work in Scala.js Javalib. All credit to Tobias for their Scala.js PRs!


unit-tests/shared/src/test/scala/utils/AssertThrows.scala
```diff
@@ -43,4 +43,8 @@ object AssertThrows {
       }
     )
   }
+
+  def assertThrowsNPEIfCompliant(code: => Unit): Unit = {
+    assertThrows(classOf[NullPointerException], code)
+  }
```
armanbilge, line 49: This is used in many tests but I commented them all out. For some reason they are all segfaulting. I didn't look into it deeper at present, but I will make a follow-up issue if it's not solved by the time this PR merges.

armanbilge: Aha, I understood the problem! It is tricks like this used in the Scala.js code.
```scala
value.getClass() // null check
```
On Native, I think that is segfaulting if `value` is `null`. Is this a bug or expected behavior?

armanbilge: Fixed by using `Objects.requireNonNull(...)` instead.

WojciechMazur (approved):

Issue #3104, opened 2023-01-24
NPE on Console


isomarcte:
I'm getting a NullPointerException when running `console` on a native project on multiple repos. I've tried [cats v2.9.0](https://github.com/typelevel/cats/releases/tag/v2.9.0) and this branch of [idna4s](https://github.com/isomarcte/idna4s/tree/scala-native-bug-report). 

```
[error] java.lang.NullPointerException
[error]         at scala.scalanative.nscplugin.NirGenFile.genPathFor(NirGenFile.scala:28)
[error]         at scala.scalanative.nscplugin.NirGenFile.genPathFor$(NirGenFile.scala:18)
[error]         at scala.scalanative.nscplugin.NirGenPhase.genPathFor(NirGenPhase.scala:14)
[error]         at scala.scalanative.nscplugin.NirGenFile.genPathFor(NirGenFile.scala:15)
[error]         at scala.scalanative.nscplugin.NirGenFile.genPathFor$(NirGenFile.scala:13)
[error]         at scala.scalanative.nscplugin.NirGenPhase.genPathFor(NirGenPhase.scala:14)
[error]         at scala.scalanative.nscplugin.NirGenPhase$NirCodePhase.$anonfun$apply$6(NirGenPhase.scala:150)
[error]         at scala.collection.StrictOptimizedMapOps.map(StrictOptimizedMapOps.scala:28)
[error]         at scala.collection.StrictOptimizedMapOps.map$(StrictOptimizedMapOps.scala:27)
[error]         at scala.collection.immutable.HashMap.map(HashMap.scala:39)
[error]         at scala.scalanative.nscplugin.NirGenPhase$NirCodePhase.apply(NirGenPhase.scala:148)
[error]         at scala.tools.nsc.Global$GlobalPhase.applyPhase(Global.scala:467)
[error]         at scala.tools.nsc.Global$GlobalPhase.run(Global.scala:414)
[error]         at scala.scalanative.nscplugin.NirGenPhase$NirCodePhase.run(NirGenPhase.scala:55)
[error]         at scala.tools.nsc.Global$Run.compileUnitsInternal(Global.scala:1530)
[error]         at scala.tools.nsc.Global$Run.compileUnits(Global.scala:1514)
[error]         at scala.tools.nsc.Global$Run.compileSources(Global.scala:1506)
[error]         at scala.tools.nsc.interpreter.IMain.$anonfun$global$3(IMain.scala:155)
[...]

PR #3105, opened 2023-01-24
Port sbt console warning from Scala.js

armanbilge:
Taken from https://github.com/scala-js/scala-js/blob/ab60f8260be87000dc60aade640ae84247747fc3/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L491-L494

Relates to https://github.com/scala-native/scala-native/issues/3104, but I'm not sure if it actually solves it. Since this PR is only adding a warning to the `console` task, it is not actually re-defining it. So I'm not sure what causes that NPE.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -158,6 +158,14 @@ object ScalaNativePluginInternal {
       }
       .tag(NativeTags.Link)
       .value,
+    console := console
+      .dependsOn(Def.task {
+        streams.value.log.warn(
+          "Scala REPL doesn't work with Scala Native. You " +
+            "are running a JVM REPL. Native things won't work."
+        )
+      })
+      .value,
     run := {
       val env = (run / envVars).value.toSeq
       val logger = streams.value.log


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -158,6 +158,14 @@ object ScalaNativePluginInternal {
       }
       .tag(NativeTags.Link)
       .value,
+    console := console
```
ekrich, line 161: It might be nice if this was pushed to the bottom of these settings. Did it work for you? I tried it in my branch but had some problems but it might be due to other changes. When I type `console` it compiles everything and then fails.

ekrich: Weird, when I try this it just goes to a normal console. I put the code and the bottom of that block but I wouldn't think that would matter.

WojciechMazur: The warning worked for me. Make sure to run this command in the scope of project with enabled ScalaNativePlugin, eg. `sandbox3/console`. When using `console` you would run it in the root project which does not enable this plugin

WojciechMazur (approved):
Thank you for this change. The problem with NPE is due to the usage of virtual (in-memory) abstract files when using repl. I have a fix for that, I would publish a follow up soon

ekrich:
I can pull down the changes now but is this expected?
```
sbt:scala-native> sandbox2_12/console
[warn] Scala REPL doesn't work with Scala Native. You are running a JVM REPL. Native things won't work.
[info] Starting scala interpreter...
Welcome to Scala 2.12.17 (OpenJDK 64-Bit Server VM, Java 11.0.15).
Type in expressions for evaluation. Or try :help.

scala> java.lang.NullPointerException
	at scala.scalanative.nscplugin.NirGenFile.genPathFor(NirGenFile.scala:28)
	at scala.scalanative.nscplugin.NirGenFile.genPathFor$(NirGenFile.scala:18)
```

Issue #3013, opened 2022-11-26
0.4.9 Release Date is Wrong in https://scala-native.org


ShapelessCat:
I'm not sure if I should raise this issue here. It seems the source code of the website https://scala-native.org is not here, but I can't find other place to report this.

I believe the 0.4.9 release date should be 2022-11-23 instead of 2022-12-23, right? The release date of 0.4.9 in https://scala-native.org/en/stable/changelog/0.4.9.html is 2022-12-23.

PR #3106, opened 2023-01-25
Fix the 0.4.9 version release date in its changelog

ShapelessCat:
I raised this issue #3013, and there was a fix #3029. However, that fix is incomplete: the release date in this changelog title was fixed, but the release date inside this changelog has not been fixed.

By the way, the website https://scala-native.org/en/stable/changelog/0.4.9.html has not been updated after the fix #3029.


WojciechMazur (approved):
Thank you for fixing the typo!

PR #3110, opened 2023-01-26
Delay discovery for Build, take 2

ekrich:
The previous attempt to do what was suggested did not well. See https://github.com/scala-native/scala-native/pull/2380

This is a new attempt.

- gets rid of the build logic out of the plugin
- solves a whole bunch of complaints about settings on the build.
- uses delayed discovery and validation but hopefully in a nicer way
- caches the build ~ 200ms the first time to 0ms on subsequent builds

At least one test is failing for me so we will see how it does in CI. Probably many more.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -38,49 +38,6 @@ object ScalaNativePlugin extends AutoPlugin {
     val nativeLink =
       taskKey[File]("Generates native binary without running it.")
 
-    // 10 raw settings for minimal sbt projects
-
-    val nativeClang =
-      settingKey[File]("Location of the clang compiler.")
-
-    val nativeClangPP =
-      settingKey[File]("Location of the clang++ compiler.")
-
-    val nativeCompileOptions =
-      settingKey[Seq[String]](
-        "Additional options are passed to clang during compilation."
-      )
-
-    val nativeLinkingOptions =
-      settingKey[Seq[String]](
-        "Additional options that are passed to clang during linking."
-      )
-
-    val nativeLinkStubs =
-      settingKey[Boolean]("Whether to link `@stub` methods, or ignore them.")
-
-    val nativeMode =
-      settingKey[String](
-        "Compilation mode, either \"debug\", \"release-size\", \"release-fast\", or \"release-full\"."
-      )
-
-    val nativeGC =
-      settingKey[String](
-        "GC choice, either \"none\", \"boehm\", \"immix\" or \"commix\"."
-      )
-
-    val nativeLTO =
-      settingKey[String](
-        "LTO variant used for release mode, either \"none\", \"thin\", or \"full\" (legacy)."
-      )
-
-    val nativeCheck =
-      settingKey[Boolean]("Shall native toolchain check NIR during linking?")
-
-    val nativeDump =
-      settingKey[Boolean](
-        "Shall native toolchain dump intermediate NIR to disk during linking?"
-      )
   }
 
   override def globalSettings: Seq[Setting[_]] =
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -31,7 +31,7 @@ object ScalaNativePlugin extends AutoPlugin {
     }
 
     val nativeConfig =
-      taskKey[build.NativeConfig](
+      settingKey[build.NativeConfig](
         "User configuration for the native build, NativeConfig"
       )
 
@@ -41,23 +41,23 @@ object ScalaNativePlugin extends AutoPlugin {
     // raw settings minimal sbt projects
 
     val nativeClang =
-      taskKey[File]("Location of the clang compiler.")
+      settingKey[File]("Location of the clang compiler.")
 
     val nativeClangPP =
-      taskKey[File]("Location of the clang++ compiler.")
+      settingKey[File]("Location of the clang++ compiler.")
 
     val nativeCompileOptions =
-      taskKey[Seq[String]](
+      settingKey[Seq[String]](
         "Additional options are passed to clang during compilation."
       )
 
     val nativeLinkingOptions =
-      taskKey[Seq[String]](
+      settingKey[Seq[String]](
         "Additional options that are passed to clang during linking."
       )
 
     val nativeLinkStubs =
-      taskKey[Boolean]("Whether to link `@stub` methods, or ignore them.")
+      settingKey[Boolean]("Whether to link `@stub` methods, or ignore them.")
 
     val nativeMode =
       taskKey[String](
@@ -65,20 +65,20 @@ object ScalaNativePlugin extends AutoPlugin {
       )
 
     val nativeGC =
-      taskKey[String](
+      settingKey[String](
         "GC choice, either \"none\", \"boehm\", \"immix\" or \"commix\"."
       )
 
     val nativeLTO =
-      taskKey[String](
+      settingKey[String](
         "LTO variant used for release mode, either \"none\", \"thin\", or \"full\" (legacy)."
       )
 
     val nativeCheck =
-      taskKey[Boolean]("Shall native toolchain check NIR during linking?")
+      settingKey[Boolean]("Shall native toolchain check NIR during linking?")
 
     val nativeDump =
-      taskKey[Boolean](
+      settingKey[Boolean](
         "Shall native toolchain dump intermediate NIR to disk during linking?"
       )
   }
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -35,6 +35,11 @@ object ScalaNativePlugin extends AutoPlugin {
         "User configuration for the native build, NativeConfig"
       )
 
+    val nativeLink =
+      taskKey[File]("Generates native binary without running it.")
+
+    // raw settings minimal sbt projects
+
     val nativeClang =
       taskKey[File]("Location of the clang compiler.")
 
@@ -53,9 +58,6 @@ object ScalaNativePlugin extends AutoPlugin {
 
     val nativeLinkStubs =
       taskKey[Boolean]("Whether to link `@stub` methods, or ignore them.")
-
-    val nativeLink =
-      taskKey[File]("Generates native binary without running it.")
 
     val nativeMode =
       taskKey[String](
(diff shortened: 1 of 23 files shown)


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -31,52 +31,54 @@ object ScalaNativePlugin extends AutoPlugin {
...
+      settingKey[build.NativeConfig](
         "User configuration for the native build, NativeConfig"
       )
 
+    val nativeLink =
+      taskKey[File]("Generates native binary without running it.")
+
+    // 10 raw settings for minimal sbt projects
+
     val nativeClang =
```
WojciechMazur, line 43: All of the settings below were assumed to be deprecated and replaced with NativeConfig. Removing them would significantly simplify the whole initialization - we no longer need to generate nativeConfig, assign parts of it's to settingKeys with default value, and finally read these settings to populate a final nativeConfig

ekrich: Don't we need them for `sbt` synthesized root project settings?

```scala
logLevel := Level.Debug

nativeMode := "release-fast"

nativeLinkStubs := true
```

ekrich:
I have a failure here - must be related but not seeing it.
```
[info] Linking native code (immix gc, none lto) (80 ms)
[info] Total (3009 ms)
[info] - should generate the llvm IR of object A
[info] The test framework
Config(
 - basedir:         /var/folders/f8/vpd97qv1621_b7zt0z957cjc0000gn/T/native-test-out15774223344696575621
 - testConfig:      false
 - workdir:         /var/folders/f8/vpd97qv1621_b7zt0z957cjc0000gn/T/native-test-out15774223344696575621/native
 - defaultBasename: 
 - artifactPath:    /var/folders/f8/vpd97qv1621_b7zt0z957cjc0000gn/T/native-test-out15774223344696575621/out
 - classPath:       List(/Users/eric/workspace/scala-native/scalalib/.2.12/target/scala-2.12/scalalib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT.jar
                     /Users/eric/workspace/scala-native/auxlib/.2.12/target/scala-2.12/auxlib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT.jar
                     /Users/eric/workspace/scala-native/nativelib/.2.12/target/scala-2.12/nativelib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT.jar
                     /Users/eric/workspace/scala-native/javalib/.2.12/target/scala-2.12/javalib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT.jar
                     /Users/eric/workspace/scala-native/posixlib/.2.12/target/scala-2.12/posixlib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT.jar
                     /Users/eric/workspace/scala-native/clib/.2.12/target/scala-2.12/clib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT.jar
                     /Users/eric/workspace/scala-native/windowslib/.2.12/target/scala-2.12/windowslib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT.jar
                     /Users/eric/.sbt/boot/scala-2.12.17/lib/scala-library.jar
                     /Users/eric/.sbt/boot/scala-2.12.17/lib/scala-reflect.jar
                     /var/folders/f8/vpd97qv1621_b7zt0z957cjc0000gn/T/native-test-out15774223344696575621)
 - compilerConfig:  NativeConfig(
 - clang:                  /opt/homebrew/opt/llvm/bin/clang
 - clangPP:                /opt/homebrew/opt/llvm/bin/clang++
 - linkingOptions:         List(-L/usr/local/lib, -L/opt/homebrew/lib, -L/opt/homebrew/Cellar/llvm/15.0.7_1/lib)
 - compileOptions:         List(-I/opt/homebrew/include, -I/opt/homebrew/Cellar/llvm/15.0.7_1/include, -Qunused-arguments)
 - targetTriple:           None
 - buildTarget             Application
 - GC:                     immix
 - mode:                   debug
 - LTO:                    none
 - linkStubs:              false
 - check:                  false
 - checkFatalWarnings:     false
 - dump:                   false
 - asan:                   false
 - optimize                true
 - linktimeProperties:     
 - embedResources:         false
 - incrementalCompilation: true
 - optimizerConfig:        OptimizerConfig(
   - maxInlineDepth: None
   - maxInlineSize:  default
   - maxCallerSize:  default
   )
 - basename:               out
)
)
[info] Cache config, discover and validate (0 ms)
[error] Found 3 missing definitions while linking
[error] Not found Top(native.native-code-javalib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT-1.java.lang.System$$$Lambda$9)
[error] Not found Top(native.native-code-javalib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT-1.java.lang.ref.WeakReferenceRegistry$$$Lambda$3)
[error] Not found Top(native.native-code-javalib_native0.5.0-SNAPSHOT_2.12-0.5.0-SNAPSHOT-1.scala.scalanative.runtime.Shutdown$$$Lambda$2)
[info] - should generate the llvm IR of object A and B *** FAILED ***
[info]   scala.scalanative.linker.LinkingException: Undefined definitions found in reachability phase
[info]   at scala.scalanative.linker.Reach.fail(Reach.scala:995)
[info]   at scala.scalanative.linker.Reach.reportMissing(Reach.scala:990)
[info]   at scala.scalanative.linker.Reach.result(Reach.scala:52)
[info]   at scala.scalanative.linker.Reach$.apply(Reach.scala:1008)
[info]   at scala.scalanative.linker.Link$.apply(Link.scala:13)
[info]   at scala.scalanative.build.ScalaNative$.$anonfun$link$1(ScalaNative.scala:31)
[info]   at scala.scalanative.build.Logger.time(Logger.scala:35)
[info]   at scala.scalanative.build.Logger.time$(Logger.scala:32)
[info]   at scala.scalanative.build.Logger$$anon$1.time(Logger.scala:47)
[info]   at scala.scalanative.build.ScalaNative$.link(ScalaNative.scala:31)
[info]   ...
```

ekrich:
I am seeing an issue where a project can have 3 main files for example and when you call `run` it will compile that main but write it to the defaultBasename or basename based output file. With the current setup we would need to use perhaps the main class name to be able to not recompile the executable. The current setup always recompiles even if the source files don't change. Even worse you can have 2 mains named the same thing in different packages.

```
Multiple main classes detected. Select one to run:
 [1] Test
 [2] main
 [3] test.Test

Enter number: 
```
Are there any other strange cases we could consider. Either way we need to think about what we want to do. I would like to cache the config so maybe we will need to put our heads together.

There is also a tension to make things just work, discovery and such and being able to handle more complicated setups.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
ekrich: Yes, I see that you can use the following for example in a  basic `build.sbt` file.

```scala

nativeConfig ~= {
    _.withMode(Mode.releaseFast)
}
```

ekrich: Removing raw setting breaks some scripted tests so that is in the works.

ekrich:
I am having no luck with the scripted settings. Do you think we could up the minimum `sbt` version? Version `1.1.6` is terribly old. From An old PR where I was working on things I found this `1.0.4, 1.1.6 (May 27, 2018), 1.2.8 (Dec 28, 2018), 1.3.x`. It must have been at the time `1.3.x` was newest.
The rest are `1.3.13 (Jun 27, 2020), 1.4.9 (Mar 9, 2021), 1.5.8 (Dec 20, 2021), 1.6.2 (Jan 31, 2022), 1.7.3 (Oct 30, 2022), and 1.8.2 (current)`.

Any hints appreciated on how to fix some of the scripted tests.

ekrich:
ekrich
 — 
Today at 8:35 AM
Is the minimum sbt for Scala.js, scripted etc. still 1.1.6?
sjrd
 — 
Today at 9:20 AM
We only test with 1.3.x+ nowadays, but we still compile against the 1.0.0 API. 🤷‍♂️
For cross project you need 1.2.x+, though

ekrich:
There are two scripted tests failing that I have not found obvious solutions for yet.

- `link-order` - this one does not seem to create the tmp directories, not sure though.
- `java-net-socket` - this one creates 4 applications but all in the same `basename`, `java-net-socket` and since we cache via that name, the 2nd to 4th applications don't compile a binary.

ekrich:
@WojciechMazur I am not working on this PR anymore but would appreciate you thoughts on the sbt version for scripted discussed above. We get this error I don't really like plus sbt 1.1.6 uses scala 2.12.6. It looks like it uses the latest compiler bridge.

```
[info] [info] Non-compiled module 'compiler-bridge_2.12' for Scala 2.12.17. Compiling...
[info] [info]   Compilation completed in 3.594s.
[info] [info] Done compiling.
[error] WARNING: An illegal reflective access operation has occurred
[error] WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/var/folders/f8/vpd97qv1621_b7zt0z957cjc0000gn/T/sbt_20f090d6/java-net-socket/global/boot/scala-2.12.6/org.scala-sbt/sbt/1.1.6/protobuf-java-3.3.1.jar) to field java.nio.Buffer.address
[error] WARNING: Please consider reporting this to the maintainers of com.google.protobuf.UnsafeUtil
[error] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
[error] WARNING: All illegal access operations will be denied in a future release
```

WojciechMazur:
@ekrich Thank you for your efforts in improving the build! In the case of sbt, we can update to any more recent version, I think that upgrade to 1.5.x would be good enough - that's the minor release that added support for Scala 3, so we no longer would need a custom version for tests executed with Scala 3. 

I think that if we would like to cache configs, the key should be a combination of moduleId and main class - the might be multiple main classes with the same name in different modules. Good examples here are Dotty Vulpix tests, where we define hundreds of small snippets, where most of them are defined in a top package with the main class named `Test`.

One thing we should also do is to allow for cleaning the cache when running `reload` in sbt. I'm not sure if would it create a new instance of the `Build` after executing that command, but it's possible that it would not. It's important since any changes to the nativeCofnig would be applied after the reload/restart sbt. In the same way we should check that changing config in the current sessions is actually affecting the cached build, e.g. when using `set nativeConfig ~= (_.withOptimize(false))`

ekrich:
Caching the build is a good idea and if we get the cache key correct it could work. The other suggested idea was to cache the config after `sbt` shutdown so we could avoid discovery if nothing changes. There are some edge cases as well where a clean may be needed that is not handled. Each portion of the tool chain should be able to handle changes and clean.

PR #3115, opened 2023-01-29
Add bindings and helpers for C atomics

WojciechMazur:
* Add `C` atomic.h bindings for all primitive types and fences. 
* Restore usage of C atomics in `java.lang.Thread` and `java.lang.impl.PosixThread` implementation
* Add subset of c atomics in `nativelib`s `scala.scalanative.runtime.libc` to implement required fences in `NativeThread`


clib/src/main/resources/scala-native/atomic.cpp
```diff
@@ -0,0 +1,13 @@
+#include <atomic>
+
+// Due to some bug (in GCC?) it might not be possible to access
```
ekrich, line 3: I may be missing something but is this GCC or glibc?

javalib/src/main/scala/java/lang/impl/PosixThread.scala
```diff
@@ -116,11 +118,7 @@ private[java] class PosixThread(val thread: Thread, stackSize: Long)
       isAbsolute: Boolean
   ): Unit = if (isMultithreadingEnabled) {
     // fast-path check, return if can skip parking
-    // TODO: atomics
-    val prev = counter
-    counter = 0
-    if (prev > 0) return
-    //  if (counterAtomic.exchange(0) > 0) return
+    if (counterAtromic.exchange(0) > 0) return
```
bishabosha, line 121:
```suggestion
    if (counterAtomic.exchange(0) > 0) return
```

javalib/src/main/scala/java/lang/impl/PosixThread.scala
```diff
@@ -263,10 +261,9 @@ private[java] class PosixThread(val thread: Thread, stackSize: Long)
         fromRawPtr(elemRawPtr(base, offset))
     }
 
-// TODO: atomics
-//  @alwaysinline private def counterAtomic = new CAtomicInt(
-//    fromRawPtr(classFieldRawPtr(this, "counter"))
-//  )
+  @alwaysinline private def counterAtromic = new CAtomicInt(
```
bishabosha, line 264: not sure how many references there are to this definition
```suggestion
  @alwaysinline private def counterAtomic = new CAtomicInt(
```

javalib/src/main/scala/java/lang/impl/PosixThread.scala
WojciechMazur: Fixed, thanks for catching the typo

clib/src/main/resources/scala-native/atomic.cpp
WojciechMazur: To be honest I don't remember what was the original cause, I was having these problems ~ 1-1.5 years ago. I've tried again to move these 2 definitions to `.c` file and it worked, so I've removed `atomic.cpp` entirely.

PR #3120, opened 2023-01-31
Safe Zone: Memory-safe zone ensured by capture checking and GC.

yawen-guan:
TBA


nativelib/src/main/scala-3/scala/scalanative/safe/SafeZone.scala
```diff
@@ -0,0 +1,59 @@
...
+  def close(): Unit
+
+  /** Return this zone allocator is open or not. */
+  def isOpen: Boolean
+
+  /** Return this zone allocator is closed or not. */
+  def isClosed: Boolean = !isOpen
+
+  /** Return the handle of this zone allocator. */
+  def handle: RawPtr
```
bishabosha, line 21: I feel like this shouldn't be user accessible without some explicit "unsafe" signal, e.g. perhaps
```scala
scala.scalanative.unsafe.handleOfSafeZone(sz)
```

yawen-guan:
I notice that many jobs related to testing with different scala compiler versions failed since the use of the experimental capture checking syntax, and some fails when testing in different environments. So I'm going to convert it to draft pr and fix these problems first.

nativelib/src/main/scala-3/scala/scalanative/safe/SafeZone.scala
bishabosha: or alternatively instead of `RawPtr`, maybe an opaque type alias is used?

ekrich:
@yawen-guan Welcome to the Scala Native team/project.

nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -78,7 +79,35 @@ trait NirGenExpr(using Context) {
...
+      if tree.hasAttachment(SafeZoneHandle) then
+        tree.getAttachment(SafeZoneHandle).get
+      else Val.Null
+    }
+
     def genApply(app: Apply): Val = {
+      app match {
+        case Apply(_, List(sz, tree))
+            if app.fun.symbol == defnNir.SafeZoneCompat_withSafeZone && !tree
+              .hasAttachment(SafeZoneHandle) =>
```
WojciechMazur, line 96: Is there a case when we would get into `withSafeZone` and tree would have this attachment? 
If no, maybe we can simplify this condition, and add a warning/error when overriding the attachment

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -1190,6 +1209,10 @@ object Lower {
   val largeAllocName = extern("scalanative_alloc_large")
   val largeAlloc = Val.Global(largeAllocName, allocSig)
 
+  val zoneAllocSig = Type.Function(Seq(Type.Ptr, Type.Ptr, Type.Size), Type.Ptr)
+  val zoneAllocName = extern("memorypoolzone_alloc")
```
WojciechMazur, line 1213: Maybe we should prefix the SN internal methods with `scalanative_`, to prevent potential name clashes with user code similarly as it is done `dyndispatchName` few lines below

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -1260,14 +1283,31 @@ object Lower {
...
+      )
+  }.toMap
+  val arrayAllocInZoneSig = Type.typeToArray.map {
+    case (ty, arrname) =>
+      val Global.Top(id) = arrname: @unchecked
+      ty -> Type.Function(
+        Seq(Type.Ref(Global.Top(id + "$")), Type.Int, Type.Ptr),
+        Type.Ref(arrname)
+      )
+  }.toMap
```
WojciechMazur, line 1310: These definitions are duplicating the `arrayAllocInHeap` and `arrayAllocInHeapSig` with just different signatures. We can probably refactor it a bit, to remove the duplicates, eg. using a common function

tools/src/main/scala/scala/scalanative/interflow/State.scala
```diff
@@ -260,11 +266,17 @@ final class State(block: Local) {
...
           }
-        emit.arrayalloc(elemty, init, Next.None)
-      case VirtualInstance(BoxKind, cls, Array(value)) =>
+        emit.arrayalloc(elemty, init, escapedVal(zoneHandle), Next.None)
+      case VirtualInstance(
+            BoxKind,
+            cls,
+            Array(value),
+            zoneHandle
+          ) => // TODO-yawen: is boxed value always allocated in heap?
```
WojciechMazur, line 275: Yes, boxed values are always allocated in heap

nativelib/src/main/scala/scala/scalanative/runtime/CMemoryPool.scala
```diff
@@ -0,0 +1,38 @@
...
+  def free(rawpool: RawPtr): Unit = extern
+}
+
+object CMemoryPool {
+  val defaultMemoryPoolHandle = _CMemoryPool.open() // TODO: use lazy val
+
+  def freeDefaultMemoryPool(): Unit = {
+    _CMemoryPool.free(defaultMemoryPoolHandle)
+  }
+}
```
WojciechMazur, line 20: You can probably make the extern bindings an private inner object, this way we can make it a bit more cleaner and prevent leaking the access to the internals

Also, we should be able to see lazy val safely here, but it does not make a big difference - it would be allocated on the first usage anyway.

```suggestion
object CMemoryPool {
  lazy val defaultMemoryPoolHandle = Impl.open() // TODO: use lazy val

  def freeDefaultMemoryPool(): Unit =  Impl.free(defaultMemoryPoolHandle)
  
  @extern
  private object Impl {
    @name("memorypool_open")
    def open(): RawPtr = extern

    @name("memorypool_free")
    def free(rawpool: RawPtr): Unit = extern
  }
}
```

nativelib/src/main/scala/scala/scalanative/runtime/Arrays.scala.gyb
```diff
@@ -218,7 +221,9 @@ object ${T}Array {
       if (is32BitPlatform) castIntToRawSize((castRawSizeToInt(sizeOfPtr) + 8) + ${sizeT} * length)
       else castLongToRawSize((castRawSizeToLong(sizeOfPtr) + 8L) + ${sizeT}.toLong * length.toLong)
     )
-    val arr     = ${alloc}(arrcls, arrsize)
+    val arr     =
```
WojciechMazur, line 224: We can lint it a bit
```suggestion
    val arr = 
```

nativelib/src/main/scala-3/scala/scalanative/safe/SafeZoneCompat.scala
```diff
@@ -0,0 +1,7 @@
+package scala.scalanative.safe
+
+import language.experimental.captureChecking
+
+object SafeZoneCompat {
+	def withSafeZone[T](sz: {*} SafeZone, obj: T): {sz} T = obj
```
WojciechMazur, line 6: The formating seems off, probably using tabs instead of spaces

nativelib/src/main/scala-3/scala/scalanative/safe/SafeZone.scala
WojciechMazur: I think that making it `private[scalanative] def handle: RawPtr` should do the job. In general `RawPtr` by definition is `super-unsafe` and it should be used only by the internals. `RawPtr` can wrapped by `Ptr[T]` which is just `unsafe` ;)

nativelib/src/main/scala-3/scala/scalanative/safe/SafeZone.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  /** Require this zone allocator is open. */
+  def checkOpen: Unit = {
+    if (!isOpen)
+      throw new IllegalStateException(s"Zone ${this} is already closed.")
+  }
+}
+
+final class MemorySafeZone (private[this] val zoneHandle: RawPtr) extends SafeZone {
```
WojciechMazur, line 30: If we go with the previous comment we can simplify it a bit
```suggestion
final class MemorySafeZone (private[scalanative] val handle: RawPtr) extends SafeZone {
```
and remove the public accessor

nativelib/src/main/resources/scala-native/zone/MemoryPool.c
```diff
@@ -0,0 +1,128 @@
...
+    // Free pages.
+    MemoryPage *page = pool->page, *pre_page = NULL;
+    while (page != NULL) {
+        pre_page = page;
+        page = page->next;
+        free(pre_page);
+    }
+    // Free the pool.
+    free(pool);
+}
```
WojciechMazur, line 128: There is plenty of files without a blank line at the EOF (warned by GitHub with red sign that can be seen here). In most cases its not a problem, but it's a good practice to add this spare blank line, otherwise some legacy programs might have some issues, e.g. git pathes AFAIK

nativelib/src/main/resources/scala-native/gc/immix_commix/GCRoots.c
```diff
@@ -15,28 +16,50 @@ void GC_Roots_Add(GC_Roots **head, AddressRange range) {
...
+// void GC_Roots_Print(GC_Roots *head) {
+//     printf("== GC Roots start ==\n");
+//     GC_Roots *current = head;
+//     while (current != NULL) {
+//         printf("GC_Roots[%p, %p)\n", current->range.address_low,
+//                current->range.address_high);
+//         current = current->next;
+//     }
+//     printf("== GC Roots end ==\n");
+// }
```
WojciechMazur, line 65: I dont see any usage of that, so probably it can be removed

WojciechMazur (commented):
Great work on that, I have a few suggestions

nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenExpr.scala
yawen-guan: `withSafeZone(sz, new T(...))` (this `new T(...)` tree has no attachment) is translated to `{ sz.checkOpen(); new T(...)}` (this `new T(...)` tree has attachment). So the condition is needed here.

nativelib/src/main/scala-next/scala/scalanative/safe/SafeZone.scala
```diff
@@ -0,0 +1,51 @@
...
+    CZone.close(handle)
+  }
+
+  override def isOpen: Boolean = flagIsOpen
+}
+
+
+object SafeZone {
+  /** Run given function with a fresh zone and destroy it afterwards. */
+  final def apply[T](f: ({*} SafeZone) => T): T = {
```
WojciechMazur, line 46: I think context function would make it way more user friendly in the long run. In most of the cases we only want to allocate in the current Zone scope. It would match perfectly with the additional alloc method defined lower
```suggestion
  final def apply[T](f: ({*} SafeZone) ?=> T): T = {
```
here's how users code can then look like: 
```scala
  SafeZone:
    val str = alloc(String("fooBar"))
    val zone = summon[SafeZone]
    val str2= zone.alloc(String("foo"))
    val str3 = summon[SafeZone] alloc String("baz")

    SafeZone{
      val str2 = alloc(new String("bar"))
    }
    ```

File: nativelib/src/main/scala-next/scala/scalanative/safe/SafeZone.scala [hunk 13]
```diff
@@ -0,0 +1,51 @@
+package scala.scalanative.safe
```
WojciechMazur, line 1: I'm not sure if `safe` package would be the best name. Maybe we can change it to `scala.scalanative.memory` or even move it  to `scala.scalanative` to make it easily accessible.

nativelib/src/main/resources/scala-native/zone/LargeMemoryPool.c
```diff
@@ -0,0 +1,86 @@
+#include <stdio.h>
```
WojciechMazur, line 1: In theory, most of these C sources could be implemented in Scala Native instead of C, I think we should be able to have binding to all the needed headers. However, I don't require you to rewrite it. 
I guess that one of the reasons to implement them in C is to make them a bit faster and independent from GC and it's allocation. Is there maybe some other reason?

nscplugin/src/main/scala-next/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -0,0 +1,2642 @@
...
+          // Extract the handle of `sz` and put it into the attachment of `new T(...)`.
+          val handle = genExpr(Select(sz, termName("handle")))
+          if tree.hasAttachment(SafeZoneHandle) then
+            report.warning(
+              s"Safe zone handle is already attached to ${tree}, which is unexpected.",
+              tree.srcPos
+            );
+          tree.putAttachment(SafeZoneHandle, handle)
+          // Translate `withSafeZone(sz, new T(...))` to `{ sz.checkOpen(); new T(...) }`.
+          val checkOpen = Apply(Select(sz, termName("checkOpen")), List())
```
WojciechMazur, line 110: In theory, we could pass `SafeZone` instance instead of its handle to classalloc. This way we should've able to call `checkOpen` from within of actual allocation

nativelib/src/main/scala-next/scala/scalanative/safe/SafeZone.scala
bishabosha: you could even add
```scala
transparent inline def zone(using sz: {*} SafeZone): {sz} SafeZone = sz
```
then the zone can be summoned more elegantly

nscplugin/src/main/scala-next/scala/scalanative/nscplugin/NirGenExpr.scala
```diff
@@ -0,0 +1,2642 @@
...
+          // For new expression with a specified safe zone, e.g. `new {sz} T(...)`,
+          // it's translated to `withSafeZone(sz, new T(...))` in TyperPhase.
+          tree match {
+            case Apply(Select(New(_), nme.CONSTRUCTOR), _) =>
+            case Apply(fun, _) if fun.symbol == defn.newArrayMethod =>
+            case _ => 
+              report.error(s"Unexpected tree in withSafeZone: `${tree}`", tree.srcPos)
+          }
+          // Extract the handle of `sz` and put it into the attachment of `new T(...)`.
+          val handle = genExpr(Select(sz, termName("handle")))
```
bishabosha, line 102: isn't this a violation of the access modifiers on `handle` when someone tries to allocate outside of the `scalanative` package?

unit-tests/native/src/test/scala-next/scala/scalanative/safe/SafeZoneTest.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  @Test def `objects allocated in safe zone is accessible`(): Unit = {
+    
+    def assertAccessible(n: Int): Unit = {
+      case class A(v: Int) {}
+      SafeZone { sz => 
+        val ary = new Array[{sz} A](n)
+        for i <- 0 until n do
+          ary(i) = withSafeZone(sz, new A(i))
```
bishabosha, line 95: I think this should also be tested outside of the `scalanative` package, to make sure there are no access restrictions

unit-tests/native/src/test/scala-next/scala/scalanative/safe/SafeZoneTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      SafeZone { sz => 
+        sz.close()
+        Try[{sz} A].apply(withSafeZone(sz, new A())) match {
+          case Success(_) => fail("Should not allocate instances in a closed safe zone.")
+          case Failure(e: IllegalStateException) => ()
+          case Failure(_) => fail("Unexpected error.")
+        }
+      }
+    )
+  }
```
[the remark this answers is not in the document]
yawen-guan: I think both the `open` and `close` operation of safe zone should be hidden from users, if both accessible to users,  users may create zones that are overlapped but not nested: 
```scala
val sz0 = SafeZone.open()
val sz1 = SafeZone.open()
sz0.close()
sz1.close()
```
And it doesn't make sense if only one operation is accessible to users. 

For the test here, I just wanna make sure that the inner usage (i.e. within package `scala.scalanative`) of safe zone is correct. Maybe I don't need to test this but just assume any inner usage is correct?

nativelib/src/main/scala-next/scala/scalanative/safe/SafeZoneCompat.scala
```diff
@@ -0,0 +1,7 @@
+package scala.scalanative.safe
+
+import language.experimental.captureChecking
+
+object SafeZoneCompat {
+  def withSafeZone[T](sz: {*} SafeZone, obj: T): {sz} T = obj
```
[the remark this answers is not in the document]
yawen-guan: Instead of moving `withSafeZone` to `SafeZone`, maybe move it to `scala.scalanative.runtime` to indicate that it's an inner function and only the compiler should use it?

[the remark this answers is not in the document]
WojciechMazur: Sure, we can do that as well, good point. It would be easiest to just move it there. At that point, we can also rename to `scala.scalanative.runtime.SafeZone`. 

Alternatively, we can make it a part of the `runtime` package object by usage of trait mixins and different sources in scala (empty trait) and scala-next (actual implementation available only with cc), but I don't think it's worth it.

unit-tests/native/src/test/scala-next/scala/scalanative/safe/SafeZoneTest.scala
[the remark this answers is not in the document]
WojciechMazur: Ok, good point, let's leave it as it is.

nativelib/src/main/scala/scala/scalanative/runtime/CZone.scala
```diff
@@ -0,0 +1,15 @@
+package scala.scalanative.runtime
```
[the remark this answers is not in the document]
yawen-guan: `CZone` is also used in `Array` in `scala.scalanative.runtime` so it can't be `private[safe]`. Still, I could move it to `scala.scalanative.safe`, what do you think?

nativelib/src/main/resources/scala-native/zone/LargeMemoryPool.c
yawen-guan: At the beginning I simply wanted to make the zone backend as fast as possible and be maintained the same way as GC does. Hmm, implementing it in Scala Native should be better for maintenance, I'll consider rewriting it later.

nativelib/src/main/scala-next/scala/scalanative/safe/SafeZone.scala
yawen-guan: Changing `apply` to context function will disallow users to explicit declare the SafeZone value, i.e. write code like `SafeZone { sz => ...}`, right? I don't think it worths.

nscplugin/src/main/scala-next/scala/scalanative/nscplugin/NirGenExpr.scala
yawen-guan: Hmm, yes and no. If someone calls `sz.handle` outside the `scalanative` package, he will get a ReferenceError saying the handle method is not accessible, which is generated in typer phase by the typer (the type assigner, more exactly. I might be wrong since I'm not familiar enough with the typer..). For this internal function `withSafeZone` (whose name is now updated to `allocate`), the call to `sz.handle` is directly manipulated as a typed tree and skips the typer phase, so it won't trigger the ReferenceError.

nscplugin/src/main/scala-next/scala/scalanative/nscplugin/NirGenExpr.scala
yawen-guan: That sounds cool! But I'm wonder how to do so. Like, somehow call the `checkOpen` when lowering `classalloc`?

armanbilge:
Thanks for this very exciting PR!! If you don't mind, I have a question about this:
https://github.com/scala-native/scala-native/blob/2c03183df62870399790e53031b6619ea41d8547/nativelib/src/main/resources/scala-native/zone/MemoryPool.c#L49-L50

If I understand correctly, this is registering the Safe Zone with the GC, so that it can properly trace any live references to GC-managed objects that are originating from this zone.

Apologies if I am missing something, but doesn't this negate much of the performance win? Because even though the memory in the Safe Zone is "manually" managed (i.e., not managed by GC), the GC still has to _trace_ all the memory in the zone anyway ... it's just not allowed to free it.

I understand there are implementation challenges if the GC does not trace the Safe Zone. Then, anything in the safe zone must not hold a reference to any GC-managed objects. It can only reference global objects, or perhaps objects belonging to some other Safe Zone with a broader scope than the current Safe Zone.

Is it possible to express such a constraint with capture checking? If so, I think this would make it possible to safely allocate completely independently of the GC, which would decrease pressure on the GC and I expect would be a non-trival win for performance.

Thanks! :)

nativelib/src/main/scala-next/scala/scalanative/safe/SafeZone.scala
bishabosha: you can still have `SafeZone { sz ?=> ...}`

nscplugin/src/main/scala-next/scala/scalanative/nscplugin/NirGenExpr.scala
bishabosha: If it works that way currently then that's cool, I still wonder about potential future proofing if a future scala native version implements strict access checking at the linking phase e.g.

nativelib/src/main/scala/scala/scalanative/runtime/CZone.scala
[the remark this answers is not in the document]
yawen-guan: Now SafeZone is moved to `scala.scalanative` and the `safe` package doesn't exist anymore.

yawen-guan:
> Thanks for this very exciting PR!! If you don't mind, I have a question about this:
> 
> https://github.com/scala-native/scala-native/blob/2c03183df62870399790e53031b6619ea41d8547/nativelib/src/main/resources/scala-native/zone/MemoryPool.c#L49-L50
> 
> If I understand correctly, this is registering the Safe Zone with the GC, so that it can properly trace any live references to GC-managed objects that are originating from this zone.
> 
> Apologies if I am missing something, but doesn't this negate much of the performance win? Because even though the memory in the Safe Zone is "manually" managed (i.e., not managed by GC), the GC still has to _trace_ all the memory in the zone anyway ... it's just not allowed to free it.
> 
> I understand there are implementation challenges if the GC does not trace the Safe Zone. Then, anything in the Safe Zone must not hold a reference to any GC-managed objects. It can only reference global objects, or perhaps objects belonging to some other Safe Zone with a broader scope than the current Safe Zone.
> 
> Is it possible to express such a constraint with capture checking? If so, I think this would make it possible to safely allocate completely independently of the GC, which would decrease pressure on the GC and I expect would be a non-trival win for performance.
> 
> Thanks! :)

Hi Arman, thanks for this good question! 
> Is it possible to express such a constraint with capture checking?

Yes and no: not with the capture checking currently implemented in dotty, but a modified version could make it possible. We may add a special capability `Heap` to the system which every object allocated in heap should capture. Then a rule should be added to reject objects capturing `Heap` and any SafeZone instance directly or indirectly at the same time. The program will look like: 
```scala
case class A() {}
case class B(a: {*} A) {}

val a: {Heap} A = new A()
SafeZone { sz => 
  val b0: {sz, a} B = new {sz} B(a) // reject
  val b1: {sz} B = new {sz} B(new {sz} A)) // allow
}
```
> Apologies if I am missing something, but doesn't this negate much of the performance win?

It's true that GC still have to trace everything in the zone. As for the performance, a preliminary experiment shows that the current safe zone brings good performance improvement on a specific benchmark, compared to using only the GC. We're working on a more detailed experiment with more benchmarks, and we'll profile how much time the GC used to give an evidence for the question whether it negates much of the performance win. 

---

Currently, we may violate memory safety if we don't allow references from zones to heap. One can always upcast the pure type `A` to `{sz} A` since pure type is subtype of capture type, which means we can always treat a heap-allocated object as a zone-allocated object. It's safe because heap has longer lifetime than zones.  
```scala
val a: A = new A() // Allocated in heap.
SafeZone { sz => 
  val array: Array[{sz} A] = new Array[{sz} A](2)
  array(0) = a // Allow. 
  array(1) = new {sz} A()
}
```
What if `array(1)` tries to reference to `array(0)` here? The type system already forget `array(0)` is allocated in heap. It's violation of memory safety if the compiler allows it without notifying GC; otherwise, how to we reject it?

The above example shows that we can actually view the heap as a outermost zone except that objects in it are dynamically deallocated. We want to support sub-regioning in the future: if `sz0` is the outer zone and `sz1` is the inner zone, then `{sz0} <: {sz1}`. It means every time an object allocated in a inner zone is expected, we can pass an object allocated in the outer zone. It's always safe because outer zone has longer lifetime. The heap fits perfectly in the sub-regioning if we can treat it as the outermost zone. 

---

In this specific case, another question is "do we really wanna sacrifice the expressiveness gain from allowing references from zones to heap?" I think it needs futhur discussion.

armanbilge:
Wow, thanks for that detailed response, I really appreciate it! 😁 that was very helpful and informative.

---

> Currently, we may violate memory safety if we don't allow references from zones to heap. One can always upcast the pure type `A` to `{sz} A` since pure type is subtype of capturing type

Ah, I didn't realize that's how capture-checking is working 😅 in that case, I see that the current implementation is indeed required for safety.

> As for the performance, a preliminary experiment shows that the current safe zone brings good performance improvement on a specific benchmark, compared to program that only uses GC. We're working on a more detailed experiment with more benchmarks, and we'll profile how much time the GC used to give an evidence for the question whether it negates much of the performance win.

Well that's good news! I'll look forward to seeing those benchmark results.

--- 

> In this specific case, another question is "do we really wanna sacrifice the expressiveness gain from allowing references from zones to heap?" I think it needs further discussion.

Yes, well this is quite the question 😁 I guess I am not sure yet how important this is. Heap should certainly be able to reference zones, but I wonder if anything referenced by a zone can simply be encapsulated in another, outer zone.

When I first saw your PR, it inspired a thought-experiment: can we run a Scala Native HTTP microservice, that actually doesn't need GC at all? This is interesting, because the unpredictability of GC pauses creating latency is often a major concern for developers.

The hypothesis is that there are three lifecycles of allocations in such an application:
1. Request-specific allocations, that can be completely thrown away after the response is sent. This is essentially arena allocation.
2. Allocations that may be reused across many requests. For example, establishing a TCP socket connection or caching some data for fast retrieval.
3. Global allocations, whose lifecycle is the same as the application.

(2) is the interesting one, because the lifecycle of those allocations is ambiguous. However, the observation is that these objects _already_ require some kind of manual resource management. For example, sockets are managed by a connection pool. Similarly, an in-memory cache (essentially some kind of `Map`) should have some strategy to remove stale entries that are no longer needed, otherwise it will become a memory leak even in a GC runtime. So since these resources are already managed, it seems possible that their memory could be manually managed as well.

The bigger picture here is that I think that many applications are already composed of manually-managed resources, with well-defined lifecycles (request lifecycles, connection lifecycles, cache lifecycles), and we can probably take advantage of that.

> we can actually view the heap as a outermost zone except that objects in it are dynamically deallocated.

Out of curiosity: can we view the heap as the _innermost_ zone, for when you want to allocate some objects that are allowed to be dynamically deallocated _before_ their parent zone is deallocated? This can be useful if you have some long-lived zone (e.g. connection/cache lifecycle) that requires some short-term allocations with an unclear lifecycle to do something (e.g. start some tasks to refresh/reap values in the cache).

The downside of course is that the developer has to reason about whether or not some allocation can belong to the lifecycle of the zone, or whether this will become a memory-leak because the zone is long-lived. But I _think_ this is something they have to be aware of anyway, in the current scheme as well.

nscplugin/src/main/scala-next/scala/scalanative/nscplugin/NirGenExpr.scala
WojciechMazur: Yes, It might look like this: 
```patch
diff --git a/tools/src/main/scala/scala/scalanative/codegen/Lower.scala b/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
index 9f8d083b0..066739c11 100644
@@ -770,8 +770,32 @@ object Lower {
       assert(size == size.toInt)
 
       zoneHandle match {
-        case Some(zoneHandle) =>
-          assert(zoneHandle != Val.Null, "ZoneHandle is null")
+        case Some(instance) =>
+          assert(instance != Val.Null, "ZoneHandle is null")
+          val checkOpenMethod, handleAccessor = Val.Local(fresh(), Type.Ptr)
+          // Perform virtual dispatch of checkOpen method and call it
+          genMethodOp(
+            buf,
+            checkOpenMethod.name,
+            Op.Method(instance, SafeZoneCheckOpen.sig)
+          )
+          buf.call(SafeZoneCheckOpenSig, checkOpenMethod, Seq(instance), unwind)
+        
+          // Perform virtual dispatch of handle accessor and call it
+          // Currently it's not working correctly - optimizer might replace method dispatch op with direct fieldload becouse there is only 1 implementation of SafeZone
+          genMethodOp(
+            buf,
+            handleAccessor.name,
+            Op.Method(instance, SafeZoneHandle.sig)
+            )
+          val zoneHandle = buf.call(
+            SafeZoneHandleSig,
+            handleAccessor,
+            Seq(instance),
+            unwind
+          )
+
+          // Call actual zone allocation using resolved zoneHandle
           buf.let(
             n,
             Op.Call(
@@ -1422,6 +1446,19 @@ object Lower {
   val RuntimeNull = Type.Ref(Global.Top("scala.runtime.Null$"))
   val RuntimeNothing = Type.Ref(Global.Top("scala.runtime.Nothing$"))
 
+  val SafeZone = Type.Ref(Global.Top("scala.scalanative.SafeZone"))
+
+  val SafeZoneCheckOpenSig = Type.Function(Seq(SafeZone), Type.Unit)
+  val SafeZoneCheckOpen =
+    SafeZone.name.member(Sig.Method("checkOpen", Seq(Type.Unit)))
+
+  val SafeZoneHandleSig = Type.Function(Seq(SafeZone), Type.Ptr)
+  val SafeZoneHandle =
+    SafeZone.name.member(Sig.Method("handle", Seq(Type.Ptr)))
+
   val injects: Seq[Defn] = {
     implicit val pos = Position.NoPosition
     val buf = mutable.UnrolledBuffer.empty[Defn]
@@ -1467,6 +1504,9 @@ object Lower {
     buf += throwNoSuchMethod
     buf += RuntimeNull.name
     buf += RuntimeNothing.name
+    buf += SafeZoneCheckOpen
+    buf += SafeZoneHandle
     buf.toSeq
   }
 }

```

But it would be even better to add an additional method to `trait SafeZone` which would be responsible for all the required logic.
```scala
  private[scalanative] def allocImpl[T](size: CSize, cls: Class[T]): Ptr[T] = {
    this.checkOpen()
    CZone.alloc(handle, castObjectToRawPtr(cls), size)
  }
```
or 
```scala
  private[scalanative] def allocImpl[T](size: RawSize, cls: RawPtr): RawPtr = {
    this.checkOpen()
    CZone.alloc(handle, cls, size)
  }
```
Then we can add a Op.Method dispatch like in the snippet above to get correct implementation of allocImpl if there would more than 1 implementation or direct call to the only used implementation otherwise.

The one biggest improvement of moving generation of additional logic (e.g. checkOpen) to lowering phase, possibly combined with virtual method dispatch is the fact that we would be able to easily add changes to the logic if they would be needed. This would improve backward and forward compatibility of the produced NIR.

nativelib/src/main/scala-next/scala/scalanative/safe/SafeZone.scala
yawen-guan: > you could even add
> 
> ```scala
> transparent inline def zone(using sz: {*} SafeZone): {sz} SafeZone = sz
> ```
> 
> then the zone can be summoned more elegantly

Sorry I didn't get it, what's the benefit we get from calling such a `zone` function that we can't get from calling `summon[SafeZone]`?

bishabosha: would you rather type `summon[Context]` in the dotty compiler or type `ctx` :) because its the same thing

ekrich:
> 3. Global allocations, whose lifecycle is the same as the application.

This is also an interesting use case and is what is used for real time systems that JPL was requesting. They are looking for predictability which is why they were also requesting stack allocation of objects.

nativelib/src/main/resources/scala-native/gc/commix/ComixGC.c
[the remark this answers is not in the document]
yawen-guan: Thanks, I'll leave it for a separate PR since this PR didn't update the filename.

yawen-guan:
FYI: The description is updated ;)

nativelib/src/main/scala/scala/scalanative/runtime/CZone.scala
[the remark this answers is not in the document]
WojciechMazur: Let me reopen this conversation. 
We still should be able to hide this internal API from the users, the common package for both  `scala.scalanative.memory.SafeZone` and `scala.scalanative.runtime.Array` would be `private[scalanative]`. This should be enough.

WojciechMazur (commented):
Looks good. Great job! 
I have no comments on technical aspects, however, I've left some minor styling comments, which should be worth considering before merging. 

After merging this PR we should sync the dotty-experimental branch with the current main (I will do it). Then we could check how the safe zones would compose in multithreaded runtime, but that's the issue for another day.

yawen-guan:
@WojciechMazur Sorry for leaving for a long time. The original `CZone` is moved as `SafeZoneAllocator.Impl`. I currently keep `SafeZoneAllocator` in `scala.scalanative.runtime` because I need it not private to package `scalanative` for unit tests outside the scalanative package. However, we can move it to `scala.scalanative.memory` and make it `private[scalanative]` after dotty supports the `new {sz} T(...)` syntax (so we don't need to directly call `SafeZoneAllocator.allocate` to test its functionality). The dotty PR have also been created for a long time;  it'll be merged after this one.

WojciechMazur (approved):

PR #3122, opened 2023-01-31
Refactor providing memory layout and platform information

WojciechMazur:
* Information about memory layout for `ObjectHeader`, `ArrayHeader`, `Rtti`, `ClassRtti` is now centralized to a single file and calculated based on the `Metadata` 
* Magic values in tools are replaced with constants defined in layouts for mentioned types
* Defined `scala.scalanative.runtime.MemoryLayout` as a counterpart to the toolchain memory layout structure. The magic values used in code are replaced with pseudo-constants (functions that are always inlined, they shall be fully evaluated to constant by the optimizer)
* Added `codegen.PlatformInfo` containing platform specific information. It it used in `Interflow` and `CodeGen` to provide required metadata. It is implicitly used by `MemoryLayout` instead of explicit information about usage of 32bit architecture.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Arrays.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Arrays.scala
@@ -29,8 +29,7 @@ package runtime
 
 import scalanative.unsafe._
 import scalanative.unsigned._
-import scalanative.runtime.Intrinsics._
-
+import scalanative.runtime.Intrinsics.{castIntToRawSizeUnsigned => intToUSize, _}
 
 sealed abstract class Array[T]
     extends java.io.Serializable
@@ -39,7 +38,7 @@ sealed abstract class Array[T]
   /** Number of elements of the array. */
   @inline def length: Int = {
     val rawptr = castObjectToRawPtr(this)
-    val lenptr = elemRawPtr(rawptr, sizeOfPtr)
+    val lenptr = elemRawPtr(rawptr, intToUSize(MemoryLayout.Array.LengthOffset))
     loadInt(lenptr)
   }
 
@@ -164,11 +163,7 @@ final class BooleanArray private () extends Array[Boolean] {
       throwOutOfBounds(i)
     } else {
       val rawptr = castObjectToRawPtr(this)
-      elemRawPtr(
-        rawptr,
-        if (is32BitPlatform) castIntToRawSize((castRawSizeToInt(sizeOfPtr) + 8) + 1 * i)
-        else castLongToRawSize((castRawSizeToLong(sizeOfPtr) + 8L) + 1.toLong * i.toLong)
-      )
+      elemRawPtr(rawptr, intToUSize(MemoryLayout.Array.ValuesOffset + 1 * i))
     }
 
   @inline def apply(i: Int): Boolean = loadBoolean(atRaw(i))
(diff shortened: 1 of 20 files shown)


ekrich:
This looks really good and it is great that the RTTI is co-located.

If I understand correctly each Object has a pointer to the Class RTTI so that is the only overhead for object layout in memory?

Also, does the runtime refer to the RTTI vtable to dispatch the correct overridden method?

Asking because I was wondering if these are memory and/or performance advantages compared to the JVM and maybe even C++ memory due to erasure vs templates - we also do some specialization which is more like templates?

Just trying to understand a little better.

WojciechMazur:
> If I understand correctly each Object has a pointer to the Class RTTI so that is the only overhead for object layout in memory?

Yes, each object (instance of a class) starts with a header, which currently is only a pointer to class rtti, so an overhead of (4/8 bytes depending on arch), similarly each array has an overhead of 8/16 bytes (rtti + length + padding). With this change its easier to understand that. Previously we had similar headers defined only in Immix/Comix GC as c structs. 

> Also, does the runtime refer to the RTTI vtable to dispatch the correct overridden method?
Yes, vtable is used for dispatching methods of a class, it's used internally in the lowering phase (when replacing nir.Op.Method with the calculated pointer to method in vtable) https://github.com/scala-native/scala-native/blob/9a0569b139a09fd88089484d75bfe394b0ff7194/tools/src/main/scala/scala/scalanative/codegen/Lower.scala#L687-L705
For traits there is used a dedicated dispatch table  https://github.com/scala-native/scala-native/blob/9a0569b139a09fd88089484d75bfe394b0ff7194/tools/src/main/scala/scala/scalanative/codegen/Lower.scala#L707-L724

I'm not sure is it a big advantage over JVM in terms of memory usage. In case of x64 architectures, our object header is 8 bytes, and theirs AFAIK 16 bytes. 
However, with upcoming changes implementing ObjectMonitors and `synchronized` clauses we would be required to add additional 8 bytes for thin/fat lock (intptr with bitset, or pointer for object monitor instance when we would detect concurrent access to the same object). 
An alternative for that would be to allow for `synchronized` blocks only for types extending some special trait (but this would break our jvm cross-compiling) or detecting which types are using monitors (fragile and imperfect, since even a single usage of `synchronized` on `Object` / `new {}` would require us to add this field to all subtypes) and would make GC integration a lot more difficult.

tarsa:
> I'm not sure is it a big advantage over JVM in terms of memory usage. In case of x64 architectures, our object header is 8 bytes, and theirs AFAIK 16 bytes. However, with upcoming changes implementing ObjectMonitors and `synchronized` clauses we would be required to add additional 8 bytes for thin/fat lock (intptr with bitset, or pointer for object monitor instance when we would detect concurrent access to the same object). An alternative for that would be to allow for `synchronized` blocks only for types extending some special trait (but this would break our jvm cross-compiling) or detecting which types are using monitors (fragile and imperfect, since even a single usage of `synchronized` on `Object` / `new {}` would require us to add this field to all subtypes) and would make GC integration a lot more difficult.

note that there's ongoing OpenJDK project with aim to reduce object headers: https://openjdk.org/projects/lilliput/
> [Lilliput](https://openjdk.org/projects/lilliput)
> The goal of this [Project](https://openjdk.org/projects/index.html) is to explore techniques to downsize Java object headers in the Hotspot JVM from 128 bits to 64 bits or less, reducing Java's memory footprint. Improved performance across most, if not all, workloads is also expected.

mailing list: https://mail.openjdk.org/pipermail/lilliput-dev/
project is lead by Roman Kennke: https://openjdk.org/census#lilliput

as seen on the openjdk wiki page https://wiki.openjdk.org/display/lilliput (and also judging by mailing list contents) there are also some ideas about potentially shrinking the header to 32 bit or less:
> Goals
> Reduce the object header to 64 bits. It may be possible to shrink it down to 32 bits as a secondary goal.

WojciechMazur:
@tarsa thanks for the links. We had some ideas for implementing of similar ideas in the past but never had enough time to actually implement that. Maybe in the future, we would be able to learn from the Lilliput project experience and improve our memory layouts as well, especially since for us it could be a bit easier as we don't have as many requirements/backward compatibilities as they do

PR #3123, opened 2023-01-31
Define null guards for methods using `this`

### LINKED ISSUE #3014: Unexpected `MatchError` instead of `NullPointerException` when matching on this
The null guard for method parameters and especially `this` argument might be skipped. It then might lead to an unexpected `MatchError`. This for example can be witnessed when using Scala 3 enums with a method using `this match ...` and a single layer of indirection.

Usually, we would create a null guard in the Lowering phase for `Op.Method` to make sure that the method argument is not null, however, the optimizer might replace `Op.Method` with a direct call.
```
def @"M13Test$package$D7useUnitL8TimeUnituEO" : (@"T13Test$package$", @"T8TimeUnit") => unit {
%3(%2 : @"T13Test$package$", %1 : @"T8TimeUnit"):
  %4 = module @"T13scala.Predef$"
  %5 = method %1 : @"T8TimeUnit", "D7toNanosjjEO"
  %6 = call[(@"T8TimeUnit", long) => long] %5 : ptr(%1 : @"T8TimeUnit", long 1)
```

```
didopt def @"M13Test$package$D7useUnitL8TimeUnituEO" : (@"T13Test$package$", @"T8TimeUnit") => unit {
%30000(%1 : @"T13Test$package$", %2 : @"T8TimeUnit"):
  %30002 = call[(@"T8TimeUnit", long) => long] @"M8TimeUnitD7toNanosjjEO" : ptr(%2 : @"T8TimeUnit", long 1)
  %30004 = module @"T13scala.Predef$"
```

Reproducer: 
```scala
enum TimeUnit {
  case Millis
  case Nanos

  def toNanos(value: Long): Long = this match {
    case Millis => value * 1000000
    case Nanos => value
  }
}

def useUnit(unit: TimeUnit) = {
  println(unit.toNanos(1L))
}

@main def Test = {
  useUnit(TimeUnit.Nanos)
  useUnit(null)
}

```

Results in: 
```scala 
1
scala.MatchError: null
        at java.lang.StackTrace$.currentStackTrace$$anonfun$1(Unknown Source)
        at java.lang.StackTrace$$$Lambda$2.applyVoid(Unknown Source)
        at scala.runtime.function.JProcedure1.apply(Unknown Source)
        at TimeUnit.toNanos(Unknown Source)
        at Test$package$.useUnit(Unknown Source)
        at Test$package$.Test(Unknown Source)
        at Test.main(Unknown Source)
```
expected (on JVM): 
```
1
Exception in thread "main" java.lang.NullPointerException
        at Test$package$.useUnit(Test.scala:12)
        at Test$package$.Test(Test.scala:17)
        at Test.main(Test.scala:15)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at dotty.tools.scripting.ScriptingDriver.compileAndRun(ScriptingDriver.scala:36)
        at dotty.tools.scripting.Main$.main(Main.scala:45)
        at dotty.tools.MainGenericRunner$.run$1(MainGenericRunner.scala:249)
        at dotty.tools.MainGenericRunner$.main(MainGenericRunner.scala:268)
        at dotty.tools.MainGenericRunner.main(MainGenericRunner.scala)
```

WojciechMazur:
In multiple functions `this` value is replaced with simple `null`, eg. stor statics. However, there are cases in which we can produce code using `this` without checking if it's not null. It can lead to unexpected behaviour, e.g. matching on `this` could result in `MatchException` instead of expected `NullPointerException`

Fixes #3014

* Added utility for traversal of `nir` structures
* Added null guards for methods using `this` in function body. No check would be emmited for function which does not take or use `this` value to not produce additional overhead
* Fixed resolving of static and private function in Sig. Parsing of the name, was replaced with more stable unmangling of the whole symbol.

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Sig.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Sig.scala
@@ -38,29 +38,9 @@ final class Sig(val mangle: String) {
 
   final def isVirtual = !(isCtor || isClinit || isExtern)
   final def isPrivate: Boolean = privateIn.isDefined
-  final def isStatic: Boolean = {
-    def isPublicStatic = mangle.last == 'o'
-    def isPrivateStatic = {
-      val sigEnd = mangle.lastIndexOf('E')
-      val scopeIdx = sigEnd + 1
-      def hasScope = mangle.length() > scopeIdx
-      sigEnd > 0 && hasScope && mangle(sigEnd + 1) == 'p'
-    }
-    isPublicStatic || isPrivateStatic
-  }
-  final lazy val privateIn: Option[Global.Top] = {
-    val sigEnd = mangle.lastIndexOf('E')
-    val scopeIdx = sigEnd + 1
-    def hasScope = mangle.length() > scopeIdx
-    def isPrivate = {
-      val scopeIdent = mangle(scopeIdx)
-      scopeIdent == 'p' || scopeIdent == 'P'
-    }
-    if (sigEnd > 0 && hasScope && isPrivate) {
-      val global = Unmangle.unmangleGlobal(mangle.substring(sigEnd + 2))
-      Some(global.top)
-    } else None
-  }
+  final def isStatic: Boolean = unmangled.sigScope.isStatic
+  final lazy val privateIn: Option[Global.Top] =
+    unmangled.sigScope.privateIn.map(_.top)
 }
 object Sig {
   sealed abstract class Scope(
--- a/nir/src/main/scala/scala/scalanative/nir/Traverse.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Traverse.scala
@@ -1,0 +1,182 @@
+package scala.scalanative.nir
+
+trait Traverse {
+  def onDefns(defns: Iterable[Defn]): Unit = defns.foreach(onDefn)
+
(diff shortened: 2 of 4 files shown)


armanbilge:
Does this also fix the issue I had with `.getClass()` in https://github.com/scala-native/scala-native/pull/3102#discussion_r1083688761?

WojciechMazur:
> Does this also fix the issue I had with `.getClass()` in [#3102 (comment)](https://github.com/scala-native/scala-native/pull/3102#discussion_r1083688761)?

I have not checked it yet, but with quite a big possibility it should also fix them.
`.getClass` is using `this` in order to get `rtti`/ `Class[_]`. I expected it tried to execute `load[ptr] null` which lead to segfault. It was not used in any other way before calling `getClass` in the comparator, so null guard was not checked in other places. Seems to be the same scenario as in the reproducer

PR #3125, opened 2023-02-01
Update build #3

ekrich:
Take a more systematic approach based on previous attempts.

Leaving current PRs open for easy reference. There are various edge cases and problems with the previous attempt. https://github.com/scala-native/scala-native/pull/3110/files#diff-a610de80dc17f463fea5440deffcec5b2b70a7dbeef002624b94f6fa894f4b77

Interested folks should follow along in case they want to review or make suggestions.


ekrich:
@WojciechMazur In the previous PR, I found that avoiding the linking if all the inputs are older that the output means that in sbt where multiple main classes are allowed in the same project it won't relink. This also could be a problem, perhaps hidden, in the scripted test `java.net.socket` as it seems to run in the same directory.

I am looking for a more elegant solution but I could allow the user to set a boolean and then use the fully qualified main class as the basename - that is not foolproof but would work for the edge case.

WojciechMazur:
> In the previous PR, I found that avoiding the linking if all the inputs are older that the output means that in sbt where multiple main classes are allowed in the same project it won't relink.

Yes, I've observed some similar issues with my local changes even in the case of the same main (when running tests), so this technique is not very effective. Especially since we can make some changes at the start of nativeLink and save them (update modification date) before linking ends. In such a case, the next linking would be skipped. 
The most reliable solution would be to collect checksum for used sources at the very beginning of the linking and compare it in the succeeding invocations. 

It would be great if we would be able to provide a generic solution, without the need of specifying a dedicated option.

ekrich:
So normally the `sbt` `moduleName.value` which is unique for a project is used for the `basename` of the output file. We added `basename` to `NativeConfig` so you could override the `moduleName`.

It seems that only `sbt` supports multiple main classes so we can use the fully specified classname as the `basename` which has to be unique so this can be enabled by a boolean.

The current `ll` and `.c` to `.o` and the linking of `.o` the the executable work well and saves time so I hate to roll that back for certain special cases. We can set `withMultipleMains` to enable cases such as the `java.net.socket` scripted test or `clean` can be called between tests.

I have tested `withMultipleMains` in an existing Scala Native test program I had and it works well but rebuilds everytime in releaseFast mode because the single `ll` file is new each time. If we change to debug mode then we get the optimizations if we rerun but not if we switch because the code generation is not deterministic but that is upstream to the portion of the build being worked on. I think if we went to incremental compilation the IR to LL could let the ll avoid being compiled again but have not tested that yet. Incremental also depends on package name so that can have an influence.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -62,6 +62,19 @@ sealed trait NativeConfig {
...
+   *  not embed files with certain extensions, including ".c", ".h", ".scala"
+   *  and ".class".
+   */
+  def embedResources: Boolean
+
+  /** Base name for executable or library, typically the project name. */
+  def basename: String
+
+  /** Allow multiple main classes in the same project */
+  def multipleMains: Boolean
```
WojciechMazur, line 76: In interactive mode, every build tool can potentially have multiple main classes: 
* `sbt` is interactive by default
* `mill` has it, here's [example how it can be used](https://alvinalexander.com/scala/mill-build-tool/how-specify-main-method-in-repl/)  
*  `scala-cli` also has an interactive mode, allowing to specify the main class when running with `-i` / `--interactive` flag

Every currently supported build tool can potentially allow multiple main classes. Requiring to explicitly allow that seems not like a good development experience.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -150,7 +193,18 @@ object Config {
     override lazy val workdir: Path =
       basedir.resolve(s"native$nameSuffix")
 
-    override lazy val artifactPath: Path = {
+    override lazy val basename: String =
+      if (compilerConfig.multipleMains &&
+          compilerConfig.buildTarget == BuildTarget.application) {
+        mainClass.get
```
WojciechMazur, line 199: It might work but would lead to very long fully qualified names, because of that user might want to override the artifact name, but that's impossible since we calculate it. 

I wonder, what if we would have internal and external artifacts:
* internal artifact is a file stored in our `workdir` ("native" directory in the target). We could potentially store there raw artifacts, e.g. named as mainClass or `moduleName` + hash of mainClass and possibly other settings. After successful linking we would copy it to the external artifact path (copy instead of symlink because it's broken on Windows and makes it easier to share the output) 
* the external ones are based on the module name, its names does not change between subsequent links.
What do you think? 

Also because we have our workdir we can store store other significant information required for detecting if we can potentially skip the build (storage is cheap, read from local IO is relatively cheap)

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
ekrich: I wasn't aware of that. I asked someone about `mill` and got the answer that it doesn't work like `sbt` so I thought it was sbt only. Agree that it make the developer experience worse.

tools/src/main/scala/scala/scalanative/build/Config.scala
ekrich: I like that idea to use `moduleName` + hash of main and then the copy. We can certainly go with storing info about the build settings and caching and stuff but that can be done later as it shouldn't affect the API. Denys suggested something like this as well.

ekrich:
@WojciechMazur Based on this as well, we will need a more unified way to fix these underlying issues - https://github.com/scala-native/scala-native/pull/2998#discussion_r1023715815

It seems like we need to run a build type clean based on changes in the Config. I would like to solve this current setup and then attack that as a follow on. The last caching we had was all in the build and the goal of this is to remove the logic from the build.

ekrich:
@WojciechMazur So this works. The current failures are unrelated. I think perhaps the best approach is to generate into native as the full classname and then copying the executable back into the final location as the basename based file name. The way it is easy for the user to tell which executable is which if it is not sha encoded. Simpler too.

ekrich:
@WojciechMazur This is ready for another look. This removes the logic out of the build and works with compilation caching so it can save a developer a lot of time. We can do more in additional PRs I would think.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -94,50 +145,54 @@ sealed trait Config {
...
       basedir: Path,
       testConfig: Boolean,
+      moduleName: String,
+      mainClass: Option[String],
+      classPath: Seq[Path],
       logger: Logger,
       compilerConfig: NativeConfig
   ) extends Config {
-    def withNativelib(value: Path): Config =
-      copy(nativelib = value)
```
WojciechMazur, line 127: I wondered if the removal of it can have a negative impact on build tools integration. 
From the mill sources, I can see that it was never used for `0.4.x` versions ([first version of its integration](https://github.com/com-lihaoyi/mill/blob/82ea8738ef31f082a4afc2b553b2bee7a550b267/scalanativelib/worker/0.4/src/ScalaNativeWorkerImpl.scala)) so there should no problems with future upgrades.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -94,50 +145,54 @@ sealed trait Config {
...
-
-    def withMainClass(value: String): Config =
-      copy(mainClass = Option(value).filter(_.nonEmpty))
-
-    def withClassPath(value: Seq[Path]): Config =
-      copy(classPath = value)
```
WojciechMazur, line 133: Spurious changed: changed order of defintions

WojciechMazur (approved):
Looks good, thank you your improvements to both functional aspects and improving the documentation! 
I only have a few suggestions

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -7,15 +7,6 @@ import scala.scalanative.nir.Val
...
-  def gc: GC
-
-  /** Compilation mode. */
-  def mode: Mode
-
-  /** Build target for current compilation */
-  def buildTarget: BuildTarget
-
   /** The path to the `clang` executable. */
   def clang: Path
```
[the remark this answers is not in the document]
ekrich: Actually, I would like to only validate the clang version if we find "clang" otherwise let the the developer use whatever the want.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -7,15 +7,6 @@ import scala.scalanative.nir.Val
...
 
-  /** The garbage collector to use. */
-  def gc: GC
-
-  /** Compilation mode. */
-  def mode: Mode
-
-  /** Build target for current compilation */
-  def buildTarget: BuildTarget
-
```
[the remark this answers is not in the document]
ekrich: I tried to make the definitions in the same order so it was easier to follow so that was deliberate.

tools/src/main/scala/scala/scalanative/build/Validator.scala
```diff
@@ -0,0 +1,50 @@
...
+  /** Runs all the individual private validators
+   *
+   *  @param config
+   *    the pre-validation original [[Config]]
+   *  @return
+   *    potentially a modified [[Config]] that is valid
+   */
+  def validate(config: Config): Config = {
+    // side effecting
+    validateMainClass(config)
```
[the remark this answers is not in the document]
ekrich: Yes, that is currently done in discovery along with extracting the reported target triple IIRC.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -15,16 +15,47 @@ sealed trait Config {
...
+   */
+  def moduleName: String
+
+  /** Base name for executable or library, typically the project/module name
+   *  from the build tool [[#moduleName]] or can be overridden by the user with
+   *  [[NativeConfig#basename]]. This must be unique over all module names and
+   *  other `basename`s in the project. Delegated method to
+   *  [[NativeConfig#basename]]
+   */
+  def basename: String
```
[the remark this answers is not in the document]
ekrich: Both `basedir` and `basename` are like like that perhaps because of the UNIX command of the same name. I can change both of those to toggle. It would be more consistent.

tools/src/main/scala/scala/scalanative/build/Config.scala
ekrich: It now matches the order above for the fields.

tools/src/main/scala/scala/scalanative/build/Config.scala
ekrich: It looks like they have versioned `0.4` in the path so maybe they can do `0.5` next, hopefully.

ekrich:
@WojciechMazur Now I went and changed `basename` to `baseName` and `basedir` to `baseDIr` and now I found `workdir` which I suppose should be `workDir` now. I will change it now - I had to go back because a missed a couple of places of the others.

PR #3129, opened 2023-02-02
Port `java.util.concurrent.Atomic` types from JSR-166

### LINKED ISSUE #753: Add support for full atomic semantics of java.util.concurrent.{Atomic*}
I'm currently looking at implementing basic atomic operations for further concurrency work such as compare and swap for java.util.concurrent.Atomic

WojciechMazur:
Ports Java Atomic types and their related tests from JSR-166 
Includes: 
* `AtomicBoolean`
* `AtomicInteger` 
* `AtomicLong`
* `AtomicReference`
* `AtomicMarkableReference`
* `AtomicStampedReference`
* `AtomicIntegerArray`
* `AtomicLongArray`
* `AtomicReferenceArray`

The remaining `AtomicFieldUpdater` types were originally based on the reflection in the JVM. The solution to support them might be added in the future (they're required by `scala.collection.concurrent.TrieMap`)
Resolves #753

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/atomic/AtomicBoolean.scala
+++ b/javalib/src/main/scala/java/util/concurrent/atomic/AtomicBoolean.scala
@@ -1,34 +1,337 @@
+/*
+ * Based on JSR-166 originally written by Doug Lea with assistance
+ * from members of JCP JSR-166 Expert Group and released to the public domain,
+ * as explained at http://creativecommons.org/publicdomain/zero/1.0/
+ */
+
 package java.util.concurrent.atomic
 
-class AtomicBoolean(private[this] var value: Boolean) extends Serializable {
-  def this() = this(false)
+import scala.annotation.tailrec
+import scala.language.implicitConversions
+import scala.scalanative.annotation.alwaysinline
+import scala.scalanative.unsafe._
+import scala.scalanative.libc.atomic.memory_order._
+import scala.scalanative.libc.atomic.CAtomicByte
+import scala.scalanative.runtime.{fromRawPtr, MemoryLayout}
+import scala.scalanative.runtime.Intrinsics
 
+@SerialVersionUID(4654671469794556979L)
+class AtomicBoolean private (private var value: Byte) extends Serializable {
+
+  // Pointer to field containing underlying Byte.
+  @alwaysinline
+  private[concurrent] def valueRef: CAtomicByte = new CAtomicByte(
+    fromRawPtr(Intrinsics.classFieldRawPtr(this, "value"))
+  )
+
+  def this() = {
+    this(0.toByte)
+  }
+
+  def this(initialValue: Boolean) = {
+    this(if (initialValue) 1.toByte else 0.toByte)
+  }
+
+  private implicit def byteToBoolean(v: Byte): Boolean = v != 0
+  private implicit def booleanToByte(v: scala.Boolean): Byte = if (v) 1 else 0
+
+  /** Returns the current value, with memory effects of volatile read
+   *
+   *  @return
+   *    the current value
+   */
   final def get(): Boolean = value
 
-  final def compareAndSet(expect: Boolean, update: Boolean): Boolean = {
-    if (expect != value) false
-    else {
-      value = update
+  /** Atomically sets the value to {@code newValue} if the current value {@code
+   *  \== expectedValue}, with memory effects as specified by {@link
+   *  VarHandle#compareAndSet}.
+   *
+   *  @param expectedValue
+   *    the expected value
+   *  @param newValue
+   *    the new value
+   *  @return
+   *    {@code true} if successful. False return indicates that the actual value
+   *    was not equal to the expected value.
+   */
+  final def compareAndSet(
+      expectedValue: Boolean,
+      newValue: Boolean
+  ): Boolean = {
+    valueRef.compareExchangeStrong(expectedValue, newValue)
+  }
+
+  /** Possibly atomically sets the value to {@code newValue} if the current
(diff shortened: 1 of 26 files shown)


armanbilge:
> The remaining `AtomicFieldUpdater` types were originally based on the reflection in the JVM.

We are also using them in Cats Effect. But I assume they can be replaced with `Intrinsics`?

WojciechMazur:
> But I assume they (AtomicFieldUpdater) can be replaced with Intrinsics? 

To some extent yes, similarly as it is done in atomics implementation we can get the underlying field is use C atomic ops on it: 
```
  private[concurrent] def fieldRef: CAtomicByte = new CAtomicByte(
    fromRawPtr(Intrinsics.classFieldRawPtr(this, "fieldName"))
  )
```
However, I would like to experiment a bit so we can have `AtomicFieldUpdater` available somehow. `classFieldRawPtr` requires us to have the field name defined as a string literal, so we can hijack this value in the compiler plugin. Maybe we would need to come up with some kind up compat layer to make field updaters work.

WojciechMazur:
@armanbilge I've tried to experiment a bit, and it looks like we would be required to create some compat library, which would either dispatch to the reflection-based mechanism on the JVM, give stubs for Scala.js, or would use compile time resolution + some macros/inlines to work with `classFieldRawPtr`

POC
```scala
import scala.scalanative.runtime.fromRawPtr
import scala.scalanative.runtime.Intrinsics
import scala.scalanative.libc.atomic._

case class Bar(v: Int)

class Foo {
  var field: Bar = Bar(0)
}

object Test {
    val update = FieldUpdater.newUpdater(
      classOf[Foo],
      classOf[Bar],
      "field"
    )
    val inst = new Foo()
    
    println(inst.field)
    
    update.update(inst, Bar(42))
    
    println(inst.field)
  }
}

object FieldUpdater {
  inline def newUpdater[T <: AnyRef, U <: AnyRef](
      owner: Class[T],
      field: Class[U],
      inline fieldName: String with Singleton
  ) = new FieldUpdater[T, U]((ref: T) => fromRawPtr(Intrinsics.classFieldRawPtr(ref, fieldName)))
}

class FieldUpdater[T <: AnyRef, U <: AnyRef](selector: T => Ptr[U]) {
  def update(v: T, value: U) = new CAtomicRef[U](selector(v)).store(value)
}
```
In the debug mode there will be some overhead, because the closure was not optimized, but after optimizer run in releaseFastmode we get: 
```LLVM
didopt def @"M5Test$D4mainLAL16java.lang.String_uEO" : (@"T5Test$", array[@"T16java.lang.String"]) => unit {
%3000000(%1 : @"T5Test$", %2 : array[@"T16java.lang.String"]):
  %8000001 = module @"T14scala.Console$"

  // init and allocate instance
  %8000002 = classalloc @"T3Bar"
  %8000003 = call[(@"T14scala.Console$", @"T16java.lang.Object") => unit] @"M14scala.Console$D7printlnL16java.lang.ObjectuEO" : ptr(%8000001 : !?@"T14scala.Console$", %8000002 : !?@"T3Bar")
  %11000002 = classalloc @"T3Foo"
  %11000003 = fieldstore[@"T3Bar"] %11000002 : !?@"T3Foo", @"M3FooF5fieldO", %8000002 : !?@"T3Bar"
  %11000004 = bitcast[@"T3Foo"] %11000002 : !?@"T3Foo"
  
  // Here the selector was inlined into the `field` op
  %10000001 = field %11000004 : @"T3Foo", @"M3FooF5fieldO"
  
  // Init new value
  %10000002 = classalloc @"T3Bar"
  %10000003 = fieldstore[int] %10000002 : !?@"T3Bar", @"M3BarF1vPT3Bar", int 42
  %10000004 = call[(ptr, @"T16java.lang.Object") => unit]
  
  // actual update
   @"M30scala.scalanative.libc.atomic$C31scalanative_atomic_store_intptr" : ptr(%10000001 : ptr, %10000002 : !?@"T3Bar")
 
 // print result
  %12000001 = fieldload[@"T3Bar"] %11000002 : !?@"T3Foo", @"M3FooF5fieldO"
  %13000001 = call[(@"T14scala.Console$", @"T16java.lang.Object") => unit] @"M14scala.Console$D7printlnL16java.lang.ObjectuEO" : ptr(%8000001 : !?@"T14scala.Console$", %12000001 : @"T3Bar")
  ret unit
}
```

Alternatively we might think about adding new intrinsic for getting offset to the field for given class

armanbilge:
Thanks for experimenting with that!

---

> Alternatively we might think about adding new intrinsic for getting offset to the field for given class

Hmm, I thought this is what `Intrinsics.classFieldRawPtr(this, "fieldName")` is doing already? Are there some limitations to that?

---

> and it looks like we would be required to create some cross-platform compat library

FWIW I'm not sure how useful it is to strive for cross-compilability here. For example in Cats Effect, we are actually using the `Atomic*FieldUpdater` in Java code, because it's difficult/impossible to generate fields with the right access semantics in Scala.

https://github.com/typelevel/cats-effect/blob/7507b3119d873966d33d58e518c30a58d946b89e/core/jvm/src/main/java/cats/effect/unsafe/LocalQueue.java#L46-L49

And I see that Scala `TrieMap` is also relying on Java code for this purpose.
https://github.com/scala/scala/blob/de9fde542e0a8e76fa4c755cfaf2549896c387c6/src/library/scala/collection/concurrent/MainNode.java#L20-L21

WojciechMazur:
`classFieldRawPtr` is actually giving us ready-to-use a pointer to the underlying field in the instance of a class, rather than offset to that field. At the point when it was being added it sounded more reasonable and a bit more runtime-safe.

PR #3144, opened 2023-02-06
Refactor `scala.scalanative.runtime.ExecutionContext` 

WojciechMazur:
* Refactor `scala.scalanative.runtime.ExecutionContext`
  * Rename to `NativeExecutionContext` to match Scala.js naming conventions (JsExecutionContext)
  * Deprecated `global` accessor, use `queue` instead 
  * Add capability of running single task from `QueueExecutionContext` based on @lolgab work in #1852 
  * Add accessor for `hasNext`, `sheduled`, `runNext`
* Add tests for new interface of NativeExecutionContext
* Adapt `junit-async` and `test-interface` modules to use new interface 
* Depreacate package method `scala.scalanative.runtime.loop` 

Based on that change we would provide the `scala.concurrent.ExecutionContext` changes allowing to use real concurrent execution context in multithreading mode


nativelib/src/main/scala/scala/scalanative/runtime/NativeExecutionContext.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  /** Exuecute next available runnable in the queue. No-op if the queue is empty
+   */
+  private[runtime] def runNext(): Unit = QueueExecutionContext.runNext()
+
+  /** Returns the number of currently scheduled tasks */
+  private[runtime] def scheduled: Int = QueueExecutionContext.scheduled
+
+  /** Execute all the tasks in the queue until there is none left */
+  private[runtime] def loop(): Unit = QueueExecutionContext.loop()
```
armanbilge, line 59: I'm glad we are considering these APIs, but maybe can this happen in a separate PR? I think this deserves a discussion. Cats Effect might also benefit from this, but we have also been successful without it—it is not a blocker.

In Scala Contributors[^1] we discussed adding some new APIs to Scala standard library that will better support event loops. For example a `Scheduler` API. This appears to have support from Scala Center, and Cats Effect team is also motivated for this change.

If those changes land in Scala 2.13 standard library, then it means single-thread Scala Native will be able to support tasks and timers out-of-the-box. So we should be careful to design this API so that we can make those changes without breaking backwards-compatibly.

[^1]: https://contributors.scala-lang.org/t/return-of-the-scala-actor/6041/8

armanbilge (changes requested):

nativelib/src/main/scala/scala/scalanative/runtime/NativeExecutionContext.scala
WojciechMazur: Oh, that's a good point. In such a case, I'll be putting this on hold, and I would use the old API instead in further integration of multithreading.

WojciechMazur: I've reverted all the changes that expose new methods, I'll create a seperate PR with API extension

PR #3149, opened 2023-02-08
Allow to use `scala.collection.concurrent.TrieMap`

WojciechMazur:
* Implements java dependencies and adapts TrieMap implementation to use Intrinsic based atomic field updaters
* Removed to longer required Scala 2 sources filtering for `scala.collection.parallel` 
* Updated ignored lists of Scala2 partest junit tests to test TrieMap and other concurrent types
* Use TrieMap instead of not concurrent aware implementation of `java.util.concurrent.CouncurrentHashMap`

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
@@ -7,7 +7,7 @@ import scala.scalanative.unsafe._
 import scala.scalanative.meta.LinktimeInfo.isMultithreadingEnabled
 import scala.scalanative.runtime.libc.atomic_thread_fence
 import scala.scalanative.runtime.libc.memory_order._
-import java.util.concurrent.ConcurrentHashMap
+import scala.collection.concurrent.TrieMap
 
 trait NativeThread {
   import NativeThread._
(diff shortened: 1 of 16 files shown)


nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
```diff
@@ -94,7 +94,7 @@ object NativeThread {
     fromRawPtr[scala.Byte](castObjectToRawPtr(thread))
 
   object Registry {
-    private val _aliveThreads = new ConcurrentHashMap[Long, NativeThread]()
+    private val _aliveThreads = TrieMap.empty[Long, NativeThread]
```
armanbilge, line 97: Is there still a plan to port `ConcurrentHashMap` at a later point? I'd have to read the paper again but I think it makes different performance trade-offs than `ConcurrentHashMap` to support some special features.

WojciechMazur: Yes, at some point I wish we port all the collections from JSR-166. The implementation ported from Scala.js is great, but I'm worried that it's not safe enough for usage in a multithreading environment.

armanbilge: > The implementation ported from Scala.js is great, but I'm worried that it's not safe enough for usage in a multithreading environment.

Oh, definitely cannot use the current implementation in multi-threading :) we rely a lot on `ConcurrentHashMap` in Cats Effect.

PR #3179, opened 2023-02-22
Add libstdc++-12-dev to Ubuntu setup instructions

sideeffffect:
https://github.com/scala-native/scala-native/issues/3098


sideeffffect:
I think the tests aren't red because of my change.

WojciechMazur (approved):
Thanks for this update. Yes, the issues in the CI are due to experimental windows support and probably some changes to CI env on MacOS, it can be ignroed.

PR #3180, opened 2023-02-23
fix cygwin and msys build problems

philwalk:
These changes add support for building on Windows in `cygwin` or `msys`  environments.

the most significant changes include:

- convert backslashes to forward slashes in source file paths
- add methods to detect cygwin and msys targets
- for cygwin or msys targets, clang compiles are consecutive rather than parallel

The check in `LLVM.scala` for disabling parallel `clang` builds is a generally useful feature, so it should eventually be configurable for any environment.   It's helpful when diagnosing a failed `clang` command line, because avoid interleaving multiple `clang` compiles in the log output.

These changes don't affect other OS or targets.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Platform.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Platform.scala
@@ -33,12 +33,4 @@ object Platform {
 
   @name("scalanative_platform_is_msys")
   def isMsys(): Boolean = extern
-
-  @deprecated("Use windows.WinNlsApi to retrieve locale info instead", "0.4.1")
-  @name("scalanative_windows_get_user_lang")
-  def windowsGetUserLang(): CString = extern
-
-  @deprecated("Use windows.WinNlsApi to retrieve locale info instead", "0.4.1")
-  @name("scalanative_windows_get_user_country")
-  def windowsGetUserCountry(): CString = extern
 }
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Platform.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Platform.scala
@@ -30,4 +30,15 @@ object Platform {
 
   @name("scalanative_wide_char_size")
   final def SizeOfWChar: CSize = extern
+
+  @name("scalanative_platform_is_msys")
+  def isMsys(): Boolean = extern
+
+  @deprecated("Use windows.WinNlsApi to retrieve locale info instead", "0.4.1")
+  @name("scalanative_windows_get_user_lang")
+  def windowsGetUserLang(): CString = extern
+
+  @deprecated("Use windows.WinNlsApi to retrieve locale info instead", "0.4.1")
+  @name("scalanative_windows_get_user_country")
+  def windowsGetUserCountry(): CString = extern
 }
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -35,30 +35,27 @@ private[scalanative] object LLVM {
   def compile(config: Config, paths: Seq[Path]): Seq[Path] = {
     implicit val _config: Config = config
 
+    def compileIfNeeded(srcPath: Path): Path = {
+      val inpath = srcPath.abs
+      val outpath = inpath + oExt
+      val objPath = Paths.get(outpath)
+      // compile if out of date or no object file
+      if (needsCompiling(srcPath, objPath)) {
+        compileFile(srcPath, objPath)
+      } else objPath
+    }
     // generate .o files for included source files
     if (config.targetsMsys || config.targetsCygwin) {
       // TODO: should this be configurable in build.sbt?
       // sequentially; produces correct clang command lines in sbt -debug mode
       // clang command lines needed for quickly diagnosing failed compiles.
       paths.map { srcPath =>
-        val inpath = srcPath.abs
-        val outpath = inpath + oExt
-        val objPath = Paths.get(outpath)
-        // compile if out of date or no object file
-        if (needsCompiling(srcPath, objPath)) {
-          compileFile(srcPath, objPath)
-        } else objPath
+        compileIfNeeded(srcPath)
       }
     } else {
       // generate .o files for all included source files in parallel
       paths.par.map { srcPath =>
-        val inpath = srcPath.abs
-        val outpath = inpath + oExt
-        val objPath = Paths.get(outpath)
-        // compile if out of date or no object file
-        if (needsCompiling(srcPath, objPath)) {
-          compileFile(srcPath, objPath)
-        } else objPath
+        compileIfNeeded(srcPath)
       }.seq
     }
   }
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -36,7 +36,7 @@ private[scalanative] object LLVM {
     implicit val _config: Config = config
 
     // generate .o files for included source files
-    if (config.targetsMsys || config.targetsCygwin){
+    if (config.targetsMsys || config.targetsCygwin) {
       // TODO: should this be configurable in build.sbt?
       // sequentially; produces correct clang command lines in sbt -debug mode
       // clang command lines needed for quickly diagnosing failed compiles.
@@ -49,7 +49,7 @@ private[scalanative] object LLVM {
           compileFile(srcPath, objPath)
         } else objPath
       }
-    } else { 
+    } else {
       // generate .o files for all included source files in parallel
       paths.par.map { srcPath =>
         val inpath = srcPath.abs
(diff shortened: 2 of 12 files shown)


nativelib/src/main/scala/scala/scalanative/runtime/Platform.scala
```diff
@@ -30,4 +30,15 @@ object Platform {
 
   @name("scalanative_wide_char_size")
   final def SizeOfWChar: CSize = extern
+
+  @name("scalanative_platform_is_msys")
```
WojciechMazur, line 34: There is no foreign method defined with that name

tools/src/main/scala/scala/scalanative/build/Platform.scala
```diff
@@ -41,4 +41,21 @@ object Platform {
...
+
+  /** Test for the target type
+   *
+   *  @return
+   *    true if `cygnus`, false otherwise
+   */
+  lazy val isCygwin: Boolean = target.endsWith("cygnus")
+
+  private lazy val target =
+    System.getProperty("target.triple", "unknown").toLowerCase(Locale.ROOT)
```
WojciechMazur, line 60: That's a good idea, but `target.triple` is currently never set in system properties. 
We might use `java.lang.System.setProperty("target.triple", ???)` somewhere early in the build, but I'm not sure what would be the best place to put it. Probably in the `scala.scalanative.build.NativeConfig.Impl` constructor, because we would need to propagate this information on every update to NativeConfig. 
Because that `nativeConfig` might change, eg. when setting it manually in sbt, the target and the values depending on it should not by `lazy val` but `def`

WojciechMazur (changes requested):
Thank you for this PR, there're a fix issues to fix before merging it. 
Also, it would be great if we would be able to test it in the CI - do you know about some workflow examples that we could use to set it up for ScalaNative CI? I've never used neither cygwin or msys and have no idea what would be required.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -35,16 +35,32 @@ private[scalanative] object LLVM {
...
-      val inpath = srcPath.abs
-      val outpath = inpath + oExt
-      val objPath = Paths.get(outpath)
-      // compile if out of date or no object file
-      if (needsCompiling(srcPath, objPath)) {
-        compileFile(srcPath, objPath)
-      } else objPath
-    }.seq
+    // generate .o files for included source files
+    if (config.targetsMsys || config.targetsCygwin) {
```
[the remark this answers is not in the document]
philwalk: The problem is that, in order to efficiently diagnose `clang` or `clang++` compile errors, the input and output of the compile session must be examined, by launching `sbt` with the `-debug` argument.  However, the output of multiple parallel compilations are interleaved and unreadable.     In some cases (for example, `sandbox3/run`), the time savings of running in parallel are insignificant and not worth the confused log output.

However, it would be better if we could enable and disable parallel execution by a setting in the `sbt` configuration, but I don't know enough about how to do it that way, so this is an expedient that will permit contributing developers to more easily diagnose problems when resolving the remaining issues with `MSYS2` and `CYGWIN` builds.

tools/src/main/scala/scala/scalanative/build/Platform.scala
philwalk: What about something like this (in scala.scalanative.build.NativeConfig.Impl line 255)?
```scala
    def withTargetTriple(value: Option[String]): NativeConfig =
      copy(targetTriple = value)
      System.setProperty("target.triple", value.getOrElse(""))
```

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -67,6 +83,7 @@ private[scalanative] object LLVM {
       } else Seq("-std=gnu11")
     }
     val platformFlags = {
+      if (config.targetsMsys) msysExtras
       if (config.targetsWindows) Seq("-g")
       else Nil
```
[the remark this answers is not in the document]
philwalk: done, with minor tweak required by `scripts/check-lint.sh`

tools/src/main/scala/scala/scalanative/build/Platform.scala
philwalk: I can set the target.triple like this:

```bash
sbt 'set nativeConfig ~= {_.withTargetTriple("x86_64-pc-linux-gnu")}; show nativeConfig'
```

but the value reported by `clang` would be a good default value.

nativelib/src/main/scala/scala/scalanative/runtime/Platform.scala
```diff
@@ -30,4 +30,15 @@ object Platform {
...
+  def isMsys(): Boolean = extern
+
+  @deprecated("Use windows.WinNlsApi to retrieve locale info instead", "0.4.1")
+  @name("scalanative_windows_get_user_lang")
+  def windowsGetUserLang(): CString = extern
+
+  @deprecated("Use windows.WinNlsApi to retrieve locale info instead", "0.4.1")
+  @name("scalanative_windows_get_user_country")
+  def windowsGetUserCountry(): CString = extern
```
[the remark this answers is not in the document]
philwalk: removed

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -35,16 +35,32 @@ private[scalanative] object LLVM {
...
+      // sequentially; produces correct clang command lines in sbt -debug mode
+      // clang command lines needed for quickly diagnosing failed compiles.
+      paths.map { srcPath =>
+        val inpath = srcPath.abs
+        val outpath = inpath + oExt
+        val objPath = Paths.get(outpath)
+        // compile if out of date or no object file
+        if (needsCompiling(srcPath, objPath)) {
+          compileFile(srcPath, objPath)
+        } else objPath
```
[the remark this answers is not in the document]
philwalk: Am in the process of re-testing the change, should be able to commit sometime today.

nativelib/src/main/scala/scala/scalanative/runtime/Platform.scala
philwalk: method has been added to `./nativelib/src/main/resources/scala-native/platform/platform.c`

tools/src/main/scala/scala/scalanative/build/Platform.scala
philwalk: had to remove the default `targetTriple`, it was causing test failures that I was unable to diagnose.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -251,8 +251,10 @@ object NativeConfig {
     def withCompileOptions(value: Seq[String]): NativeConfig =
       copy(compileOptions = value)
 
-    def withTargetTriple(value: Option[String]): NativeConfig =
+    def withTargetTriple(value: Option[String]): NativeConfig = {
+      System.setProperty("target.triple", value.getOrElse(""))
```
WojciechMazur, line 255: Let's set this system property only when it is defined instead of setting empty string. Even better lets clear it when changing target triple Some(???) -> None, this way we would not get into troubles when modifying config
```suggestion
      val propertyName = "target.triple"
      value match { 
         case Some(triple) => System.setProperty(propertyName, triple)
         case None =>         System.clearProperty(propertyName)
       }
```

WojciechMazur (commented):
Thank you for applying the fixes, I only have a few last comments before merging.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
philwalk: makes sense.
I will run tests and hope to be able to push these changes before too long ...

philwalk: done

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -251,8 +251,14 @@ object NativeConfig {
     def withCompileOptions(value: Seq[String]): NativeConfig =
       copy(compileOptions = value)
 
-    def withTargetTriple(value: Option[String]): NativeConfig =
+    def withTargetTriple(value: Option[String]): NativeConfig = {
+      val propertyName = "target.triple"
```
ekrich, line 255: If this data is already in the `targetTriple`, why are we setting a system property? I am a little late reviewing this so I may have missed something.

WojciechMazur (approved):

philwalk:
> Also, it would be great if we would be able to test it in the CI - do you know about some workflow examples that we could use to set it up for ScalaNative CI?
I'm investigating, although I don't know how soon I might have something

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
philwalk: Sorry for the delayed reply.
This is not quite right, as you suspected.   I hope to submit another PR soon.

PR #3186, opened 2023-02-25
Allow to use libunwind on Windows

WojciechMazur:
This change adapts libunwind sources to allow for compilation on Windows - which should fix some of the issues related to empty expectation stack traces which were using build Windows build-in tools. The sources have been adapted to silence warnings and fixes allowing for compilation without CMake config (used in the upstream LLVM project)

This PR should be rebased on top of #3180 and hopefully also tested on Cygwin/MSys2 when it would be available in our CI workflow


WojciechMazur:
Seems like libunwind is producing empty stack traces on Windows. I'm converting it into WIP until it is solved

PR #3187, opened 2023-02-25
Various fixes to runtime and compile time multithreading issues

WojciechMazur:
* Use `STACK_SIZE_PARAM_IS_A_RESERVATION` attribute for created Windows threads to replace creating threads with committed memory equal to stack size
* Add explicit `WINAPI` preprocessor definition to enforce correct calling convention of signal handlers and thread routines
* `setjmp` uses `jmp_buf` allocated in MutatorThread, instead of creating local value and copying it's content
* Use `VectoredExceptionHandler` with the highest priority instead of `UnhandledExceptionFilter` to yield safepoints
* Process runner prints both decimal and hexadecimal exit code of the process for easier resolving of unhandled Windows exception codes
* Fix to `FilesTestOnJDK11` test not containing `new` keyword when allocating `java.lang.StringBuilder`
* Fix end conditions of `ObjectMonitorTest` and remove no longer needed exclusion for testing on JVM with Windows

### DIFF
--- a/unit-tests/shared/src/test/require-jdk11/org/scalanative/testsuite/javalib/nio/file/FilesTestOnJDK11.scala
+++ b/unit-tests/shared/src/test/require-jdk11/org/scalanative/testsuite/javalib/nio/file/FilesTestOnJDK11.scala
@@ -205,7 +205,7 @@ class FilesTestOnJDK11 {
     val ioPath = getCleanIoPath("StringBuilderSmall")
     val dataOut = getDataOut()
 
-    val output = jl.StringBuilder(dataOut)
+    val output = new jl.StringBuilder(dataOut)
     Files.writeString(ioPath, output)
 
     verifySmallUtf8Payload(ioPath)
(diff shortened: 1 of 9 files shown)


unit-tests/shared/src/test/require-jdk11/org/scalanative/testsuite/javalib/nio/file/FilesTestOnJDK11.scala
```diff
@@ -205,7 +205,7 @@ class FilesTestOnJDK11 {
     val ioPath = getCleanIoPath("StringBuilderSmall")
     val dataOut = getDataOut()
 
-    val output = jl.StringBuilder(dataOut)
```
ekrich, line 208: How did this work at all? Is there a problem the Java API?

ekrich: `AbstractStringBuilder` has this but no args.
```scala
object AbstractStringBuilder {
  final val INITIAL_CAPACITY = 16
}
```
Can't imagine that it would compile.

PR #3199, opened 2023-03-07
Restore stack state after executing inlined function to prevent stack overflows

### LINKED ISSUE #3195: Stack memory might leak when allocating inside of loop
Let's take a look at the following snippet of code: 

```scala
    val base = stackalloc[Int]()
    while(true){
      val inner = stackalloc[Int]()
    }
```
We might expect that the stack of the frame would effectively allocate 2 integers on the stack. That's e.g. what would happen in C. However, our code would allocate `base` once as expected, however `innner` would be allocated on each iteration of the loop. Eventually, our program would fail due to stack overflow. That's one of the reasons for failures in Windows CI, when threads are trying to enter locked object monitors.

The problem here actually has 2 cases: 
* <del> [ ] - in fixed size allocation all `stackalloc` instructions should be moved to the entry of the method (I have a working prototype with a fix for that, but also requiring other changes)</del>
#edit: Actually that one is incorrect. If we would move all stackallocation to the top, we would not allow for initialization of multiple objects inside a while loop
* [ ] - in  dynamic size allocations we should use LLVM intrinsic  to restore state of the stack from before the allocation. C uses for that `llvm.stack. Here example of C code emitted to LLVM IR:  
```LLVM 
  %8 = alloca ptr, align 8
  %14 = call ptr @llvm.stacksave()
  store ptr %14, ptr %8, align 8
  ... 
  %20 = load ptr, ptr %8, align 8
  call void @llvm.stackrestore(ptr %20)
```

WojciechMazur:
Fixes #3195 responsible for CI failures of Windwos build with enabled multithreaidng. 

Usage of `stackalloc` in inlined functions called inside a loop were not resetting the stack pointer after exiting the inlined function. This behavior, combined with calling the inlined function in loop could have lead to stack overflow. 

This change modifies interflow optimizer to check if inlined function uses `stackalloc`. If it does, a body of inlined function would be guarded with set of LLVM intrinsic calls `llvm.stacksave` and `llvm.stackrestore` ensuring the stack pointer would behave as the function would not be inlined.

### DIFF
--- a/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
+++ b/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
@@ -592,7 +592,7 @@ class IssuesTest {
   }
 
   @Test def i3195(): Unit = {
-    // Make sure that inlined calls are reseting the stack upon returning
+    // Make sure that inlined calls are resetting the stack upon returning
     // Otherwise calls to functions allocating on stack in loop might lead to stack overflow
     @alwaysinline def allocatingFunction(): CSize = {
       import scala.scalanative.unsafe.{CArray, Nat}
--- a/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
+++ b/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
@@ -7,6 +7,7 @@ import org.scalanative.testsuite.utils.AssertThrows.assertThrows
 import scalanative.unsigned._
 import scalanative.unsafe._
 import scala.annotation.nowarn
+import scala.scalanative.annotation.alwaysinline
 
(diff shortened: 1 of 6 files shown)


unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
```diff
@@ -590,6 +591,25 @@ class IssuesTest {
     assertEquals(Foo.fooLiteral, Foo.fooLazy)
   }
 
+  @Test def i3195(): Unit = {
+    // Make sure that inlined calls are reseting the stack upon returning
```
ekrich, line 595:
```suggestion
    // Make sure that inlined calls are resetting the stack upon returning
```

Issue #3219, opened 2023-03-12
Runtime#gc() and System.gc() should do the same thing


JD557:
From https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#gc--

> The method [System.gc()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#gc--) is the conventional and convenient means of invoking this method.

And from https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#gc--

> The call System.gc() is effectively equivalent to the call: `Runtime.getRuntime().gc()`

However, currently, `Runtime#gc()` does nothing:

https://github.com/scala-native/scala-native/blob/e6fdf3ef3ce0763f0d60082438f6fe1af86e0872/javalib/src/main/scala/java/lang/Runtime.scala#L24

While `System.gc()` calls `GC.collect()`:

https://github.com/scala-native/scala-native/blob/e6fdf3ef3ce0763f0d60082438f6fe1af86e0872/javalib/src/main/scala/java/lang/System.scala#L136

PR #3224, opened 2023-03-14
Trigger GC collction on Runtime#gc()

JD557:
Fix https://github.com/scala-native/scala-native/issues/3219

### DIFF
--- a/javalib/src/main/scala/java/lang/System.scala
+++ b/javalib/src/main/scala/java/lang/System.scala
@@ -6,7 +6,7 @@ import java.util.{Collections, HashMap, Map, Properties, WindowsHelperMethods}
 import scala.scalanative.posix.pwdOps._
 import scala.scalanative.posix.{pwd, unistd}
 import scala.scalanative.meta.LinktimeInfo.isWindows
-import scala.scalanative.runtime.{GC, Intrinsics, Platform, time}
+import scala.scalanative.runtime.{Intrinsics, Platform, time}
 import scala.scalanative.unsafe._
 import scala.scalanative.unsigned._
 import scala.scalanative.windows.FileApi._
@@ -133,7 +133,8 @@ object System {
   def setErr(err: PrintStream): Unit =
     this.err = err
 
-  def gc(): Unit = GC.collect()
+  def gc(): Unit =
+    Runtime.getRuntime().gc()
 
   private def getCurrentDirectory(): Option[String] = {
     val bufSize = 1024.toUInt
--- a/javalib/src/main/scala/java/lang/System.scala
+++ b/javalib/src/main/scala/java/lang/System.scala
@@ -6,7 +6,7 @@ import java.util.{Collections, HashMap, Map, Properties, WindowsHelperMethods}
 import scala.scalanative.posix.pwdOps._
 import scala.scalanative.posix.{pwd, unistd}
 import scala.scalanative.meta.LinktimeInfo.isWindows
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/lang/System.scala
```diff
@@ -133,7 +133,8 @@ object System {
   def setErr(err: PrintStream): Unit =
     this.err = err
 
-  def gc(): Unit = GC.collect()
```
ekrich, line 136: I reviewed the documentation and can't see a reason for this change.

ekrich: Looks like it needs a `scalafmt` as well.

JD557: > I reviewed the documentation and can't see a reason for this change.

The reason for this is to make sure that both methods are always in sync (and from the documentation, `Runtime.getRuntime().gc()` seems to be the authoritary method).
Who knows, maybe in the future there's a `GC.collectEden()` or a `GC.collectFull()` and only one of the methods gets updated?
Not that I think anyone will change in the near future, but doing it this way seemed a bit more future proof (and I don't think having an extra function call will be a bottleneck when forcing the GC)

> Looks like it needs a `scalafmt` as well.

Are you saying because of the `check-lint`? I think it failed with a `scripts/.coursier: line 1: `<h1>Internal Server Error</h1>'` (if not, maybe I messed up somewhere, but I think I reformatted both files)

ekrich: We have a `scripts/scalafmt` you can run.

It is a little unclear - the is that hint in the javadoc - https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#gc() I think rather than create and instance of `Runtime` maybe `System.gc()` can be the reference and call the GC and the Runtime can call that method. This seems simpler and avoids the allocation. If you use the `Runtime` method, you have to get the instance. WDYT?

JD557: Makes sense, reversed the logic in ee52dec

ekrich:
I am wondering if you tested both methods in a small app the did some allocations and `gc()` calls in a loop? You can use sandbox to just make sure it works - technically we could just add a test to `System` and since `Runtime` calls system, maybe we don't need one there. Looks good though, to me at least.

https://github.com/scala-native/scala-native/tree/main/unit-tests/shared/src/test/scala/org/scalanative/testsuite/javalib/lang

javalib/src/main/scala/java/lang/System.scala
ekrich: Thanks for this PR :-)

JD557:
> I am wondering if you tested both methods in a small app the did some allocations and `gc()` calls in a loop? You can use sandbox to just make sure it works - technically we could just add a test to `System` and since `Runtime` calls system, maybe we don't need one there. Looks good though, to me at least.
> 
> https://github.com/scala-native/scala-native/tree/main/unit-tests/shared/src/test/scala/org/scalanative/testsuite/javalib/lang

I did test this in an app (that's how I noticed the problem). I tried to call the GC on each frame to see if it would smoothen the frame pacing.
The app slowed down to a crawl, so I think the GC is being called correctly 😬

Unfortunately, I am not sure how to write an unit test to check this

WojciechMazur (approved):

ekrich:
@WojciechMazur Will this be merged to 0.5.x too?

PR #3226, opened 2023-03-16
Fix regression in handling opaque pointers on Windows

WojciechMazur:
Windows was not tested with LLVM 15+ allowing to introduce regression with new handling for opaque pointers

### DIFF
--- a/.github/actions/windows-setup-env/action.yml
+++ b/.github/actions/windows-setup-env/action.yml
@@ -7,6 +7,9 @@ inputs:
   java-version:
     description: "Java version to use in tests"
     default: "8"
+  llvm-version:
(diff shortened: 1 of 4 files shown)


.github/actions/windows-setup-env/action.yml
```diff
@@ -7,6 +7,9 @@ inputs:
   java-version:
     description: "Java version to use in tests"
     default: "8"
+  llvm-version:
+    description: "Java version to use in tests"
```
ekrich, line 11: Does this description need changing?

WojciechMazur: Thank you, I've forget to edit copied setting

PR #3227, opened 2023-03-16
Update sbt to 1.5.8 for scripted tests

ekrich:
As discussed in #3110, update minimum `sbt` version. Version `1.5.0` is the first that supports Scala 3 and `1.5.8` is the last in the series and also has the log4f vulnerability fixed. `sbt` `1.5.8` was released on Dec 20, 2021.


ekrich:
@WojciechMazur You are more confident than I am - I will believe it if all the tests pass.

PR #3230, opened 2023-03-16
Improve resolving repeated parameters in Scala3

### LINKED ISSUE #3228: Constant value definition in `@extern object` raises warning in Scala 3
```scala
//> using platform "native"
//> using nativeVersion "0.4.11"

@extern
object Foo {
  final val bar = 42
}
```

```
Compiling project (Scala 3.2.2, Scala Native)
[warn] ./test.scala:8:13
[warn] Unable to resolve method sig params for symbol, extern VarArgs would not work
[warn]   final val bar = 42
[warn]             ^
```

The same code compiles without warning in 2.13.

WojciechMazur:
Fixes #3228 

Typically extern objects should not be allowed to contain any non-extern fields of methods. However, `final val` was accepted in the previous version for some reason.

### DIFF
--- a/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenType.scala
+++ b/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenType.scala
@@ -311,27 +311,18 @@ trait NirGenType(using Context) {
     import core.Phases._
     val repeatedParams = if (sym.isExtern) {
       atPhase(typerPhase) {
-        sym.paramInfo match {
+        sym.paramInfo.stripPoly match {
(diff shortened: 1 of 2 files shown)


armanbilge:
> However, `final val` was accepted in the previous version for some reason.

I always assumed this worked because `final val` with a literal and no type annotation is a "constant value definition" that the compiler inlines at use-site.

https://www.scala-lang.org/files/archive/spec/2.13/04-basic-declarations-and-definitions.html

ekrich:
Just a couple of nits not on the PR. The title should have `repeated` in the description. I think it is late there :-)

This as well.
> In general extenr object

PR #3243, opened 2023-04-02
Synchronise ForkJoinPool port with JSR-166 changes for JDK 19

WojciechMazur:
JSR-166 had a recent update changing the implementation of ForkJoinPool with a goal of better support for VirtualThreads, it also adds new methods and interfaces, eg.: 

* `ExecutorService` now extends `AutoClosable`
* `java.util.concurrent.Future` now defines a new enum type `Future.State` and  methods `state(): State`, `resultNow(): V`, `exceptionNow(): Throwable`
* New protected constructor for `ForkJoinPool`

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/Future.scala
+++ b/javalib/src/main/scala/java/util/concurrent/Future.scala
@@ -1,3 +1,4 @@
+// revision 1.47
 /*
  * Written by Doug Lea with assistance from members of JCP JSR-166
  * Expert Group and released to the public domain, as explained at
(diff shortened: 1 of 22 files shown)


javalib/src/main/scala/java/util/concurrent/Future.scala
```diff
@@ -23,4 +24,100 @@ trait Future[V] {
...
+      throw new IllegalStateException("Task was cancelled")
+    var interrupted = false
+    try
+      while (true) {
+        try {
+          get()
+          throw new IllegalStateException("Task completed with a result")
+        } catch {
+          case _: InterruptedException => interrupted = true
+          case e: ExecutionException   => e.getCause()
```
bishabosha, line 64: it seems this never returns?
```suggestion
          case e: ExecutionException   => return e.getCause()
```

WojciechMazur: Fixed and ensured by adding tests

PR #3253, opened 2023-04-19
Remove most of the lint warnings in main sources when compiling with `-WwarnUnused:all`

WojciechMazur:
Removes most of the warnings in the main sources detected in Scala 3.3.0-RC4 -WwarnUnused, there are still ~250 warnings in tests, and ~50 warnings in main sources, due to either false positives (2), unusused parameters in constructors (for compliance with JDK API), parameters in annotations, and a few cases witch should be annotated with `@nowarn`

Notes: 
At this moment it's too early to enable `-WwarnUnused` settings in default scalacOptions if we continue to compile with `-Xfatal-warnings`.

### DIFF
--- a/clib/src/main/scala/scala/scalanative/libc/atomic.scala
+++ b/clib/src/main/scala/scala/scalanative/libc/atomic.scala
@@ -1,10 +1,9 @@
 // format: off
 package scala.scalanative.libc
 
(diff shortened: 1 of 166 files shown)


ekrich:
Pretty impressive 👍

Issue #2802, opened 2022-08-27
Add more platform checks to `LinktimeInfo`


armanbilge:
Currently only `isWindows` is offered.
https://github.com/scala-native/scala-native/blob/2f2d322f29239bd77220e52c4ecd5833fbe0132c/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala#L10-L11

I would find more of these useful: specifically I am seeking `isLinux` and `isBSD` so that I can conditionally call either the [`epoll`](https://man7.org/linux/man-pages/man7/epoll.7.html) or [`kqueue`](https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2) APIs.

But there are an infinite number of platforms, so there are important questions here:
1. how do we decide which are important enough to include in `LinktimeInfo`?
2. or, can we support userland-extensibility? as teased in the [0.4.1 release notes](https://scala-native.readthedocs.io/en/latest/changelog/0.4.1.html).
  > Currently, the list of linktime resolved expressions is limited and users cannot create new ones, but it might change in the future.

Alternatively: is there any way we could expose the target-triple string in the `LinktimeInfo` for useful link-time checks to be written against it in user code?

armanbilge:
Here are some additional ideas from Wojciech on Discord:

> Working with the raw target triple might be really hard, if we would like to allow for ops, like .contains or .startsWith . It would require changes in compiler plugin or in the linker. I think maybe a better idea would be parsing the target triple and exposing it's fragments in the linktime infos. Then we would be able to easily match them with == operator. Here, we can find llvm API including all possible parts of the target triple, parsing them manually might be really error prone https://llvm.org/doxygen/classllvm_1_1Triple.html

> One option would be to use JNI to access the LLVM parser directly. I don't have an experience with that, and I worry about the portability. The safest way would be reimplementing parser in the Scala, but then it might require a maintance in the future. In theory it's should be that hard

armanbilge:
And from Denys:
> In my experience it’s always easier to launch a C/C++ subprocess than use JNI. (For one, there is no easy way to distribute JNI code portably.) 
For example we could write a tiny cli tool in C++ that calls llvm APIs and prints it to stdout in some structured format (json?). SN sbt plugin will compile it on user machine and invoke it as a sub process to extract all the platform info and parse the output on JVM side.

JD557:
I think this can be closed now (it was closed by #2809)

armanbilge:
[...]

PR #3258, opened 2023-04-26
Expose parsed target triple in `LinktimeInfo`

armanbilge:
Closes https://github.com/scala-native/scala-native/issues/2802. Closes https://github.com/scala-native/scala-native/issues/2987.

1. Ports the target triple parser/normalizer from LLVM.
2. Extracts and normalizes the arch, vendor, os, and env components of a triple.
3. Exposes this under `LinktimeInfo.target` for link-time conditionals.

---

_Original description._

<details>

This is a first step towards https://github.com/scala-native/scala-native/issues/2802#issuecomment-1230342862, where we discussed exposing the parsed target triple in `LinktimeInfo`, so that the components can be compared with `==`.

I was not sure how to write a useful test for this. Since I think the architecture in the target triple may not match the architecture detected at runtime, in general 🤔 

**Motivation** in the Linux epoll API, the layout of `epoll_event` struct is different for `x86_64`. I'd like to be able to access this info at linking time (and not runtime).

https://github.com/torvalds/linux/blob/61d325dcbc05d8fef88110d35ef7776f3ac3f68b/include/uapi/linux/eventpoll.h#L71-L81

</details>

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -24,9 +24,6 @@ object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isFreeBSD")
   def isFreeBSD: Boolean = resolved
 
-  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.arch")
-  def arch: String = resolved
-
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.is32BitPlatform")
   def is32BitPlatform: Boolean = resolved
 
@@ -42,4 +39,15 @@ object LinktimeInfo {
     "scala.scalanative.meta.linktimeinfo.isMultithreadingEnabled"
   )
   def isMultithreadingEnabled: Boolean = resolved
+
+  object target {
+    @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.target.arch")
+    def arch: String = resolved
+    @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.target.vendor")
+    def vendor: String = resolved
+    @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.target.os")
+    def os: String = resolved
+    @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.target.env")
+    def env: String = resolved
+  }
 }
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -12,17 +12,17 @@ object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.releaseMode")
   def releaseMode: Boolean = resolved
 
-  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isWindows")
-  def isWindows: Boolean = resolved
+  @resolvedAtLinktime
+  def isWindows: Boolean = target.os == "windows"
 
-  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isLinux")
-  def isLinux: Boolean = resolved
+  @resolvedAtLinktime
+  def isLinux: Boolean = target.os == "linux"
 
-  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isMac")
-  def isMac: Boolean = resolved
+  @resolvedAtLinktime
+  def isMac: Boolean = target.vendor == "apple" && target.os == "darwin"
 
-  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isFreeBSD")
-  def isFreeBSD: Boolean = resolved
+  @resolvedAtLinktime
+  def isFreeBSD: Boolean = target.os == "freebsd"
 
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.is32BitPlatform")
   def is32BitPlatform: Boolean = resolved
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -24,6 +24,9 @@ object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isFreeBSD")
   def isFreeBSD: Boolean = resolved
 
+  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.arch")
+  def arch: String = resolved
+
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.is32BitPlatform")
   def is32BitPlatform: Boolean = resolved
 
--- a/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
+++ b/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
@@ -76,7 +76,7 @@ sealed trait NativeConfig {
   def optimizerConfig: OptimizerConfig
 
   private[scalanative] lazy val configuredOrDetectedTriple =
-    targetTriple.getOrElse(Discover.targetTriple(this))
+    TargetTriple.parse(targetTriple.getOrElse(Discover.targetTriple(this)))
 
   /** Are we targeting a 32-bit platform?
    *
@@ -85,10 +85,7 @@ sealed trait NativeConfig {
    *  architecture for a name that is not found seems excessive perhaps?
    */
   def is32BitPlatform = {
-    configuredOrDetectedTriple
-      .split('-')
-      .headOption
-      .getOrElse("") match {
+    configuredOrDetectedTriple.arch match {
       case "x86_64"  => false
       case "aarch64" => false
       case "arm64"   => false
--- a/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
+++ b/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
@@ -75,7 +75,8 @@ sealed trait NativeConfig {
   /** Configuration when doing optimization */
   def optimizerConfig: OptimizerConfig
 
-  private lazy val detectedTriple = Discover.targetTriple(this)
+  private[scalanative] lazy val configuredOrDetectedTriple =
+    targetTriple.getOrElse(Discover.targetTriple(this))
 
   /** Are we targeting a 32-bit platform?
    *
@@ -84,8 +85,7 @@ sealed trait NativeConfig {
    *  architecture for a name that is not found seems excessive perhaps?
    */
   def is32BitPlatform = {
-    targetTriple
-      .getOrElse(detectedTriple)
+    configuredOrDetectedTriple
       .split('-')
       .headOption
       .getOrElse("") match {
--- a/tools/src/main/scala/scala/scalanative/build/TargetTriple.scala
+++ b/tools/src/main/scala/scala/scalanative/build/TargetTriple.scala
@@ -28,10 +28,10 @@ private[scalanative] object TargetTriple {
     val components = triple.split("-", 4).toList
     val unknown = "unknown"
     TargetTriple(
-      components.unapply(0).map(Arch.parse).getOrElse(unknown),
-      components.unapply(1).map(Vendor.parse).getOrElse(unknown),
-      components.unapply(2).map(OS.parse).getOrElse(unknown),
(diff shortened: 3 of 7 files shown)


nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
```diff
@@ -24,6 +24,9 @@ object LinktimeInfo {
   @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isFreeBSD")
   def isFreeBSD: Boolean = resolved
 
+  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.arch")
+  def arch: String = resolved
```
armanbilge, line 28: Another possible name: `targetArch`?

ekrich:
I am not sure what the best solution is for this. We removed the "real" target triple detection to allow for setting the target triple and the detected target triple does not necessarily pick the exact target triple so not sure what problems we will get into. Would it make sense to have an object for the components? <arch><sub>-<vendor>-<sys>-<abi> Not sure what the parsing would be exactly.

I also think we need to experiment along these lines if the code has to be in Scala on the extern side. I am not seeing how to get around these types of problems. Lee has some asserts on the C side but not sure that is your use case.

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
WojciechMazur: how about a dedicated object?
```scala
object target {
 def arch: String = resolved 
 // in the future
 def subArch: String = resolved
 def vendor: String = resolved
 def environment: String = resolved
}
```

tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala
```diff
@@ -17,6 +17,12 @@ trait LinktimeValueResolver { self: Reach =>
       s"$linktimeInfo.isLinux" -> Platform.isLinux,
       s"$linktimeInfo.isMac" -> Platform.isMac,
       s"$linktimeInfo.isFreeBSD" -> Platform.isFreeBSD,
+      s"$linktimeInfo.arch" -> {
+        config.compilerConfig.configuredOrDetectedTriple
+          .split('-')
```
WojciechMazur, line 22: This might be problematic in the case of architectures that define subparts, e.g. arm, aarch64, but at this point when we mainly target x86 it should be enough. However we might need to ensure stability of `x86` which on some platforms is defined as `x86` on others as `i386` based on the source code from the LLVM I would opt for the latter

WojciechMazur:
Currently it might be just enough, but in the future we might port the triple parsing from the LLVM, it shouldn't be that hard. https://github.com/llvm/llvm-project/blob/main/llvm/lib/TargetParser/Triple.cpp

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
ekrich: I fixed my comments below - we could see what LLVM names they to keep consistent.

tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala
ekrich: I see you are way ahead of me.

tools/src/main/scala/scala/scalanative/build/TargetTriple.scala
```diff
@@ -0,0 +1,453 @@
...
+  def parse(triple: String): TargetTriple = {
+    val components = triple.split("-", 4).toList
+    val unknown = "unknown"
+    TargetTriple(
+      components.lift(0).map(Arch.parse).getOrElse(unknown),
+      components.lift(1).map(Vendor.parse).getOrElse(unknown),
+      components.lift(2).map(OS.parse).getOrElse(unknown),
+      components.lift(3).map(Env.parse).getOrElse(unknown)
+    )
+  }
```
armanbilge, line 36: This parser is a bit naive e.g. it will not work correctly for e.g. `x86_64-linux-gnu` where the vendor is omitted. However it seems to work fine for the discovered target triple emitted by LLVM. If the user configures it manually they should be careful to specify all the parts e.g. `x86_64-unknown-linux-gnu`.

In the future we can do something more complex if necessary.

tools/src/test/scala/scala/scalanative/build/TargetTripleTest.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import org.scalatest._
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class TargetTripleTest extends AnyFlatSpec with Matchers {
+
+  val cases = List(
+    "aarch64-unknown-linux-gnu" ->
+      TargetTriple("aarch64", "unknown", "linux", "gnu"),
```
armanbilge, line 11: I discovered all these test cases by running `clang --version` on different machines and CI runners.

tools/src/test/scala/scala/scalanative/build/TargetTripleTest.scala
```diff
@@ -0,0 +1,37 @@
...
+    "x86_64-apple-darwin21.6.0" ->
+      TargetTriple("x86_64", "apple", "darwin", "unknown"),
+    "x86_64-apple-darwin22.4.0" ->
+      TargetTriple("x86_64", "apple", "darwin", "unknown"),
+    "x86_64-pc-linux-gnu" ->
+      TargetTriple("x86_64", "pc", "linux", "gnu"),
+    "x86_64-pc-windows-msvc" ->
+      TargetTriple("x86_64", "pc", "windows", "msvc"),
+    "x86_64-portbld-freebsd13.1" ->
+      TargetTriple("x86_64", "unknown", "freebsd", "unknown")
```
armanbilge, line 27: Not sure why `portbld` is not a recognized vendor in LLVM.

tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala
```diff
@@ -9,14 +9,15 @@ trait LinktimeValueResolver { self: Reach =>
...
       s"$linktimeInfo.debugMode" -> (conf.mode == Mode.debug),
       s"$linktimeInfo.releaseMode" -> (conf.mode == Mode.releaseFast || conf.mode == Mode.releaseFull || conf.mode == Mode.releaseSize),
-      s"$linktimeInfo.isWindows" -> Platform.isWindows,
-      s"$linktimeInfo.isLinux" -> Platform.isLinux,
-      s"$linktimeInfo.isMac" -> Platform.isMac,
-      s"$linktimeInfo.isFreeBSD" -> Platform.isFreeBSD,
+      s"$linktimeInfo.isWindows" -> (triple.os == "windows"),
+      s"$linktimeInfo.isLinux" -> (triple.os == "linux"),
+      s"$linktimeInfo.isMac" -> (triple.vendor == "apple" && triple.os == "darwin"),
+      s"$linktimeInfo.isFreeBSD" -> (triple.os == "freebsd"),
```
armanbilge, line 20: Actually I was hoping that we can remove these completely and instead define them in the code. e.g. replace
```scala
  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isLinux")
  def isLinux: Boolean = resolved
```
with
```scala
  def isLinux: Boolean = target.os == "linux"
```
but unfortunately the `getLinktimeCondition` does not recognize that properly. I hope we can find a way to fix it, because otherwise it is impossible for users to define their own helper function for checking triples, they must always inline at use-site.

tools/src/main/scala/scala/scalanative/build/TargetTriple.scala
```diff
@@ -0,0 +1,453 @@
...
+package build
+
+import java.nio.ByteOrder
+
+private[scalanative] final case class TargetTriple(
+    arch: String,
+    vendor: String,
+    os: String,
+    env: String
+) {
```
armanbilge, line 21: I omitted sub-architecture at this time, since parsing it seems complicated, using it also seems complicated, and I wasn't sure what the usecase will be anyway.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -84,11 +85,7 @@ sealed trait NativeConfig {
    *  architecture for a name that is not found seems excessive perhaps?
    */
   def is32BitPlatform = {
-    targetTriple
-      .getOrElse(detectedTriple)
-      .split('-')
-      .headOption
-      .getOrElse("") match {
+    configuredOrDetectedTriple.arch match {
```
ekrich, line 88: Is there a way to determine 32 bit via the ported code or maybe we could match the known 32 bit and like 386 etc. and `contains("32")` and let the default be 64 bit?

armanbilge: > let the default be 64 bit?

That is currently the default. The difference is whether we print a warning when we default to 64-bit because we are defaulting, not because we actually know it's 64 bit.

ekrich: Well, now that you have the entire list you could Unknown warn to 64bit if you can't find an arch and then list the known 32bit ones and also find ones that have 32 in the name so it has a chance of working on all known processors.

tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala
WojciechMazur: I was not aware of that issue, we should be able to workaround it by combining both `@alwaysinline` annotation and making rhs and if-else
This snippet would link if os is not set to `windows`
```
import scala.scalanative.annotation._

@extern
object bindings {
  def nonExistingMethod(): Unit = extern
}
object check {
  @alwaysinline def isWindows =
    if (scala.scalanative.meta.LinktimeInfo.os == "windows") true
    else false
}
object Test {

    println("Hello, World!")
    // if( scala.scalanative.meta.LinktimeInfo.os == "windows")
    if (check.isWindows)
      bindings.nonExistingMethod()
    else println("Ok!")

  }
}
```

I'll try to fix it in the compiler plugin, so we would not require a boilerplate if-else

armanbilge: Thanks, I tried some tricks like that, but was having problems here:
https://github.com/scala-native/scala-native/blob/9a8a782f22486136b2ab9c3fa893f9d200182b77/test-interface/src/main/scala/scala/scalanative/testinterface/signalhandling/SignalConfig.scala#L152

WojciechMazur: I have an idea of how we might fix it, I will try it tomorrow. Anyway we should be able to resolve it in follow up PR

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
ekrich: Overall, what I am trying to say is that this feature is really for the 32bit support with the normal case being 64bit. So we could use this as the basic approach.

1. If it is a 32 bit arch and found in the list then return 32 bit (my idea was x86 plus ones with 32 in the name unless that is a bad assumption)
2. If it is a 64 bit arch and found then return 64 bit
3. If it is not found, assume 64 bit.

Note 1: If it is a 32bit os running on 64 bit then the user needs to supply a target triple.
Note 2: We most likely don't expect to add that many 32bit so hopefully maintaining a list should not be hard where needed.

The Swift code has some code for pointer width / 32bit - https://github.com/apple/swift-driver/blob/main/Sources/SwiftDriver/Utilities/Triple.swift

tools/src/test/scala/scala/scalanative/linker/LinktimeConditionsSpec.scala
```diff
@@ -58,11 +58,15 @@ class LinktimeConditionsSpec extends OptimizerSpec with Matchers {
     Set(
       s"$linktimeInfo.asanEnabled",
       s"$linktimeInfo.is32BitPlatform",
-      s"$linktimeInfo.isFreeBSD",
-      s"$linktimeInfo.isMac",
-      s"$linktimeInfo.isWindows",
+      "M36scala.scalanative.meta.LinktimeInfo$D9isFreeBSDzEO",
+      "M36scala.scalanative.meta.LinktimeInfo$D5isMaczEO",
+      "M36scala.scalanative.meta.LinktimeInfo$D9isWindowszEO",
```
armanbilge, line 63: This is a bit weird 😅 didn't have chance to investigate further.

WojciechMazur (approved):
Look good, thank you for these changes. Probably it needs scalafmt run before merging

PR #3259, opened 2023-04-26
Backport `uioOps` to 0.4.x

armanbilge:


ekrich (approved):
LGTM. I didn't know you could deal with a struct directly - is that a GC object then?

armanbilge:
Good eye, I actually know that either, I just copied this from the main branch. I think @LeeTibbert wrote that, maybe they can shed some insight on how it works?

WojciechMazur (approved):

Issue #3220, opened 2023-03-13
`CFuncPtr.fromScalaFunction` should not require `scalanative.unsafe.Tag`


WojciechMazur:
Current implementation of this intrinsic functionally is based on `unsafe.Tag` evidence types, however their handling is flaky and can be improved. Instead of current parsing of materialization methods in `NirGenUtil` which is error prone, we might use exact type information about the type of called function. At this point, it was not implemented like that in the first place (maybe Scala 2.11 compact?). However, currently we might use 1 of two available approaches to mitigate this issue. The exact type information would be erased when generating CFuncPtr, but we can use 1 of the workarounds: 
- add special attachment to the tree which would be presevered during the AST transformations, eg. sizeOf intrinsics #3198
- use time travel mechanism on the full tree of the function, similarly as in for resolving varargs parameters #3204 (simpler, but might be fragile for AST transformations in the compiler)

tanishiking:
Hey, @WojciechMazur!
I've been exploring this issue as a good starting point to understand the core of Scala Native. However, I'm having some difficulty grasping the specific details of the problem and the desired outcome, due to my lack of compiler backend knowledge. I have a few questions that I hope you could help clarify:

- How are `unsafe.Tag`s used in Scala Native in general, and for `CFuncPtr.{apply, fromScalaFunction}`?
  - Based on [NirGenExpr#genCFuncFromScalaFunction](https://github.com/scala-native/scala-native/blob/c98eeeba8fa757f204b1730c2a0bca11f9bae4de/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenExpr.scala#L2419), I assumed that generated NIR functions would have type names in [call[$type]](https://scala-native.org/en/stable/contrib/nir.html#call) or somewhere similar, but I couldn't find this in my experiment below.
- When you mentioned that "their handling is flaky", could you provide some more context or examples of this issue?
- Regarding the suggestion to `"add a special attachment to the tree which would be preserved during the AST transformations, e.g., sizeOf intrinsics Intrinsic based sizeOf[T] / alignmentOf[T] resolution #3198",`
  - Do you mean that we should add a special attachment that allows us to retrieve the exact type information at a later stage (`NirGenPhase`), **similar to** how Scala Native handles `sizeOf[T]` and `alignmentOf[T]`?
  - Not attaching information of `sizeOf[T]` or `alignmentOf[T]` for `alloca` in LLVM
[...]

PR #3270, opened 2023-05-05
Remove unsafe.Tag based type resolution for CFuncPtr.{fromScalaFunction,apply}

tanishiking:
Fix https://github.com/scala-native/scala-native/issues/3220

This pull request refacts the handling of `CFuncPtr.{fromscalaFunction,apply}` by preserving the exact type information during the type erasure phase and referencing it in the NIR generation phase. This is the same approach as `sizeOf[T]` and `alignmentOf[T]` https://github.com/scala-native/scala-native/pull/3198.

Previously, we restored the type information based on the `unsafe.Tag' proof types, which is a bit buggy.

This PR includes the following changes

- Addition of special attachments to the AST for `CFuncPtr#apply` and `CFuncPtr.fromScalaFunction` whose values are their type information (parameter types and return type).
- At the NIR generation phase (`NIRGenExpr.scala`), get the attached exact type information and use it to generate external forwarder methods (previously we used the type information restored based on `unsafe.Tag`).
- Remove the dead code for tag parsing in `NirGenUtil.unwrap{ClassTag,Tag}` and Tags related code in `NirDefinitions`.
- Update `nativelib/src/main/scala/scalanative/unsafe.CFuncPtr.scala.gyb` to remove unsed `evidences` types.

<details>
<summary> finished todo </summary>

## TODOs

- [x] Remove unsafe.Tag from CFuncPtrX and remove `unwrapTag `from NirGenUtil`
- [x] Migrate scalalib-patch-tool to scala-cli
  - It required to re-generate patch files to publish locally during development, but it turned out we don't need re-creation now 🤔 Maybe we can separate https://github.com/scala-native/scala-native/commit/f86daba792b075bf6c4bda96752aaa8c625a0ba7 to another PR.
- [x] Testing
  - [x] `tests2_13 / Test`
  - [x] `test3 / Test`
  - [x] `test2_12 / Test`
    - ~auxlib failed to build with this patch~ nevermind, it seems like it my local environment probem.
- [x] manual testing (testing on https://github.com/tanishiking/scala-native-sbt)
  - [x] `++3.2.2 publishLocal` ok
  - [x] `++2.12.17 publishLocal`
  - [x] `++2.13.10 publishLocal`~ it seems like nativeLink always fail with the following error while tests pass.~ it's also my local environment problem. nevermind

```
[info] ZipFile invalid LOC header (bad signature)
[error] Found 2 missing definitions while linking
[error] Not found Top(Main)
[error] Not found Member(Top(Main),D4mainLAL16java.lang.String_uEo)
[error] Undefined definitions found in reachability phase
[error] (Compile / nativeLink) Undefined definitions found in reachability phase
[error] Total time: 1 s, completed May 5, 2023, 10:11:12 PM
```

</details>

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
+++ b/nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
@@ -29,16 +29,23 @@ object CFuncPtr {
 % for N in range(0, 23):
 %   args      = ", ".join("arg" + str(i) + ": T" + str(i) for i in range(1, N+1))
 %   allTps    = ", ".join(["T" + str(i) for i in range(1, N+1)] + ["R"])
+%   wildcards = ", ".join(["_" for i in range(1, N+1)] + ["_"])
+%   tps       = ", ".join(["paramTy" + str(i) + ": Class[_]" for i in range(1, N+1)] + ["retTy: Class[_]"])
 %   evidences = ", ".join(["ev{}: Tag[T{}]".format(i, i) for i in range(1, N+1)] + ["evRet: Tag[R]"])
 %   CFuncPtrN = "CFuncPtr{}[{}]".format(N, allTps)
 %   FunctionN = "Function{}[{}]".format(N, allTps)
+%   CFuncPtrNWildcard = "CFuncPtr{}[{}]".format(N, wildcards)
+%   FunctionNWildcard = "Function{}[{}]".format(N, wildcards)
+%   evidencesWildcard = ", ".join(["ev{}: Tag[_]".format(i) for i in range(1, N+1)] + ["evRet: Tag[_]"])
 
 final class ${CFuncPtrN} private (rawptr: RawPtr) extends CFuncPtr(rawptr) {
-  def apply(${args})(implicit ${evidences}): R = intrinsic
+  def apply(${args}): R = intrinsic
 }
 
 object CFuncPtr${N} {
-  implicit def fromScalaFunction[${allTps}](fn: ${FunctionN})(implicit ${evidences}): ${CFuncPtrN} = intrinsic
+  implicit def fromScalaFunction[${allTps}](fn: ${FunctionN}): ${CFuncPtrN} = intrinsic
+
+  def _fromScalaFunction(fn: ${FunctionNWildcard}): ${CFuncPtrNWildcard} = intrinsic
 
   private[scalanative] def fromRawPtr[${allTps}](ptr: RawPtr): ${CFuncPtrN} = {
     new ${CFuncPtrN}(ptr)
--- a/nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
+++ b/nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
@@ -30,13 +30,11 @@ object CFuncPtr {
 %   args      = ", ".join("arg" + str(i) + ": T" + str(i) for i in range(1, N+1))
 %   allTps    = ", ".join(["T" + str(i) for i in range(1, N+1)] + ["R"])
 %   wildcards = ", ".join(["_" for i in range(1, N+1)] + ["_"])
-%   tps       = ", ".join(["paramTy" + str(i) + ": Class[_]" for i in range(1, N+1)] + ["retTy: Class[_]"])
 %   evidences = ", ".join(["ev{}: Tag[T{}]".format(i, i) for i in range(1, N+1)] + ["evRet: Tag[R]"])
 %   CFuncPtrN = "CFuncPtr{}[{}]".format(N, allTps)
 %   FunctionN = "Function{}[{}]".format(N, allTps)
 %   CFuncPtrNWildcard = "CFuncPtr{}[{}]".format(N, wildcards)
 %   FunctionNWildcard = "Function{}[{}]".format(N, wildcards)
-%   evidencesWildcard = ", ".join(["ev{}: Tag[_]".format(i) for i in range(1, N+1)] + ["evRet: Tag[_]"])
 
 final class ${CFuncPtrN} private (rawptr: RawPtr) extends CFuncPtr(rawptr) {
   def apply(${args}): R = intrinsic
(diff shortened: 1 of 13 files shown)


scripts/scalalib-patch-tool.sc
```diff
@@ -1,6 +1,11 @@
-import $ivy.`com.lihaoyi::ammonite-ops:2.3.8`, ammonite.ops._, mainargs._
```
ekrich, line 1: Maybe this could be in another PR so you could get it in faster. It is not related correct?

ekrich: Looks like you need to run `scripts/scalafmt` for CI happiness.

nscplugin/src/main/scala-3/scala/scalanative/nscplugin/PrepNativeInteropLate.scala
```diff
@@ -48,6 +48,34 @@ class PostInlineNativeInterop extends PluginPhase {
...
+        val idx = defnNir.CFuncPtr_fromScalaFunction.indexOf(fun.symbol)
+        val transformed = defnNir._CFuncPtr_fromScalaFunction(idx)
+        val cls = defnNir.CFuncPtrNClass(idx)
+        val tys = tArgs.map(t => dealiasTypeMapper(t.typeOpt))
+        cpy
+          .Apply(tree)(
+            ref(transformed),
+            args
+          )
+          .withAttachment(NirDefinitions.NonErasedTypes, tys)
```
WojciechMazur, line 67: We can add attachment to outer `Apply` without transformation of inner `TypeApply`. This combined with the fact that on Scala 2 it's not needed as well, would allow to eliminate additional `_fromScalaFunction` method. 
In Scala2 this transformation was unused, but in Scala3 we need to get original type and save it in attachment, because the target of a lambda might get replaced with adapted (boxed and erased) variant of a function. https://github.com/lampepfl/dotty/blob/abf9a25d5f27a41c11c11c11cf671ed3cb33d10d/compiler/src/dotty/tools/dotc/transform/Erasure.scala#L409-L439


```suggestion
      case appy @ Apply(TypeApply(fun, tArgs), _) if defnNir.CFuncPtr_fromScalaFunction.contains(fun.symbol) =>
        val tys = tArgs.map(t => dealiasTypeMapper(t.typeOpt))
        app
          .withAttachment(NirDefinitions.NonErasedTypes, tys)
```

scripts/scalalib-patch-tool.sc
```diff
@@ -19,27 +24,26 @@ def main(
...
-    @arg(
-      doc =
-        "Path to directory containing overrides, defaults to scalalib/overrides-$scalaBinaryVersion"
-    )
-    overridesDir: Option[os.Path] = None
+    // @arg(
+    //   doc =
+    //     "Path to directory containing overrides, defaults to scalalib/overrides-$scalaBinaryVersion"
+    // )
+    // overridesDir: Option[os.Path] = None
```
WojciechMazur, line 31: It does not compile/work or it's just commented out by mistake?

nativelib/src/main/scala/scala/scalanative/unsafe/CFuncPtr.scala.gyb
```diff
@@ -29,16 +29,21 @@ object CFuncPtr {
...
 final class ${CFuncPtrN} private (rawptr: RawPtr) extends CFuncPtr(rawptr) {
-  def apply(${args})(implicit ${evidences}): R = intrinsic
+  def apply(${args}): R = intrinsic
 }
 
 object CFuncPtr${N} {
-  implicit def fromScalaFunction[${allTps}](fn: ${FunctionN})(implicit ${evidences}): ${CFuncPtrN} = intrinsic
+  implicit def fromScalaFunction[${allTps}](fn: ${FunctionN}): ${CFuncPtrN} = intrinsic
+
+  def _fromScalaFunction(fn: ${FunctionNWildcard}): ${CFuncPtrNWildcard} = intrinsic
```
WojciechMazur, line 46: We can get rid of these, see comments below

WojciechMazur (commented):
Looks good so far, I've left some comments how we can simplify it a bit.

Regarding to problems with local publishing, it is probably due to usage of `nativeLink` task in root project. I'm not sure where does it comes from, probably we could inspect the build and override it. Due to complexity of the build (we don't have typical crossVersion but use project duplicates for each Scala binary version), and the fact that there are cases where we need to either always depend on Scala 2.12 (sbtPlugin) or Scala 2.13 (scalalib for Scala 3) running tasks in root project is not recommended. To target a single project you can use one of dedicated subprojects, eg. `tests2_13`, `tests3`, etc. 
If you want to publish all projects locally you can use a custom command `publish-local-dev <scala-version>`

scripts/scalalib-patch-tool.sc
tanishiking: > Maybe this could be in another PR so you could get it in faster. It is not related correct?

Yeah :+1:
It was required to update patches during development for some reason https://github.com/scala-native/scala-native/issues/3220#issuecomment-1521588637 🤔 
But now it turned out it's no longer needed, I'll remove this change and open a separate PR.

scripts/scalalib-patch-tool.sc
tanishiking: I commented out for now because it didn't work with scala-cli for some reason, I'll check out in a separate PR 👍

nscplugin/src/main/scala-3/scala/scalanative/nscplugin/PrepNativeInteropLate.scala
tanishiking: Thanks! `_fromScalaFunction` was a thing I don't like in this change and I'm glad we can get rid of that!
done https://github.com/scala-native/scala-native/pull/3270/commits/cc62ca2486fb2c7cc855440bdd709ebf45a99f32

>  but in Scala3 we need to get original type and save it in attachment, because the target of a lambda might get replaced with adapted (boxed and erased) variant of a function.

Oh, that's why we didn't need this in Scala2, while Scala3 does. That makes sense, thank you for the explanation!

tanishiking:
> Regarding to problems with local publishing, it is probably due to usage of nativeLink task in root project. I'm not sure where does it comes from, probably we could inspect the build and override it

> If you want to publish all projects locally you can use a custom command `publish-local-dev <scala-version>`

It seems like the problems (both `auxlib` and problem with linking error for 2.13) are caused by my contaminated local environment (by `++x.x.x publishLocal` 😅), removing the local cache, and republish by `publish-local-dev x.x.x` fixed the problem. Thanks!

tanishiking:
Regarding the auxlib, it seems like the same problem as https://github.com/scala-native/scala-native/issues/3265 while we can go around with it by `publish-local-dev` command.

WojciechMazur (approved):

PR #3279, opened 2023-05-08
Use `Array[Int]` instead of `Array[Array[Boolean]]` for `[class,trait]_has_trait`

lolgab:
## Results

Here the results with the Ember hello world.
```
 31056 -rwxr-xr-x@ 1 lorenzo  staff   15898802  8 Mag 14:31 after*
675624 -rw-r--r--@ 1 lorenzo  staff  345918875  8 Mag 15:07 after.ll
     0 drwxr-xr-x@ 4 lorenzo  staff        128  1 Mag 16:30 b/
 40632 -rwxr-xr-x@ 1 lorenzo  staff   20803106  8 Mag 14:52 before*
783056 -rw-r--r--@ 1 lorenzo  staff  400921751  8 Mag 14:40 before.ll
```

the binary size is reduced by ~`4.9 MB` and the generated `out.ll` file is reduced by `55 MB`
The runtime perfomance seems unchanged:

```
# Before
> hey -z 60s http://localhost:8081

Summary:
  Total:        60.0026 secs
  Slowest:      0.1061 secs
  Fastest:      0.0003 secs
  Average:      0.0034 secs
  Requests/sec: 14778.1655

# After
> hey -z 60s http://localhost:8080

Summary:
  Total:        60.0031 secs
  Slowest:      0.1002 secs
  Fastest:      0.0002 secs
  Average:      0.0034 secs
  Requests/sec: 14824.3753
```

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/BitMatrix.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/BitMatrix.scala
@@ -8,8 +8,7 @@ private[scalanative] class BitMatrix private (
 
   def set(row: Int, col: Int): Unit = {
     val bitIndex = row * columns + col
-    val len = (bitIndex >> AddressBitsPerWord) + 1
-    bits(len - 1) |= 1L << (bitIndex & RightBits)
+    bits(bitIndex >> AddressBitsPerWord) |= 1L << (bitIndex & RightBits)
   }
 
   def toSeq = bits.toSeq
--- a/tools/src/main/scala/scala/scalanative/codegen/BitMatrix.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/BitMatrix.scala
@@ -1,26 +1,26 @@
 package scala.scalanative
 
 private[scalanative] class BitMatrix private (
-    bits: Array[Long],
+    bits: Array[Int],
     columns: Int
 ) {
   import BitMatrix.{AddressBitsPerWord, ElementSize, RightBits}
 
   def set(row: Int, col: Int): Unit = {
     val bitIndex = row * columns + col
-    bits(bitIndex >> AddressBitsPerWord) |= 1L << (bitIndex & RightBits)
+    bits(bitIndex >> AddressBitsPerWord) |= 1 << (bitIndex & RightBits)
   }
 
   def toSeq = bits.toSeq
 }
 private[scalanative] object BitMatrix {
-  private[scalanative] final val AddressBitsPerWord = 6 // Int Based 2^6 = 64
+  private[scalanative] final val AddressBitsPerWord = 5 // Int Based 2^5 = 32
   private[scalanative] final val ElementSize = 1 << AddressBitsPerWord
   private[scalanative] final val RightBits = ElementSize - 1
 
   def apply(rows: Int, columns: Int): BitMatrix = {
     val nbits = rows * columns
     val length = (nbits + RightBits) >> AddressBitsPerWord
-    new BitMatrix(new Array[Long](length), columns)
+    new BitMatrix(new Array[Int](length), columns)
   }
 }
--- a/tools/src/main/scala/scala/scalanative/codegen/BitMatrix.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/BitMatrix.scala
@@ -1,0 +1,27 @@
+package scala.scalanative
+
+private[scalanative] class BitMatrix private (
+    bits: Array[Long],
+    columns: Int
+) {
+  import BitMatrix.{AddressBitsPerWord, ElementSize, RightBits}
+
+  def set(row: Int, col: Int): Unit = {
+    val bitIndex = row * columns + col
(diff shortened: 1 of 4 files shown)


tools/src/main/scala/scala/scalanative/codegen/Generate.scala
```diff
@@ -504,7 +592,7 @@ object Generate {
     }
   }
 
-  private object Impl {
+  object Impl {
```
lolgab, line 595: I need to move the constats that I need in to use in `Lower` outside of the `Impl` object and make them public.

tools/src/main/scala/scala/scalanative/codegen/BitMatrix.scala
```diff
@@ -0,0 +1,26 @@
+package scala.scalanative
+
+private[scalanative] class BitMatrix private (
```
lolgab, line 3: This is a minimal version of our java BitSet implementation with fixed size and no bound checks.

WojciechMazur:
Even though in typical usage the new implementation might not introduce any regression in performance when we isolate the benchmark to mostly type checks, then the difference can be actually observed: 

Anyway, I think this change should be included into the main. Probably in the future we might provide users with option to use the old implementation while acknowledging the higher binary size. In fact we might need to do something similar for virtual methods dispatch in which we use a large array to ensure constant lookup time, which however blocks capabilities of incremental compilation.

### Benchmark
Tested with releaseFast + LTO thin
Additional read/write ops were introduce to ensure that code would not be dead-code eliminate/partially evaluated at compile-time. 
 
```scala
trait T1
trait T2 extends T1
trait T3 extends T2
trait T4 extends T3

class C1
class C2 extends C1
class C3 extends C2
class C4 extends C3

object Test {
    var c = 0
    var b = false
    var i = 0
    var x = 0
    while (i < 1000000) {
      val obj = (i % 3) match {
        case 0 => new C4
        case 1 => new T3 {}
        case _ => new {}
      }
      assert(obj != null)
      while (c < 1000000) {
        b &&= obj.isInstanceOf[T1]
        b &&= obj.isInstanceOf[T2]
        b &&= obj.isInstanceOf[T3]
        b &&= obj.isInstanceOf[T4]
        b &&= obj.isInstanceOf[C1]
        b &&= obj.isInstanceOf[C2]
        b &&= obj.isInstanceOf[C3]
        b &&= obj.isInstanceOf[C4]
        c += 1
        if (b) x += 1
      }
      i += 1
    }
  }
}

```

### Results
```bash
gitpod /workspace/scala-native (main) $ time ./before 

real    0m0.006s
user    0m0.001s
sys     0m0.006s
gitpod /workspace/scala-native (main) $ time ./after 

real    0m0.014s
user    0m0.009s
sys     0m0.005s
```

lolgab:
@WojciechMazur Great that you wrote a benchmark program, so we can test the performance of a bitset with 32 bit integers and also the performance of a one-dimensional array of Booleans, which are two other options.
I was thinking, do the matrix need to contain all the possible classes and traits in the application?
Couldn't at link time flag the classes and traits that appear in `isInstanceOf` calls and materialize the matrix only for those?
I'm not sure I understood how these runtime checks work so if I'm saying something completely wrong let me know :)

WojciechMazur:
> Couldn't at link time flag the classes and traits that appear in isInstanceOf calls and materialize the matrix only for those? 

That might be an interesting experiment, but I'm not sure if that would make a lot of difference. Becouse the NIR is generated after the erasure, the Scala AST might be already full of `.asInstanceOf` and `.isInstanceOf` calls, which probably might by present for most of the types. However we could check that. Probably it would be easiest to handler it in `Lower` phase, there `Op.Is` and `Op.As` would be transformed to actual low level calls. I mention `.asInstanceOf` here as well, because it's lowering included runtime isInstanceOf check as well.

armanbilge:
> I mention `.asInstanceOf` here as well, because it's lowering included runtime isInstanceOf check as well.

Btw, not a priority, but eventually I would be happy to see an option so that in release mode all `.asInstanceOf` is no-op and `ClassCastException` becomes undefined behavior. [It is already undefined behavior in Scala.js](https://www.scala-js.org/doc/semantics.html#undefined-behaviors) so cross-platform code should not be encountering that, and idiomatic Scala code is also not relying on `ClassCastException` for correct semantics (certainly not anything in Typelevel stack). Maybe this can be good for performance and even better if we can use it to reduce the size of `{class,trait}_has_trait`.

lolgab:
Converted the matrixes to `Array[Int]` and it seems that the performances improved on my Apple ARM64 machine. In my benchmark is always on par with baseline. But I couldn't test with `LTO=thin`. I'm converting the matrixes to `Array[Int]` since it should also improve performance on 32 bit machines.

WojciechMazur:
Thank you for introducing me to `hyperfine`. I can confirm that when using thin lto i can get similar results.

WojciechMazur (approved):

PR #3283, opened 2023-05-12
Revive caching build config

ekrich:
Well this only saves 2 or 3ms on my machine so not sure it is worth it. 

Not sure all the tests will pass either because the build config has to be well formed and not reused for different things.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -2,11 +2,43 @@ package scala.scalanative
 package build
 
 import java.nio.file.{Files, Path, Paths}
+import scala.collection.mutable
 import scala.scalanative.util.Scope
-import scala.util.Try
 
 /** Utility methods for building code using Scala Native. */
 object Build {
+
+  // holds project basename and the related Config
+  private val cache = mutable.HashMap.empty[String, Config]
+
+  /** Reuse config from cache or cache the config after running any needed
+   *  discovery and then validation. This avoids the costs on subsequent builds.
+   *  Reloading or restarting the build will clear the cache as this class will
+   *  be reloaded. This is required to change either settings or environment
+   *  variables that change the build.
+   *
+   *  Note: Runtime environment variables that control the GC can be changed
+   *  between runs and will be applied.
+   *
+   *  @param config
+   *    starting config from build
+   *  @return
+   *    config with discovery if needed and validation
+   */
+  private def checkCache(config: Config): Config = {
+    // Requires that defaultBasename or basename is set
+    // and unique across sub-projects in the same project
+    val name = config.artifactName
+    // always use a fresh logger
+    cache.get(name) match {
+      case Some(value) =>
+        value.withLogger(config.logger)
+      case None =>
+        val vconfig = Validator.validate(config)
+        cache.put(name, vconfig)
+        vconfig
+    }
+  }
 
   /** Run the complete Scala Native pipeline, LLVM optimizer and system linker,
    *  producing a native binary in the end.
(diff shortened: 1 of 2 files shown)


ekrich:
Result:
```
sbt:scala-native> sandbox2_13/run
[info] compiling 1 Scala source to /Users/eric/workspace/scala-native/sandbox/.2.13/target/scala-2.13/classes ...
[info] Cache config, discover and validate (3 ms)
[info] Linking (multithreading false) (1041 ms)
[info] Checking intermediate code (120 ms)
[info] Dumping intermediate code (linked) (279 ms)
[info] Optimizing (debug mode) (827 ms)
[info] Checking intermediate code (200 ms)
[info] Dumping intermediate code (optimized) (144 ms)
[info] Dumping intermediate code (lowered) (343 ms)
[info] Generating intermediate code (1273 ms)
[info] Produced 39 files
[info] Compiling to native code (1985 ms)
[info] Linking native code (immix gc, none lto) (86 ms)
[info] Total (6017 ms)
Hello, World!
[success] Total time: 8 s, completed May 12, 2023 11:47:54 AM
sbt:scala-native> sandbox2_13/run
[info] Cache config, discover and validate (0 ms)
[info] Linking (multithreading false) (345 ms)
[info] Checking intermediate code (18 ms)
[info] Dumping intermediate code (linked) (182 ms)
[info] Optimizing (debug mode) (416 ms)
[info] Checking intermediate code (11 ms)
[info] Dumping intermediate code (optimized) (124 ms)
[info] Dumping intermediate code (lowered) (199 ms)
[info] Generating intermediate code (399 ms)
[info] Produced 39 files
[info] Compiling to native code (145 ms)
[info] Linking native code (immix gc, none lto) (2 ms)
[info] Total (1656 ms)
Hello, World!
[success] Total time: 3 s, completed May 12, 2023 11:48:01 AM
```

ekrich:
Since `sbt` at least allows you to have multiple applications in the same module then we create the application name based on the package/classname which needs to be unique but we copy the result to the artifact name.

The failing scripted tests both have multiple apps and the combination of this along with the cache key for the `Config` is conflicting in a bad way. Not sure if we should reconsider the above or figure out how to fix scripted - I am pretty clueless on that front. Not sure how the config gets created.

tanishiking (commented):
Thanks @ekrich for the demonstration!

It seems that this cache allows scala-native to skip the validation step if it hasn't been changed (because changing the configuration requires restarting the build tools in the current implementation).

- sbt https://github.com/scala-native/scala-native/blob/89f8ed5391db31e37c4e3670ee504c92ae84711e/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L138-L146
- mill https://github.com/com-lihaoyi/mill/blob/19ec00b3484e952b073b5cba2219829907aa7a80/scalanativelib/worker/0.4/src/mill/scalanativelib/worker/ScalaNativeWorkerImpl.scala#L68-L80

I can't see that it's worth caching the configuration just to skip the validation (2-3ms), considering the effort and risk of implementing the cache.

ekrich:
>I can't see that it's worth caching the configuration just to skip the validation (2-3ms), considering the effort and risk of implementing the cache.

I agree but it does show some rough edges in the build system. The executable rename and the fact that the scripted test(s) fail point to other issue in the current design.

2 ms is not much but it is less.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -2,12 +2,44 @@ package scala.scalanative
...
 import java.nio.file.{Files, Path, Paths}
+import scala.collection.mutable
 import scala.scalanative.util.Scope
-import scala.util.Try
 
 /** Utility methods for building code using Scala Native. */
 object Build {
 
+  // holds project basename and the related Config
+  private val cache = mutable.HashMap.empty[String, Config]
```
WojciechMazur, line 12: To be honest I don't think caching Config itself is going to make any significant improvement. The only slow part of its creation is the `Discovery` for `NativeConfig`. There is no other IO or process invocation.  
The next possible IO operation is happening in the `Validator` by checking if each file on the classpath exists. In that case, it would be safer to ensure that all the files exist, and e.g. downloaded dependency was not removed between scalac compilation and the beginning of the linking. That would typically be equal to cost IO for `File.exists` invocation for each jar + a single `Files.isDirectory` compilation outputs of current project.

tools/src/test/scala/scala/scalanative/IncCompilationTest.scala
```diff
@@ -77,7 +77,7 @@ class IncCompilationTest extends codegen.CodeGenSpec with Matchers {
       )
       val entry = "A"
       val changedTop = Set[String]("A", "A$")
-      val outDir = Files.createTempDirectory("native-test-out")
+      val outDir = Files.createTempDirectory("native-test-out1")
```
WojciechMazur, line 80: The need to modify the output directory is for me a signal something very wrong was happening here. Yes, it can admit that perfectly each test should be executed in a different, clean environment, but in that case, it also perfectly shows the case of incremental editing and rebuilding the sources. We should never be forced to create new temp directory each time.

tools/src/main/scala/scala/scalanative/build/Build.scala
ekrich: Yes, I didn't remember how much time it took so I agree that it is not needed. I also wanted to see if the approach would be able to handle the scripted tests and such but clearly it doesn't. The last PR I was mixing too many things together. I think we could use the structure of the idea though to make the config naming better like is is now with the `buildImpl`.

tools/src/test/scala/scala/scalanative/IncCompilationTest.scala
ekrich: This is just showing a crack in the caching approach but really we shouldn't be reusing build directories I don't think - the multiple main class works but that is a special case. However, if our build was super good it might clean as needed to keep all the artifacts up to date and in sync.

ekrich:
This was a good experiment but saving 2ms is probably not worth it given the other issues it introduces.

Issue #3154, opened 2023-02-10
Character.isWhitespace(-1) throws an exception


JD557:
On scala-native `Character.isWhitespace(-1)` throws a `java.lang.ArrayIndexOutOfBoundsException: -1`, while on the JVM it simply returns `false`.

The reason for this is a lack of a check in https://github.com/scala-native/scala-native/blob/main/javalib/src/main/scala/java/lang/Character.scala#L551. Note that `getType` already checks for this: https://github.com/scala-native/scala-native/blob/main/javalib/src/main/scala/java/lang/Character.scala#L431, but `Character.isWhitespace` calls `getTypeLT256` directly. Skipping the check.

This problem seems to not be only on `isWhitespace`, there are other methods calling `getTypeLT256` without checking for negative numbers.

The stdlib does call `isWhitespace(-1)` at least in one place (https://github.com/scala/scala/blob/v2.13.10/src/library/scala/sys/process/Parser.scala#L28-L109), triggering this bug.

PR #3284, opened 2023-05-15
Character.isWhitespace(-1) returns false

tanishiking:
Fix https://github.com/scala-native/scala-native/issues/3154

Previously, `Character.isWhitespace(-1)` and
`Character.isLowerCase(-1)` throw
`java.lang.ArrayIndexOutOfBoundsException: -1`.
However, they should return `false` as other java std libraries do.

We may want to fix `getTypeLT256`, but this commit guaranteed that the codepoint is gte 0 on call-site of `getTypeLT256` for not introducing the duplicated check of `codePoint < 0`.


ekrich (approved):
LGTM

Issue #3282, opened 2023-05-10
Restore skipping the build functionality


WojciechMazur:
In the 0.5.x branch [one of the changes](https://github.com/scala-native/scala-native/pull/2942/files#diff-a610de80dc17f463fea5440deffcec5b2b70a7dbeef002624b94f6fa894f4b77) removed skipping the subsequent builds in the case when previous artifacts were present and there were no changes in both classpath and config. That functionality should be restored, but this time it should be implemented with the Scala Native linker, instead of the sbt plugin, though allowing to reuse of that feature in other build tools.

tanishiking:
Note:

Previously, scala-native sbt plugin's `nativeLInk` task does do the job only when

- input classpath content (`classpathFilesInfo`) is unchanged from the previous build
- previous artifacts were present (`outputTracker`)

Otherwise, scala-native return the previous artifact as a task's output.

https://github.com/scala-native/scala-native/blob/cf79a295cb36a05a8fb96922d49116608a54e3f9/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L141-L203


This behavior is removed in https://github.com/scala-native/scala-native/pull/2942 and it should be restored in somewhere like https://github.com/scala-native/scala-native/blob/ed64789e367f1a206c7c300d0ff60fc969218304/tools/src/main/scala/scala/scalanative/build/Build.scala#L56

 so that every build tools (other than sbt) can use the feature.

---

The previous one was implemented using sbt's `Tracked` API, and we can't use these utility this time.
https://github.com/sbt/sbt/blob/6dbb6fef3a3a9a1845bae29d2157ecbd4916a090/util-tracking/src/main/scala/sbt/util/Tracked.scala#L188

However, the things to do is the same:

- Calculate the classpath content hash
- Compare it (+ config) with the previous one
  - We need to dump the classpath content hash + config for the build at somewhere
- Checking there's an output artifact, and it's generated from the previous input
  - we may want to store the inputs' hash and the corresponding output's hash
- Return the Path to the previous artifact

ekrich:
The previous version of this was in the build so we want it in tools which is already mentioned. My impression was it didn't work correctly although I didn't capture the deficiencies, I just did a clean.

I recommend the following general approach.

- Optimize each phase in the chain to allow for local reasoning and optimization of each phase.

For example, right now there are various optimizations

- NIR generation by package that avoids regeneration
[...]

PR #3285, opened 2023-05-15
In-memory cache for NIR linking

tanishiking:
This commit enables scala-native to cache the NIR link results
based on input: the hash of

- `config.artifactName`
- `entires: Seq[Global]`
- timestamps of the classPath

Since we use `config.artifactName`, we have the same problem as
https://github.com/scala-native/scala-native/pull/3283#issuecomment-1546274083

Related https://github.com/scala-native/scala-native/issues/3282

Before

```
[info] Linking (multithreading false) (596 ms)
[info] Checking intermediate code (74 ms)
[info] Dumping intermediate code (linked) (230 ms)
[info] Optimizing (debug mode) (546 ms)
[info] Checking intermediate code (33 ms)
[info] Dumping intermediate code (optimized) (125 ms)
[info] Dumping intermediate code (lowered) (173 ms)
[info] Generating intermediate code (512 ms)
[info] Produced 39 files
[info] Compiling to native code (209 ms)
[info] Linking native code (immix gc, none lto) (3 ms)
[info] Total (2370 ms)
[success] Total time: 4 s, completed May 15, 2023, 9:43:49 PM
sbt:scala-native> sandbox2_13/nativeLink
[info] Linking (multithreading false) (323 ms)
[info] Checking intermediate code (49 ms)
[info] Dumping intermediate code (linked) (120 ms)
[info] Optimizing (debug mode) (357 ms)
[info] Checking intermediate code (15 ms)
[info] Dumping intermediate code (optimized) (96 ms)
[info] Dumping intermediate code (lowered) (182 ms)
[info] Generating intermediate code (363 ms)
[info] Produced 39 files
[info] Compiling to native code (166 ms)
[info] Linking native code (immix gc, none lto) (4 ms)
[info] Total (1507 ms)
[success] Total time: 2 s, completed May 15, 2023, 9:43:54 PM
```

After

```
sbt:scala-native> sandbox2_13/nativeLink
[info] Linking (multithreading false) (717 ms)
[info] Checking intermediate code (105 ms)
[info] Dumping intermediate code (linked) (223 ms)
[info] Optimizing (debug mode) (549 ms)
[info] Checking intermediate code (36 ms)
[info] Dumping intermediate code (optimized) (129 ms)
[info] Dumping intermediate code (lowered) (207 ms)
[info] Generating intermediate code (510 ms)
[info] Produced 39 files
[info] Compiling to native code (172 ms)
[info] Linking native code (immix gc, none lto) (3 ms)
[info] Total (2486 ms)
[success] Total time: 4 s, completed May 15, 2023, 9:41:43 PM
sbt:scala-native> sandbox2_13/nativeLink
[info] NIR linking skipped due to cache hit.
[info] Linking (multithreading false) (12 ms)
[info] Checking intermediate code (22 ms)
[info] Dumping intermediate code (linked) (134 ms)
[info] Optimizing (debug mode) (275 ms)
[info] Checking intermediate code (15 ms)
[info] Dumping intermediate code (optimized) (110 ms)
[info] Dumping intermediate code (lowered) (160 ms)
[info] Generating intermediate code (301 ms)
[info] Produced 39 files
[info] Compiling to native code (165 ms)
[info] Linking native code (immix gc, none lto) (1 ms)
[info] Total (1045 ms)
```

### DIFF
--- a/tools/src/main/scala/scala/scalanative/linker/Link.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/Link.scala
@@ -4,11 +4,12 @@ package linker
 import scalanative.nir._
 import scalanative.util.Scope
 import scala.scalanative.io.VirtualDirectory
+import java.nio.file.Path
 
 object Link {
 
   private val cache =
-    collection.mutable.HashMap[(String, Seq[Global], Seq[Long]), Result]()
+    collection.mutable.HashMap[(Path, Seq[Global], Seq[Long]), Result]()
 
   /** Load all clases and methods reachable from the entry points. */
   def apply(config: build.Config, entries: Seq[Global])(implicit
@@ -18,7 +19,7 @@ object Link {
       ClassPath(VirtualDirectory.real(path))
     }
     val classPathMtime = classpath.map(_.lastModifiedMillis)
-    val key = (config.artifactName, entries, classPathMtime)
+    val key = (config.buildPath, entries, classPathMtime)
     cache.get(key) match {
       case None =>
         val res = Reach(config, entries, ClassLoader.fromClasspath(classpath))
--- a/tools/src/main/scala/scala/scalanative/linker/Link.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/Link.scala
@@ -3,14 +3,32 @@ package linker
 
 import scalanative.nir._
 import scalanative.util.Scope
+import scala.scalanative.io.VirtualDirectory
 
 object Link {
+
+  private val cache =
+    collection.mutable.HashMap[(String, Seq[Global], Seq[Long]), Result]()
 
   /** Load all clases and methods reachable from the entry points. */
   def apply(config: build.Config, entries: Seq[Global])(implicit
       scope: Scope
-  ): Result =
-    Reach(config, entries, ClassLoader.fromDisk(config))
+  ): Result = {
+    val classpath = config.classPath.map { path =>
+      ClassPath(VirtualDirectory.real(path))
+    }
+    val classPathMtime = classpath.map(_.lastModifiedMillis)
+    val key = (config.artifactName, entries, classPathMtime)
+    cache.get(key) match {
+      case None =>
+        val res = Reach(config, entries, ClassLoader.fromClasspath(classpath))
+        cache.update(key, res)
+        res
+      case Some(res) =>
+        config.logger.info(s"NIR linking skipped due to cache hit.")
+        res
+    }
+  }
 
   /** Run reachability analysis on already loaded methods. */
   def apply(
--- a/tools/src/main/scala/scala/scalanative/linker/ClassLoader.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/ClassLoader.scala
@@ -14,6 +14,10 @@ sealed abstract class ClassLoader {
 
 object ClassLoader {
 
+  def fromClasspath(classPath: Seq[ClassPath]): ClassLoader = {
+    new FromDisk(classPath)
+  }
+
   def fromDisk(config: build.Config)(implicit in: Scope): ClassLoader = {
     val classpath = config.classPath.map { path =>
       ClassPath(VirtualDirectory.real(path))
--- a/tools/src/main/scala/scala/scalanative/linker/ClassPath.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/ClassPath.scala
@@ -7,8 +7,10 @@ import scala.collection.mutable
 import scala.scalanative.io.VirtualDirectory
 import scala.scalanative.nir.serialization.deserializeBinary
 import scala.scalanative.nir.{Defn, Global, Prelude => NirPrelude}
+import java.nio.file.attribute.FileTime
 
 sealed trait ClassPath {
+  private[scalanative] def lastModifiedMillis: Long
 
   /** Check if given global is present in this classpath. */
   private[scalanative] def contains(name: Global): Boolean
@@ -30,6 +32,8 @@ object ClassPath {
     new Impl(directory)
 
   private final class Impl(directory: VirtualDirectory) extends ClassPath {
+    override def lastModifiedMillis = directory.lastModified.toMillis()
+
     private val files =
       directory.files
         .filter(_.toString.endsWith(".nir"))
--- a/util/src/main/scala/scala/scalanative/io/VirtualDirectory.scala
+++ b/util/src/main/scala/scala/scalanative/io/VirtualDirectory.scala
@@ -8,6 +8,7 @@ import java.nio.file._
 import java.nio.channels._
 import java.util.HashMap
 import scalanative.util.{Scope, acquire, defer}
+import java.nio.file.attribute.FileTime
 
 sealed trait VirtualDirectory {
 
@@ -35,6 +36,9 @@ sealed trait VirtualDirectory {
 
   /** Merges content of source paths into single file in target */
   def merge(sources: Seq[Path], target: Path): Unit
+
+  /* Returns a file's last modified time. */
+  def lastModified: FileTime
 }
 
 object VirtualDirectory {
@@ -147,6 +151,8 @@ object VirtualDirectory {
           .walk(path, Integer.MAX_VALUE, FileVisitOption.FOLLOW_LINKS)
           .iterator()
       }.map(fp => path.relativize(fp))
+
+    override def lastModified: FileTime = Files.getLastModifiedTime(path)
   }
 
   private final class JarDirectory(path: Path)(implicit in: Scope)
@@ -176,6 +182,8 @@ object VirtualDirectory {
           }
         }
     }
+
+    override def lastModified: FileTime = Files.getLastModifiedTime(path)
   }
 
   private object EmptyDirectory extends VirtualDirectory {
@@ -200,5 +208,9 @@ object VirtualDirectory {
     override def merge(sources: Seq[Path], target: Path): Unit =
       throw new UnsupportedOperationException("Can't merge in empty directory.")
 
+    override def lastModified: FileTime =
+      throw new UnsupportedOperationException(
+        "Can't get mtime attribute of an empty directory."
+      )
   }
 }


tools/src/main/scala/scala/scalanative/linker/Link.scala
```diff
@@ -3,14 +3,33 @@ package linker
...
   def apply(config: build.Config, entries: Seq[Global])(implicit
       scope: Scope
-  ): Result =
-    Reach(config, entries, ClassLoader.fromDisk(config))
+  ): Result = {
+    val classpath = config.classPath.map { path =>
+      ClassPath(VirtualDirectory.real(path))
+    }
+    val classPathMtime = classpath.map(_.lastModifiedMillis)
+    val key = (config.buildPath, entries, classPathMtime)
```
WojciechMazur, line 22: The current implementation would leak the memory (each  result would probably add tens of MBs which would never be released). Java WeakHashMap should help with that to some extent (in fact SoftReference based solution would work much better here, but there seems to HashMap implementation based on that in the stdlib). 
Alternatively, the cache should be cleaned every time we know that it could never be reused again. I think that we might keep the cache when changing the main class, but the classpath is the same. In any other case, we should do a clean build. 

If the classpath has not changed and the main class is the same, we would always get exactly the same list entries `Seq[Global]` equal to mainClass and fixed amount of CodeGen/Interflow/Lower dependencies. The only thing that might have changed is the main class, and it already covered  by the `config.buildPath`. Similary if we produce a dynamic library instead of application, then if the classpath has not changed, then the list of entries would not change as well.
Also as soon as at least 1 file of classpath was modified, the outputs of the program might have changed, and we need to rebuild it. Because of that we can keep just the most recent modification time for all the classpath files.

Based on that the key could be simplified to `(buildPath: Path, lastModifiedTime: Long)`

WojciechMazur: In general we should be very fragile when it comes to increasing memory usage. The problems with OOMs while build Scala Native projects is already a known issue in some of OSS  projects.

tools/src/main/scala/scala/scalanative/linker/Link.scala
```diff
@@ -3,14 +3,33 @@ package linker
...
+    cache.get(key) match {
+      case None =>
+        val res = Reach(config, entries, ClassLoader.fromClasspath(classpath))
```
WojciechMazur, line 25: I wonder if we could change cache granularity from the whole `linker.Result` for the whole build, to a classpath entry (single jar/directory). But that might require lots of other changes

tools/src/main/scala/scala/scalanative/linker/Link.scala
armanbilge: > In general we should be very fragile when it comes to increasing memory usage.

Well, it's worth distinguishing between local development (where caching is very valuable for incremental re-building and rapid development) and CI (where we only build once and memory limits are very strict). Scala.js linker config has a flag to distinguish these situations, which we configure in our build plugins.

https://github.com/scala-js/scala-js/blob/5323e68c2daeb2937f807f0fd2559cbd3e1b6e71/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala#L55-L66

tanishiking: > The current implementation would leak the memory (each result would probably add tens of MBs which would never be released). Java WeakHashMap should help with that to some extent (in fact SoftReference based solution would work much better here, but there seems to HashMap implementation based on that in the stdlib).

Oh, yes, that's correct.
I attempted to use `WeakHashMap`, but it appears that the cache entries can be easily garbage collected because no strong references to the keys remain (It would be more ideal if there was a SoftReference-based HashMap in the standard library, as you pointed out...).

As an alternative, perhaps we don't need to store multiple cache entries, but just retain the most recent one, especially considering the incremental compilation scenario.

Like in the form of `private var cache: (Key, Result) = ...`.?



> Based on that the key could be simplified to (buildPath: Path, lastModifiedTime: Long)

Thanks! Indeed, we don't need to use `entries` as we already use the timestamps of classPaths + buildPath 👍 

---

> In general we should be very fragile when it comes to increasing memory usage. The problems with OOMs while build Scala Native projects is already a known issue in some of OSS projects.

That makes sense.
Hmm, we can reduce the number of cache entries, decrease the object size of keys, and make the cache easier to garbage collect. However, the in-memory cache still increases memory usage by storing the `Linker.Result` (I'm not sure how much it could increase though) for some extent.

It would be better to check how much the `Linker.Result` eats up the memory if we store in a cache, or should we store the Result in file so we don't store them in-memory?

<details>
<summary>I attempted to check the cache size using jol but somehow I couldn't use it 🤔 </summary>

```diff
diff --git a/project/Deps.scala b/project/Deps.scala
index 69a0f761..9d75f86e 100644
--- a/project/Deps.scala
+++ b/project/Deps.scala
@@ -20,6 +20,7 @@ object Deps {
   lazy val ScalaCheck          = "org.scalacheck"         %% "scalacheck"                 % "1.15.4"
   lazy val ScalaTest           = "org.scalatest"          %% "scalatest"                  % "3.2.9"
   lazy val ScalaParCollections = "org.scala-lang.modules" %% "scala-parallel-collections" % "1.0.3"
+  lazy val Jol                 = "org.openjdk.jol"         % "jol-core"                   % "0.17"
   lazy val SbtPlatformDeps     = "org.portable-scala"      % "sbt-platform-deps"          % "1.0.1"
   lazy val SbtTestInterface    = "org.scala-sbt"           % "test-interface"             % "1.0"
   lazy val JUnitInterface      = "com.novocode"            % "junit-interface"            % "0.11"
@@ -28,8 +29,10 @@ object Deps {
   def Tools(scalaVersion: String) = {
     List(ScalaCheck % "test", ScalaTest % "test") ++
       scalaVersionsDependendent(scalaVersion) {
@@ -28,8 +29,10 @@ object Deps {
   def Tools(scalaVersion: String) = {
     List(ScalaCheck % "test", ScalaTest % "test") ++
       scalaVersionsDependendent(scalaVersion) {
-        case (2, 12) => Nil
-        case _       => ScalaParCollections :: Nil
+        case (2, 12) => Jol :: Nil
+        case _       => ScalaParCollections :: Jol :: Nil
       }
   }
   def NativeLib(scalaVersion: String) = scalaVersionsDependendent(scalaVersion) {
```

```
[info] compiling 1 Scala source to /Users/tanishiking/ghq/github.com/tanishiking/scala-native/project/target/scala-2.12/sbt-1.0/classes ...
[error] /Users/tanishiking/ghq/github.com/tanishiking/scala-native/tools/src/main/scala/scala/scalanative/linker/Link.scala:8:20: object jol is not a member of package org.openjdk
[error] import org.openjdk.jol.info.GraphLayout;
[error]                    ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

</details>

---

> Scala.js linker config has a flag to distinguish these situations, which we configure in our build plugins.

I'm not sure how much things we store in memory for incremental compilation in scala-native at this moment, but it seems good to have for scala-native too if we store a lot things in memory for inc-compilation.

WojciechMazur: To fix problem with jol you would need to add that dependency to `project/build.sbt` That's because to start sbt project we need to build sbt-plugin and it's dependencies first. So, to break that cyclic dependency we bootstrap them when starting the sbt. The only downside is the fact that we need to duplicate some dependencies config.

tanishiking: Thanks! I printed debug using `jol` and it turned out the build cache has around 35MB (with 1 cache entry) even for the `sandbox` project IIUC. Which seems not a good idea to keep in memory.

Now I'm wondering which is the way to go for build cache

- Cache the NIR linking/optimization in memory (which takes more than 30-40MB, it may go more for the real projects)
- Dump the `LInker.Result` into a disk for cache
  - I'm not sure this is really a good idea, reading / writing that size of stuff may take some CPU and memory.
- Or, maybe we should (if cache hits) skip the whole output of NIR linking ~ CodeGen

<details>

```diff
diff --git a/project/build.sbt b/project/build.sbt
index f161c052..27ba3856 100644
--- a/project/build.sbt
+++ b/project/build.sbt
@@ -18,6 +18,7 @@ addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.2.0")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.4")

 libraryDependencies += "org.eclipse.jgit" % "org.eclipse.jgit.pgm" % "5.10.0.202012080955-r"
+libraryDependencies += "org.openjdk.jol" % "jol-core" % "0.17"

 // scalacOptions used to bootstrap to sbt prompt.
 // In particular, no "-Xfatal-warnings"
diff --git a/tools/src/main/scala/scala/scalanative/linker/Link.scala b/tools/src/main/scala/scala/scalanative/linker/Link.scala
index a8acf077..a2338442 100644
--- a/tools/src/main/scala/scala/scalanative/linker/Link.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/Link.scala
@@ -5,6 +5,8 @@ import scalanative.nir._
 import scalanative.util.Scope
 import scala.scalanative.io.VirtualDirectory
 import java.nio.file.Path
+import org.openjdk.jol.info.GraphLayout
+

 object Link {

@@ -15,6 +17,7 @@ object Link {
   def apply(config: build.Config, entries: Seq[Global])(implicit
       scope: Scope
   ): Result = {
+    println(GraphLayout.parseInstance(cache).totalSize())
     val classpath = config.classPath.map { path =>
       ClassPath(VirtualDirectory.real(path))
     }
```

```
sbt:scala-native> sandbox2_13/nativeLink
# WARNING: Unable to get Instrumentation. Dynamic Attach failed. You may add this JAR as -javaagent manually, or supply -Djdk.attach.allowAttachSelf
# WARNING: Unable to attach Serviceability Agent. Unable to attach even with module exceptions: [org.openjdk.jol.vm.sa.SASupportException: Sense failed., org.openjdk.jol.vm.sa.SASupportException: Sense failed., org.openjdk.jol.vm.sa.SASupportException: Sense failed.]
120
[info] Linking (multithreading false) (867 ms)
[info] Checking intermediate code (79 ms)
[info] Dumping intermediate code (linked) (212 ms)
[info] Optimizing (debug mode) (594 ms)
[info] Checking intermediate code (37 ms)
[info] Dumping intermediate code (optimized) (140 ms)
[info] Dumping intermediate code (lowered) (230 ms)
[info] Generating intermediate code (587 ms)
[info] Produced 39 files
[info] Compiling to native code (189 ms)
[info] Linking native code (immix gc, none lto) (3 ms)
[info] Total (2745 ms)
[success] Total time: 4 s, completed May 17, 2023, 6:18:15 PM
sbt:scala-native> sandbox2_13/nativeLink
36837256
[info] NIR linking skipped due to cache hit.
[info] Linking (multithreading false) (1055 ms)
[info] Checking intermediate code (24 ms)
[info] Dumping intermediate code (linked) (127 ms)
[info] Optimizing (debug mode) (466 ms)
[info] Checking intermediate code (22 ms)
[info] Dumping intermediate code (optimized) (215 ms)
[info] Dumping intermediate code (lowered) (195 ms)
[info] Generating intermediate code (398 ms)
[info] Produced 39 files
[info] Compiling to native code (178 ms)
[info] Linking native code (immix gc, none lto) (3 ms)
[info] Total (2496 ms)
[success] Total time: 3 s, completed May 17, 2023, 6:18:34 PM
```

</details>

WojciechMazur: I think we start with going with the Scala.js approach - they do not cache the results of linking, but they read IR instead. That might make more sense, because any change of inputs, would lead to modification of inputs. However, by keeping the IR in memory they reduce the costs of IO and reading from the disk every time. 
I'm in the progress of rebasing Denys's changes to compact the NIR file size. With this in mind we might use one of the two approaches in the future: keep the deserialized Defns in-memory (scala.js seems to be doing that) or keep the binary NIR in-memory and deserialize it on demand (no IO overhead, lower memory usage, but would require additional cycles for deserialization).

Scala.js implementation is a bit convoluted, but that might be a good starting point: the cache is based on `IRContainers` (jars or directories) and results in `IRFile`s mapping to a single `ir.Tree`. In our case that would be a `Seq[nir.Defn]` or `Map[nir.Global, nir.Defn]` because we keep the flat structure at that point.
https://github.com/scala-js/scala-js/blob/5323e68c2daeb2937f807f0fd2559cbd3e1b6e71/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/IRFileCache.scala#L23

Despite the previous comments, I think we might have 3 levels of caching: 
- config + classpath based - allowing to skip the whole build process if nothing has changed, just return previously linked binary
- NIR cache - to skip unnecessary IO for parts of the classpath that have not changed
- Compilation outputs cache - (we already have that when building with experimental `NativeConfig.incrementalCompilation)), it skips a generation of LLVM IR for parts of the code that have not changed (currently package scoped)

Everything between linking (Reach) and emitting LLVM IR cannot be safely cached unless we would invest heavily in implementing incremental optimizer, and incremental code lowering (involves reimplementation of virtual method dispatch). Otherwise, all these steps are very fragile to any change of the inputs. The next step after that might be incremental linking, but that might be harder to achieve.

ekrich: >config + classpath based - allowing to skip the whole build process if nothing has changed, just return previously linked binary

We have the additional complication that we allow native code in jars and sub-projects that is not present in Scala.js. 

- If the classpath itself does not change and no Scala files in the sub-projects change then no classes or NIR are created then the front end of the build can be skipped as appropriate. This is handled by the build system and is before the compile itself starts.
- We still need to run the backend as native code can change. We currently hash the jar and directories and store the result but there is some work involved to check the hash. We only unpack the code and copy to the build area the first time or if the hash differs. Perhaps at that point we could short circuit the rest of the build rather than calculating time stamps which does take some time. I think that would have to be passed along from phase to phase.

Not sure how much of the is the hash checks vs the timestamps. We also could keep a running max timestamp rather than having to recheck all the objects again to compare for the final check on the final output freshness.

I could work on those optimizations.

I think this is the part of the previous setup that was not correct that I complained about.

tools/src/main/scala/scala/scalanative/linker/Link.scala
ekrich: That was a change that was made to the backend where it goes by each classpath entry. We also started passing the artifacts from the `ll` phase to the `clang` backend as part of that change as well as the artifacts from each native lib to collect for the final link phase.

tools/src/main/scala/scala/scalanative/linker/Link.scala
WojciechMazur: > We still need to run the backend as native code can change.  
The native code is shipped in jars + optionally a single directory for the current project. So if the native code changes, then the classpath have needed to change as well. In such case we would always trigger a new build, and the content of the jars on the classpath would override the previous dirs in our workdir. The only exception from that would involved modification of included headers on the path, and the changed libraries in e.g. `/usr/lib`. These two cases seem to be very unlikely, and even now they're out of our control. We cannot detect their modification and force a fresh linking. 
Based on that it looks to me like the backend could be safely skipped if there are no changes on the classpath and the config has not changed. However, maybe I might be missing something here.

ekrich: The project could have many sub projects with native code and jars as well just like the Scala Native project. If you are checking every jar / directory for changes then the build could be short circuited - shouldn't that be done upfront in the build not in the NIR linking? We could even skip the validation and directory creation as long as the final artifact still exists (no deletion of the build dir).

Also, then how do we know which classpath elements changed so we can re-extract native code downstream and avoid doing the hash that is being done now? Plus there is the jar/dir detection code that needs to be run. Maybe it is better to keep track of new/update/removed to just make downstream detection easier?

tanishiking: Thanks @WojciechMazur @ekrich 

[[So, the initial step would be to establish a "config + classpath based" cache as we initially described in https://github.com/scala-native/scala-native/issues/3282 (If cache hit + there exists the previously generated native binary, skip the whole process and return the path to previous native binary)]], correct?

I'm thinking we can detect the change by recursively checking the mtime of the `config.classPath`, but ...
> We currently hash the jar and directories and store the result but there is some work involved to check the hash. We only unpack the code and copy to the build area the first time or if the hash differs.

do we already have a place that hashes the contents of `config.classPath`? @ekrich I'm not sure how much the timestamp is faster, but I can compare 👍 

The next step would be to implement a "NIR cache", similar to the Scala.js approach (While, I still need to fully understand this idea, need  to review the Scala.js implementation).

> Also, then how do we know which classpath elements changed so we can re-extract native code downstream and avoid doing the hash that is being done now? Plus there is the jar/dir detection code that needs to be run. Maybe it is better to keep track of new/updated/removed to just make downstream detection easier?

This should also be studied in the next step?

---

> We still need to run the backend as native code can change.
>
>  If you are checking every jar / directory for changes then the build could be short circuited - shouldn't that be done upfront in the build not in the NIR linking?
>
> The native code is shipped in jars + optionally a single directory for the current project. So if the native code changes, then the classpath have needed to change as well.

I agree with @WojciechMazur  that if we check for changes in the classpath, we should be able to detect all native code changes (by "native code", we mean NIR rather than native binary, correct?).

Native code input changes when the upstream source (sub-projects or third-party libraries) changes. In such cases, the build system should detect the changes and the scala-native front-end should re-run. This will also trigger a re-linking (NIR -> binary).

As @WojciechMazur mentioned, we cannot detect changes in C libraries located in `LD_LIBRARY_PATH` or similar locations, and users have to manually force a re-run of the linking. However, we may compromise considering that shouldn't be the case so often.

---

> We currently hash the jar and directories and store the result but there is some work involved to check the hash. We only unpack the code and copy to the build area the first time or if the hash differs.

tanishiking: Skipping the whole build process if the input configuration and class paths mtime has unchanged (and output remains the same as the previous build)

tanishiking:
Closing for https://github.com/scala-native/scala-native/pull/3287

PR #3287, opened 2023-05-19
Skip nativeLink if the input hasn't changed from the previous build

tanishiking:
Based on the conversation in https://github.com/scala-native/scala-native/pull/3285

This PR

- Adds an new endpoint `buildCached` to `Build.scala`
- Makes sbt plugin to use `buildCached` instead of `build`.

`buildCached` will skip the whole build process of `nativeLink` if the followings are unchanged from previous build (only 1 previous build, so we don't use much memory to remember the cache).

- Build configuration
- The output native binary('s mtime)
- class path contents(' mtime)

One thing we cannot detect is the change in c library somewhere in `/usr/lib/...`, and if there's a change users may have to explicitly run `sbt clean` to re-run

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -150,8 +150,8 @@ object Build {
     // One thing we miss is, we cannot detect changes in c libraries somewhere in `/usr/lib`.
     (
       config,
-      config.classPath.map(getNewestMtime(_)),
-      getLastModifiedTimeMillis(config.artifactPath)
+      config.classPath.map(getLatestMtime(_)),
+      getLastModified(config.artifactPath)
     ).hashCode()
   }
 
@@ -183,17 +183,18 @@ object Build {
     }
   }
 
-  private def getLastModifiedTimeMillis(path: Path): Long =
-    if (Files.exists(path)) Files.getLastModifiedTime(path).toMillis()
-    else 0L
+  private def getLastModified(path: Path): FileTime =
+    if (Files.exists(path))
+      Try(Files.getLastModifiedTime(path)).getOrElse(FileTime.fromMillis(0L))
+    else FileTime.fromMillis(0L)
 
-  /** Get the newest file's last modified time in millis under the given path.
+  /** Get the latest last modified time under the given path.
    */
-  private def getNewestMtime(path: Path): Optional[FileTime] =
+  private def getLatestMtime(path: Path): Optional[FileTime] =
     if (Files.exists(path))
       Files
         .walk(path, FileVisitOption.FOLLOW_LINKS)
-        .map[FileTime](Files.getLastModifiedTime(_))
+        .map[FileTime](getLastModified(_))
         .max(_.compareTo(_))
     else Optional.empty()
 
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -4,7 +4,6 @@ package build
 import java.nio.file.{Files, Path, Paths}
 import scala.scalanative.util.Scope
 import scala.util.Try
-import java.security.MessageDigest
 import java.nio.file.FileVisitOption
 import java.util.Optional
 import java.nio.file.attribute.FileTime
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -12,7 +12,7 @@ import java.nio.file.attribute.FileTime
 /** Utility methods for building code using Scala Native. */
 object Build {
 
-  private var previousBuildInputHash: Int = 0
+  private var prevBuildInputCheckSum: Int = 0
 
   /** Run the complete Scala Native pipeline, LLVM optimizer and system linker,
    *  producing a native binary in the end.
@@ -69,7 +69,7 @@ object Build {
       val inputHash = checkSum(fconfig)
 
       if (Files.exists(fconfig.artifactPath) &&
-          previousBuildInputHash == inputHash) {
+          prevBuildInputCheckSum == inputHash) {
         fconfig.logger.info(
           "Build skipped: No changes detected in build configuration and class path contents since last build."
         )
@@ -136,12 +136,11 @@ object Build {
     // - the output native binary ('s mtime)
     // Since the NIR code is shipped in jars, we should be able to detect the changes in NIRs.
     // One thing we miss is, we cannot detect changes in c libraries somewhere in `/usr/lib`.
-    val key = (
-      config.toString, // we use toString because `config` object contains Logger instance that changes for every invocation
+    (
+      config,
       config.classPath.map(getNewestMtime(_)),
       getLastModifiedTimeMillis(config.artifactPath)
-    )
-    key.hashCode()
+    ).hashCode()
   }
 
   /** Convenience method to combine finding and compiling native libaries.
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -4,9 +4,14 @@ package build
 import java.nio.file.{Files, Path, Paths}
 import scala.scalanative.util.Scope
 import scala.util.Try
+import java.security.MessageDigest
+import java.nio.file.FileVisitOption
+import scala.collection.JavaConverters._
 
 /** Utility methods for building code using Scala Native. */
 object Build {
+
+  private var previousBuildInputHash: Int = 0
 
   /** Run the complete Scala Native pipeline, LLVM optimizer and system linker,
    *  producing a native binary in the end.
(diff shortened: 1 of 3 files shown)


ekrich:
This is definitely along the lines of what I was thinking. Does this cover the case of the multiple mains where we generate the classname binary and then copy or rename to the artifactName? If the config changes in that case we are good. You can put 2 main classes in sandbox to test I would guess.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -53,60 +58,87 @@ object Build {
...
+  def build(config: Config)(implicit scope: Scope): Path = {
+    def makeKey(config: Config): Int = {
+      // skip the whole nativeLink process if the followings are unchanged since the previous build
+      // - build configuration
+      // - class paths' mtime
+      // - the output native binary ('s mtime)
+      // Since the NIR code is shipped in jars, we should be able to detect the changes in NIRs.
+      // One thing we miss is, we cannot detect changes in c libraries somewhere in `/usr/lib`.
+      val key = (
+        config.toString, // we use toString because `config` object contains Logger instance that changes for every invocation
```
WojciechMazur, line 70: We can make `logger` an internal field of `Config`, which would not be a field of the constructor. Any interaction with it should happen using `withLogger` accessor. That should simplify and possibly save a few cycles of the CPU.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -53,60 +58,87 @@ object Build {
    *  @return
    *    [[Config#artifactPath]], the path to the resulting native binary.
    */
-  def build(config: Config)(implicit scope: Scope): Path =
+  def build(config: Config)(implicit scope: Scope): Path = {
+    def makeKey(config: Config): Int = {
```
WojciechMazur, line 62: This function seems to be big enough to make a private a in Build instead of nesting it here. Also, we could rename it a bit - We do calculate hashCode and assign it to the `inputHash` field,  so `configHashCode`, `getHashCode`, `calcHashCode`, `checkSum` or other similiars might be more appropriate.

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -53,60 +58,87 @@ object Build {
...
-        ScalaNative.logLinked(fconfig, linked)
-        linked
-      }
+      if (Files.exists(fconfig.artifactPath) &&
+          previousBuildInputHash == inputHash) {
+        fconfig.logger.info(
+          "Build skipped: No changes detected in build configuration and class path contents since last build."
+        )
+        fconfig.artifactPath
+      } else {
```
WojciechMazur, line 90: It might be useful to refactor the `else` branch into a separate method. That way we still would be able to access it while testing/benchmarking to ensure caching would not interfere with the results. We can make it public to allow other build tools to access it directly if they don't want to relay on our caching. So base on that we can rename current `build` to `buildCached`, and refactor the else part into the `build`

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -136,4 +168,18 @@ object Build {
...
+
+  /** Get the newest file's last modified time in millis under the given path.
+   */
+  private def getNewestMtime(path: Path): Long =
+    Files
+      .walk(path, FileVisitOption.FOLLOW_LINKS)
+      .iterator()
+      .asScala
+      .map(p => Files.getLastModifiedTime(p).toMillis())
+      .max
```
WojciechMazur, line 183: There seemed to be some deprecation warning elevated to the errors due to `-Xfatal-warnings` flag. 
However, in this case we can just fallback to Java streams and skip converting `Stream->Iterator->scala.Iterator`: 

```scala
Files
  .walk(path, FileVisitOption.FOLLOW_LINKS)
  .map(Files.getLastModifiedTime(_))
  .max(_.compareTo(_)) // : j.u.Optional[j.n.f.a.FileTime]
```
Probably we can use resulting `Optional[FileTime]` directly - when I've tested it in REPL it had a stable hashCode

tools/src/main/scala/scala/scalanative/build/Build.scala
ekrich: You might consider using the following `buildImpl` code as it also solves the confusion of not using the correct config reference after filtering as well as having the ability to return with an `if` in the short code section. https://github.com/scala-native/scala-native/pull/3283

tanishiking:
> This is definitely along the lines of what I was thinking. Does this cover the case of the multiple mains where we generate the classname binary and then copy or rename to the artifactName? If the config changes in that case we are good. You can put 2 main classes in sandbox to test I would guess.


@ekrich Thank you for your help! Yes, (because `artifactPath` should aware of the multiple main) if we have two main like this

```scala
object Test {
    println("Hello, World!")
  }
}

object Test2 {
    println("Hello, World!")
  }
}
```

It works like this (build `Test` subsequently skips the build, and when we build `Test2` it builds from scratch).

```scala
sbt:scala-native> sandbox2_13/run
[info] compiling 1 Scala source to /Users/tanishiking/ghq/github.com/tanishiking/scala-native/sandbox/.2.13/target/scala-2.13/classes ...

Multiple main classes detected. Select one to run:
 [1] Test
 [2] Test2

Enter number: 1

[info] Linking (multithreading false) (749 ms)
[info] Checking intermediate code (104 ms)
[info] Dumping intermediate code (linked) (235 ms)
[info] Optimizing (debug mode) (709 ms)
[info] Checking intermediate code (44 ms)
[info] Dumping intermediate code (optimized) (167 ms)
[info] Dumping intermediate code (lowered) (218 ms)
[info] Generating intermediate code (560 ms)
[info] Produced 39 files
[info] Compiling to native code (231 ms)
[info] Linking native code (immix gc, none lto) (3 ms)
[info] Total (2859 ms)
Hello, World!
[success] Total time: 10 s, completed May 22, 2023, 6:17:31 PM
sbt:scala-native>
sbt:scala-native> sandbox2_13/run

Multiple main classes detected. Select one to run:
 [1] Test
 [2] Test2

Enter number: 1
[info] Build skipped: No changes detected in build configuration and class path contents since last build.
Hello, World!
[success] Total time: 2 s, completed May 22, 2023, 6:17:34 PM
sbt:scala-native> sandbox2_13/run

Multiple main classes detected. Select one to run:
 [1] Test
 [2] Test2

Enter number: 2
[info] Linking (multithreading false) (362 ms)
[info] Checking intermediate code (18 ms)
[info] Dumping intermediate code (linked) (126 ms)
[info] Optimizing (debug mode) (323 ms)
[info] Checking intermediate code (25 ms)
[info] Dumping intermediate code (optimized) (122 ms)
[info] Dumping intermediate code (lowered) (207 ms)
[info] Generating intermediate code (595 ms)
[info] Produced 39 files
[info] Compiling to native code (343 ms)
[info] Linking native code (immix gc, none lto) (85 ms)
[info] Total (2012 ms)
Hello, World!
[success] Total time: 4 s, completed May 22, 2023, 6:17:40 PM
```

tools/src/main/scala/scala/scalanative/build/Build.scala
tanishiking: I added `buildCache` endpoint, and left the `build` same, so build tools can choose which endpoint to use

tools/src/main/scala/scala/scalanative/build/Build.scala
tanishiking: Renamed as `checkSum` and made it as a private method 👍 https://github.com/scala-native/scala-native/pull/3287/commits/0ed856ef6511b8e4b762f1aeb4fcdf212cb7e359

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -136,4 +184,18 @@ object Build {
...
+    if (Files.exists(path)) Files.getLastModifiedTime(path).toMillis()
+    else 0L
+
+  /** Get the newest file's last modified time in millis under the given path.
+   */
+  private def getNewestMtime(path: Path): Optional[FileTime] =
+    if (Files.exists(path))
+      Files
+        .walk(path, FileVisitOption.FOLLOW_LINKS)
+        .map[FileTime](Files.getLastModifiedTime(_))
```
WojciechMazur, line 197: It's possible that this call might throw an exception if the file was removed in the meantime or we don't have permissions. I think it would be best to refer to the getLastModifiedTimeMillis defined above (possibly both can reduce FileTime (and FileTime.fromMilllis(0)) for non existing files)

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -136,4 +184,18 @@ object Build {
...
+        .max(_.compareTo(_))
```
WojciechMazur, line 198: I forget we can use prediefined comparator here: java.util.Comparator.naturalOrder

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -136,4 +184,18 @@ object Build {
     }
   }
 
+  private def getLastModifiedTimeMillis(path: Path): Long =
+    if (Files.exists(path)) Files.getLastModifiedTime(path).toMillis()
+    else 0L
+
+  /** Get the newest file's last modified time in millis under the given path.
+   */
+  private def getNewestMtime(path: Path): Optional[FileTime] =
```
WojciechMazur, line 193: Probably `latest` would be better name then `newest`.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -180,43 +180,43 @@ object Config {
...
       classPath: Seq[Path],
-      logger: Logger,
       compilerConfig: NativeConfig
+  )(
+      val logger: Logger // Exclude logger from hashCode calculation https://stackoverflow.com/questions/10373715/scala-ignore-case-class-field-for-equals-hascode
   ) extends Config {
 
     def withBaseDir(value: Path): Config =
-      copy(baseDir = value)
+      copy(baseDir = value)(logger)
```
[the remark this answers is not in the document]
tanishiking: Good idea 👍 done https://github.com/scala-native/scala-native/pull/3287/commits/5dbbeec6c26133964f5d873254d3fe885b2331c1

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -4,10 +4,44 @@ package build
 import java.nio.file.{Files, Path, Paths}
 import scala.scalanative.util.Scope
 import scala.util.Try
+import java.security.MessageDigest
```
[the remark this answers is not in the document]
tanishiking: oops, right https://github.com/scala-native/scala-native/pull/3287/commits/ac270eb817331a21c0ae127fe3bc426c5c762c55

WojciechMazur (approved):

tools/src/main/scala/scala/scalanative/build/Build.scala
tanishiking: Remained same as `_max(_.compareto(_))` since `java.util.Comparator.naturalOrder` doesn't work for `FileTime`, or do you mean we can make this function return `Long` instead of `FileTime` (I think we can because we just calculate hashCode from `FileTime` instance).

tools/src/main/scala/scala/scalanative/build/Build.scala
tanishiking: Right, I made a change like this https://github.com/scala-native/scala-native/pull/3287/commits/5a4ee264a5b5d966fb2600c25ce5c44f5feda537

Maybe we should print warning for Exceptions in `getLastModified`?

ekrich:
Well, I just think it is very weird to have two entry points and I thought we would proceed to `0.5.0` shortly. I am not sure why there couldn't have been some alternative. Really, `buildCached` should be an implementation detail.

WojciechMazur:
My reasoning here is that the build tool which is going to be integrated might have some better context to reason about when to invalidate the cache. For example, it might use the in-memory information directly from the incremental compiler, allowing to skip I/O involved in checking the classpath for changes. Also, if we would make the default `build` optionally cached, then we would silently break the current contract between the linker and build tool. No change in signature, but a difference in runtime, which might be unexpected, especially in terms of mill and scala-cli they which might support multiple minors of Scala Native at the same time. In the case of sbt, we have full control over that and work with only 1 version of SN. 
If my reasoning is flawed, then we could modify the current implementation, to make `build` cached by default, and introduce alternative uncached version available for benchmarks/build tools.

@lolgab AFAIK you're maintaining the mill Scala Native plugin. Do you have some thoughts about this topic?

ekrich:
I was hoping the SN build after the NIR could be pretty self contained. Current we only have the `Config` and the `implicit Scope` as inputs. If we could pass what has changed from the Scala Compiler phase or from the classpath  / `Config` perhaps the downstream code could use that smartly. Passing the artifacts from phase to phase was the first logical step in that direction.

> Also, if we would make the default build optionally cached, then we would silently break the current contract between the linker and build tool.

I am not sure exactly what you are getting at here.

tanishiking:
I added a new endpoint `buildCached` instead of making change on the `build`  endpoint, which is already used from various build tools, so that build tools authors can gradually migrate to the `buildCached`.

> Really, buildCached should be an implementation detail.

What is your concern for having two endpoints? (Perhaps because we will have to maintain two endpoints' binary compatibility?)
I'm not sure why `buildCached` should be an implementation detail, and it's wired to have multiple endpoint in Build. Maybe I'm missing something.

PR #3313, opened 2023-06-05
Try to fix resolving methods using LinktimeIf produced using 0.4.12-

WojciechMazur:


armanbilge (approved):
Can confirm this fixes https://github.com/armanbilge/epollcat. But I still recommend that libraries should re-publish to take advantage of the new mechanism that is based on the target triple.

PR #3314, opened 2023-06-05
Use conditional compilation for GC selection

ekrich:
This is part of a larger idea / experiment for the backend compilation where we can finally have the C code and glue code in the library it belongs in. The cool thing here is that we can use `clangd` and it resolves and can help development by making the header navigable and shows syntax errors etc. See `compile_flags.txt` file.

This could also be used to solve the linking problem when the library is not used or not installed. The conditional compilation trick for the header works https://github.com/scala-native/scala-native/issues/2778 like `#if __has_include(<unicode/uidna.h>)` but this is slightly different where the front end NIR compiler can let us know if the `@link` library is needed or not.

Currently there are two special rules to compile the `nativelib`. 

1. The first is the optional lib `z` for the zip support that is really should be in `javalib`.
2. The second is the GC compilation which this experiment shows how it might work. I still think it might be a good idea to have it in a separate project in the future to make it more contained.

I started to look at some other build tools that could be combined with the library but really didn't find anything that really could make it very easy on the library developer. We have already went down the path of build plugins and this doesn't seem like a way forward either. I think the more contained the back end compilation can be even by making it a separate object may be a good option.

The proposal consists of two parts.

1. We need to use the fact that the front end knows whether we need to compile the file like `z.c` when `libz` is needed in the `javalib` along with setting a `-DLINK_Z` for example. The backend code would have matching conditional compilation and then use that to make sure the `z.c` gets compiled.
2. We need to allow each library to have a `scala-native.properties` file to tell the build code that we should be looking for a library to link.

The downside is that this introduces a new build descriptor. I have struggled with this a long time but it does separate the compilation front end to the back end nicely so if we find a better way it could be replaced. Here are some ideas that could be included.

- Open up the build options so different options could be included for different libraries. This would allow `clib` to not use `gnu` option and to allow POSIX lib to define its own conformance level.
- This could also allow a Scala Native library target to be compiled as a portion of a larger project.
- We could also consider allowing compiler options for different target triple components like architecture.

The property based language could get quite complex so deciding what is reasonable to support would be good.

I am really a big fan or trying to do everything this way https://github.com/scala-native/scala-native/pull/3217 because it is complex and I think that getting to the point where the libraries are self contained would be good.


nativelib/src/main/resources/scala-native/compile_flags.txt
```diff
@@ -0,0 +1,33 @@
+# GC setup
```
WojciechMazur, line 1: This file should not be included

tools/src/main/scala/scala/scalanative/build/Filter.scala
```diff
@@ -61,9 +61,14 @@ private[scalanative] object Filter {
...
 
+      val gcFlag = {
+        val gc = config.compilerConfig.gc.toString
+        s"-DSCALANATIVE_GC_${gc.toUpperCase}"
+      }
+
       val projectConfig = config.withCompilerConfig(
         _.withCompileOptions(
-          config.compileOptions ++ gcIncludePaths.map("-I" + _)
+          config.compileOptions ++ gcIncludePaths.map("-I" + _) :+ gcFlag
```
WojciechMazur, line 71: If we use C preprocessor conditions we could potentially get rid of the `includes` paths passed here and just compile everything. It would also make sure that is coverts everything.

WojciechMazur:
> We need to allow each library to have a scala-native.properties file to tell the build code that we should be looking for a library to link. 

I don't really understand this part. Would this file contain information on which libraries to include? If so, it is already converted by a global list of `@link` annotations. If it is meant to inform if a given jar contains foreign sources to compile, then it is covered by the existence of `resources/scala-native`, while `scala-native.properties` currently is not mentioned anywhere in the logic. 
The only thing that it might be useful would be containing a list of clang options allowing to tune how to compile sources. That one might be really useful, but would require refactoring of our current pipeline.  

> We need to use the fact that the front end knows whether we need to compile the file like z.c when libz is needed in the javalib along with setting a -DLINK_Z for example. The backend code would have matching conditional compilation and then use that to make sure the z.c gets compiled. 

So if I understand it correctly, based on the collected `@link` we would get the list of required libs to link and based on them list of preprocessor definitions. If by backend you refer to Scala Native toolchain at the point of compiling sources then, this information would not be enough, because multiple files might require given library. Otherwise, if you meant LLVM backend then files would be always compiled, but sources would need to be wrapped in `ifdef LINK_LIB`. The problem with that we quickly get to the coloring of the native sources - if you use anything that is defined by other source using `lib` then you also need to wrap the current file as well. Maybe that is not a big issue, but that would definitely be important DX aspect when compiling bigger code bases spanned across multiple source files. 

> Open up the build options so different options could be included for different libraries. This would allow clib to not use gnu option and to allow POSIX lib to define its own conformance level. 

I'm not sure what `opening up the build options` would mean. But, if we would like to allow to define custom compilation options, why not go one step further, and allow to define a Makefile in sources? In that case, each library's foreign sources would be self-contained, allowing to fully define how to compile sources. As a contract, each makefile may create a file listing all object files, which should be linked by SN backend. Also these Makefiles could take list of capabilities allowing to enable or disable some features. These could be collected similarly to `@link`/`@compile`, and might work somehow like Rust features. 
Alternatively, we would need to have a definition of options defined in foreign sources. In such I'm not sure if Java properties files would be the best format for that, but that might be discussed other day. 

One thing that would be the most important for me, is moving responsibility on the author of foreign sources,a not on user. This is what `@link` was originally meant to do - you didn't need to think about explicitly adding list of libraries as a user. Similarly, I wish that conditional compilation of foreign sources should work similarly. It should be responsibility of author to ensure the thing compiles. User should only be responsible for providing the implementation in the form of having library installed on the path

nativelib/src/main/resources/scala-native/compile_flags.txt
ekrich: Have you seen what that can do with `clangd`? It could really help developers. We could always ignore it and have some docs so Scala Native developers could use it.

tools/src/main/scala/scala/scalanative/build/Filter.scala
ekrich: The latest update uses one include path and cleans up the build.

ekrich:
I was trying to let you know my ideas more than anything. I don't see any solutions that are perfect but I wanted to try and make improvements.

My goal was to remove the special logic for `optional` and to solve the issue of glue code trying to compile link when it is not used - generalize a method. My idea was that if the library via `scala-native.properties` identifies needed libraries and then the filter code could search for the `@link` and if found then add e.g. `-DNEEDS_LIB_Z` so the glue code can conditionally compile. It could also be for not found. This puts all the burden on the library author not the user.

Overall, I was starting to realize that using the features in C, conditional compilation is more direct and understandable. Maybe, in the future the interopt can get updated so we don't need glue code.

Since this is a large change already, maybe we can just do the GC piece first?

nativelib/src/main/resources/scala-native/compile_flags.txt
ekrich: ok, removed it and added a .gitignore

tools/src/main/scala/scala/scalanative/build/Filter.scala
ekrich: I left the `gc` path because we have so many things in the `nativelib`.

ekrich:
@WojciechMazur Ready for another review.

I can add some docs for the `clangd` setup which is much simpler now or I can followup with another PR.

```
# GC setup
# Boehm path on mac arm
-I
/opt/homebrew/include
-I
gc/
-DSCALANATIVE_GC_BOEHM
-DSCALANATIVE_GC_IMMIX
-DSCALANATIVE_GC_COMMIX
-DSCALANATIVE_GC_NONE
-DSCALANATIVE_GC_EXPERIMENTAL
-DSCALANATIVE_MULTITHREADING_ENABLED
-DENABLE_GC_STATS
-DENABLE_GC_STATS_SYNC
-DDEBUG_PRINT
-DDEBUG_ASSERT
# end GC
```

WojciechMazur (approved):

PR #3324, opened 2023-06-10
Port `ConcurrentNavigableMap.java` from JSR-166

mox692:
Ported from jsr166 ([revision 1.20](https://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/jsr166/src/main/java/util/concurrent/ConcurrentNavigableMap.java?revision=1.20&view=markup)).
`ConcurrentNavigableMap` is required for `ConcurrentSkipListMap`.

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/ConcurrentNavigableMap.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ConcurrentNavigableMap.scala
@@ -11,29 +11,10 @@ import java.util.NavigableSet;
  * http://creativecommons.org/publicdomain/zero/1.0/
  */
 
-/** A {@link ConcurrentMap} supporting {@link NavigableMap} operations, and
- *  recursively so for its navigable sub-maps.
- *
- *  <p>This interface is a member of the <a href="{@docRoot
- *  }/java.base/java/util/package-summary.html#CollectionsFramework"> Java
- *  Collections Framework</a>.
- *
- *  @author
- *    Doug Lea
- *  @param <K>
- *    the type of keys maintained by this map
- *  @param <V>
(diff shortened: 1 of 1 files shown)


javalib/src/main/scala/java/util/concurrent/ConcurrentNavigableMap.scala
```diff
@@ -0,0 +1,125 @@
...
+  override def headMap(toKey: K): ConcurrentNavigableMap[K, V]
+
+  override def tailMap(fromKey: K): ConcurrentNavigableMap[K, V]
+
+  /** Returns a reverse order view of the mappings contained in this map. The
+   *  descending map is backed by this map, so changes to the map are reflected
+   *  in the descending map, and vice-versa.
+   *
+   *  <p>The returned map has an ordering equivalent to {@link
+   *  java.util.Collections# reverseOrder ( Comparator )
```
WojciechMazur, line 65: Scalafmt does not work well with JavaDoc comments - because the `{@link java.util.Collections#}` is spanned across multiple lines Scaladoc would fail to build it. In most of JSR166 stuff (e.g. ForkJoinPool) javadoc comments were completely removed because of multiple issues related to that. Alternative is to use ScalaDoc syntax using brackets, however in case of JavaLib we can just get rid of documentation. 
We still have official documentations, which would be typically used, while our generation of javadocs could be mostly used to compare which classess/methods are available in our implementation.

mox692: thanks, I omitted document in 4d6911c.

WojciechMazur (approved):

PR #3342, opened 2023-06-19
Port Striped64, LongAdder from JSR-166

mox692:
This PR Ports Striped64 and LongAdder, which is needed for ConcurrentSkipListMap.

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
+++ b/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
@@ -1,3 +1,5 @@
+// Ported from JSR 166 revision 1.23
+
 package java.util.concurrent.atomic
 
 import java.io.Serializable
--- a/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
+++ b/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
@@ -12,7 +12,8 @@ import java.io.Serializable
 @SerialVersionUID(7249069246863182397L)
 object LongAdder {
 
-  // for sufficient serialization, without unnecessary parent class info.
+  // This SerializationProxy provides sufficient serialization for LongAdder,
+  // without unnecessary parent class info.
   @SerialVersionUID(7249069246863182397L)
   private class SerializationProxy(a: LongAdder) extends Serializable {
     final private var value = a.sum
--- a/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
+++ b/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
@@ -12,8 +12,9 @@ import java.io.Serializable
 @SerialVersionUID(7249069246863182397L)
 object LongAdder {
 
+  // for sufficient serialization, without unnecessary parent class info.
   @SerialVersionUID(7249069246863182397L)
-  private class SerializationProxy(val a: LongAdder) extends Serializable {
+  private class SerializationProxy(a: LongAdder) extends Serializable {
     final private var value = a.sum
 
     private def readResolve = {
--- a/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
+++ b/javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
@@ -34,7 +34,7 @@ class LongAdder() extends Striped64 with Serializable {
     var m = 0
     var c: Striped64.Cell = null
     if ({ cs = cells; cs != null || !casBase({ b = base; b }, b + x) }) {
-      val index = Striped64.getProbe(this)
+      val index = Striped64.getProbe()
       var uncontended = true
       if ({
         m = cs.length;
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/util/concurrent/atomic/Striped64.scala
```diff
@@ -0,0 +1,288 @@
...
+ * Written by Doug Lea with assistance from members of JCP JSR-166
+ * Expert Group and released to the public domain, as explained at
+ * http://creativecommons.org/publicdomain/zero/1.0/
+ */
+
+@SuppressWarnings(Array("serial"))
+object Striped64 {
+
+  final private[atomic] class Cell private[atomic] (var value: Long) {
+    @volatile private[concurrent] var _value = value
```
armanbilge, line 25: Why do we need `value` and `_value`? If they need to be separate, we should remove `var` from `value`, otherwise we have two fields instead of one field and a ctor argument..

Also, I think the original Java version has some padding here? I'm not sure if that trick works in Scala Native, but if it does, we should port it as well. If it doesn't, we should figure out a way to support that kind of memory layout, because it's important for concurrent code.

For a discussion about padding and how it prevents "false-sharing" on the JVM, see https://github.com/typelevel/cats-effect/pull/2543#issuecomment-968362106.
> That would create false sharing, i.e. two independent fields residing on the same cache line while different threads independently write to each variable. Each write to either field in this case invalidates the cache of the other thread, slowing everyone down.

armanbilge: Relevant JSR166 code.
https://github.com/armanbilge/jsr166/blob/abd155f72243cc2066c31ee732d21894f0edf7bd/src/jsr166e/Striped64.java#L88-L92

javalib/src/main/scala/java/util/concurrent/atomic/Striped64.scala
```diff
@@ -0,0 +1,288 @@
...
+    @alwaysinline def valueAtomic() = new CAtomicLongLong(
+      fromRawPtr(Intrinsics.classFieldRawPtr(this, "_value"))
+    )
+
+    final private[atomic] def cas(cmp: Long, `val`: Long) =
+      valueAtomic().compareExchangeWeak(
+        cmp,
+        `val`,
+        memory_order.memory_order_release
+      )
```
armanbilge, line 36: Dumb question: in JSR166 I see the following comment, is that what you've done here with `memory_order_release`?

```
     * JVM intrinsics note: It would be possible to use a release-only
     * form of CAS here, if it were provided.
```

armanbilge: Ah, sorry I am looking at an old version. This matches the latest code :)

javalib/src/main/scala/java/util/concurrent/atomic/Striped64.scala
armanbilge: Sorry, just realized I was looking at an old version :)

However, I see the latest code uses `@jdk.internal.vm.annotation.Contended`. This is an important annotation that relates to the false-sharing issue. @WojciechMazur I wonder if we can port its semantics to Scala Native?

> At the VM’s discretion, those fields annotated as contended are given extra padding so they do not share cache lines with other fields that are likely to be independently accessed. This helps to avoid cache contention across multiple threads.

https://docs.azul.com/prime/Contended

javalib/src/main/scala/java/util/concurrent/atomic/Striped64.scala
mox692: yeah I refered to [VarHandle.weakCompareAndSetRelease](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/invoke/VarHandle.html#weakCompareAndSetRelease(java.lang.Object...)), and implemented it for scala-native.

javalib/src/main/scala/java/util/concurrent/atomic/Striped64.scala
armanbilge: I opened an issue to track this.
- https://github.com/scala-native/scala-native/issues/3359

javalib/src/main/scala/java/util/concurrent/atomic/LongAdder.scala
```diff
@@ -1,55 +1,95 @@
...
 
-class LongAdder extends Number with Serializable {
-  private[this] var value: Long = 0L
+@SerialVersionUID(7249069246863182397L)
+object LongAdder {
 
-  final def add(x: Long): Unit =
-    value = value + x
+  @SerialVersionUID(7249069246863182397L)
+  private class SerializationProxy(val a: LongAdder) extends Serializable {
```
WojciechMazur, line 11: Should `a: LongAdder` be only constructor argument and not a field? Otherwise it looks like a cyclic dependency for me, becouse we would again want to serialize field of LongAdder type in it's serializer. 

```suggestion
  private class SerializationProxy(a: LongAdder) extends Serializable {
```

Also you can point why it's needed (I guess that's it's done to hide the implementation detail of Stripped64 transient fields)

WojciechMazur (commented):
Thank you for this contribution, I've added a few comments.

WojciechMazur (approved):

Issue #97, opened 2016-05-17
CBT tooling support


RomanHargrave:
I have a few projects I have been wanting to do, and I would like to use SN for them. They aren't urgent, so chances are that I will wait for some level of SN maturity. But nonetheless, it would be nice to have good tooling for SN.

For instance, C is pretty easy to write builds for due to it's simple and well documented tooling. Due to the modular nature of SN, tooling more like D would have to be used, but nonetheless, there is still a single go-to utility that can locate a default runtime library, build object files, and link libraries and executables. 

Currently, SN requires invoking the scala compiler with a path to the plugin, and the build automation support is just not there (I don't really like SBT all that much).

densh:
This is a great discussion topic @RomanHargrave, thanks for bringing this up.

Our current tooling strategy is based on sbt. We're working on a plugin that should make it extremely easy to build/run and publish Scala Native code. This build tool was chosen as it's most familiar to existing Scala developers. I understand your gripes with `sbt`, it has a bit of a learning curve, even if you're just using the basic features. We'll do our best to document the ways we expect people to build Scala Native with `sbt` on our side. 

An alternative build tool support is borderline out-of-scope for the project but I would gladly help anyone exploring this space. I'm leaving this ticket open for discussion purposes.

RomanHargrave:
If an SBT plugin is in the pipeline, that may well serve as good enough documentation for alternative build methods.

My dislike for SBT stems mainly from the issues that plagued older versions (e.g. newlines, etc, didn't behave like scala even though it is scala). Nonetheless, what would be needed for implementing other systems is documentation on how the compiler is invoked and what parameters must be fed to it. That's the most important part.

frgomes:
SBT is complex internally and difficult to understand even if you are not interested on its internals. Despite SBT wosks really well and its very flexible, I cannot stop thinking that its over-engineered.

Recently I've discovered CBT, which embraces the lemma that "a task is a def". Looks pretty simple and flexible enough.

i will try CBT on my http://github.com/frgomes/poc-scala-native

I've also asked in CBT's gitter about the idea of porting CBT to SN. No feedback yet... maybe I even try that...

The point is: I find SBT needlessly complex. Being so dependent on macros, it diverges from future advancements in the language and basically anchors itself on Scala 2.10, which obviously depends on the JVM, which is exactly what we intend to get rid of with SN.

Milyardo:
[...]

PR #3343, opened 2023-06-21
Add filename and line number to backtrace

tanishiking:
Based on https://github.com/scala-native/scala-native/pull/2869

## TL;DR
This PR adds filename and line number to scala-native backtrace using DWARF debug information.

```
java.lang.Error: test
        at java.lang.StackTrace$.$anonfun$currentStackTrace$1(Throwables.scala:56)
        at java.lang.StackTrace$$$Lambda$3.apply(Throwables.scala:56)
        at scala.scalanative.unsafe.Zone$.apply(Zone.scala:27)
        at java.lang.StackTrace$.currentStackTrace(Throwables.scala:50)
        at java.lang.Throwable.fillInStackTrace(Throwables.scala:126)
        at Test$.error(Test.scala:9)
        at Test$.g(Test.scala:7)
        at Test$.f(Test.scala:5)
        at Test$.main(Test.scala:1)
        at Test.main(Test.scala:1)
```

Read/parse myself executable at runtime (when `fillInStacktrace` is called), and extract debug information and add the filename and line number information to backtrace. Also, the runtime will read `/proc/self/maps` or call `vmmap <pid>` to calculate the ASLR offset.

- [ ] Support various executable format
  - [x] MachO
  - [ ] ELF
  - [ ] PE/COFF
- [ ] Enable to work under PIE/ASLR
  - [x] OSX/MACHO (`vmmap`)
  - [ ] ELF/Linux (`/proc/self/maps`)
  - [ ] Windows (?)
- [ ] Read separate debug file
  - [x] dSYM
  - [ ] build id
  - [ ] gnu debug link
- [x] optional: binary search DIE

## Background
Current `scala-native` backtrace doesn't have filename and line number information.

For example:

```scala
object Test {
    f()
  }

  def f() = g()

  def g() = error()

  def error() = throw new Error("test")
}
```

scala-native will emit the following backtrace, that doesn't have `Unknown Source` instead of filename and line number.

```
java.lang.Error: test
        at java.lang.StackTrace$.$anonfun$currentStackTrace$1(Unknown Source)
        at java.lang.StackTrace$$$Lambda$2.apply(Unknown Source)
        at Test$.error(Unknown Source)
        at Test$.g(Unknown Source)
        at Test$.f(Unknown Source)
        at Test$.main(Unknown Source)
        at Test.main(Unknown Source)
```

scala-native construct backtrace using [(llvm) libunwind](https://bcain-llvm.readthedocs.io/projects/libunwind/en/latest/) around [here](https://github.com/scala-native/scala-native/blob/e5ee79a0fc39bc86bbf4a5de0c211be0ed1c2019/javalib/src/main/scala/java/lang/Throwables.scala#L45-L55) and it gives us the (mangled) symbol name and the call address, but nothing beyond.

However, thanks to @keynmol's recent effort https://github.com/scala-native/scala-native/pull/2869 scala-native will be able to generate DWARF debug information. And they (filename and line number) are in the DWARF information of the binary, and given the call address we can extract the exact call location.

Also, he developed DWARF parser (+ macho parser) in Scala, which can be re-used in scala-native implementation https://github.com/indoorvivants/macho-elf-coff-parser/ :tada:

## Algorithm overview

- When exception is raised and called `fillInStackTrace`
- Extract the DWARF debug information by parsing myself executable (currently macho is supported)
  - It takes approximately 300ms (for sandbox project, on my local machine). It's fast as something like `llvm-dwarfdump` AFAIK, and it's only one time. However, still it hurts performance if the program contains a Throwable object 🤔 
- Calculate the ASLR slide by running `vmmap` command against for myself PID (for OSX, for Linux we will need to check `/proc/<pid>/maps`).
- Search for a Debug Information Entry that corresponds to the address of the subprogram (retrieved from `unwind.get_reg(cursor, unwind.UNW_REG_IP, ip)` and adjusted for ASLR offset)
- We should be able to get the subprogram (function) name and line number from DIEs.

## What this PR does

This PR 

- Migrate https://github.com/indoorvivants/macho-elf-coff-parser/ to this repository
  - Added some improvement to speed up parsing DWARF information
- Read/parse myself executable (`argv[0]`) and read/parse DWARF information
- Read `vmmap` (in OSX) / `/proc/self/maps` to calculate ASLR offset
  - I have an WIP small blog post why this is needed https://gist.github.com/tanishiking/3fbacf36a12dfd0991127f3db6df9e53
- Search for an DIE that has filename and line number
- Add those information to `StacktraceElement`

As a result, scala-native's backtrace (for the above Scala snippet) looks like this:

```
java.lang.Error: test
        at java.lang.StackTrace$.$anonfun$currentStackTrace$1(Throwables.scala:56)
        at java.lang.StackTrace$$$Lambda$3.apply(Throwables.scala:56)
        at scala.scalanative.unsafe.Zone$.apply(Zone.scala:27)
        at java.lang.StackTrace$.currentStackTrace(Throwables.scala:50)
        at java.lang.Throwable.fillInStackTrace(Throwables.scala:126)
        at Test$.error(Test.scala:9)
        at Test$.g(Test.scala:7)
        at Test$.f(Test.scala:5)
        at Test$.main(Test.scala:1)
        at Test.main(Test.scala:1)
```

## References

- libunwind basics
  - [Programmatic access to the call stack in C++ - Eli Bendersky's website](https://eli.thegreenplace.net/2015/programmatic-access-to-the-call-stack-in-c/)
  - [libunwind documentation](https://www.nongnu.org/libunwind/docs.html) (though scala-native uses llvm one)
- Confirm we're on a right direction
  - [ianlancetaylor/libbacktrace: A C library that may be linked into a C/C++ program to produce symbolic backtraces](https://github.com/ianlancetaylor/libbacktrace)
  - [Backtrace line numbers by ysbaddaden · Pull Request #3303 · crystal-lang/crystal](https://github.com/crystal-lang/crystal/pull/3303)
- Workaround for PIE/ASLR
  - [Not getting line numbers with addr2line · Issue #97 · boostorg/stacktrace](https://github.com/boostorg/stacktrace/issues/97)
  - (Chinese blog post about Mach-O and ASLR)  [Mach-O 文件之进程(虚拟)地址空间、ASLR - 简书](https://www.jianshu.com/p/7ad3d3d868f9)
  - [How do you read the memory maps of a Mac process?](https://jvns.ca/blog/2018/01/26/mac-memory-maps/)
- DWARF
  - [How debuggers work: Part 3 - Debugging information - Eli Bendersky's website](https://eli.thegreenplace.net/2011/02/07/how-debuggers-work-part-3-debugging-information)
  - (Japanese blog post about debug info) [デバッグ情報の歩き方 - Qiita](https://qiita.com/mhiramat/items/8df17f5113434e93ff0c)


tanishiking:
I found reading DWARF especially `readByte` from `RandomAccessFile` adds up to around 3000-3500ms for `sandbox` project in my local environment, I guess we have to find some ways to read file efficiently with random access. (FileChannel may help?)

Also, going to write scripted tests

keynmol:
re: RandomAccessFile

It's _horribly slow_: https://bugs.openjdk.org/browse/JDK-4219374 apparently because there is no buffering.

On the JVM that is. Not sure about Scala Native but probably same, for the same reasons

Regardless, we should use a buffered version of that, i.e. when you read a location:

1. Check if this location falls into some small (5 elements) set of pre-buffered locations (see below)
2. If so, load the values from the buffer, go to 5
3. If not, then pre-load a buffer of some size ahead of that location (like, 5KB), put it into your set of buffers (if necessary, evict an older one using LRU)


Given that most DWARF access is sequential, I imagine that will speed up the metadata reading, and smaller buffers will speed up accessing the metadata for a particular source code location

tanishiking:
Thanks @keynmol !
Although it's a bit hacky, I managed to use the `BufferedInputStream` in the `BinaryFile` https://github.com/scala-native/scala-native/pull/3343/commits/ba8b21b98d7b4f29e588d7d9084576c9c1ca9371 and speed up the `readDWARF` from about 4500ms -> 1300ms on my local machine :+1:
(I'm not sure if there is a better way to implement BufferedStreamInput with seek).

> (evict an older one with LRU if necessary)

Oh yes, I think that's a good idea, it seems we are `seek`ing forward and back to the original position, e.g. in `DIE.parse`, we can avoid reading the file when coming back to the original position by caching the old buffer.

armanbilge:
Sorry, random question: in the example I see just the source filename, however I believe the entire URI is actually available in the NIR. Can / should we use this for stacktraces (like Scala.js does) or will this be too horrible for binary size, performance, etc.?

- https://github.com/scala-native/scala-native/pull/3109

FWIW I like the URIs because they can make it clear which dependency a source file is in and also link to the specific tag/hash.

tanishiking:
@armanbilge 

> Can / should we use this for stacktraces (like Scala.js does) or will this be too horrible for binary size, performance, etc.?

We can display URIs in stacktraces with minimal additional cost. The directory information is already included in the debug information of the binary, which can be retrieved from the DW_AT_comp_dir in the compile unit Debug Information Entry (which is the same entry we retrieve the file name from).

However, I'm not sure whether we should implement this feature. The reason is that the URI would reflect the directory structure **at compile time** instead at runtime (probably as Scala.js does). For example, if the binary is built on my local computer, the URI would appear as follows: `/Users/tanishiking/ghq/github.com/tanishiking/scala-native/sandbox/src/main/scala/Test.scala`.

Perhaps it would be more appropriate to display the URI only if the code file comes from a third-party libraries and the decl_file begins with http. In such cases, the URI might look something like: `https://raw.githubusercontent.com/scala/scala/v2.13.11/src/library/scala/runtime/AbstractPartialFunction.scala`.

armanbilge:
> The reason is that the URI would reflect the directory structure **at compile time** instead at runtime (probably as Scala.js does).

Well, yes 😁 the sources may not even be available at runtime or even linking time! So by definition it it can only reflect the compile time structure, because compile time is the last phase in the toolchain where the sources exist.

> Perhaps it would be more appropriate to display the URI only if the code file comes from a third-party libraries and the decl_file begins with http.

I don't have a strong opinion. I think it's fine to work the same way as Scala.js here: IMHO a longer path/URI is _always_ more helpful than a single filename, even if it's based on some arbitrary location on someone's dev machine.

ekrich:
I don't think it is good idea to leave traces of the client machine compile path structure. So why if the path is relative in NIR would it then become a full path?

tanishiking:
We can make the URI relative I guess :+1:
Maybe

- Show full path if the path if it's scheme is `http(s)`
- Show relative path (from the project root) if it's scheme is `file:`

would be good, but we can do that in a separate PR :)

javalib/src/main/scala/java/lang/Throwables.scala
```diff
@@ -7,25 +7,31 @@ import scalanative.runtime.unwind
...
+      ip: CUnsignedLong
   )(implicit zone: Zone): StackTraceElement = {
     val nameMax = 1024
     val name = alloc[CChar](nameMax.toUSize)
     val offset = stackalloc[scala.Long]()
 
     unwind.get_proc_name(cursor, name, nameMax.toUSize, offset)
 
+    val fileline: Option[(String, Int)] =
+      Backtrace.decodeFileline(ip.toLong)
```
tanishiking, line 26: The problem is, if `Backtrace.decodeFileline` throws an exception (even if we catch it), the exception will call `fillInStackTrace` (when `Throwable` is instantiated) which leads to here and endless loop :\

I'm wondering is there a way to suppress stacktraces (`writableStackTraces = false`) inside this function 🤔

nativelib/src/main/scala/scala/scalanative/runtime/Backtrace.scala
```diff
@@ -0,0 +1,236 @@
...
+
+  private val MACHO_MAGIC = "cffaedfe"
+  private val ELF_MAGIC = "7f454c46"
+
+  private val cache = TrieMap.empty[String, Option[DwarfInfo]]
+
+  def decodeFileline(pc: Long): Option[(String, Int)] = {
+    cache.get(filename) match {
+      case Some(None) =>
+        None // cached, there's no debug section
```
tanishiking, line 50: cache the debug information, so we don't need to try to parse the executable more than once.

nativelib/src/main/scala/scala/scalanative/runtime/Backtrace.scala
```diff
@@ -0,0 +1,236 @@
...
+          // `DW_AT_decl_file` attribute of the `subprogram` DIE.
+          filenameAt = die.getName
+          None
+        } else None
+      }
+      .sortBy(_.lowPC)
+      .toIndexedSeq
+  }
+
+  private def processFile(
```
tanishiking, line 174: This is the endpoint for parsing the executable, and returns `DwarfInfo` instance.

nativelib/src/main/scala/scala/scalanative/runtime/dwarf/BinaryFile.scala
```diff
@@ -0,0 +1,84 @@
...
+import java.io.RandomAccessFile
+import java.nio.channels.Channels
+import scalanative.unsigned._
+import MachO._
+import scala.collection.mutable.ArrayBuffer
+import java.io.File
+import java.io.BufferedInputStream
+import java.io.InputStream
+
+class SeekableBufferedInputStream(
```
tanishiking, line 15: Custom BufferedInputStream that has an capability of seeking the position manually.
I developed this since `RandomAccessFile` is extremely slow due to it's lack of buffer.

nativelib/src/main/scala/scala/scalanative/runtime/dwarf/DWARF.scala
```diff
@@ -0,0 +1,726 @@
...
+        case v
+            if v._1.at == DWARF.Attribute.DW_AT_high_pc &&
+              DWARF.Form.isConstantClass(v._1.form) =>
+          val value = v._2.asInstanceOf[UInt]
+          lowPC + value.toLong
+        case v
+            if v._1.at == DWARF.Attribute.DW_AT_high_pc &&
+              DWARF.Form.isAddressClass(v._1.form) =>
+          v._2.asInstanceOf[Long]
+      }
```
tanishiking, line 180: Except these getter methods, there's no change from https://github.com/indoorvivants/macho-elf-coff-parser/blob/20cf7c473ff7be9b15b44a2de91b9d133efda1f0/DWARF.scala 👍

nativelib/src/main/scala/scala/scalanative/runtime/dwarf/DWARF.scala
```diff
@@ -0,0 +1,726 @@
...
+    val dies = Vector.newBuilder[DIE]
+    while (!stop) {
+      val die = DIE.parse(debug_info, debug_abbrev)
+      dies += die
+    }
+    dies.result()
+  }
+
+  object DIE {
+    private val abbrevCache = mutable.Map.empty[Long, Vector[Abbrev]]
```
tanishiking, line 224: We cache the result of parsing `.debug_abbrev` section, without it, we parse the same `.debug_abbrev` section over and over again for the number of `CompileUnit` DIE.

nativelib/src/main/scala/scala/scalanative/runtime/dwarf/macho.scala
```diff
@@ -0,0 +1,254 @@
+package scala.scalanative.runtime.dwarf
```
tanishiking, line 1: No change from https://github.com/indoorvivants/macho-elf-coff-parser/blob/20cf7c473ff7be9b15b44a2de91b9d133efda1f0/macho.scala

nativelib/src/main/scala/scala/scalanative/runtime/package.scala
```diff
@@ -62,6 +66,7 @@ package object runtime {
       c += 1
     }
 
+    _filename = fromCString(argv(0))
```
tanishiking, line 69: Remember the own executable path so we can parse it for DWARF later.

tanishiking:
There's still things to do + need to fix CI for windows + Linux (to ignore the backtrace test), but marking it as a ready for review 👍

javalib/src/main/scala/java/lang/Throwables.scala
armanbilge: > I'm wondering is there a way to suppress stacktraces (`writableStackTraces = false`) inside this function

Maybe with sort of thread local?

tanishiking: Thanks! I end up doing like this https://github.com/scala-native/scala-native/pull/3343/commits/0e2c67a56ea02568604bb5b7b9f0e0ef857a3f4a :)

tanishiking:
One last thing to fix is `run/build-library-static` on Windows 🤔 but I don't have windows machine at hand right now, will check later

javalib/src/main/scala/java/lang/Throwables.scala
```diff
@@ -39,23 +43,61 @@ private[lang] object StackTrace {
...
+    // infinite loop, ultimately resulting in a stack overflow exception.
+    //
+    // To prevent excessive recursion, it's preferable to minimize multiple consecutive calls to the "currentStackTrace" method.
+    //
+    // The "currentStackTrace" process is straightforward and typically doesn't trigger exceptions.
+    // On the other hand, the "BackTrace.decodeFileline" is intricate, and if an exception thrown, it's likely to originate from that method.
+    //
+    // Consequently, to mitigate the risk of cascading recursive exceptions,
+    // skip executing "BackTrace.decodeFileline" if "currentStackTrace" is already being invoked recursively.
+    val recur = buffer.count(e => e._1.getMethodName == "currentStackTrace") > 1
```
WojciechMazur, line 73: At some point to would be good to eliminate this check, becouse it would happen each time we create stack trace. 
Based on the information where the exception in `decodeFileline` is thrown we should be able to guard it with runtime checks and not allow to throw the exception in the first place.  

An alternative would be also to use ThreadLocal boolean set at the beginning of `currentStacktrace` This way if we renter this method recursively we would know that the current thread throwed the exception.

javalib/src/main/scala/java/lang/Throwables.scala
```diff
@@ -39,23 +43,61 @@ private[lang] object StackTrace {
...
+    buffer.map { e =>
+      val elem = e._1
+      val ip = e._2
+      val maybeFileline =
+        if (recur) None
+        else
+          Try(Backtrace.decodeFileline(ip.toLong)) match {
```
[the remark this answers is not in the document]
tanishiking: Originally, we used to call Backtrace.decodeFileline from within the makeStackTraceElement function. However, I intentionally moved this functionality out to a different location, as can be seen in this commit: https://github.com/scala-native/scala-native/pull/3343/commits/0e2c67a56ea02568604bb5b7b9f0e0ef857a3f4a

This change was made because we were checking for recursive calls to `currentStackTrace` by examining the stack trace and counting the occurrences of `currentStackTrace` entries. 

---

I thought we can call it from within `makeStackTraceElement` in other way (which I found doesn't work)
The idea was that the first call to `currentStackTrace` would record the stack pointer's address at that moment. Subsequent calls to `currentStackTrace` would then compare their stack pointer addresses to the initial one. If the address was lower in the subsequent call, it would indicate a recursive invocation, and thus could be treated as such.

However, I realized that this approach doesn't always work as expected, particularly if the exception object is generated multiple times within the application.

In order to tell we have recursive call to `currentStackTrace`, we have to know the whole stacktrace.

javalib/src/main/scala/java/lang/Throwables.scala
```diff
@@ -11,6 +11,7 @@ import scala.scalanative.runtime.Backtrace
 import scala.util.Try
 import scala.util.Failure
 import scala.util.Success
+import scala.scalanative.runtime.Platform
```
armanbilge, line 11: Instead of `Platform` can we use a linktime condition?
- https://github.com/scala-native/scala-native/pull/3335

tanishiking: Thanks! Changed to use linktime condition https://github.com/scala-native/scala-native/pull/3343/commits/11d3e465cfc872244bbd5e49d22ad843d5982921 :+1:

javalib/src/main/scala/java/lang/Throwables.scala
```diff
@@ -39,23 +43,61 @@ private[lang] object StackTrace {
...
+          case Some(v) =>
+            new StackTraceElement(
+              elem.getClassName,
+              elem.getMethodName,
+              v._1,
+              v._2
+            )
+        }
+      // Update cache with the updated stacktrace element
+      cache.update(ip, updated)
```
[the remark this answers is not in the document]
tanishiking: Right, now we check the stacktrace already has filename before decoding their filename and line number by

```scala
        if (...
            elem.getFileName != null // Skip decoding if we already have filename information
        ) elem
```

tanishiking:
The rest thing to do is to work on https://github.com/scala-native/scala-native/pull/3343#discussion_r1284431759 by using `ThreadLocal` boolean instead. But maybe we can work on it in another PR? Otherwise, it's ready for review :)

scripted-tests/run/backtrace/build.sbt
```diff
@@ -0,0 +1,31 @@
...
+}
+
+lazy val debugBuild = taskKey[Unit]("Compile and run dsymutil if exists")
+debugBuild := {
+  val path = (Compile / nativeLink).value
+  if (System
+        .getProperty("os.name")
+        .toLowerCase(Locale.ROOT)
+        .startsWith("mac")) {
+    Process(s"dsymutil ${path.getAbsolutePath()}") !
```
WojciechMazur, line 29: This step is necessary to make the positions available. Can we try to integrate it with the main workflow? We can try to run it, and then if it's not present we should log a warning for the users that dsumutil and missing and stacktraces would not contain positions, and potentially if executiono fit fails we should give another warning. 
Streaminline this into the main workflow would actually make this feature the most usefull in tests where frequent exceptions would guide us in resolving bugs and slower startup is not an issue.  

`dsymutil` is part of LLVM toolchain so we can try to discover it similary to `llvm-ar` here: https://github.com/WojciechMazur/scala-native/blob/1cdb0c71d9dcd1b527d70d0d36a0ce81855977af/tools/src/main/scala/scala/scalanative/build/LLVM.scala#L232

WojciechMazur (commented):
Looks good. When testing this out I've made some changes in my local branch which could be potentially integrated into this PR: https://github.com/WojciechMazur/scala-native/commit/1cdb0c71d9dcd1b527d70d0d36a0ce81855977af 

1 remaining thing would be the automatic invoke fo `dsymutils`, I've added a bigger comment below

scripted-tests/run/backtrace/build.sbt
tanishiking: > This step is necessary to make the positions available. Can we try to integrate it with the main workflow?  ...
> dsymutil is part of LLVM toolchain so we can try to discover it similary to llvm-ar here:

Ah, that sounds better! :+1: Actually `clang` also integrate `dsymutil` into its workflow.

ekrich: Just as a note, Apple toolchain does supply `dsymutil` but not `llvm-ar` but I understand `ar` will work. I am thinking the archive won't work on Apple toolchain with our current setup.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -150,6 +150,16 @@ private[scalanative] object LLVM {
...
   }
 
+  def dsymutil(config: Config, path: Path) = {
+    val dsymutil = Discover.discover("dsymutil", "LLVM_BIN")
+    val command = Seq(dsymutil.abs, path.abs)
+    val proc = Process(command, config.workDir.toFile())
+    val result = proc ! Logger.toProcessLogger(config.logger)
+    if (result != 0) {
+      throw new BuildException(s"Failed to run dsymutil ${path}")
+    }
```
WojciechMazur, line 160: If the `dsymutil` fails it should rather be a warning saying that we failed to postprocess binary and the stack traces would not work correctly.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -150,6 +150,16 @@ private[scalanative] object LLVM {
     copyOutput(config, buildPath)
   }
 
+  def dsymutil(config: Config, path: Path) = {
```
WojciechMazur, line 153: It's a public method so explicit result type would be recommended

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -150,6 +150,16 @@ private[scalanative] object LLVM {
...
+    val dsymutil = Discover.discover("dsymutil", "LLVM_BIN")
```
WojciechMazur, line 154: Currently `discover` is always failing if it cannot find given binary. Maybe let's introduce an additional variant of it `def tryDiscover(String, String): Option[Path]`. If we won't find it should not be a fatal error, but warning instead. It would be great if in case when we don't find dsymutil we point to the user that we couldn't find this program, possibly show the paths we tried, and point him up that he can use the `LLVM_BIN` env to try fix this issue.

scripted-tests/run/backtrace/build.sbt
```diff
@@ -0,0 +1,31 @@
...
+    )
+  else scalaVersion
+}
+
+nativeConfig ~= { c =>
+  c.withDebugMetadata(true)
+    .withMode(Mode.debug) // otherwise, clang O2 inlines the call stack in Linux
+}
+
+lazy val debugBuild = taskKey[Unit]("Compile and run dsymutil if exists")
```
WojciechMazur, line 22: This should not be required anymore

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -118,7 +118,13 @@ object Build {
...
             .codegen()
             .flatMap(backend.compile)
-            .map(backend.link)
+            .map { p =>
+              val linked = backend.link(p)
+              if (Platform.isMac && config.compilerConfig.debugMetadata) {
+                backend.dsymutil(linked)
+              }
+              linked
+            }
```
WojciechMazur, line 127: Maybe let's create an additional method `backend.postprocess` instead which would call the dsymutil and possibly other non-crucial post processing operations.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -254,6 +264,10 @@ private[scalanative] object LLVM {
     Process(command, config.workDir.toFile()) #< MIRScriptFile
   }
 
+  private def prepareDSymUtilCommand() = {
+    val dsymutil = Discover.discover("dsymutil", "LLVM_BIN")
+  }
+
```
WojciechMazur, line 270: Unused method?

tools/src/main/scala/scala/scalanative/build/Build.scala
```diff
@@ -172,6 +178,12 @@ object Build {
     ) {
       LLVM.link(config, linkerResult, compiled)
     }
+
+    def dsymutil(linked: Path): Unit = time(
+      s"Running dsymutil on ${linked.getFileName}"
```
WojciechMazur, line 183: We should not require users to now what dsymutil is and what it does. Maybe let's just introduce a single logging of time spent in whole postprocessing stage (based on suggestion to move all postprocessing to seperate step). At that point can remove this method and use `LLVM.dsymutil` directly in postprocessing stage.

WojciechMazur (approved):

PR #605, opened 2017-04-03
Fix incremental compilation in javalib

Duhemm:
Fixes #599

PR #3344, opened 2023-06-22
Improve IDE development experience

WojciechMazur:
* Limits the amount of created underlying bloop projects to only Scala 3 projects. This change should allow for faster imports, and more reliable IDE diagnostics (no-more classpath duplicate entries or compilation errors not allowing to finish indexing Metals)
* Disables usage of Scala Native compiler plugins when compiling for IDE to speedup compilaiton
* Sets scalafmt default dialect to scala3 (was scala213). Scala 3 is the main platform we're going to develop and maintaine. 
* Removes redundant javabootclasspath settings leading to indexing failures
* Don't create bloop projects for scalalib and auxlib - depend on Scala stdlib artifacts 

Most of these changes are ported from Scala.js, but modified to adjust our goals and allow to target Scala 3 by default and potentially Scala 3 unstable versions on demand.

Known issues: 
* Javalib might raise numerous `problems` (compilation errors), which are present due to incremental compilation. In normal compilation (in sbt) we force full compilation on any change to mitigate problem of double classpath definitions (javalib + JDK). This is however not possible to overcome in IDE/bloop

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -485,6 +485,14 @@ object Settings {
         }
         val newSources = experimentalSources.values.toList.diff(updatedSources)
         updatedSources ++ newSources
+      },
+      // Adjustment for bloopInstall which tries to add whole source directory leading to double definitions
+      scope / sourceDirectories --= {
+        val sourcesDir = (scope / sourceDirectory).value
+        lazy val experimentalSources = allScalaFromDir(sourcesDir / baseDir)
+        if (isGeneratingForIDE && experimentalSources.nonEmpty)
+          Seq((scope / scalaSource).value)
+        else Nil
       }
     )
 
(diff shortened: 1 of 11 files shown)


ekrich:
Sounds very cool :-)

project/Settings.scala
```diff
@@ -577,14 +586,6 @@ object Settings {
     // We put our classes on scalac's `javabootclasspath` so that it uses them
     // when compiling rather than the definitions from the JDK.
     recompileAllOrNothingSettings,
-    Compile / scalacOptions := {
```
tanishiking, line 580: Just out of curiosity, why this was needed, and now we can remove it?

tanishiking (commented):
I left 1 nitpick (typo) + question, but otherwise looks good to me 🎉

project/Settings.scala
WojciechMazur: The history tracks back to #605, however I'm not sure why it's no longer an issue. Maybe it was fixed in Scala 2.12

PR #3346, opened 2023-06-23
Fix encoding of nir.Positions

WojciechMazur:
During rebase of #2869 I've discovered a flow in encoding nir.Positions. Due to combination of interning and incremental encoding of positions using deltas, we could have received a NIR position defined with negative line or column value. 

This change removes the encoding of nir.Positions using deltas, and instead always stores full positions. The encoded NIR size it might increase, but it's not significent: 
ScalaLib 2.13 classess size has increase by ~200 bytes, from 23,748 kB to 23,910 kB 
It also fixes setting nir.Positions for some of Scalac ASTs


tanishiking (commented):
The diffs looks good to me, but I would check in the downstream changes (debug info + backtrace) just in case 👍

WojciechMazur:
I've failed to reproduce the behaviour with positions returning nulls. Have you cleaned and reload project after rebase? Maybe there was something left from the previous version of the nir encoding. I'm going to merge this PR now, if we'll obserbve this issue again we'll fix it in the follow up. As you mention we always use `Position.NoPosition` so we should never get null

tanishiking:
With https://github.com/scala-native/scala-native/pull/2869 there's no problem, I'm not sure what caused the issue in my local environment, but now it works totally fine 👍 Thank you for the follow up work in #2869 !

Issue #3341, opened 2023-06-17
`WeakReference` does not work if subclassed


armanbilge:
It seems that if `WeakReference` is sub-classed then it is not correctly handled by the GC. Probably due to the logic here?

https://github.com/scala-native/scala-native/blob/e5ee79a0fc39bc86bbf4a5de0c211be0ed1c2019/nativelib/src/main/resources/scala-native/gc/immix_commix/headers/ObjectHeader.h#L88-L90

Consider the following program, which allocates large arrays and stores them in weak references. It terminates when it detects the that a weak reference has been GCed.

```scala
import java.lang.ref.WeakReference

class MyWeakRef[A](a: A) extends WeakReference(a)

@main def main =
  val refs = new Array[MyWeakRef[Array[Long]]](Int.MaxValue / 8)

  var i = 0
  while (i < refs.length) {
    refs(i) = new MyWeakRef(new Array[Long](Int.MaxValue / 64))
    var j = 0
    while (j < i) {
      if (refs(j).get == null) {
        println((i, j))
        System.exit(0)
      }
      j += 1
    }
    i += 1
  }
```

As written, that program never prints or terminates for me. This suggests that none of the arrays are being GCed.

However, if I replace `MyWeakRef` with `WeakReference` then it terminates quickly with `(3,0)`. If I run the program on the JVM, it also terminates quickly.

### Motivation

In Cats Effect we sub-class `WeakReference` for a custom data structure. We use this data structure to power a "fiber dump" feature (similar to thread dump).

https://github.com/typelevel/cats-effect/blob/e9aeb8ccb53016a37588256f4c8086f8ef1612bc/core/shared/src/main/scala/cats/effect/unsafe/WeakBag.scala#L108-L110

mox692:
I will work on this.
To me, it looks good issue to delve into scala-native internals and gc stuff.

PR #3347, opened 2023-06-24
Fix Subclassed `WeakReference` to be GCed

mox692:
Fix https://github.com/scala-native/scala-native/issues/3341

Instead of using `WeakReferenceId` directory to determine WeakReference, we use the `ranges` of WeakReference to make the determination, as `Class.is` does.


tanishiking (commented):
I took a look at the changes (mostly for learning purposes), and it looks good to me. I can indeed confirm that this change  address the specific examples mentioned in the original issue. 🎉 
However, since I'm not very familiar with this particular area, I would appreciate it if someone else could review it as well 😃

Issue #3214, opened 2023-03-10
Get rid of unneccessary `@stub` annotations


WojciechMazur:
`@stub` is harmful to the native ecosystem, maybe at some point in the past it was useful, but right now it makes no sense to use. 
Option `withLinkStubs(true)` is typically unnecessarily enabled in most of the projects, even when it's not required. If for some reason, the stub annotated methods are used, their usage can lead to unexpected runtime failures, which could have been detected at compile (link) time. 
With current support for link-time properties, it is also easy to eliminate code flows that for some reason could not be implemented of linked otherwise. 

The only justified use case for `@stub` is the definition of abstract members which should never have been used. In such case access to stub would fail to link preventing us from a runtime error. 
https://github.com/scala-native/scala-native/blob/e6fdf3ef3ce0763f0d60082438f6fe1af86e0872/javalib/src/main/scala/java/lang/process/PipeIO.scala#L122-L124

The change should include the removal of stub methods/classes annotated with `@stub`  annotations in the code (e.g. `java.lang.reflect.Field.get`).

ekrich:
Just to confirm, you are not saying `withLinkStubs(true) is going away but that the `@stub` will be removed?

WojciechMazur:
Currently, it's the most important to remove abuses for that annotation. Later we might think if it should remain for the usage like in the snippet above. If we remove `@stub` we can deprecate `withLinkStubs` and make it a no-op. It's only used for disabled stubs behavior.

ekrich:
I see, the default would be true like it is in Scala.js?

I think as long as we can still use `???`, I think that is fine. I need to have `true` set to link in `ekrich/sconfig` so I am thinking the default is `false`. https://github.com/scala-native/scala-native/blob/main/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala#L208

Do you plan to backport to the 0.4.x series? Eventhough is might be breaking, it might be a gateway.

WojciechMazur:
As far as I know, Scala.js has no stubs. No, it should default to `false` to protect from the usage of nonported methods. Otherwise it would always just fail at runtime. You should enable linking stubs only when you explicitly need it

ekrich:
If you take a look at the build here in this project you can see that I have no options for Scala.js but add the `withLinkStubs(true)` for the Native options. https://github.com/ekrich/sconfig

[...]

PR #3348, opened 2023-06-27
Remove  unnecessary `@stub` annotations

tanishiking:
https://github.com/scala-native/scala-native/issues/3214

- Deprecated `nativeLinkStubs`
  - The link behavior is the same as `nativeLinkStubs = false` (even if the build tools set `nativeLinkStubs = true`), meaning access to `@stub` methods will fail at link time.
- Implement `java.lang.File#toURL` and `URI#toURL`
- Removed `@stub` methods
  - `java.lang.ClassLoader#...` everything
  - `java.lang.Runtime#shutdownHook`
  - `java.lang.reflect.Constructor#getParameterTypes`, `java.lang.reflect.Constructor#newInstance`
  - `java.lang.reflect.Field#{get, getName, getType}`
  - `java.util.Calendar#...`
  - `java.lang.Class#...`
  - **We still have `@stub` annotations for "abstract members which should never have been used".**
- Add some documents on release note
  - Maybe it's worth writing more detailed instruction for workaround?

~Should we remove `scalanative.annotations.stub` too? I guess it's still useful for users to make scala-native code compile (while access to them fail at link-time).~ Let's keep the `stub` annotation for a while, it is still useful for https://github.com/scala-native/scala-native/pull/3348#issuecomment-1608762269 cases, we could do in another ways, but let's not do that in this PR.

Also, let's not backport to 0.4.x


javalib/src/main/scala/java/lang/ClassLoader.scala
```diff
@@ -1,15 +1,9 @@
...
-  @stub
-  def getParent(): ClassLoader = ???
-  @stub
-  def getResourceAsStream(name: String): java.io.InputStream = ???
-  @stub
-  def getResources(name: String): java.util.Enumeration[_] = ???
+  // def loadClass(name: String): Class[_] = ???
+  // def getParent(): ClassLoader = ???
+  // def getResourceAsStream(name: String): java.io.InputStream = ???
+  // def getResources(name: String): java.util.Enumeration[_] = ???
```
tanishiking, line 8: Should we remove this class itself?
The `ClassLoader` class with no members might make it hard to stub by users?

project/Build.scala
```diff
@@ -172,6 +172,9 @@ object Build {
             case (3, _)  => scala213StdLibDeprecations
           }
       },
+      Test / scalacOptions --= Seq(
+        "-Xfatal-warnings"
+      ),
       // Running tests in parallel results in `FileSystemAlreadyExistsException`
```
tanishiking, line 178: To make `toolsX/test` build pass with the `@deprecated` `withLinkStubs`

armanbilge:
> Should we remove `scalanative.annotations.stub` too? I guess it's still useful for users to make scala-native code compile (while access to them fail at link-time).

I'm using this in epollcat for precisely this reason.

https://github.com/armanbilge/epollcat/blob/c5c2347bc9c59a90ce68aa5588b0ce91ce6da52f/core/src/main/scala/epollcat/internal/ch/EpollAsyncSocketChannel.scala#L96-L108

ekrich:
@armanbilge I don't think you need the `@stub` annotation. I just use `withLinkStubs(true)`. In Scala.js it seems that linking stubs is default with no options or annotations.

Hopefully in 0.5.0, all this can go away. It would be nice to back port it to get people ready as well.

javalib/src/main/scala/java/lang/ClassLoader.scala
ekrich: We do have a `Class.getResourceAsStream` so normally you can do that with a `ClassLoader` too and also it is not clear why the other `getResource` methods are not implemented. Not sure if that would hurt porting efforts. Normally I would say we should remove it but not 100% sure because `FileSystems` has a couple of methods needing a `ClassLoader`.

ekrich: Well, I answered my own question - they are not implemented because the return `URL` or `Enumeration[URL]`.

armanbilge:
@ekrich I'm not sure I understand. I don't want to implement those methods, but I have to because they are declared by an abstract class that I am extending. So I mark them as `@stub`. When the user correctly sets `.withLinkStubs(false)`, then they will be warned at linking-time if they attempt to use those methods, because they should not use those methods. This is the desired behavior.

tanishiking:
I agree with @armanbilge 

Indeed `EpollAsyncSocketChannel#read` doesn't need a `@stub` annotation to compile and link as @ekrich mentioned.
https://github.com/armanbilge/epollcat/blob/c5c2347bc9c59a90ce68aa5588b0ce91ce6da52f/core/src/main/scala/epollcat/internal/ch/EpollAsyncSocketChannel.scala#L96-L108

The issue arises when the `@stub` annotation is missing, and scala-native code attempts to access the `EpollAsyncSocketChannel#read` method. In such cases, a runtime failure occurs.

However, if we add the `@stub` annotation, we can identify that the scala-native code has access to non-implemented code (annotated as `@stub`) during the linking phase. This information can still be valuable even after deprecating `nativeLinkStubs`.

>  When the user correctly sets .withLinkStubs(false), then they will error at linking-time if they attempt to use those methods, because they should not use those methods. This is the desired behavior.

ekrich:
If you don't need the stubs for linking then you shouldn't include them at all. If you need them for linking they should link just as they do on the JVM or JS. On JVM, JS, or Native, if you call the method you get `scala.NotImplementedError: an implementation is missing` as expected. I don't think Scala Native needs the annotation or any other special treatment so that it works just like the other platforms. Besides, the compiler / linker doesn't need any additional work.

armanbilge:
@ekrich 

> if you call the method you get `scala.NotImplementedError: an implementation is missing` as expected

That's the point, from the user perspective this is highly _unexpected_ :) which is precisely why we want to fail their build _before_ they reach runtime.

tanishiking:
Hm, maybe `@deprecated` helps?

armanbilge:
> Hm, maybe `@deprecated` helps?

I'm not sure if that would work. In my case, the user never directly calls the stub methods, they are on a private implementation class. Instead, the user interfaces with a public API that is defined in the JDK and that I cannot add deprecations to.

---

> It would be better to fail at link time, but I understand @ekrich 's point that don't wanna scala-native only annotations.

I'm not sure I understand the issue with native-only annotations. This is a native-specific issue. For Scala JVM, linking happens at runtime in the JVM—there is no equivalent phase in the compiler.

tanishiking:
TBH, I don't have a strong opinion on removing/keeping `@stub` annotation.
I thought maybe aligning annotations across platforms might be useful, but now I have second thought that `@stub` is still only in scala-native code, so it doesn't matter. I'm a bit confused 😓 

@ekrich  Could you elaborate on what is the point of removing `@stub` annotations?

ekrich:
> TBH, I don't have a strong opinion on removing/keeping `@stub` annotation. (I lean to keeping the annotation) I thought maybe aligning annotations across platforms might be useful, but now I have second thought that `@stub` is still only in scala-native code, so it doesn't matter. I'm a bit confused 😓
> 
> @ekrich Could you elaborate on what is the point of removing `@stub` annotations?

My point is that JVM and JS don't have `@stub`. To make all the platforms easier to deal with and simplify Scala Native, I think it would be better to drop it and the cognitive load it adds. Developers should probably expect as much standardization as possible between the different platforms.

som-snytt:
I was going to joke, keep the annotation but call it `???`. However TIL you have to write
```
@ ??? class C
```
Maybe that is a bug, but you could freak out your coworkers by writing
```
@ deprecated class C
```
another reason to avoid the Ammonite prompt.

The missing implementation message calls the examples "stubs", but to me, stub implies not throwing. Oh, I guess the Scala 2 message no longer calls them stubs, because it is obvious. Scala 3 has the old message:
```
For convenience, these are usable as stub implementations.
```
They are not usable in the sense of callable. And are they implementations if the body is `???`?

So many questions, and so many question marks.

On chat, I floated an old idea about using `@elidable` for stubs. I never fully embraced this workflow, before `???` was introduced, but stubs or partial implementations would be marked elidable, and features which must be implemented first would have a higher elide level. That is, you hit a milestone M1, then no M1 features should be elided in the build.

For a given platform, where a feature might be a stub, I may want to simply ignore usages (stub is a no-op or returns the default value, same as an elided method), or be warned, or also turn them into `???` (unelided, not-implemented stubs).

tanishiking:
Conversation from discord

```
armanbilge — Today at 4:56 AM
@ekrich ok great, so I think we are agreeing
if all internal uses of stub in SN can go away, I'd support its entire removal in 0.5.x 

ekrich — Today at 4:58 AM
I think that is the idea - if we can't implement the method we should not have it implemented with ??? (no implementation)

armanbilge — Today at 4:58 AM
right, but in some cases a method can't be implemented at all but it's part of a class that would be useful to implement

ekrich — Today at 4:59 AM
User libraries are another story - in sconfig's case it could probably be mitigated but it would be a ton of work.

ekrich — Today at 5:01 AM
I think that the javalib falls into this area but you really can't have a partial implementation - toURL is a special case and Scala.js does the same thing. 
ekrich — Today at 5:07 AM
Maybe you could add this comment to the ticket - I think that was what I was getting at - I may have been spouting about 0.4.x but that was not my view or goal.

armanbilge — Today at 5:09 AM
my comment there requested it not be removed in 0.4.x specifically
I can add a clarification later 

ekrich — Today at 5:30 AM
I think leaving it in 0.4.x is appropriate.
```
https://discord.com/channels/632150470000902164/635668881951686686/1124065775469203456

WojciechMazur (commented):
The is also `java.security.MessageDigest` implemented with `DummyMessageDigest` It belive these would should be removed as well, becouse it's inheritlly unsafe. In my free time I'm slowly trying to port java.util.security package from Apache Harmony. If sucessfull we might publish it as seperate artifact, and remove all current java.security stuff from core repo. 

Based on some of the opinions in comments I currently also don't have a strong opinions about removal of `@stubs` annotation. As @armanbilge mentioned there are some valid use cases to make it stay, however, in such case we should warn the user about that, instead of silently ignoring the fact of using stubs. In case of stubs implemented with `???` the runtime behaviour is pretty straightforward as it would throw error and print stack trace with the context of invalid usage. However, when using `MessageDigest` which is providing a working, but dummy implementation the user of linked program, might be completly unaware of possibly dangerous runtime behaviour. Maybe printing a warning to stdout at program init would not be a bad idea?

I don't think it should be backported to 0.4 branch, it rather should be 0.5.x and forward change. In my opinion annotation my stay for some more time, however we should remove most of stubs from our core code. We might reevaulte this decision when moving to version 1.0 someday in the future.

javalib/src/main/scala/java/net/URI.scala
```diff
@@ -924,6 +924,11 @@ final class URI private () extends Comparable[URI] with Serializable {
 
   def isOpaque(): Boolean = opaque
 
+  def toURL(): URL = {
```
ekrich, line 927: Is it really needed to implement URL? This would make it impossible to have a third party URL implementation.  I have already confirmed that removing URL works downstream and I would prefer that we match Scala.js.

ekrich (changes requested):

javalib/src/main/scala/java/net/URI.scala
ekrich: I really view this like `Locale` or `java.time` so that third party implementations can occur if needed. We also have the https://github.com/scala-native/scala-native-java-stubs or we could have something else outside of core if we need it.

javalib/src/main/scala/java/io/File.scala
```diff
@@ -603,8 +603,7 @@ class File(_path: String) extends Serializable with Comparable[File] {
 
   def deleteOnExit(): Unit = DeleteOnExit.addFile(this.getAbsolutePath())
 
-  @stub
-  def toURL(): java.net.URL = ???
+  // def toURL(): java.net.URL = ???
```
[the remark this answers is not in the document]
tanishiking: > That also means that we would need to implement URI::toURL as well. In my opinion it's fine, as long as we would only the public constructors of URL known from javadocs.

Do you mean something like this: https://github.com/scala-native/scala-native/pull/3348/commits/799464649fc2ae7f39e992597a328e9b2949f83f ?
If I understand correctly, as @ekrich mentioned in https://github.com/scala-native/scala-native/pull/3348#discussion_r1280619728 having `public constructors of URL` in scala-native makes impossible to have a third party URL implementation. And`URI#toURL` will be an URL we defined, not a third-party one.
Am I having a misunderstanding?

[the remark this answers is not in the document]
tanishiking: Ah, you mean, like this? https://github.com/scala-native/scala-native/pull/3348/commits/0f6002a0c6b673695f3f5f7d0ed804c09dfe7527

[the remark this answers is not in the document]
tanishiking: I though it doesn't compile, but actually

- it does compile (to NIR) because the symbol to `java.net.URL` resolves, it is defined in javalib (implemented in Java).
- At linking phase, Reachability check will fail if we call `URI#toURL` because `Not found Top(java.net.URL)`

Am I correct?

javalib/src/main/scala/java/net/URI.scala
tanishiking: Thanks!
I think I was misunderstood @WojciechMazur comments in https://github.com/scala-native/scala-native/pull/3348#discussion_r1284162278 😅

project/Build.scala
```diff
@@ -200,6 +200,9 @@ object Build {
             case (3, _)  => scala213StdLibDeprecations
           }
       },
+      Test / scalacOptions --= Seq(
+        "-Xfatal-warnings"
```
WojciechMazur, line 204: Is it still needed?

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -154,7 +154,9 @@ sealed trait NativeConfig {
   def withASAN(value: Boolean): NativeConfig
 
   /** Create a new config with given behavior for stubs. */
-  def withLinkStubs(value: Boolean): NativeConfig
+  @deprecated(
+    "This option has been deprecated and no longer have effects on the build"
+  ) def withLinkStubs(value: Boolean): NativeConfig
```
WojciechMazur, line 159: I think we've discussed somewhere that it should remain in it's current form, without deprecation, but we should remove all unnecesery usages of `@stub`

WojciechMazur: That also means we should restore previous behaviour from linker.Reach and tests and scaladoc comments for annotation

WojciechMazur (commented):
Looks good! Just a few notes

test-interface/src/main/scala/java/lang/ClassLoader.scala
```diff
@@ -1,19 +0,0 @@
-package java.lang
-
-class ClassLoader protected (parent: ClassLoader) {
```
WojciechMazur: It seems like like actually need the ClassLoader type, even though it's never actually used in our code (only passed in arguments). 
It think we should put it back to javalib, but let's remove the stubs from it. We can leave only the constructors like Scala.js does: 
https://github.com/scala-js/scala-js/blob/2da5417f692955dfc9311e276294d6fa4964f05d/javalib/src/main/scala/java/lang/ClassLoader.scala#L13-L19

tanishiking: Yeah, it seems like so, I moved it to `test-interface` and see it works but it doesn't. We at least need a name to pass the reachability check.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
tanishiking: I somehow thought the issue is about making `withLinkStubs` no-op, but it seems to be extended goal. Let's just focus on removing the abuse of `@stub` annotation in this PR as mentioned in https://github.com/scala-native/scala-native/issues/3214#issuecomment-1464336249

WojciechMazur (commented):
One last spurious change, beside that it's ready to merge

WojciechMazur (approved):

PR #3233, opened 2023-03-20
Allow cross compiling of native tools

ekrich:
The idea here is to cross compile the compiler backend for Scala Native. The front end is a plugin to the Scala Compiler (or could be built in like Scala.js for Scala 3) that would require the Scala Compiler to be cross compiled which is much harder lift because of the use of reflection and Java libraries that would require a big effort.

Why would we want to cross compile the `tools`, `nir`, and `util` project which constitutes the backend?

1. `scala-cli` could call a native application using the native artifacts directly rather than using the `scala-native-cli` project.
2. Scala Native could create an application to call as the backend which might be an option for the `sbt` and Mill plugins.
3. Scala Native could provide tighter integration with `clang` to call it directly rather that create a subprocess to call `clang`.
4. Scala Native could use other lower level LLVM tools for other unforeseen features.

Numbers 3 and 4 above would mean that a JVM version and a Native version could diverge.

So far the Sandbox works but not tests. The tools project will need the ability to run Native and JVM tests. We will also need to publish artifacts and perhaps work on better documentation.

Co-authored-by: @WojciechMazur

PR #3383, opened 2023-07-11
Remove dependency on Scala parallel collections in toolchain

WojciechMazur:
This change removes dependency on Scala parallel collections. These were removed from stdlib and were needed to be included as external dependency since 2.13. Fact of usage external dependencies is a blocker for #3233. The only remaining dependecy after this change would be ScalaTest.

Scala parallel collections have been replaced with direct usage of `Future`, especially `Future.traverse`.  However, this change requires us, to change the API of build. Now, `scalanative.Build.build` does require implicit `ExecutionContext` and does return `Future`. This now makes the API more similar to the API of Scala.js.  Because of viral nature of `scala.Future`, the `ExecutionContext` is propagated in all usages of code, that might execute in parallel.


WojciechMazur:
Change seems to not have a negative impact on toolachin performance, it might even in fact slightly improve it: 

```
# Before refactor
[info] Benchmark                                        Mode  Cnt     Score    Error  Units
[info] LinkerBench.link                                 avgt   10   381.692 ± 21.700  ms/op
[info] OptimizeDebug.optimize                           avgt   10   340.831 ± 18.205  ms/op
[info] OptimizeReleaseFast.optimize                     avgt   10  1118.887 ±  8.960  ms/op
[info] testinterface.CodeGen.codeGen                    avgt   10   201.963 ±  9.069  ms/op
[info] testinterface.CodeGenWithDebugMetadata.codeGen   avgt   10   227.957 ± 19.301  ms/op
[info] testinterface.CodeGenWithMultithreading.codeGen  avgt   10   278.591 ±  1.369  ms/op

# After refactor
[info] Benchmark                                        Mode  Cnt     Score    Error  Units
[info] LinkerBench.link                                 avgt   10   362.773 ±  9.951  ms/op
[info] OptimizeDebug.optimize                           avgt   10   308.945 ± 13.501  ms/op
[info] OptimizeReleaseFast.optimize                     avgt   10  1095.197 ± 14.564  ms/op
[info] testinterface.CodeGen.codeGen                    avgt   10   159.182 ± 16.184  ms/op
[info] testinterface.CodeGenWithDebugMetadata.codeGen   avgt   10   161.618 ±  5.701  ms/op
[info] testinterface.CodeGenWithMultithreading.codeGen  avgt   10   230.890 ±  4.489  ms/op

```

PR #3384, opened 2023-07-11
Drop dependency on ScalaTest. Rewrite tests to JUnit

### LINKED ISSUE #3371: Convert tools tests to use JUnit to support cross compilation
In order to allow `clang` compilation and linking to be a Scala Native application or to allow use of native LLVM code we are making our `tools` project cross compile. One advantage of cross compilation is that we could use the project as a native app but still allow the current JVM approach to work and thus avoid any sort of bootstrap issues.

See https://github.com/scala-native/scala-native/pull/3233 for the cross compilation.

WojciechMazur:
Resolves #3371


ekrich (approved):
Looked through it all sort of quickly but it LGTM.

ekrich:
Do you think our AssertThrows utility could be added to the JUnit artifacts somehow so the could be used downstream and also not duplicated between `unit-tests` and `unit-tests-ext`?

WojciechMazur:
> Do you think our AssertThrows utility could be added to the JUnit artifacts somehow so the could be used downstream and also not duplicated between `unit-tests` and `unit-tests-ext`?

Not really, we do provide runtime for JUnit 4 and we should stay compliant with their API in junit artifacts https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html 
We could potentially provide JUnit5 API, but that's not in the scope of the current work, and probably could be provided as 3rd party library https://junit.org/junit5/docs/5.9.1/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html. 
What's important most of `AssertThrows` could be easily replaced by usage of `() => ` with the Junit `assertThrows` method. It's nice to use Scala toys to make the code cleaner, but so far it's not a big hassle to use standard junit api. Ths also makes seperating failing tests into the sandbox for faster feedback loop way easier. 

> Also, this got put in there but probably should be somewhere else. https://github.com/scala-native/scala-native/blob/main/unit-tests/shared/src/test/scala/utils/AssertThrows.scala#L47-L49

This one was ported from Scala.js along with the Scala.js tests. The Scala.js null pointers compliance has no sense in Scala Native, though our implementation is more straightforward. https://github.com/scala-js/scala-js/blob/037fa855a636a521721f6b126f915744136cd47e/test-suite/shared/src/test/scala/org/scalajs/testsuite/utils/AssertThrows.scala#L25-L28

Issue #3349, opened 2023-06-27
Enhance debug information for improved debugging experience


tanishiking:
Follow up of https://github.com/scala-native/scala-native/issues/29

Currently, there is an [open PR that introduces basic debug information generation](https://github.com/scala-native/scala-native/pull/2869), but we can further enhance this feature to provide more comprehensive debug information.

There are a few things to do to improve the debugging information, and any of them can be done separately (so feel free to split this into smaller issues if anyone thinks it's useful).

## [ ] Declaring Local Variables
Following the guidelines outlined in the [LLVM documentation on source-level debugging](https://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-declare), Scala-Native can include local variable declarations `!DILocalVariable`.

This allows

- Debugger to display the variable name, type, and location while debugging, enabling us to inspect the state of variables during program execution.
- Debuggers to associate variables with their corresponding memory locations or register values.

## [ ] Demangled Name in DISubprogram
Currently, the debug information generated by Scala-Native does not include demangled names in the `DISubprogram`.

Adding this would make it easier to
- identify functions and methods during debugging sessions
- **set a break point to function entry points** (currently, users have to know the linkageName to set a break point).

## [ ] Nested Scoping to Match Scala Scopes
Align the nested scoping in the debug information with the scoping used in Scala.

This would ensure that the debug information accurately reflects the structure of the code, allowing us to navigate and understand the codebase more effectively during debugging.

---

PR #3387, opened 2023-07-13
Unmangle procedure names in DISubprogram

tanishiking:
https://github.com/scala-native/scala-native/issues/3349

In the following Scala program,
to set a break point to a `foo` method,

```scala
object Test {
    foo()
  }

  def foo() = {
    println("Hello World")
  }
}
```

Previously we had to use a mangled name like `_SM5Test$D3foouEO`. Which is quite difficult for users to guess the mangled name.

```
❯ lldb sandbox/.2.13/target/scala-2.13/sandbox
(lldb) target create "sandbox/.2.13/target/scala-2.13/sandbox"
Current executable set to '/Users/tanishiking/ghq/github.com/tanishiking/scala-native/sandbox/.2.13/target/scala-2.13/sandbox' (arm64).
(lldb) b _SM5Test$D3foouEO
Breakpoint 1: where = sandbox`_SM5Test$D3foouEO + 4 at Test.scala:6:11, address = 0x00000001000788c4
```

This commit unmangle the method name, and now users can set a breakpoint using the function name.

```
scala-native on  unmangle-method [$!+] via ☕ v17.0.6 via 🆂 v3.2.2
❯ lldb sandbox/.2.13/target/scala-2.13/sandbox
(lldb) target create "sandbox/.2.13/target/scala-2.13/sandbox"
Current executable set to '/Users/tanishiking/ghq/github.com/tanishiking/scala-native/sandbox/.2.13/target/scala-2.13/sandbox' (arm64).
(lldb) b foo
Breakpoint 1: where = sandbox`_SM5Test$D3foouEO + 4 at Test.scala:6:11, address = 0x00000001000788c4
(lldb) run
Process 61145 launched: '/Users/tanishiking/ghq/github.com/tanishiking/scala-native/sandbox/.2.13/target/scala-2.13/sandbox' (arm64)
sandbox was compiled with optimization - stepping may behave oddly; variables may not be available.
Process 61145 stopped
* thread #1, stop reason = breakpoint 1.1
    frame #0: 0x00000001000788c4 sandbox`_SM5Test$D3foouEO at Test.scala:6:11 [opt]
   3        foo()
   4      }
   5
-> 6      def foo() = {
   7        println("Hello World")
   8      }
   9    }

```

---

I'm wondering how can we write a test for debug metadata 🤔

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/llvm/MetadataCodeGen.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/llvm/MetadataCodeGen.scala
@@ -9,6 +9,9 @@ import scala.scalanative.util.unsupported
 
 import scala.language.implicitConversions
 import scala.scalanative.codegen.llvm.MetadataCodeGen.Writer.Specialized
+import scala.scalanative.nir.Unmangle
+import scala.scalanative.nir.Global.Member
+import scala.scalanative.nir.Sig.Method
 
 // scalafmt: { maxColumn = 100}
 trait MetadataCodeGen { self: AbstractCodeGen =>
@@ -68,9 +71,24 @@ trait MetadataCodeGen { self: AbstractCodeGen =>
       producer = Constants.PRODUCER,
       isOptimized = attrs.opt == nir.Attr.DidOpt
     )
+    // unmangle method name
+    // see: https://scala-native.org/en/stable/contrib/mangling.html
+    val linkageName = mangled(name)
+    val defnName = if (linkageName.startsWith("_S")) linkageName.drop(2) else linkageName
+    val unmangledName = if (defnName.startsWith("M")) { // subprogram should be a member
+      Unmangle.unmangleGlobal(defnName) match {
+        case Member(owner, sig) =>
+          Unmangle.unmangleSig(sig.mangle) match {
+            case Method(id, _, _) => Some(id)
+            case _                => None
+          }
+        case _ => None
+      }
+    } else None
+
     DISubprogram(
-      name = name.top.id,
-      linkageName = mangled(name),
+      name = unmangledName.getOrElse(linkageName),
+      linkageName = linkageName,
       scope = unit,
       file = file,
       unit = unit,


tools/src/main/scala/scala/scalanative/codegen/llvm/MetadataCodeGen.scala
```diff
@@ -68,9 +71,24 @@ trait MetadataCodeGen { self: AbstractCodeGen =>
       producer = Constants.PRODUCER,
       isOptimized = attrs.opt == nir.Attr.DidOpt
     )
+    // unmangle method name
+    // see: https://scala-native.org/en/stable/contrib/mangling.html
+    val linkageName = mangled(name)
+    val defnName = if (linkageName.startsWith("_S")) linkageName.drop(2) else linkageName
```
WojciechMazur, line 77: I'm not sure, but I think that on Windows there can be just a 'S` prefix, however we should probably check that. We had same special handling for genreating stacktraces on Windows, but I'm not sure if it applies to debug metadata

https://github.com/scala-native/scala-native/blob/4e9ca7d0b478765ee81e3a698e33d037ac0b407a/nativelib/src/main/scala/scala/scalanative/runtime/SymbolFormatter.scala#L31-L49

tools/src/main/scala/scala/scalanative/codegen/llvm/MetadataCodeGen.scala
```diff
@@ -68,9 +71,24 @@ trait MetadataCodeGen { self: AbstractCodeGen =>
...
+    val unmangledName = if (defnName.startsWith("M")) { // subprogram should be a member
+      Unmangle.unmangleGlobal(defnName) match {
+        case Member(owner, sig) =>
+          Unmangle.unmangleSig(sig.mangle) match {
+            case Method(id, _, _) => Some(id)
```
WojciechMazur, line 82: Wouldn't use just a method name would be ambiguous?  Let's say we have classes `A` and `B` both defining method `foo`. Then we would trigger break point in each of them. Probably we should use a fully qualified name. `A.foo` or `B.foo`
```suggestion
            case Method(id, _, _) => Some(s"{$owner.id}.${id}")
```

tools/src/main/scala/scala/scalanative/codegen/llvm/MetadataCodeGen.scala
tanishiking: Oh, thanks, I'll check that.

tanishiking: I haven't confirmed on my Windows machine (since I'm struggling to setup dev environment on Win), it looks like we definitely have `_S` prefix for mangled name at least on NIR -> LLVM IR https://github.com/scala-native/scala-native/blob/439188535a050ae87f49aa2913cd9f69757a4196/tools/src/main/scala/scala/scalanative/codegen/llvm/AbstractCodeGen.scala#L594-L595

Maybe Windows linker or Windbg's `get_proc_name` drops the `_` prefix? https://github.com/scala-native/scala-native/pull/2287/commits/9f7f24c607bc10ff6e953886ccd4638fca3e33a1

WojciechMazur: Maybe, that's possible. Let's merge this change, and potentially fix it later if we'd get a bug report

WojciechMazur (approved):

PR #3388, opened 2023-07-13
Sync documentation with 0.4.x branch and publish nightly snapshots in CI

WojciechMazur:
* Syncs documentation with 0.4.x branch
* Sets up CI job for publishing nightly snapshots every second day of the week


.github/workflows/publish.yml
```diff
@@ -2,24 +2,34 @@ name: Publish
...
     uses: ./.github/workflows/run-jdk-compliance-tests.yml
   test-multiarch:
+    if: github.event_name != 'schedule' && github.repository == 'scala-native/scala-native'
     uses: ./.github/workflows/run-tests-linux-multiarch.yml
 
   publish:
     name: Publish
     runs-on: ubuntu-22.04
     needs: [test-linux, test-windows, test-macos, test-jdk-compliance, test-multiarch]
+    if: github.repository == 'scala-native/scala-native'
```
armanbilge, line 32: I don't think this is working? The `publish` job is skipped because it needs the `test-*` jobs, but those are skipped if the workflow runs on schedule.

https://github.com/scala-native/scala-native/actions/runs/5593698843

WojciechMazur: Yes, you're correct. I really thought I fixed it already. Oh well, I will fix it soon

WojciechMazur: Oh, right, it's already on main: https://github.com/scala-native/scala-native/commit/61f4230a0bb7fe6c0fc58496ffddf4fd319debf0

PR #3397, opened 2023-07-18
Add partial implementation of java.util.EnumSet

### LINKED ISSUE #3385: Implement `java.util.EnumSet`

WojciechMazur:
Adds simple implementation of EnumSet for parts not requiring reflection support.
Resolves #3385

### DIFF
--- a/javalib/src/main/scala/java/util/EnumSet.scala
+++ b/javalib/src/main/scala/java/util/EnumSet.scala
@@ -2,27 +2,113 @@ package java.util
 
 import java.lang.Enum
 
-final class EnumSet[E <: Enum[E]] private (values: Array[E])
+final class EnumSet[E <: Enum[E]] private (values: Set[E])
     extends AbstractSet[E]
     with Cloneable
     with Serializable {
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/util/EnumSet.scala
```diff
@@ -2,27 +2,113 @@ package java.util
 
 import java.lang.Enum
 
-final class EnumSet[E <: Enum[E]] private (values: Array[E])
+final class EnumSet[E <: Enum[E]] private (values: Set[E])
```
bishabosha, line 5: this could still be an array where the index corresponds to ordinal, and the max ordinal + 1 is the size of the array? (never shrink, only grow by powers of 2?)

WojciechMazur: Sure, it could have been, but it's not going to be frequently used, and that way we get the Set ops for free. At some point it could be rewriten to be more optimal, but for current usages its good enough, especially since we cannot support the relection based semantics to populate it with all values

PR #3399, opened 2023-07-19
Refactor `tools` tests to remove dependency to `nscplugin`

tanishiking:
Related https://github.com/scala-native/scala-native/pull/3233#issuecomment-1632374110

- [x] Move tests (that are not related to `tools`) from `tools` to `nscplugin` and `nir`
  - [x] scala2
  - [x] scala3
- [ ] Invoke Scala compiler with provided classpath and sources - probably by invoking process -> let's do in another PR
- ~[ ] Retrieve list of errors from compile~

---

**solved**

When I try to run `nscplugin_3 / test`, we get the following error

`
[error] Test scala.scalanative.NIRCompilerTest.genericExternMethod failed: java.lang.LinkageError: loader constraint violation: when resolving method 'dotty.tools.dotc.util.SourceFile dotty.tools.dotc.util.SourceFile$.apply(dotty.tools.io.AbstractFile, scala.io.Codec)' the class loader java.net.URLClassLoader @79739ff4 of the current class, scala/scalanative/NIRCompiler, and the class loader 'app' for the method's defining class, dotty/tools/dotc/util/SourceFile$, have different Class objects for the type scala/io/Codec used in the signature (scala.scalanative.NIRCompiler is in unnamed module of loader java.net.URLClassLoader @79739ff4, parent loader scala.scalanative.FilteredClassLoader @55a49b8c; dotty.tools.dotc.util.SourceFile$ is in unnamed module of loader 'app'), took 0.001 sec
`

I'm not sure why this happens only in Scala3 🤔


nscplugin/src/test/scala/scala/scalanative/NIRCompiler.scala
```diff
@@ -0,0 +1,150 @@
+package scala.scalanative
+
+import java.nio.file.{Files, Path}
+import java.io.{File, PrintWriter}
+import java.net.URLClassLoader
+
+object NIRCompiler {
+
+  private val allow: String => Boolean =
+    n => n.startsWith("scala.scalanative.api.") || !n.startsWith("scala.")
```
WojciechMazur, line 10: This change should fix the linkage errors. In general we need to filter out both `scala.*` and `dotty.*` classes
```suggestion
  private val allow: String => Boolean = {
    case s"scala.scalanative.api.${_}" => true
    case s"scala.${_}"                 => false
    case s"dotty.${_}"                 => false
    case _                             => true
  }
```

project/Build.scala
```diff
@@ -131,6 +135,26 @@ object Build {
...
+          val scalalibCp = (scalalib / Compile / fullClasspath).value.files
+            .map(_.getAbsolutePath)
+            .mkString(pathSeparator)
+          Seq(
+            "-Dscalanative.nscplugin.jar=" + nscCompilerJar,
+            "-Dscalanative.testingcompiler.cp=" + testingCompilerCp,
+            "-Dscalanative.nativeruntime.cp=" + scalalibCp
+          )
+        },
+    }
```
WojciechMazur, line 157: I think we can potentially just use natilib instead of javalib here. This way we would not need to build javalib/auxlib/clib/posixlib/windowslib which has lots of sources. 
```suggestion
    .zippedSettings(Seq("testingCompiler", "nativelib")) {
      case Seq(testingCompiler, nativelib) =>
        Test / javaOptions ++= {
          val nscCompilerJar =
            (Compile / Keys.`package`).value.getAbsolutePath()
          val testingCompilerCp =
            (testingCompiler / Compile / fullClasspath).value.files
              .map(_.getAbsolutePath)
              .mkString(pathSeparator)
          val nativelibCp = (scalalib / Compile / fullClasspath).value.files
            .map(_.getAbsolutePath)
            .mkString(pathSeparator)
          Seq(
            "-Dscalanative.nscplugin.jar=" + nscCompilerJar,
            "-Dscalanative.testingcompiler.cp=" + testingCompilerCp,
            "-Dscalanative.nativeruntime.cp=" + nativelibCp
          )
        },
    }
```

tanishiking: Indeed it works 👍

tanishiking:
Oops, we probably need to fix CI to run those moved tests

tanishiking:
Run `nscPlugin / test` and `nir / test` on `test-tools` for now https://github.com/scala-native/scala-native/pull/3399/commits/2d5da9191bfe6afac34d857249e69dc348812f65

WojciechMazur (approved):

PR #3260, opened 2023-04-27
2-parameter (num/size) `stackalloc` intrinsic overload

natsukagami:
`stackalloc(num: Long, size: RawSize)` allocates `num` * `size` bytes as memory.
The `num` parameter is `Long` to maximize constant folding.

PR #3411, opened 2023-07-25
Overhaul of `scala.scalanative.unsafe.{stackalloc,alloc}`

WojciechMazur:
Deprecation and further removal of stackalloc/alloc methods taking a primitive Int was a mistake - it made the Scala Native code less readable and did not improved typesafety in any way (-1.toUnsigned = 0xFFFFFFF...). In this PR we revert some of this changes, and provide new mechanisms for safer interop. 

* Restore `stackalloc[T](Int)` and `alloc[T](Int)` methods. The now have special handling for literals integers and detect zero and negative literals at compile time. Non-literal values now have a runtime check to ensure that provide integer is positive integer. Unsigned variants are still present and have the same sementics as before. 
* `Intrinsic.stackalloc`signature has changed, methods take now generic `T` type used to resolve size of required memory. The `size: RawSize` argument no longer describes size of the allocated memory, instead it describe number of elements of sizeOf[T] to allocate. This change allows to skip some arithmetic operations at runtime/optimizer and allows to use the `nir.Type` argument of `nir.Op.Stackalloc` (previously always Byte). This work is based on #3260 
* 0-initialization of stackallocated memory is now moved to `Lowering` stage. Thanks to having a full information about allocated type, we can calculate exactly the required amount of memory at linktime after optimizer run, and allows to add optimizer runs designed to removal of zero-initialization of manually initialized memory. 
* All special intrinsic operations used as intermiediete step of compilation were moved to `Intrinsics.internal` object
* Removed most of no longer required `toUnsigned` conversion

### DIFF
--- a/nativelib/src/main/scala-2/scala/scalanative/unsafe/UnsafePackageCompat.scala
+++ b/nativelib/src/main/scala-2/scala/scalanative/unsafe/UnsafePackageCompat.scala
@@ -34,6 +34,11 @@ private[scalanative] trait UnsafePackageCompat { self =>
 
   /** Stack allocate and zero-initialize n values of given type */
   def stackalloc[T](n: CSize): Ptr[T] = macro MacroImpl.stackallocN[T]
+
+  /** Stack allocate and zero-initialize n values of given type. Better static
+   *  version.
+   */
+  def stackalloc[T](n: Long): Ptr[T] = macro MacroImpl.stackallocNLong[T]
 }
 
 private object MacroImpl {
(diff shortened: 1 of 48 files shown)


nativelib/src/main/scala-2/scala/scalanative/unsafe/UnsafePackageCompat.scala
```diff
@@ -114,32 +146,71 @@ private object MacroImpl {
...
           $runtime.fromRawPtr[$T]($rawptr)
         }"""
   }
 
+  private def validateSize(c: Context)(size: c.Tree): c.Tree = {
+    import c.universe._
+    size match {
+      case lit @ Literal(Constant(size: Int)) =>
+        if (size == 0)
+          c.error(c.enclosingPosition, "Allocatation of size 0 is fruitless")
```
ekrich, line 164:
```suggestion
          c.error(c.enclosingPosition, "Allocation of size 0 is fruitless")
```

nativelib/src/main/scala-3/scala/scalanative/unsafe/UnsafePackageCompat.scala
```diff
@@ -80,4 +99,36 @@ private[scalanative] trait UnsafePackageCompat {
...
+
+private object UnsafePackageCompat {
+  import scala.quoted.*
+  def validatedSize(size: Expr[Int])(using Quotes): Expr[RawSize] = {
+    import quotes.*
+    import quotes.reflect.*
+    val validatedSize = size.asTerm match {
+      case lit @ Literal(IntConstant(n)) =>
+        if n == 0 then
+          report.errorAndAbort("Allocatation of size 0 is fruitless", size)
```
ekrich, line 119:
```suggestion
          report.errorAndAbort("Allocation of size 0 is fruitless", size)
```

nscplugin/src/test/scala/scala/scalanative/unsafe/StackallocTest.scala
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.Assert._
+import org.junit.Assume._
+
+import scala.scalanative.api.CompilationFailedException
+import scala.scalanative.linker.StaticForwardersSuite.compileAndLoad
+import scala.scalanative.buildinfo.ScalaNativeBuildInfo._
+import scala.scalanative.NIRCompiler
+
+class StackallocTest {
+  val StackallocConcreateType = "Stackalloc requires concreate type"
```
ekrich, line 15:
```suggestion
  val StackallocConcreteType = "Stackalloc requires concrete type"
```
And places used.

nscplugin/src/test/scala/scala/scalanative/unsafe/StackallocTest.scala
```diff
@@ -0,0 +1,113 @@
...
+          |object Test {
+          |  val x = stackalloc()
+          |}""".stripMargin
+          )
+        )
+    )
+    assertTrue(err.getMessage().contains(StackallocConcreateType))
+  }
+
+  @Test def stackallocInferedType(): Unit = NIRCompiler(
```
ekrich, line 32:
```suggestion
  @Test def stackallocInferredType(): Unit = NIRCompiler(
```

ekrich (changes requested):
This looks like it will really help, especially with some math code I have.

PR #3424, opened 2023-08-09
Fix deprecations in `scala.scalanative.junit.AssertEqualsDoubleTest.scala`

Bensonater:
Hi all, 

When using Scala 3, `assertEquals(expected: Double, actual: Double)` throws a deprecation warning. Using `assertEquals(expected: Double, actual: Double, delta: Double)` suppresses the warning. This PR applies such fix to `AssertEqualsDoubleTest.scala`

Fix https://github.com/scala-native/scala-native/issues/3107 : New deprecations in AssertEqualsDoubleTest.scala


WojciechMazur (commented):
The tests defined in `junit-test` are designed to compare outputs of JUnit on JVM with the native implementation.  
To make the tests pass you'd need to first run JVM tests with field pointed below to true, and then commit the changed expected outputs.

https://github.com/scala-native/scala-native/blob/59f5b1200ad0b5119cbac2979b038d3edf6f39fb/junit-test/shared/src/test/scala/scala/scalanative/junit/utils/JUnitTest.scala#L21-L22

WojciechMazur (approved):

WojciechMazur:
> I wonder if there is a way to turn on that system property for the CI pipeline. So that CI runs the JVM tests first, write results to the output folder, and then runs native tests to compare against them. Relying on contributors to manually ensure the output folder is up-to-date seems more prone to error?

That might be possible, but this module is rarely changed. If the outputs of either JVM or Native would change for some reason we would see that. But typically, they don't

PR #3443, opened 2023-08-21
Emit local values for the debugger

WojciechMazur:
This uses local names and lexical scopes added in #3438  and #3386 to create LLVM `DILocalVariable` metadata, thus allowing to see local variables in the debuger. Scope of this PR is limited only to emitting local values using `llvm.dbg.value` intrinsic, and does not yet enrich generated metadata with the informations about reference types and their layout. These changes can be added in the follow up PRs. 
For local allocas we also doe define a metadata usign `llvm.dbg.declare` allowing to track state of local mutable variables, however, most of these is erased by the compiler. 
This PR does also fix the positions (line/column) in the backend to match offsets expected by the LLVM. We do not change logic of emiting positions in the frontend to make them compliant with JVM and Scala.js. 

During this works some new issues have been found and should be addressed in the future: 
* `llvm.dbg.value` is used to emit local state of variables, however, LLVM optimizer can erase some of the these information if it can prove that given value stored in the register is never used. To improve correctness of debug informations we might always allocate a memory on stack which would hold the state of local val, but this solution might be less efficient then current approach. 
* Optimizer can inline constants though it can eliminate any traces of local variables. We could mitigate this issue by introducing backend-only instructions, which would only hold debug info of partially-evaluated code. This new kinds of instructions would not be serializable in the NIR. 
* Function inlining can mess up iterating in the debugger. Stepping to the next instruction, can result in expected steping-into the inlined call. We would need to check if `inlinedAt` parameter for `DILocation` would be able to mitigate this issue. (It is emmited by the clang for inlined calls in C)
* Usage of URL's in `DIFile` is fruitless. LLVM cannot handle remote source files. To mitigate this issue we would need to download original sources (based on `mapSourceURI`) and used these instead if we want to allow for debugging foreign code.

### DIFF
--- a/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenExpr.scala
+++ b/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenExpr.scala
@@ -1446,7 +1446,9 @@ trait NirGenExpr[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
     def negateBool(value: nir.Val)(implicit pos: nir.Position): Val =
       buf.bin(Bin.Xor, Type.Bool, Val.True, value, unwind)
 
-    def genUnaryOp(code: Int, rightp: Tree, opty: nir.Type)(implicit pos: nir.Position): Val = {
+    def genUnaryOp(code: Int, rightp: Tree, opty: nir.Type)(implicit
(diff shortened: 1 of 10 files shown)


WojciechMazur:
CI reveals that this PR is block until #3444 is merged

Issue #3435, opened 2023-08-14
`@struct` with `SCALANATIVE_OPTIMIZE=false` and `withDebugMetadata(true)` leads to "No idea how to dwarfise"


tanishiking:
From scheduled build https://github.com/scala-native/scala-native/actions/runs/5851070393/job/15863485291

```scala
// Main.scala
import scala.scalanative.runtime.struct

object Main {
  @struct class SC1(val b: Int)

  def main(args: Array[String]): Unit = {
    val x = new SC1(5)
    println(x.b)
  }
}

// build.sbt
nativeConfig ~= { c =>
  c.withDebugMetadata(true)
}
```

```sh
export SCALANATIVE_OPTIMIZE=false

sbt nativeLink

...
[error] Caused by: scala.NotImplementedError: No idea how to dwarfise StructValue(List(Int))
[error]         at scala.scalanative.codegen.llvm.MetadataCodeGen.$anonfun$toMetadataTypeOpt$1(MetadataCodeGen.scala:166)
```

`nativeLink` will fail by falling into https://github.com/scala-native/scala-native/blob/36d32f015d5a024edf307cbd3e68e8ee90753a5d/tools/src/main/scala/scala/scalanative/codegen/llvm/MetadataCodeGen.scala#L162

We would need to somehow convert `StructValue` to DWARF type (or ignore at least).

tanishiking:
Gonna see can we convert StructValue to DWARF, or is it okay to ignore types.

tanishiking:
```cpp
struct car {
  int year;
};

int main() {
  car myCar1;
  myCar1.year = 1999;
}
```

```sh
$ clang++ -g -S -emit-llvm struct.cpp
$ cat struct.ll

...
%struct.car = type { i32 }

; Function Attrs: mustprogress noinline norecurse nounwind optnone ssp uwtable
define noundef i32 @main() #0 !dbg !13 {
  %1 = alloca %struct.car, align 4
  call void @llvm.dbg.declare(metadata %struct.car* %1, metadata !18, metadata !DIExpression()), !dbg !22
  %2 = getelementptr inbounds %struct.car, %struct.car* %1, i32 0, i32 0, !dbg !23
  store i32 1999, i32* %2, align 4, !dbg !24
  ret i32 0, !dbg !25
}

...
[...]

Issue #3442, opened 2023-08-21
nativeLink fails with `SCALANATIVE_OPTIMIZE=false` saying "Lowering ClassOf needs nir.Buffer"


tanishiking:
It looks like https://github.com/scala-native/scala-native/pull/3386 introduces a bug when `SCALANATIVE_OPTIMIZE=false` which occurs for `sandbox` project: blocks https://github.com/scala-native/scala-native/issues/3435

https://github.com/scala-native/scala-native/actions/runs/5898012257/job/15999032485

```sh
export SCALANATIVE_OPTIMIZE=false
sbt
sbt> sandbox3 / nativeLink
...
[error] scala.scalanative.util.UnsupportedException: Lowering ClassOf needs nir.Buffer
[error]         at scala.scalanative.util.package$.unsupported(package.scala:18)
[error]         at scala.scalanative.codegen.Lower$Impl.onVal(Lower.scala:250)
[error]         at scala.scalanative.nir.Transform.onOp(Transform.scala:100)
[error]         at scala.scalanative.nir.Transform.onOp$(Transform.scala:58)
[error]         at scala.scalanative.codegen.Lower$Impl.onOp(Lower.scala:17)
[error]         at scala.scalanative.nir.Transform.onInst(Transform.scala:41)
[error]         at scala.scalanative.nir.Transform.onInst$(Transform.scala:32)
[error]         at scala.scalanative.codegen.Lower$Impl.onInst(Lower.scala:244)
[error]         at scala.scalanative.codegen.Lower$Impl.$anonfun$onInsts$7(Lower.scala:236)
[error]         at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
[error]         at scala.collection.mutable.UnrolledBuffer$Unrolled.foreach(UnrolledBuffer.scala:242)
[error]         at scala.collection.mutable.UnrolledBuffer.foreach(UnrolledBuffer.scala:148)
[error]         at scala.collection.TraversableLike.map(TraversableLike.scala:286)
[error]         at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
[error]         at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[error]         at scala.scalanative.codegen.Lower$Impl.onInsts(Lower.scala:236)
[error]         at scala.scalanative.nir.Transform.onDefn(Transform.scala:19)
[error]         at scala.scalanative.nir.Transform.onDefn$(Transform.scala:8)
[...]

PR #3444, opened 2023-08-22
Fix emitting debug information for structe types with `SCALANATIVE_OPTIMIZE=false`

tanishiking:
fix https://github.com/scala-native/scala-native/issues/3435

Now, the following Scala code without optimization will emit

```scala
import scala.scalanative.runtime.struct

object Main {
  @struct class SC1(val b: Int, val c: Boolean)

    val x = new SC1(5, true)
    println(x.b)
  }
}
```

```
!22 = !DIBasicType(name: "int", size: 32, align: 32, encoding: DW_ATE_signed)
!23 = !DIDerivedType(tag: DW_TAG_member, baseType: !22, size: 32)
!24 = !DIBasicType(name: "bool", size: 8, align: 8, encoding: DW_ATE_boolean)
!25 = !DIDerivedType(tag: DW_TAG_member, baseType: !24, size: 8)
!26 = !{!23, !25}
!27 = !DICompositeType(tag: DW_TAG_structure_type, size: 64, elements: !26)
```

Since we don't have `name` attributs in `DICompositeType` and `DIBasicType`, we will debugger can't show their type name and members' names.
In order to preserve those information, we need to somehow change the way of traversing NIR or adding name information into NIR

```
(__lldb_invalid_typedef_name) oldCar = {
   = 1980
   = false
   = 0xb96d000190b47e3c
}
(__lldb_invalid_typedef_name) myCar1 = {
   = 16260
   = true
   = 0x0000000100009c40
}
```

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/llvm/AbstractCodeGen.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/llvm/AbstractCodeGen.scala
@@ -17,8 +17,6 @@ import scala.scalanative.codegen.{Metadata => CodeGenMetadata}
 
 import scala.language.implicitConversions
 import scala.scalanative.codegen.llvm.Metadata.conversions._
-import scala.scalanative.nir.Global.Member
-import scala.scalanative.nir.Global.Top
 
 private[codegen] abstract class AbstractCodeGen(
     env: Map[Global, Defn],
--- a/tools/src/main/scala/scala/scalanative/codegen/llvm/AbstractCodeGen.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/llvm/AbstractCodeGen.scala
@@ -17,6 +17,8 @@ import scala.scalanative.codegen.{Metadata => CodeGenMetadata}
 
 import scala.language.implicitConversions
 import scala.scalanative.codegen.llvm.Metadata.conversions._
+import scala.scalanative.nir.Global.Member
+import scala.scalanative.nir.Global.Top
 
 private[codegen] abstract class AbstractCodeGen(
     env: Map[Global, Defn],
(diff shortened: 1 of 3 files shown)


tanishiking:
Due to the issue https://github.com/scala-native/scala-native/issues/3442 this PR is based on `36d32f0`

tools/src/main/scala/scala/scalanative/codegen/llvm/AbstractCodeGen.scala
```diff
@@ -17,6 +17,8 @@ import scala.scalanative.codegen.{Metadata => CodeGenMetadata}
 
 import scala.language.implicitConversions
 import scala.scalanative.codegen.llvm.Metadata.conversions._
+import scala.scalanative.nir.Global.Member
+import scala.scalanative.nir.Global.Top
```
WojciechMazur, line 21: Spourious change

WojciechMazur (approved):
Overall looks good. We potentially can use indexes if we would stumble upon unnamed structs, but overall it would be better to have an actual names in nir. Currently most of the structs are used for intrinsic stuff (eg. generated values). We potentially can use `scalanative.linker.Class` / `scalanative.codegen.FieldLayour` for the user defined types which are also struct types, but let's leave it for another dedicated PR

tanishiking:
> We potentially can use indexes if we would stumble upon unnamed structs, but overall it would be better to have an actual names in nir.

Maybe it's better to add field name info into `StructValue` https://github.com/tanishiking/scala-native/blob/f0b390e4dacdefa74c04a8f663f93f6402ae70a5/nir/src/main/scala/scala/scalanative/nir/Types.scala#L70

like 

```
final case class StructValue(
  name: String, // name of this struct, SC1 in the example above,
  tys: Seq[(String, Type)] // tuple of member name and type
) extends AggregateKind
```

before release of 0.5.0?

anyway, we can work on this in another PR :+1:

WojciechMazur:
> Maybe it's better to add field name info into StructValue 

That's one of solutions, but I don't think we should mix metadata with actual values. Also, the @struct is one of the remaining special use cases just to be able to use functions which return struct by value. All other StructValues are completly internal, and emited in the codegen, in most cases they represent the field layout of a class. 
Also, in debugger we would rather be interested in the local variable of `x: SC1`. When it comes to the composite type describing `SC1` we can even make it up by ourselfs my mangling the types and assuming this is an anonymous struct (which is true for all `CStructN` types currently.

PR #3446, opened 2023-08-23
Remove static calls mitigation logic from linker

WojciechMazur:
Removes special handling for static methods from linker. That was required only for backward compatibilty with Scala Native 0.4.3-.

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/ForkJoinTask.scala
+++ b/javalib/src/main/scala/java/util/concurrent/ForkJoinTask.scala
@@ -637,7 +637,7 @@ object ForkJoinTask {
     ForkJoinPool.getSurplusQueuedTaskCount()
 
   // The next 4 methods should be defined as `protected static`, however this kind of access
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/util/concurrent/ForkJoinTask.scala
```diff
@@ -636,28 +636,32 @@ object ForkJoinTask {
   def getSurplusQueuedTaskCount(): Int =
     ForkJoinPool.getSurplusQueuedTaskCount()
 
-  protected def peekNextLocalTask(): ForkJoinTask[_] = {
+  // The next 4 methods should be defined as `protected static`, however this kind of access
+  // does not make lots of sens in Scala. The most similiar would be `protected[concurrent]`
```
ekrich, line 640:
```suggestion
  // does not make a lot of sense in Scala. The most similar access would be `protected[concurrent]`
```

PR #3450, opened 2023-08-25
Enable IDE support for nscplugin in Scala 2

tanishiking:
This PR enables IDE support for the nscplugin project in Scala 2. Previously, IDEs (especially Metals) were restricted to using Scala 3 (https://github.com/scala-native/scala-native/pull/3344), which meant that they could not provide any IDE features for sources under `nscplugin/src/main/scala-2`. This was because we did not export bloop configuration for Scala 2 projects.

This PR fixes this issue by:

- Adding the `additionalIDEScalaVersions` property to `MultiScalaProject`, which allows bloop to generate configuration for Scala 2 projects.
- Generating IDE configuration for the `nscplugin2_13` project.
- Making the `nscplugin` project depend on the `nir`, `util`, and `testingCompilerInterface` projects on source.
  - The `nscplugin2_13` project would depend on the `nir2_13` and `util2_13` projects, which are not supported by IDEs.
  - To work around this, the `nscplugin2_13` project will also depend on the `nir` and `util` projects on source in the IDE.


project/Build.scala
```diff
@@ -930,7 +930,8 @@ object Build {
 
     /** Depends on the sources of another project. */
     def dependsOnSource(dependency: MultiScalaProject): MultiScalaProject = {
-      if (isGeneratingForIDE) project.dependsOn(dependency)
+      if (isGeneratingForIDE && !project.shouldDependsOnSourceInIDE)
```
ekrich, line 933:
```suggestion
      if (isGeneratingForIDE && !project.dependsOnSourceInIDE)
```
This would probably read better.

project/Build.scala
```diff
@@ -114,7 +114,7 @@ object Build {
       )
 
   // Compiler plugins
-  lazy val nscPlugin = MultiScalaProject("nscplugin", file("nscplugin"))
+  lazy val nscPlugin = MultiScalaProject("nscplugin", file("nscplugin"), "2.13")
```
WojciechMazur, line 117: Can we use a named argument here please? Otherwise it's really hard to reason what it means. Also probably a list instead of varargs would be easier to maintaine here, especially if we would add another parameters and since most of other parameters are Strings as well

WojciechMazur (commented):
Just 1 comment, otherwise looks good to me. I was wondering why `val Scope = config(_)`  was needed, but that's actually the sbt limitation

project/Build.scala
tanishiking: Sure! done https://github.com/scala-native/scala-native/pull/3450/commits/fedb4f35c1292863175ff27372a0886e3ae02b6c

project/Build.scala
tanishiking: thanks! done https://github.com/scala-native/scala-native/pull/3450/commits/0b305f4810ca3ee31e0cb51cfd94bb2ef441e845

Issue #3470, opened 2023-09-05
javalib String.offsetByCodePoints throws if the final character in the string is an unpaired high surrogate


rayrobdod:
javalib `String.offsetByCodePoints` throws if the final character in the string is an unpaired high surrogate and the method processes this final character. 

### Example

```scala
@main def Test = {
  println("\uD800".offsetByCodePoints(0, 1))
}
```

#### Expected (on JVM)

```
1
```

#### Actual

```
java.lang.ArrayIndexOutOfBoundsException: 1
        at java.lang.StackTrace$.currentStackTrace$$anonfun$1(Unknown Source)
        at java.lang.StackTrace$$$Lambda$2.applyVoid(Unknown Source)
        at scala.runtime.function.JProcedure1.apply(Unknown Source)
        at scala.scalanative.unsafe.Zone$.apply(Unknown Source)
        at java.lang.StackTrace$.currentStackTrace(Unknown Source)
        at java.lang.Throwable.fillInStackTrace(Unknown Source)
        at scala.scalanative.runtime.package$.throwOutOfBounds(Unknown Source)
        at java.lang.Character$.offsetByCodePoints(Unknown Source)
        at java.lang.String.offsetByCodePoints(Unknown Source)
        at issue$package$.Test(Unknown Source)
        at Test.main(Unknown Source)
        at <none>.main(Unknown Source)
        at <none>.__scrt_common_main_seh:(D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288)(Unknown Source)
        at <none>.BaseThreadInitThunk(Unknown Source)
        at <none>.RtlUserThreadStart(Unknown Source)
```

tanishiking:
>  Unpaired surrogates within the text range given by index and codePointOffset count as one code point each.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#offsetByCodePoints(int,int)

```
❯ cat test.scala
@main def main = {
    println("\uD800".offsetByCodePoints(0, 1))
}

❯ scala-cli test.scala
Compiling project (Scala 3.3.0, JVM)
Compiled project (Scala 3.3.0, JVM)
1

❯ scala-cli test.scala --native
Compiling project (Scala 3.3.0, Scala Native)
Compiled project (Scala 3.3.0, Scala Native)
[info] Linking (756 ms)
[info] Discovered 680 classes and 3771 methods
[info] Optimizing (debug mode) (889 ms)
[...]

PR #3471, opened 2023-09-05
Fix String.offsetByCodePoints for unpaired surrogates

tanishiking:
fix https://github.com/scala-native/scala-native/issues/3470

Previously, String.offsetByCodePoints had an error when the offsetByCodePoints ends with the unpaired higher surrogate.

```
❯ cat test.scala
@main def main = {
    println("\uD800".offsetByCodePoints(0, 1))
}

❯ scala-cli test.scala
Compiling project (Scala 3.3.0, JVM)
Compiled project (Scala 3.3.0, JVM)
1

❯ scala-cli test.scala --native
Compiling project (Scala 3.3.0, Scala Native)
Compiled project (Scala 3.3.0, Scala Native)
[info] Linking (756 ms)
[info] Discovered 680 classes and 3771 methods
[info] Optimizing (debug mode) (889 ms)
[info] Generating intermediate code (856 ms)
[info] Produced 8 files
[info] Compiling to native code (1649 ms)
[info] Total (4257 ms)
java.lang.ArrayIndexOutOfBoundsException: 1
        at java.lang.StackTrace$.currentStackTrace$$anonfun$1(Unknown Source)
        at java.lang.StackTrace$$$Lambda$2.applyVoid(Unknown Source)
        at scala.runtime.function.JProcedure1.apply(Unknown Source)
        at scala.scalanative.runtime.package$.throwOutOfBounds(Unknown Source)
        at java.lang.Character$.offsetByCodePoints(Unknown Source)
        at java.lang.String.offsetByCodePoints(Unknown Source)
        at test$package$.main(Unknown Source)
        at main.main(Unknown Source)
```

This commit fix the bound check in the offsetByCodePoints not to check the next index is the lower surrogate after the higher surrogate.

### DIFF
--- a/javalib/src/main/scala/java/lang/Character.scala
+++ b/javalib/src/main/scala/java/lang/Character.scala
@@ -378,7 +378,7 @@ object Character {
         }
         if (isHighSurrogate(seq(i))) {
           val next = i + 1
-          if (next <= end && isLowSurrogate(seq(next))) {
+          if (next < end && isLowSurrogate(seq(next))) {
             i += 1
           }
         }
@@ -396,7 +396,7 @@ object Character {
         }
         if (isLowSurrogate(seq(i))) {
           val prev = i - 1
-          if (prev >= start && isHighSurrogate(seq(prev))) {
+          if (prev > start && isHighSurrogate(seq(prev))) {
             i -= 1
           }
         }
--- a/unit-tests/shared/src/test/scala/org/scalanative/testsuite/javalib/lang/StringTest.scala
+++ b/unit-tests/shared/src/test/scala/org/scalanative/testsuite/javalib/lang/StringTest.scala
@@ -127,6 +127,32 @@ class StringTest {
     )
   }
 
+  @Test def offsetByCodePoints(): Unit = {
+    assertTrue("abc".offsetByCodePoints(0, 3) == 3)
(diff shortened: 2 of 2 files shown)


javalib/src/main/scala/java/lang/Character.scala
```diff
@@ -378,7 +378,7 @@ object Character {
         }
         if (isHighSurrogate(seq(i))) {
           val next = i + 1
-          if (next <= end && isLowSurrogate(seq(next))) {
+          if (next < end && isLowSurrogate(seq(next))) {
```
tanishiking, line 381: Though there shouldn't be a difference, it might be more accurate by

```scala
if (next <= end && next < seq.length && isLowerSurrogate(seq(next))) {
```
?

WojciechMazur (approved):

javalib/src/main/scala/java/lang/Character.scala
tanishiking: That's a good idea. I also checked the Scala.js's implementation and it seems they don't have an implementation [Character#offsetByCodePoints](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#offsetByCodePoints-char:A-int-int-int-int-) and there's only a `String.offsetByCodePoints`.

It might be good to sync after we implement `Character.offsetByCodePoints` on scala.js side :+1:

PR #3472, opened 2023-09-05
Detect at linktime usage of unsupported features

WojciechMazur:
This change allows us to detect usage of unsupported features early at linktime based of linktime-resolved conditions. 
Thanks to this feature, we can guard against of runtime failures for some well known execution flows. One of them is running Thread with disabled multithreading support. 
By defining `scala.scalanative.runtime.UnsupportedFeature` as Java file it would never contain NIR. We can collect it's usages and adapt them to present more user friendly messages.


WojciechMazur:
@natsukagami It can be used later to guard against usage of delimited continuations on Windows. It would be great to add 1 more unsupported feature for continuations instead of reusing virtualThreads category.

PR #3500, opened 2023-09-23
+javalib Add CompletionStage

He-Pin:
Motivation:
Add CompletionStage


javalib/src/main/scala/java/util/concurrent/CompletableFuture.scala
```diff
@@ -0,0 +1,3 @@
+package java.util.concurrent
+
+abstract class CompletableFuture[T] extends Future[T] with CompletionStage[T] {}
```
He-Pin, line 3: Will continue it tomorrow.

PR #3509, opened 2023-09-26
Factor out common definitions in compiler plugin phases (DRY)

kyouko-taiga:
This PR factors out `defn` and `defnNir` from the `PrepNativeInterop` and `PostInlineNativeInterop` phases of the Scala Native compiler plugin.


kyouko-taiga:
Thanks for letting me know. I'll keep that in mind. I don't think the code duplication was intentional here, though. The two passes from which the methods were factored out are exclusive to SN AFAICT.

WojciechMazur (approved):

nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NativeInteropUtil.scala
```diff
@@ -0,0 +1,16 @@
...
+import dotty.tools.dotc.core.Contexts.ctx
+import dotty.tools.dotc.core.Definitions
+
+trait NativeInteropUtil { self: PluginPhase =>
+
+  /** Returns the definitions in the current context. */
+  protected def defn(using Context): Definitions = ctx.definitions
+
+  /** Returns the Native IR definitions in the current context. */
+  protected def defnNir(using Context): NirDefinitions = NirDefinitions.get
```
WojciechMazur, line 14: These are also used in `NirCodeGen` however, this is taking a Context in constructor so it can be just a val and save a few cycles on each usage

PR #3517, opened 2023-09-28
Allow to prefetch debug info to mitigate spurious failures in the CI

WojciechMazur:
Loading debug info can take almost 1s on MacOS M1 for a simple tests (1 test in sandbox) or even 3s for our current unit tests. This can influence the assertions that expect code to finish withing some time interval, especially in multithreading builds. 
Mitigate this issue by introducing env var `PREFETCH_DEBUG_INFO=1` to ensure that debug info would be parsed before starting to execute tests

### DIFF
--- a/test-interface/src/main/scala/scala/scalanative/testinterface/TestMain.scala
+++ b/test-interface/src/main/scala/scala/scalanative/testinterface/TestMain.scala
@@ -92,7 +92,9 @@ object TestMain {
     // Prefetch the debug metadata before the actual tests do start
     if (LinktimeInfo.hasDebugMetadata) {
       val shouldPrefetch =
-        sys.env.get("SCALANATIVE_TEST_PREFETCH_DEBUG_INFO").exists(v => v.isEmpty() || v == "1")
+        sys.env
+          .get("SCALANATIVE_TEST_PREFETCH_DEBUG_INFO")
+          .exists(v => v.isEmpty() || v == "1")
       if (shouldPrefetch)
         new RuntimeException().fillInStackTrace().ensuring(_ != null)
     }
--- a/test-interface/src/main/scala/scala/scalanative/testinterface/TestMain.scala
+++ b/test-interface/src/main/scala/scala/scalanative/testinterface/TestMain.scala
@@ -88,6 +88,15 @@ object TestMain {
 
     SignalConfig.setDefaultHandlers()
 
+    // Loading debug metadata can take up to few seconds which might mess up timeout specific tests
+    // Prefetch the debug metadata before the actual tests do start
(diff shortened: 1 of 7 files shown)


test-interface/src/main/scala/scala/scalanative/testinterface/TestMain.scala
```diff
@@ -88,6 +88,15 @@ object TestMain {
 
     SignalConfig.setDefaultHandlers()
 
+    // Loading debug metadata can take up to few seconds which might mess up timeout specific tests
+    // Prefetch the debug metadata before the actual tests do start
+    if (LinktimeInfo.hasDebugMetadata) {
+      val shouldPrefetch =
+        sys.env.get("PREFETCH_DEBUG_INFO").exists(v => v.isEmpty() || v == "1")
```
armanbilge, line 95: Should env be prefixed with e.g. `SN_TEST` or `SCALA_NATIVE_TEST` ?

WojciechMazur: I could be prefixed, but don't you think it is too verbose? When it's added in CI it's fine, but when testing locally it gives a lot of opportunities for typos. On the other hand the short version breaks naming consistency. What's your opinion on that?

armanbilge: If this was private to this repo, then it would be okay, but since this is a public API feature it should follow existing convention. Verbosity is annoying but hopefully we choose the right default and most user should never need to configure it.

ekrich: I know Denys always wanted it spelled out `SCALA_NATIVE_` perhaps because so many C libs use a 2 character namespacing like `tf_` for Tensorflow.

tanishiking (approved):
> Loading debug info can take almost 1s on MacOS M1 for a simple tests (1 test in sandbox) or even 3s for our current unit tests. This can influence the assertions that expect code to finish withing some time interval, especially in multithreading builds.

Nice catch! I thought it was acceptable to take 1s to load debug info when throwing an exception, but I hadn't thought about the test cases where exceptions are expected.

Regarding the env var prefix, I personally prefer ScalaNative having some kind of prefix so that the env var name doesn't crash with others, but not a strong opinion

test-interface/src/main/scala/scala/scalanative/testinterface/TestMain.scala
WojciechMazur: I've added the `SCALANATIVE_TEST_` prefix to allign it with other env variables (eg. `SCALANATIVE_GC`) and added `TEST` to point out it's only applied when testing

PR #3537, opened 2023-10-05
Preserve more information about unreachable symbols

WojciechMazur:
When working on #3532 I've found out that current unreachable symbols don't contain information about argument/result types of unreachable symbol. This made debuggin harder, when the issue were the differencing argument type lists. 
This PR improves unreachable symbols stacktraces by adding more information, and adapts the backtrace loggin to use additional informaiton. It also improves formatting of backtraces to make them more readable. 

image

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/ScalaNative.scala
+++ b/tools/src/main/scala/scala/scalanative/build/ScalaNative.scala
@@ -111,12 +111,27 @@ private[scalanative] object ScalaNative {
           buf: StringBuilder,
           backtrace: List[Reach.BackTraceElement]
       ): Unit = {
+        import scala.io.AnsiColor._
         // Build stacktrace in memory to prevent its spliting when logging asynchronously
-        val padding = backtrace.foldLeft(0)(_ max _.kind.length())
-        backtrace.foreach {
-          case Reach.BackTraceElement(_, kind, name, filename, line) =>
-            val pad = " " * (padding - kind.length())
-            buf.append(s"    $pad$kind at $name($filename:$line)\n")
+        val elems = backtrace.map {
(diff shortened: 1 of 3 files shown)


tools/src/main/scala/scala/scalanative/build/ScalaNative.scala
```diff
@@ -111,22 +111,37 @@ private[scalanative] object ScalaNative {
...
+        val elems = backtrace.map {
+          case elem @ Reach.BackTraceElement(_, symbol, filename, line) =>
+            import symbol.argTypes
+            val rendered = symbol.toString
+            val descriptorStart = rendered.indexOf(symbol.name)
+            val (modifiers, descriptor) = rendered.splitAt(descriptorStart)
+            val (name, typeInfo) =
+              if (argTypes.nonEmpty) descriptor.splitAt(descriptor.indexOf("("))
+              else (descriptor, "")
+            modifiers -> s"$BOLD$YELLOW$name$RESET$typeInfo at $BOLD$filename:$line"
```
keynmol, line 125: I don't think AnsiColor respects the various colour disabling env flags (see https://no-color.org/ and https://clig.dev/#output in general), but it'd be good to respect the env variable (+ whatever color information the build tool can give us) so that the output of the build tooling can be processed easier.

WojciechMazur: Not sure about the other build tools, but running sbt with `--color=false` is able to filter out all ANSI modifiers on its own. I'm not sure if we're able to disable colors in scala-cli at all. 
Anyway I'll the `NO_COLOR` env variable to allow for disabling coloring

PR #3545, opened 2023-10-06
Synchronize access to `WeakReferenceRegistry`

WojciechMazur:
`WeakReferenceRegistry` was fragile to data races - thread executed post GC handlers could have interfered with registration of new `WeakReference`. We add basic synchronized clouses and additionally replace Scala HashMap with it's Java counterpart


javalib/src/main/scala/java/lang/ref/WeakReferenceRegistry.scala
```diff
@@ -10,12 +10,10 @@ import scala.scalanative.runtime.GC
...
 private[java] object WeakReferenceRegistry {
-  private var weakRefList: immutable.List[WeakReference[_]] =
-    Nil
+  private var weakRefList: List[WeakReference[_]] = Nil
 
-  private val postGCHandlerMap
-      : mutable.HashMap[WeakReference[_], Function0[Unit]] =
-    new mutable.HashMap()
+  private val postGCHandlerMap: ju.HashMap[WeakReference[_], Function0[Unit]] =
+    new ju.HashMap()
```
armanbilge, line 16: Instead of synchronization, can we use an `AtomicReference` and a `TrieMap`?

ekrich: It needs to use a Java collection in the `javalib`.

armanbilge: Yes, ideally we would use `ConcurrentHashMap`, but multi-threading support has not yet been added for it 😕 

Until then, a `TrieMap` will be okay. We already depend on it in other places in `javalib` for the same reasons.
https://github.com/scala-native/scala-native/blob/92d5393dd3bf22498cc478777a04d4070a8fe1be/javalib/src/main/scala/java/lang/Throwables.scala#L8-L9

ekrich: Sounds like a good job but I know squat about concurrency. Isn't the Java atomic stuff done now?

WojciechMazur: I actually needed to go few more steps ahead. So for some reason AtomicReference or even stdatomic.AtomicRef wrappers were leading to runtime exceptions, probably due to main thread initialization. Becouse I've that I've used raw stdatomic calls, but then I've seen that we don't relaly need expensive TrieMap, we can just use linked-list of weak references. This way we can remove Scala collections dependency from WeakReference and make them more lightweight

PR #3546, opened 2023-10-06
Try to limit amount of classloaded symbols 

WojciechMazur:
This change tries to limit ammount of classloaded, always required symbols. The changes allow to reduce this number from ~10k to ~4.8k loaded methods for builds with disabled debug metadata and multithread on macos.

 The changes are limit only changes which does not significentlly change runtime behaviour. Additional removal of loaded methods might be possible in the future.

### DIFF
--- a/javalib/src/main/scala/java/lang/Throwables.scala
+++ b/javalib/src/main/scala/java/lang/Throwables.scala
@@ -5,10 +5,12 @@ import scalanative.unsafe._
 import scalanative.unsigned._
 import scalanative.runtime.unwind
 import scala.scalanative.meta.LinktimeInfo
-// TODO: Replace with j.u.c.ConcurrentHashMap when implemented to remove scalalib dependency
+// TODO: Replace with j.u.c.ConcurrentHashMap when implemented to remove scalalib dependency and save loading ~600 symbols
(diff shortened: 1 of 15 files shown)


javalib/src/main/scala/java/lang/Throwables.scala
```diff
@@ -5,20 +5,22 @@ import scalanative.unsafe._
 import scalanative.unsigned._
 import scalanative.runtime.unwind
 import scala.scalanative.meta.LinktimeInfo
-// TODO: Replace with j.u.c.ConcurrentHashMap when implemented to remove scalalib dependency
+// TODO: Replace with j.u.c.ConcurrentHashMap when implemented to remove scalalib dependency and save loading ~600 symbols
 import scala.collection.concurrent.TrieMap
 import scala.scalanative.runtime.Backtrace
 import scala.scalanative.runtime.NativeThread
+import scala.scalanative.libc.stdlib.{malloc, calloc, free}
+import java.util.concurrent.ConcurrentHashMap
```
ekrich, line 13: This looks really good. Doesn't look like we need this import.

ekrich: One other small nit "ammount" -> "amount" in the title.

PR #3560, opened 2023-10-10
Port JSR-166 `LinkedTransferQueue`

natsukagami:
Should be a complete port with tests included.

### DIFF
--- a/javalib/src/main/scala/java/util/concurrent/LinkedTransferQueue.scala
+++ b/javalib/src/main/scala/java/util/concurrent/LinkedTransferQueue.scala
@@ -1,0 +1,381 @@
+/*
+ * Written by Doug Lea with assistance from members of JCP JSR-166
+ * Expert Group and released to the public domain, as explained at
+ * http://creativecommons.org/publicdomain/zero/1.0/
+ */
+
+package java.util.concurrent
+
+import java.util.AbstractQueue
+import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.locks.LockSupport
+import scala.scalanative.libc.atomic.CAtomicRef
+import scala.scalanative.libc.atomic.memory_order._
+import scala.scalanative.runtime.{fromRawPtr, Intrinsics}
+import scala.annotation.tailrec
+
+@SerialVersionUID(-3223113410248163686L) class LinkedTransferQueue[E <: AnyRef]
+    extends AbstractQueue[E]
+    with TransferQueue[E]
+    with Serializable {
+  import LinkedTransferQueue._
+
+  @transient @volatile private var head: Node = _
+  @transient @volatile private var tail: Node = _
+  @transient @volatile private var needSweep: Boolean = _
+
+  val tailAtomic = CAtomicRef[Node](
+    fromRawPtr(Intrinsics.classFieldRawPtr(this, "tail"))
+  )
+  val headAtomic = CAtomicRef[Node](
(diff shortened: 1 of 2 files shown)


javalib/src/main/scala/java/util/concurrent/LinkedTransferQueue.scala
```diff
@@ -0,0 +1,1804 @@
...
+        if (_nanos > SPIN_FOR_TIMEOUT_THRESHOLD)
+          LockSupport.parkNanos(this, _nanos)
+      }
+
+      item = s.item
+    }
+    if (stat == 1)
+      s.waiterAtomic.store(null)
+    if (!isData)
+      s.itemAtomic.store(s) // self-link to avoid garbage
```
armanbilge, line 561: Actually `.set` on a `VarHandle` is a plain write.

> Sets the value of a variable to the newValue, with memory semantics of setting as if the variable was declared non-volatile and non-final. Commonly referred to as plain write access. 

https://github.com/armanbilge/jsr166/blob/abd155f72243cc2066c31ee732d21894f0edf7bd/src/main/java/util/concurrent/LinkedTransferQueue.java#L650-L653

unit-tests/shared/src/test/scala/org/scalanative/testsuite/javalib/util/concurrent/LinkedTransferQueueTest.scala
```diff
@@ -0,0 +1,1034 @@
...
+    mustEqual(0, q.size())
+  }
+
+  /** offer transfers elements across Executor tasks
+   */
+  @Test def testOfferInExecutor(): Unit = {
+    val q = new LinkedTransferQueue[Item]()
+    val threadsStarted = new CheckedBarrier(2)
+    val executor = Executors.newFixedThreadPool(2)
+    Using(cleaner(executor)) { cleaner =>
```
WojciechMazur, line 549: For the sake of Scala 2.12 compability there is defined a helper method for these in JSR166Test: ` usingPoolCleaner(new ForkJoinPool()) { p: ForkJoinPool =>`

javalib/src/main/scala/java/util/concurrent/LinkedTransferQueue.scala
natsukagami: Thanks for pointing that out! `relaxed` the store :)

javalib/src/main/scala/java/util/concurrent/LinkedTransferQueue.scala
```diff
@@ -0,0 +1,1804 @@
...
+import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.locks.LockSupport
+import scala.scalanative.libc.stdatomic.AtomicRef
+import scala.scalanative.libc.stdatomic.memory_order.{
+  memory_order_relaxed,
+  memory_order_release
+}
+import scala.scalanative.runtime.{fromRawPtr, Intrinsics}
+
+/** An unbounded {@link TransferQueue} based on linked nodes. This queue orders
```
WojciechMazur, line 21: You can drop the scaladoc comments, these are problematic when using `{ x }` javadoc syntax which might be splitted by scalafmt and moved partially to next line making them invalid. Because of that Scala 2 scaladoc fails the compilation. In case of javalib we use scaladoc only to have a database of implemented methods/classes. The user shall refer to JDK scaladoc instead for details.

WojciechMazur (approved):

PR #3562, opened 2023-10-11
Enable to specify include and exclude resource patterns with glob

tanishiking:
fix https://github.com/scala-native/scala-native/issues/3433

This commit enables scala-native to specify include and exclude resource patterns with glob. This will allow users to reduce the size of their executables by excluding unnecessary resources.

To use this feature, users can specify the include and exclude patterns using the `NativeConfig`'s `withResourceIncludePatterns` and `withResourceExcludePatterns` APIs. For example, to exclude the `rootdoc.txt` file, users would specify the following pattern:

```scala
_.withResourceIncludePatterns(Seq("**.txt"))
 .withResourceExcludePatterns(Seq("rootdoc.txt"))
```

The default behavior is to include all resources (`**`), excluding `/scala-native/**`. (Note that, scala-native will exclude the resources `/scala-native/**` anyway, even users update the exclude patterns).

Also, note that this featuer is using Java's PathMatcher, which behave a bit different from the posix glob.
https://docs.oracle.com/javase/tutorial/essential/io/find.html


tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -347,6 +365,14 @@ object NativeConfig {
...
     }
 
+    def withResourceIncludePatterns(value: Seq[String]): NativeConfig = {
+      copy(resourceIncludePatterns = value)
+    }
+
+    def withResourceExcludePatterns(value: Seq[String]): NativeConfig = {
+      copy(resourceExcludePatterns = value)
+    }
+
```
WojciechMazur, line 375: We can add variants for appending a single pattern, these might be sometimes easier to use, eg: 
```Scala
 config
  .withResourceIncludePattern("foo")
  .withResourceIncludePatterns(config.resourceIncludePatterns :+ "foo")
```
However, at that time we would need to handle the default `**` pattern. We can do it here by checking if `resourceIncludePatterns eq DefaultIncludePattern` and then replace the sequence, or otherwise append to it.

tools/src/main/scala/scala/scalanative/codegen/ResourceEmbedder.scala
```diff
@@ -28,11 +29,25 @@ private[scalanative] object ResourceEmbedder {
       classpathDirectory: VirtualDirectory
   )
 
+  private final val ScalaNativeExcludePattern: String = "/scala-native/**"
+
   def apply(config: Config): Seq[Defn.Var] = Scope { implicit scope =>
     val classpath = config.classPath
+    val includePatterns =
+      config.compilerConfig.resourceIncludePatterns.map(p =>
+        FileSystems.getDefault().getPathMatcher(s"glob:$p")
```
WojciechMazur, line 38: We can factor out default file system to common val

tools/src/main/scala/scala/scalanative/codegen/ResourceEmbedder.scala
```diff
@@ -49,9 +64,13 @@ private[scalanative] object ResourceEmbedder {
...
 
-              if (isInIgnoredDirectory(path)) {
+              if (shouldIgnore(path)) {
                 config.logger.debug(
-                  s"Did not embed: $pathName - file in the ignored \'scala-native\' folder."
+                  s"Did not embed: $pathName - " +
+                    s"Include patterns: ${config.compilerConfig.resourceIncludePatterns
+                        .mkString(", ")}; " +
+                    s"Exclude patterns: ${(config.compilerConfig.resourceExcludePatterns :+ ScalaNativeExcludePattern)
+                        .mkString(", ")}"
```
WojciechMazur, line 73: This part could be cached as it needs to be computed only once, so it can be appended to `s"Did not embed: $pathName - "` 
But, even better we can refactor it out check which exactly pattern was applied to exclude given path. So we'd have something like this 
```
 val notInIncludePatterns = s"Not matched by any include path ${includePatterns.mkString}"
 type IgnoreReason = String
 def shouldIgnore(path: Path): Option[IgnoreReason] =  
    includePatterns.find(_.matches(path)).fold(notInIncludePatterns){ includePattern => 
     excludePatterns.find(_.matches(path).map(excludePattern => s"Matched by $includePattern by excluded by $excludePattern")
     
  ... 
  shouldIgnore(path) match {
  case Some(reason) => config.logger.debug("Did not embed: $path - $reason")
  case None =>  ??? 
  }

**Review by WojciechMazur — COMMENTED** [review]
Overall looks good, great that you've include additional scripted tests. I have a few comments with possible improvements

**Comment by keynmol** [comment]
No comments, just super excited to see this land because my projects include a ton of useless stuff from classpath :D 

CleanShot 2023-10-11 at 11 52 24@2x

Perhaps in a future PR we can consider _by default_ excluding some known offenders, like rootdoc.txt and LICENSE from scala library jars?

**Comment by ekrich** [comment]
Weird that the `scala-native/libpq.c` would be in that list. I thought that directory was excluded.

File: tools/src/main/scala/scala/scalanative/build/NativeConfig.scala (hunk 1 above)
Comment (tanishiking) [reply tools/src/main/scala/scala/scalanative/build/NativeConfig.scala:375]: I personally feel having two APIs: "appending a single pattern" and "replace the patterns with the given seq" seems confusing, and even complicated with the special handling with replacing the default pattern.
Maybe we can compromise with the API that takes varargs instead of Seq so users can write something like `.withResourceIncludePatterns("foo")`?

Comment (WojciechMazur) [reply tools/src/main/scala/scala/scalanative/build/NativeConfig.scala:375]: Let's stay with the current solution then using Seq, that way it would be aligned with remaining settings

File: tools/src/main/scala/scala/scalanative/codegen/ResourceEmbedder.scala [hunk 4]
```diff
@@ -28,11 +29,38 @@ private[scalanative] object ResourceEmbedder {
...
+    /** If the return value is defined, the given path should be ignored. If
+     *  it's None, the path should be included.
+     */
+    def shouldIgnore(path: Path): Option[IgnoreReason] =
+      includePatterns.find(_.matches(path)).fold(Option(notInIncludePatterns)) {
+        includePattern =>
+          excludePatterns
+            .find(_.matches(path))
+            .map(excludePattern =>
+              s"Matched by $includePattern by excluded by $excludePattern"
```
WojciechMazur, line 60: Typo here? Also it might be worth to put patterns in quotes, so it's easier to spot empty pattern or containing whitespaces.  
```suggestion
              s"Matched by '$includePattern', but excluded by '$excludePattern'"
```

WojciechMazur (commented):
Looks good, I've just spotted as single typo

tools/src/main/scala/scala/scalanative/codegen/ResourceEmbedder.scala
WojciechMazur: Also I don't think we have a test case which would trigger this path

tanishiking: https://github.com/tanishiking/scala-native/blob/0f99a2a2861e5aa647595469d6ea0e80181bae73/scripted-tests/run/resource-embedding/build.sbt#L78-L79

Actually, in project G, include pattern matches with `exclude/b.txt` but it will be excluded because of `exclude/b.txt`.
I updated the comments for that tests https://github.com/scala-native/scala-native/pull/3562/commits/1b9c8defaedf298b24374b07115313a665c7e83b

WojciechMazur (approved):

TonioGela:
😍 can't wait to see it release and ported into scala-cli as well

PR #3574, opened 2023-10-17
Support for Java Service Providers via linktime resolved `java.util.ServiceLoader.load`

WojciechMazur:
In the light of recent discussions how to implement custom FileSystems (#3453 ) or possibly JDBC Drivers I want to discuss possible usage of Java ServiceProviders model and it's possible implementation for Scala Native. 

Why introduce new feature instead of using `scala.scalanative.reflect`? 
Current support for reflective instantiation of classes or loading of modules is currently widely used in different test frameworks in both Scala.js and Scala Native. However, one their limitation is eager dependency injection in the class path and though final binary produced by Scala Native. Each class marked with `@EnabledReflectiveInstantization` available on the classpath would be always loaded and processed in the toolchain, even if it is never used. In case of tests it's fine as we don't bother about increased size and compilation time that much. However, in case of custom file systems we might want to control when given provider shall be available. It's especially important in case if the service implementation contains a native dependencies. Usage of `@EnabledReflectiveInstantization` in the library would force dependency injection on the native library even if it is never used. 
Moreover to allow instantiation of class using `scala.scalanative.reflect` we need to have it's fully qualified names. It means that we still need to introduce a mechanism allowing to select which service provider shall be available. In the users code, we don't want to force knowledge about actual implementation of given service, like file system, it's fully qualified name or access modifiers. 

Proposed solution 
`java.util.ServiceLoader` is a solid framework for introducing dependency injection for service providers based on the `META-INF/services` resource files describing available implementation of given service provided by given jar. The access to the implementation of given class is available via `java.util.ServiceLoader.load(Class[_])`. 
To allow for linktime resolution of all available service providers via `java.util.ServiceLoader.load` we introduce a linktime intrinsics. Similarly as at the compile time we are able to modify behaviour of our code, we can add special handling for defined set of functions. 
When class loading upong finding a call to stubbed method `java.util.ServiceLoader.load` we would replace this call with allocation of `ServiceLoader` instance prepopulated with available implementation of given interface based on information gathered from `META-INF/services` and available symbols on the classpath. This should allow to reference them by the user - see `java.nio.charset.Charset` example of usage for `CharsetProvider`

Differences from the Java Service Providers 
The Java implementation allows for dynamic classloading of providers, which for abvious reasons is not possible in Scala Native during AoT compilation. To allow for selective disabling of unused providers found on the classpath we introduce an entry in `NativeConfig.disabledServiceProviders`. Each service provider fully qualified name added in this config entry would by skipped durring classloading. This means it would never be classloaded at the first place, unless it would be referenced by any other mechanism. This should allow to reduce amount of unwanted dependencies and the size of the final binary. The user would be notified when given entry is ignored (at debug logging level), when it's included (at info level) or when given provider defined in `META-INF/services/x` is not available on the classpath. 

Before finalising work on the service providers I'd like to hear opinions about this proposal. Especially the library authors or users of Java ServiceProviders on the JVM.


tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -208,6 +211,9 @@ sealed trait NativeConfig {
   /** Create a new [[NativeConfig]] with updated resource exclude patterns. */
   def withResourceExcludePatterns(value: Seq[String]): NativeConfig
 
+  /** Create a new [[NativeConfig]] with a updated list of service providers excluded from the final binary */
+  def withDisabledServiceProviders(value: Seq[String]): NativeConfig
```
WojciechMazur, line 215: Possibly it can be `Map[String, Seq[String]]` defined as `Map[ServiceName, ProviderPattern]` allowing to use regex/glob patterns for exclusions, and possibly to disable whole not interesing services, eg. `Map("java.charset.spi.CharsetProvider" -> Seq(".*"))`
However it would be more verbose

sandbox/src/main/scala/Test.scala
```diff
@@ -1,5 +1,30 @@
...
+  }
+  override def charsetForName(charsetName: String): Charset = null
+}
+
 object Test {
   def main(args: Array[String]): Unit = {
-    println("Hello, World!")
+    val loader = ServiceLoader.load(classOf[CharsetProvider])
+    println(loader)
+    loader.forEach(println)
```
WojciechMazur, line 28: Would print `MockCharsetProvider` instance only, `DisabledCharsetProvider` was disabled in the config

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
armanbilge: Why opt-out instead of opt-in? e.g. for the `ZipFileSystemProvider` seems more like an opt-out sort of thing.

LeeTibbert:
I think this approach is promising. Certainly better than 20 or so additional `LinktimeInfo` categories,
I need to study it more.

At the risk of embarrassing myself in front of my elders and betters, let me test out my current understanding:

1) There would be a built in list of known service provider classes (`META-INF/services`). For a concrete discussion
    say SN started with "charset" and "filesystem".  If someone wanted to add a (mostly hypothetical) "crypto" 
    service provider, they would have to modify SN build to add that provider to `META-INF/services` and
    ensure that the implementing code was on the classpath at link time (as a dependency?)

    An application builder would have to "opt-in" to pickup that Provider (probably good idea).

2) Now say an application builder wanted to switch the implementing of a built in code provider, say from   
    LeesCharsetProvider or SnDefaultCharsetProvider to ArmansCharsetProvider. They would have to ensure
   (I have not worked out the details yet.) that latter was earlier on the class path than the former two. 
   SnDefaultCharsetProvider would probably always be on the CP, and pretty early at that.

3) A given Service Provider could decide to implement, perhaps as an option,  actually dynamically load a suitable
    sharable library instead of having static code.  The gives great utility and flexibility, especially in the 3rd party
    case, but gets into the deep end of complexity. I am told that shareable libraries tend to be problematic
    on Windows.

4) IIUC, if implementations are static, that means that the size of the, say, javalib artifact provided by
    SN will increase when zipfs is implemented and then again when a hypothetical jrt is implemented.
    I have not yet measured the incremental size of zipfs, but suspect it will not be small.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
WojciechMazur: Good point. I don't yet have a strong opinion about making it opt-in or opt-out.
My initial idea was that if you have something on classpath you probably would like to use it, and you'd catch the redundant implementations from the logs and possibly eliminate them. But at the same time we kinda go back to issues of `@EnableReflectiveInstantization`
Going opt-in instead of opt-out would probably move most of the usage issues from compile time (missing deps, failed linking) to runtime (exception when there is no suitable file system providers for zip files). 
Both approaches have in my opinion some benefits and some issues. In each case a good communication with the user via logs would be a priority, especially in area of reporting which providers are available, used or discarded.

WojciechMazur:
@LeeTibbert  Thank you for comments, here's some answers for the questions

>  If someone wanted to add a (mostly hypothetical) "crypto"
service provider, they would have to modify SN build to add that provider to META-INF/services and
ensure that the implementing code was on the classpath at link time (as a dependency?)

There typically should be no need to modify SN build. In case of built-in services like FileSystems or Charsets we would in fact need to modify their code, to allow for discovery of custom extensions (custom fileSystems or charsets), just like in the example for the Charsets, where we use `ServiceLoader` to lookup for additional implementations. 
For other services, like `java.sql.Driver` which probably would not be included in the core repo, but rather implemented as 3rd-party library we can put the `ServiceLoader` usage in that library without modification of core Scala Native build.

`META-INF/services/<service-name>` are only distributed by the dependencies. They're a meta information file in jar, and contain a list of implementation for given service. For example `META-INF/services/foo.Bar` would be a plain file containing fully qualified names of classess implementing `foo.Bar` interface in this library (there can be more then one implementation provided within a single jar/dependency). 
You, as a potential library creator, are required to ensure that the information in `META-INF/services/<service-name>` of your library is correct, since it's the only way to efficently discover the available implementations.
Multiple libraries can provide their own providers, let's say for `java.nio.charset.spi.CharsetProvider`. They do not clash with each other. 

In most cases we would like to follow the JVM conventions whenever possible (see [ServiceLoader docs](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/ServiceLoader.html)), but in some cases we might be more strict due to SN limitations, eg. `ServiceLoader.laod` needs to use `classOf[X]` to provide a compile-time known const, instead of some arbitrary instance of a class, possibly passed as a parameter. 

> Now say an application builder wanted to switch the implementing of a built in code provider ... They would have to ensure
 that latter was earlier on the class path than the former two.
 
The custom service providers should be treated more like extension with additional capabilities, instead of single custom implementation. They don't override each other. Their order on the classpath does not matter. We typically would still have some default service providers providing default capabilities, like Unix/Windows FileSystems, or UTF-8/16/16LE/16BE charsets. These can be extended via custom ones.
Most of interfaces implemented as ServiceProviders are typically containing some way of runtime dispatch allowing to use the custom providers. For example, in case of FileSystems dispatch is based on the `getScheme(): String` which identifies which kind of files given service provider can operate. These are used later when via `FileSystem.getFileSystem(uri: URI)`. For charset providers on the other hand custom provider allows to register custom aliases, which latter can be referenced via `Charset.forName("name-of-my-charset")`

> A given Service Provider could decide to implement, perhaps as an option, actually dynamically load a suitable
sharable library instead of having static code 

That's correct, usage of dynamically loaded or static library is up to choose of implementor of given provider. In case if given provider is never referenced, either explicitl, via `ServiceLoader.load` or when it was explcitly discarded in the config, it would not be class loaded, so it would add additional `@link` dependencies or would not execute a code required for dynamic loading of library. 

> IIUC, if implementations are static, that means that the size of the, say, javalib artifact provided by
SN will increase 

The `javalib` artifact (a jar) would only increase by the size of it's additional .class/.nir/.tasty files. The final size of linked executable would depend, based on the fact if the given service would be include or not (see previous point)

WojciechMazur:
I've modified the native config entry to now make it selection opt-in instead of opt-out. To make checking which providers are available I've extended the logging capability 
Screenshot 2023-10-19 at 20 07 51

I'm wondering how to deal with missing entries in the `META-INF/services`. Right now logging/reachability ananalysis only tries o load entries listed in services. In theory it should be responsibility of the service creator to introduce corect entries in META-INF/services, but maybe we can help in case of typos. 
One possible improvement might be checking if some type which is implementing some `Service` was classloaded, but is missing from providers list. Altough I don;t think that should be a priority right now

LeeTibbert:
Good to see progress happening here.   I have to read through the code and the Java ServiceProvider URL provided.
That is going to take time and this PR is moving rapidly.

At the risk of being politely asked to "Do your homework, the answer is obvious", let me try to cut through the
time consuming part and ask the question that concerned me in Issue #3453.  

```
A straight forward, sight unseen, implementation of zipfs adds a line of code to
javalib/src/main/scala/java/nio/file/spi/FileSystemProvider.scala:

object FileSystemProvider {
  def installedProviders: List[FileSystemProvider] = {
    val list = new LinkedList[FileSystemProvider]
    if (isWindows)
      list.add(new WindowsFileSystemProvider())
    else
      list.add(new UnixFileSystemProvider())                 
                                   
// >>>     list.add(new jdk.nio.zipfs.ZipFileSystemProvider()) // Support ZipFileSystem               
    list
  }
```

Can  the `installedProviders` method be implemented by the code in this PR?

The `ServiceProvider` object has methods to load() but I did not see any methods to 
runtime or link time query which providers  are available without loading them.

Say zipfs was on the classpath at compile time and otherwise unused at link or runtime,
when I call `FileSystemProvider#installedProviders`, I want `jar` in the list of names but
I do not want to actually load the code.

I am off to look at the Java ServiceProvider description while I wait for an hours long
network test to run.

LeeTibbert:
As a flourish for a future evolution:

`ServiceLoader.scala` in this PR has a TODO comment.
```
 val builder = ju.stream.Stream.builder[ServiceLoader.Provider[S]]()
    // TODO: the stream shall be lazy, instanization of provider should be delayed until `get()`
   
```

`javalib/src/main/scala/java/util/stream/StreamSupport.scala`
has a method:
``` def stream[T](
      supplier: Supplier[Spliterator[T]],
      characteristics: Int,
      parallel: Boolean
  ): Stream[T] = {
    new StreamImpl[T](supplier, characteristics, parallel)
  }
```

The `ServiceLoader` class has an iterator method defined.  It is possible to go from an
iterator to a ` Supplier[Spliterator[T]]`. I believe there are examples in `unit-tests`.
The Stream  returned will not do any work (will be lazy) until a given element is accessed.
I think that is the intent of that TODO.   

The `supplier[Spliterator[T]] is a way to (semi) gurantee that the stream does not bind to the
source until first use of the stream. Any reasonable stream based on Spliterators will be lazy after that.

In order to match the Java 8 `ServiceLoader` description, the `ServiceLoader#iterator` would need to
be modified to return first cached service providers of that class, if any, then new providers, possibly loaded().

WojciechMazur:
> Can the installedProviders method be implemented by the code in this PR? 

Yes, but it would work a bit differently: 
```scala
object FileSystemProvider {
  def installedProviders: List[FileSystemProvider] = {
    val list = new LinkedList[FileSystemProvider]
    // load default providers, these are always available
    if (isWindows)  list.add(new WindowsFileSystemProvider())
    else list.add(new UnixFileSystemProvider())                 
    
   // This would add dependency on ZipfileSystem to every build                               
   //  list.add(new jdk.nio.zipfs.ZipFileSystemProvider()) // Support ZipFileSystem   
   
    // Instead use ServiceLoader to load custom providers found on classpath and allowed by the native config 
    java.util.ServiceLoader.load(classOf[FileSystemProvider]).forEach(list.add)            
    list
  }
```

The only way currently to query available services is iterating over `ServiceLoader[S]` which is an `Iterator[S]`, it would currently create eagerly new instances. Eventually we can would like to implement lazy `ServiceLoader[S].stream()` which allows to access to elements `type: Class[_ <: S]` to check if given provider is interesting for us, and `get(): S` which returns an instance of service. 

Based on Arman's comments, we now configure providers which can be queried in an opt-in manner. So for the zip filesystem we might have entry in nativeConfig 
```scala
nativeConfig ~= _.withServiceProviders{ Map( 
 "java.nio.file.spi.FileSystemProvider" -> Seq(
      "jdk.nio.zipfs.ZipFileSystemProvider"
    )
 )}
``` 
Even if we would have multiple FileSystemProvider implementation only this single one would be returned by `ServiceLoader.load(classOf[java.nio.file.spi.FileSystemProvider])`

Becouse the AheadOfTime nature of our builds, typically it should really matter if the loading would be done lazily or not. In most cases there would be only 1 `load[T]` operation per given type, which would register all the available implementations at the component requiring them, eg. `java.nio.file.FileSystemProvider.installedProviders` which actually should be (lazy) val .


> Say zipfs was on the classpath at compile time and otherwise unused at link or runtime,
when I call FileSystemProvider#installedProviders, I want jar in the list of names but
I do not want to actually load the code.

By "load the code" do you mean  "initialize the `ZipFileSystemProvider`"?  If so, when for example the provider would require initialization/ dynamic loading the native library then in the `ZipFileSystemProvider` as a form of optimization you can delay it until the first actual usage, via lazy val or other measures. Otherwise it would not be possible. Just to register support for `jar` in the FileSystems we need to access the `ZipFileSystemProvider.getSchema()` which would require us to create an instance of it. We cannot explicitlly refer to `ZipFileSystemProvider` instance, eg. to it's companion object as it would lead to class loading additional symbols.

LeeTibbert:
Sorry, I do not follow at all.  Even for Java, this seems overly complex and burdensome.
If you say that  `def installedProviders: List[FileSystemProvider]` can be provided, I will believe you.

For `FileSystemProvider` to create the list of installedProviders at compile time it
seems like the `javalib` build would need to read the NativeConfig to determine all the
providers known to the configuration, and 'know` that `mumble.zipfs` has a text name of `jar`
(and other such magic) and then write out a `foo.scala` which FileSystemProvider queries. 
This would mix all kinds of providers. `FileSystemProvider` would have to "know" about
"zip", "jar", and eventually "jrt" and skip everything else.

LeeTibbert:
```
 Just to register support for jar in the FileSystems we need to access the ZipFileSystemProvider.getSchema() which would require us to create an instance of it. 
```

_Exactly_ my concern!

LeeTibbert:
While waiting for a long `test-scripted` to complete (is there such a thing as a short `test-scripted`?)

Perhaps the "At runtime try{} to load a zipfs provider. If it succeeds, then add 'jar' to the installedProviders list,
if it fails, then leave `installedProviders` unchanged is not as hideous as it sounds, even with upcoming Halloween.

Yes, on success, the code in the zipfsProvider would be in the executable and would not be eligible for Dead Code Elimination. But, IIUC, the code which the zipfsProvider calls would be eligible for DCI if no methods in zipfsProvider
where used by the application.  My concern is that `java.util.zip` is a _lot_ of code to carry & ship around if it is
never used.

I would have to find the size of zipfsProvider. Perhaps the size of the unixFileSystemProvider would
be a first approximation/goal.

Food for thought.

docs/lib/javalib.rst
```diff
@@ -790,4 +790,32 @@ network is first used.
...
+
+All providers of service referenced by ``java.util.ServiceLoader.load`` that were reached from any of entrypoints, would be enlisted when linking.
+The providers might have 1 out 5 available statuses: 
+  * ``Loaded`` - this provider was allowed by the config and found on the classpath. It would be available at runtime.
+  * ``Available`` - this provider was found on classpath, but it was not enlisted in the config. It would not be available at runtime.
+  * ``UnknownConfigEntry`` - provider enlisted in config was not found on classpath. It might suggest typo in configuration or in ``META-INF/servies`` file.
+  * ``NotFoundOnClasspath`` - given provider was found both in config and in ``META-INF/services`` file, but it was not found on classpath. It might suggest that given provider was not cross-compiled for Scala Native.
+  * ``NoProviders`` - status assigned for services without any available implementations found on classpath and without config entries
+
+
```
LeeTibbert, line 820: Good description.

I think what is also needed is for a way to _exclude_ a provider which might be in any (or, better yet, a specified)
library `META_INF/services`.

That is, "do not bother me with the cost of stuff I know I do not want."

A use case:   Envision an application "LeanApp" where the developer of "LeanAndMeanApp" knows for certain that the zipfs is
never used by "LeanAndMeanApp".

I believe that such a developer should be able to say that
the `jar` provider known to be in javalib `META-INF/services`
should not increase application code size or runtime initialization time.

LeeTibbert:
I've spent time both actively considering a lean "installedProviders" implementation and letting
such an implementation sit on my back brain.

I think I will let this PR flow through and try a sample implementation afterwards.   My current thinking
is to start that effort by doing a design study to see if a `LinktimeInfo` can be created with a parameter
which would be the class name given to "opt-in".   

A bit of work, but then the specific case of "installedProviders" should be easy to implement without 
handstands & contortions.  That solution would also be extensible.  I strongly suspect that other
prospective providers will be able to use the same to minimize code size and initialization time.

PR #3580, opened 2023-10-22
Make implementation of `java.util.Properties` based on `j.u.c.ConcurrentHashMap`

WojciechMazur:
Make implementation of `java.util.Properties` based on `j.u.c.ConcurrentHashMap` instead of Hashtable. 
When refering to methods of `j.u.Properties` from within it's subclass when using JDK 9 or higher the linker complained about missing methods. This suggests that these methods are overriden in the JDK and are missing in our implementation of `j.u.Properties`
We take the oportunity here to remove dependency of Scala stdlib in `j.u.Properties` 
Override the `java.util.Properties` methods to allow for their static usage in subclasses of `j.u.Properties`

### DIFF
--- a/javalib/src/main/scala/java/util/Properties.scala
+++ b/javalib/src/main/scala/java/util/Properties.scala
@@ -8,10 +8,12 @@ import java.nio.charset.StandardCharsets
 import scala.annotation.switch
 import scala.annotation.tailrec
 
-import ScalaOps._
+import java.util.function.{Function, BiConsumer, BiFunction}
 
 class Properties(protected val defaults: Properties)
     extends ju.Hashtable[AnyRef, AnyRef] {
+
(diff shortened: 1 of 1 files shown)


ekrich:
Looking at the code makes me think we should fix the `ju.Hashtable` to not use Scala `mutable.HashMap` like they did in Scala.js. I don't think it is a good idea to pull `concurrent` sub package into the mix and follow Java as much as possible. Delegating ju.Hashtable to ju.HashMap like they did in Scala.js adding synchronized as appropriate should be fine. ScalaOps has been updated so using that should be fine rather than using the `function` sub package for the `forEach` stuff.

Honestly, I am not seeing all the additional missing methods unless they are in `ju.Dictionary` or other super classes.

PR #3584, opened 2023-10-27
Support universal equality of unsigned numeric types

WojciechMazur:
This change allows to use universal equality for combination of signed and unsigned numeric types. It means we can run comparisons like `2 + 2 == 4.toUInt`, or `-1.toUByte == 255`. 
To allow for that when used in erased context, we modify the NIR generated for `==` and `!=` primitives. These now would delegate to `scala.runtime.BoxesRunTime.equals` instead of delegating to custom `Object.__scala_==` which now becomes obsolete and removed.

All numeric equality comparsions are now available between any of `Byte`, `Short`, `Int`, `Long`, `UByte`, `UShort`, `UInt`, `UByte`, `Size`, `USize` and `Char`.


docs/user/interop.rst
```diff
@@ -513,4 +513,8 @@ using ``byteValue.toUByte``, ``shortValue.toUShort``, ``intValue.toUInt``, ``lon
 and conversely ``unsignedByteValue.toByte``, ``unsignedShortValue.toShort``, ``unsignedIntValue.toInt``,
 ``unsignedLongValue.toLong``, ``unsignedSizeValue.toSize``.
 
+Universal equality is supported between signed and unsigned integers, for example ``-1.toUByte == 255`` or ``65535 == -1.toUShort`` would yield ``true``, 
+However, similary to singned integers on JVM, class equality between different (boxed) integer types is not supported.
```
ekrich, line 517:
```suggestion
However, similar to signed integers on JVM, class equality between different (boxed) integer types is not supported.
```

auxlib/src/main/scala/scala/runtime/BoxesRunTime.scala
```diff
@@ -24,21 +28,53 @@ object BoxesRunTime {
...
+      case _: UByte  => INT
+      case _: UShort => INT
+      case _: UInt   => LONG
+      case _: ULong  => ULONG
+      case _: Size   => if (LinktimeInfo.is32BitPlatform) INT else LONG
+      case _: USize  => if (LinktimeInfo.is32BitPlatform) LONG else ULONG
+      case _         => OTHER
+    }
+
+  // Char is unsigned, we don't need to extended int/long
```
ekrich, line 65:
```suggestion
  // Char is unsigned, we don't need to extend int/long
```
Did you mean this?

ekrich (changes requested):
Wow, this is a great change.

auxlib/src/main/scala/scala/runtime/BoxesRunTime.scala
WojciechMazur: Yes, thank you for catching typos!

PR #3588, opened 2023-10-29
Remove usages of `Object.__scala_##`

WojciechMazur:
Removes `Object.__scala_##` which typically is not used at runtime. Scala compiler already dispatches calls to `Any.##` to dedicated `scala.runtime.Statics.*Hash` or dedicated versions for float/long/double etc. Now we dispatch remaining calls if existing to the `Statics` instead of using custom mechanism


ekrich:
Nice to see these changes. The must increase the compile speed too. :-)

PR #3590, opened 2023-10-31
Add code documentation

kyouko-taiga:
This PR simply adds documentation to classes, methods, and fields related to NIR data structures.

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Defns.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Defns.scala
@@ -2,33 +2,78 @@ package scala.scalanative
 package nir
 import scala.scalanative.nir.Defn.Define
 
+/** A definition in NIR.
+ *
+ *  Programs in NIR are represented as a sequence of definitions denoting types,
+ *  methods and fields. Definitions fall into two categories:
+ *
+ *    - Top-level definitions: these represent classes, modules, or traits.
+ *      Classes and modules inherit from a single parent with the exception of
+ *      `java.lang.Object`, which sits at the top of the hierarchy. They may
+ *      additionally implement traits.
+ *    - Member definitions: these represent fields or methods.
+ *
+ *  Definitions may carry attributes providing further information about their
+ *  semantics.
+ */
 sealed abstract class Defn {
+
+  /** Returns the name of the definition. */
   def name: Global
+
+  /** Returns the attributes of the definition. */
   def attrs: Attrs
+
+  /** Returns the site in the program sources corresponding to the definition.
+   */
   def pos: Position
-  final def show: String = nir.Show(this)
+
+  /** Returns a textual representation of `this`. */
+  final def show: String =
+    nir.Show(this)
+
+  /** Returns `true` iff `this` is considered as an entry point by reachability
+   *  analysis.
+   */
   final def isEntryPoint = this match {
     case Define(attrs, Global.Member(_, sig), _, _, _) =>
       sig.isClinit || attrs.isExtern
     case _ => false
   }
+
 }
 
 object Defn {
-  // low-level
-  final case class Var(attrs: Attrs, name: Global.Member, ty: Type, rhs: Val)(
-      implicit val pos: Position
-  ) extends Defn
-  final case class Const(attrs: Attrs, name: Global.Member, ty: Type, rhs: Val)(
-      implicit val pos: Position
-  ) extends Defn
+
+  // === Member definitions ==== //
+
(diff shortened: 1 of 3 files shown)


nir/src/main/scala/scala/scalanative/nir/Defns.scala
```diff
@@ -2,33 +2,78 @@ package scala.scalanative
...
+/** A definition in NIR.
+ *
+ *  Programs in NIR are represented as a sequence of definitions denoting types,
+ *  methods and fields. Definitions fall into two categories:
+ *
+ *    - Top-level definitions: these represent classes, modules, or traits.
+ *      Classes and modules inherit from a single parent with the exception of
+ *      `java.lang.Object`, which sits at the top of the hierarchy. They may
+ *      additionally implement traits.
+ *    - Member definitions: these represent fields or methods.
```
kyouko-taiga, line 14: Because we also have `Const` declarations, I am not sure this comment is true. Is a constant always describing a member?

nir/src/main/scala/scala/scalanative/nir/package.scala
```diff
@@ -1,13 +1,27 @@
 package scala.scalanative
 
 package object nir {
+
   type LocalName = String
```
kyouko-taiga, line 5: I couldn't document this alias because I don't know what a "local" is.

nir/src/main/scala/scala/scalanative/nir/Defns.scala
WojciechMazur: It depends from the stage.
The `Defn.Var` in the form of `Defn.Var(_, Global.Member(_, Sig.Field(_),_,_)` exists only until lowering, after which the information about the field exists only in the `MemoryLayour` or `FieldLayout` of given type. I don't think we do define `Defn.Const` for `Sig.Field` at all. 

All the remaining `Defn.Var` and `Defn.Const` are basically a global variables/constants. At this point they would typically would be defined a `Member` with `Sig.Extern` or `Sig.Generated`. They can be used for example for keeping the stack bottom of the main thread or other information required at runtime. `Defn.Const` for example is used to create instances of strings or create the instances of `java.lang.Class[T]` (actually I think now we use Var for that)

nir/src/main/scala/scala/scalanative/nir/Defns.scala
```diff
@@ -2,33 +2,78 @@ package scala.scalanative
...
-  final case class Var(attrs: Attrs, name: Global.Member, ty: Type, rhs: Val)(
-      implicit val pos: Position
-  ) extends Defn
-  final case class Const(attrs: Attrs, name: Global.Member, ty: Type, rhs: Val)(
-      implicit val pos: Position
-  ) extends Defn
+
+  // === Member definitions ==== //
+
+  /** A variable definition corresponding to a field in class or module. */
```
WojciechMazur, line 50: ... or a top level global variables.

nir/src/main/scala/scala/scalanative/nir/Defns.scala
```diff
@@ -2,33 +2,78 @@ package scala.scalanative
...
+  final case class Var(
+      attrs: Attrs,
+      name: Global.Member,
+      ty: Type,
+      rhs: Val
+  )(implicit val pos: Position)
+      extends Defn
+
+  /** A constant value. */
```
WojciechMazur, line 59: A constant, but rather in a meaning of unique read-only instance of given type. It's important not not confuse them with constant literal, which exists only as a Val.

nir/src/main/scala/scala/scalanative/nir/package.scala
WojciechMazur: Here the `LocalName` is the name of variable in the source code. The `Local` is the SSA identifier in the scope of given `Defn.Define`, so it can be either a function parameter or an identifier of an instruction. 
To preserve the original name of the variable during transformations of the code, and make it available in the backend for generating debug metadata, each name of variable is paired with the `Local` for the result of mentioned variable 

For example `val x = 2; val y = x + 2`, might be represented as folows, where the `%` is the SSA identifier`, and `<...> is the local name of given identifier
```
 %1 <x> = 2
 %2 <y> = add(%1, 2)
```

kyouko-taiga: Oh wow, I'm surprised that there can be a correspondence so direct between Scala and NIR.

WojciechMazur (approved):

PR #3593, opened 2023-11-01
Use more literal constants to skip unnecessary initialization of pure modules

WojciechMazur:


ekrich:
Can some of the constant integers be saved in a smaller size like Short or smaller?

WojciechMazur:
> Can some of the constant integers be saved in a smaller size like Short or smaller?

Unfortunately it seems to be impossible using standard Scala syntax. Usage of explicit result type or type annotations would not allow to interpret right-hand side as a literal constant. For these the workaround would be a method with inline annotation.

WojciechMazur:
Also, when the 'final val' is a valid literal constant it would not be defined as a class field. Instead at the Scala backend phase it would already be available only as a method. Because of that literal constants don't increase the memory layout of given type.

Issue #3591, opened 2023-10-31
Loading module in multithreading mode should be more aware when init fails


WojciechMazur:
Let's take a look at following code, where we create n threads which shall reference the object module. The 1st thread to access the module would try to initialise it (calculcate `val init` in our case). All the other threads would wait until the object would be initialized

```scala
object Deadlocking {
  val init = {
    println(s"start init by ${Thread.currentThread()}")
    Thread.sleep(1000)
    throw new RuntimeException()
    println("init done")
    42
  }
}

object Test {
  def main(args: Array[String]): Unit = {
    List.tabulate(8) { n =>
        val t = new Thread(() => {
          println(s"start thread $n")
          println(s"thread $n, result=${Deadlocking.init}")
          println(s"thread done $n")
        })
        t.setName(s"thread-$n")
        t
      }
      .tapEach(_.start())
      .foreach(_.join())
  }
}
```

On the JVM when the initialisation would fail, every thread waiting for the `Deadlocking` would fail with an exception: 
```
start thread 0
start thread 1
start thread 2
start thread 3
start thread 4
start thread 5
start init by Thread[thread-0,5,main]
start thread 6
start thread 7
Exception in thread "thread-0" Exception in thread "thread-4" Exception in thread "thread-2" java.lang.ExceptionInInitializerError
        at Test$.$anonfun$1(Test.scala:21)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.RuntimeException
        at Deadlocking$.<clinit>(Test.scala:7)
        ... 2 more
Exception in thread "thread-5" java.lang.NoClassDefFoundError: Could not initialize class Deadlocking$
        at Test$.$anonfun$1(Test.scala:21)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.RuntimeException [in thread "thread-0"]
        at Deadlocking$.<clinit>(Test.scala:7)
        ... 2 more
Exception in thread "thread-6" java.lang.NoClassDefFoundError: Could not initialize class Deadlocking$
        at Test$.$anonfun$1(Test.scala:21)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.RuntimeException [in thread "thread-0"]
        at Deadlocking$.<clinit>(Test.scala:7)
        ... 2 more
[...]

PR #3598, opened 2023-11-06
Make module loader in multithreading mode be more aware when init fails

tanishiking:
https://github.com/scala-native/scala-native/issues/3591

- [x] Catch an exception thrown in `ctor(instance)` (in `module_load.cpp`)
- [x] Store the thrown exception somehow and notify to other threads that module initialization is failed
  - Through `InitializationContext`
  - Waiting (for module initialization) threads will know when the initialization is failed in a thread via `InitializationContext`.
- [ ] Show meaningful error/stacktrace to users
  - Currently, we just re-throw the C++ exception, and die with `libc++abi: terminating due to uncaught exception of type void*`.
  - Maybe we need to store the thrown exception into pointer, and re-throw in Scala world? Potentially related "We potentially can provide a CFuncPtr to the native code which would be used to wrap the exception thrown by the ctor into the ExceptionInInitializerError, which later can be thrown using the scalanative_throw" https://github.com/scala-native/scala-native/issues/3591#issuecomment-1790987537
  - Or, maybe we can somehow pretty-print the error message and stacktrace of the Throwable object underlying of the `ExceptionWrapper` in C++?

```
libc++abi: terminating due to uncaught exception of type void*
[error] java.lang.RuntimeException: Nonzero exit code: 134
[error]         at scala.scalanative.sbtplugin.ScalaNativePluginInternal$.$anonfun$scalaNativeConfigSettings$18(ScalaNativePluginInternal.scala:276)
[error]         at scala.scalanative.sbtplugin.ScalaNativePluginInternal$.$anonfun$scalaNativeConfigSettings$16(ScalaNativePluginInternal.scala:276)
[error]         at scala.scalanative.sbtplugin.ScalaNativePluginInternal$.$anonfun$scalaNativeConfigSettings$16$adapted(ScalaNativePluginInternal.scala:253)
[error] stack trace is suppressed; run last sandbox3 / Compile / run for the full output
[error] (sandbox3 / Compile / run) Nonzero exit code: 134
[error] Total time: 8 s, completed Nov 6, 2023, 5:12:09 PM
```

### DIFF
--- a/nativelib/src/main/resources/scala-native/eh.cpp
+++ b/nativelib/src/main/resources/scala-native/eh.cpp
@@ -1,4 +1,5 @@
 #include <exception>
+#include "eh.h"
 
 // Scala Native compiles Scala's exception in C++-compatible
 // manner under the hood. Every exception thrown on the Scala
@@ -6,14 +7,6 @@
 // ExceptionWrapper-based exceptions can be caught by
 // Scala code. We currently do not support catching arbitrary
 // C++ exceptions.
-
-namespace scalanative {
-class ExceptionWrapper : public std::exception {
-  public:
-    ExceptionWrapper(void *_obj) : obj(_obj) {}
-    void *obj;
-};
-} // namespace scalanative
 
 extern "C" {
 void scalanative_throw(void *obj) { throw scalanative::ExceptionWrapper(obj); }
--- a/nativelib/src/main/resources/scala-native/eh.h
+++ b/nativelib/src/main/resources/scala-native/eh.h
@@ -1,0 +1,9 @@
+#include <exception>
+
+namespace scalanative {
+class ExceptionWrapper : public std::exception {
+  public:
+    ExceptionWrapper(void *_obj) : obj(_obj) {}
+    void *obj;
+};
+}
--- a/nativelib/src/main/resources/scala-native/module.h
+++ b/nativelib/src/main/resources/scala-native/module.h
@@ -1,0 +1,4 @@
+#include <stdatomic.h>
+typedef _Atomic(void **) ModuleRef;
+typedef ModuleRef *ModuleSlot;
+typedef void (*ModuleCtor)(ModuleRef);
--- a/nativelib/src/main/resources/scala-native/module_load.cpp
+++ b/nativelib/src/main/resources/scala-native/module_load.cpp
@@ -1,4 +1,9 @@
 #ifdef SCALANATIVE_MULTITHREADING_ENABLED
+
+#include "eh.h"
+
+extern "C" {
+#include "module.h"
 #include "stdatomic.h"
 #include "gc/shared/ScalaNativeGC.h"
 
@@ -19,6 +24,7 @@
 #endif
 
 #include <assert.h>
+} // extern "C"
 
 // Thread identity helpers
 #ifdef _WIN32
@@ -50,7 +56,7 @@ static void sleep_ms(int milliseconds) {
     struct timespec ts;
     ts.tv_sec = milliseconds / 1000;
     ts.tv_nsec = (milliseconds % 1000) * 1000000;
-    nanosleep(&ts, NULL);
+    nanosleep(&ts, nullptr);
 #else
     if (milliseconds >= 1000)
         sleep(milliseconds / 1000);
@@ -58,24 +64,24 @@ static void sleep_ms(int milliseconds) {
 #endif
 }
 
-typedef _Atomic(void **) ModuleRef;
-typedef ModuleRef *ModuleSlot;
-typedef void (*ModuleCtor)(ModuleRef);
 typedef struct InitializationContext {
     thread_id initThreadId;
     ModuleRef instance;
+    scalanative::ExceptionWrapper *exception;
 } InitializationContext;
 
 inline static ModuleRef waitForInitialization(ModuleSlot slot,
                                               void *classInfo) {
     int spin = 0;
     ModuleRef module = atomic_load_explicit(slot, memory_order_acquire);
-    assert(module != NULL);
+    assert(module != nullptr);
     while (*module != classInfo) {
         InitializationContext *ctx = (InitializationContext *)module;
-        // Usage of module in it's constructor, return unitializied instance
         if (isThreadEqual(ctx->initThreadId, getThreadId())) {
             return ctx->instance;
+        }
+        if (ctx->exception != nullptr) {
+            throw ctx->exception;
         }
         if (spin++ < 32)
             YieldThread();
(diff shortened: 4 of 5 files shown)


tanishiking:
## Show meaningful error/stacktrace to users
- ~Currently, we just re-throw the C++ exception, and die with `libc++abi: terminating due to uncaught exception of type void*`.~
- ~Maybe we need to store the thrown exception into pointer, and re-throw in Scala world? Potentially related "We potentially can provide a CFuncPtr to the native code which would be used to wrap the exception thrown by the ctor into the ExceptionInInitializerError, which later can be thrown using the scalanative_throw" https://github.com/scala-native/scala-native/issues/3591#issuecomment-1790987537~
- ~It should be around [here](https://github.com/scala-native/scala-native/blob/644fbd5bb9d40329cf6587bb782dec9dd49aeeaa/tools/src/main/scala/scala/scalanative/codegen/Generate.scala#L429-L455) and can pass something like `val exceptionSlot = Val.Local(fresh(), Type.Ptr)` to `__scalanative_loadModule` but should we re-throw in NIR level? -> oh, wait we can just call Scala function from NIR...~
- **update** I talked with @WojciechMazur and we would optimize error handling during initialization by introducing a custom method within `codegen.Generate` that takes an object, creates an initialization error instance, and throws it.
- https://github.com/scala-native/scala-native/pull/3598/commits/9e1f8e29fdfde843657fe3304d0b0687ad1e8bb8 generates a function that calls another Scala function that throws `ExceptionInInitializerError` in `codegen.Generate`. Now `module_load.cpp` calls that generated function.
  - However, we now get the following runtime error.
  - Since we didn't see [throw exception in initializer error](https://github.com/scala-native/scala-native/pull/3598/commits/9e1f8e29fdfde843657fe3304d0b0687ad1e8bb8#diff-7e51c3171d2c97d99351427de13c1c20d969cd3f689a234494d17e772dd07986R164), it seems we didn't even reach the `throwExceptionInInitializerError` function defined in Scala 🤔 

```
❯ lldb ./sandbox/.3/target/scala-3.3.1/sandbox
(lldb) target create "./sandbox/.3/target/scala-3.3.1/sandbox"
Current executable set to '/Users/tanishiking/ghq/github.com/tanishiking/scala-native/sandbox/.3/target/scala-3.3.1/sandbox' (arm64).
(lldb) run
Process 93413 launched: '/Users/tanishiking/ghq/github.com/tanishiking/scala-native/sandbox/.3/target/scala-3.3.1/sandbox' (arm64)
start thread 0start thread 0start thread 3
start init by Thread[5,thread-3,5,main]

tart init by Thread[5,thread-3,5,main]d 5
start thread 2
start thread 4
start thread 6

t

tart thread 6start thread 1
start thread 7
Process 93413 stopped
* thread #7, stop reason = signal SIGBUS
    frame #0: 0x00000001001c44ac sandbox`scalanative_gc_safepoint_poll at ImmixGC.c:149:20
   146      MutatorThread_switchState(currentMutatorThread, state);
   147  }
   148  void scalanative_gc_safepoint_poll() {
-> 149      void *pollGC = *scalanative_gc_safepoint;
   150  }
   151
   152  void scalanative_add_roots(void *addr_low, void *addr_high) {

(lldb) bt
* thread #7, stop reason = signal SIGBUS
  * frame #0: 0x00000001001c44ac sandbox`scalanative_gc_safepoint_poll at ImmixGC.c:149:20
    frame #1: 0x00000001001c6cd4 sandbox`waitForInitialization(slot=0x0000000100323940, classInfo=0x0000000100325d00) at module_load.cpp:94:9
    frame #2: 0x00000001001c6bc4 sandbox`::__scalanative_loadModule(slot=0x0000000100323940, classInfo=0x0000000100325d00, size=24, ctor=(sandbox`_SM12Deadlocking$RE at Test.scala)) at module_load.cpp:134:16
    frame #3: 0x0000000100050d20 sandbox`_SM12Deadlocking$G4load at Test.scala:1:8
    frame #4: 0x0000000100050fd8 sandbox`_SM5Test$D10$anonfun$1iuEPT5Test$(n$1=5) at Test.scala:16:41 [opt]
    frame #5: 0x00000001000518cc sandbox`_SM15Test$$$Lambda$4D3runuEO at Test.scala:18:10 [opt]
    frame #6: 0x0000000100140780 sandbox`_SM16java.lang.ThreadD3runuEO(this=0x00000002c000e280) at Thread.scala:247:31 [opt]
    frame #7: 0x000000010002bad8 sandbox`_SM39scala.scalanative.runtime.NativeThread$D16threadEntryPointL38scala.scalanative.runtime.NativeThreaduEPT39scala.scalanative.runtime.NativeThread$(this=0x0000000100321248, nativeThread=0x00000002c0040c90) at NativeThread.scala:133:19
    frame #8: 0x000000010002c578 sandbox`_SM39scala.scalanative.runtime.NativeThread$D24threadRoutine$$anonfun$1L28scala.scalanative.unsafe.PtrL28scala.scalanative.unsafe.PtrEPT39scala.scalanative.runtime.NativeThread$(arg=<unavailable>) at NativeThread.scala:121:36 [opt]
    frame #9: 0x000000010002c5ec sandbox`_SM49scala.scalanative.runtime.NativeThread$$$Lambda$1G17$extern$forwarder at NativeThread.scala:117:71 [opt]
    frame #10: 0x00000001001c4448 sandbox`ProxyThreadStartRoutine(args=0x0000600001ac0010) at ImmixGC.c:113:5
    frame #11: 0x0000000182ba7034

```

tanishiking:
Talked with @WojciechMazur. The `SIGBUS` (or SIGSEGV) signals are deliberately triggered to create safepoints for GC (by attempting to access `0x0` that's what `scalanative_gc_safepoint_poll()` does).
Therefore, debugger naturally gonna pause on the `scalanative_gc_safepoint_poll()`, and it's not the real problem here. In `lldb` we can ignore `SIGBUS` by `pro hand -p true -s false SIGBUS`.

We tried to re-throw an exception (in Scala, calling from `module_load.cpp`) but we got the following error and died with segmentation fault.

```
Unexpected signal 10 when accessing memory at address 0x104c08588
        at StackTrace_PrintStackTrace
        at SafepointTrapHandler
        at _sigtramp
zsh: segmentation fault  
```

(My two cent: this is probably caused because the memory address for [`module` in waiting threads](https://github.com/scala-native/scala-native/blob/644fbd5bb9d40329cf6587bb782dec9dd49aeeaa/nativelib/src/main/resources/scala-native/module_load.c#L85) are freed by the main thread (that initialize the module ... and fails) when it failed, and those waiting threads end up accessing freed memory addresses?)

Anyway, https://github.com/scala-native/scala-native/pull/3598/commits/24d5ad02d6b276dae69f3ff00973fda5ef65472e has kind of improvement

- Now, both `waitForInitialization` and `__scalanative_loadModule` returns `nullptr` when initialization failed. The thread initialize the module will notify to other threads by
  - Catch an `exception` thrown by `ctor(instance)`
  - Then assign `nullptr` to `module` and return `nullptr`.
  - Other threads will read the `module` and when it is `nullptr`, they interpret that the module initialization is failed. (it should points to `InitializationContext` while waiting for module initialization).
  - Then those threads also will return `nullptr`.
- If the return value of `load` was `nullptr`, the generated code in `genModuleOp` will throw an exception.
- This exception will be caught by https://github.com/scala-native/scala-native/blob/644fbd5bb9d40329cf6587bb782dec9dd49aeeaa/nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala#L133-L141

While this will add up a bit more instructions to the generated code (return value check for every module initialization), it anyway works.

We need some more improvements

- [ ] Throw `NoClassDefFoundError` or `ExceptionInInitializerError` instead of `NullPointerException` (which is thrown for a time being).
  - In JVM, the main thread throws `ExceptionInInitializerError` while others throw `NoClassDefFoundError`.
  - Those exception should have an original exception thrown while module initialization in `cause` field

Issue #3597, opened 2023-11-06
Add integration with JVM profiler 


WojciechMazur:
It would be really helpfull in terms of build times improvements, to have easy access to profiling info from within the sbt. We should try to create integration to async-profiler or other available profiler allowing for automatic collecting raports when starting `nativeLink` command (or a dedicated command which would call it) and would aggregate the results (prefered flamegraph or details with execution times) when upon finish or abort of `nativeLink` command

tanishiking:
That sounds nice! The introduction of the [ap-loader](https://github.com/jvm-profiling-tools/ap-loader) should significantly simplify the integration of async-profiler. (it was previously [a bit cumbersome](https://mostlynerdless.de/blog/2022/11/21/ap-loader-a-new-way-to-use-and-embed-async-profiler/)  😅) 

This new loader potentially enables us to profile specific spans of the Scala-Native process by using its [Java APIs](https://github.com/jvm-profiling-tools/ap-loader#usage-in-java-code).

(Slightly off-topic: Perhaps we could extend this concept and develop an sbt plugin to streamline the profiling of sbt tasks using async-profiler in the future.)

tanishiking:
Since [jmh has an integration with async-profiler](https://github.com/sbt/sbt-jmh#using-java-flight-recorder--async-profiler), 

Instruction:

- Download the DLL for `async-profiler` from https://github.com/async-profiler/async-profiler#download
- Enable to JMH to load `build/libasyncProfiler.so` (or `.dylib` on macos), by copying it to `java.library.path` or setting an `LD_LIBRARY_PATH`, you will see an instruction if JMH failed to load the DLL. 
  - (off topic: we can improve JMH to run async-profiler out of the box using ap-loader 😄 )
- Run benchmark with `-prof async:output=flamegraph`

```
 [error] Profilers failed to initialize, exiting.
 [error] Unable to load async-profiler. Ensure asyncProfiler library is on LD_LIBRARY_PATH (Linux), DYLD_LIBRARY_PATH (Mac OS), or -Djava.library.path. Alternatively, point to explicit library location with -prof async:libPath=<path>.
 [error] no asyncProfiler in java.library.path: /Users/tanishiking/Library/Java/Extensions:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java:.
```

```sh
sbt:scala-native> toolsBenchmarks2_13 / Jmh / run -prof async:output=flamegraph
[...]

PR #3602, opened 2023-11-10
Integrate with AsyncProfiler

tanishiking:
https://github.com/scala-native/scala-native/issues/3597

This PR will add `nativeLinkProfiling <output-type>` task to `MyScalaNativePlugin` that is a wrapper for `nativeLink` with an [async-profiler](https://github.com/async-profiler/async-profiler) using [ap-loader](https://github.com/jvm-profiling-tools/ap-loader)

While this will generate a bit biased profiling result (because it doesn't warm up JVM for benchmark/profile as JMH does), but still we'll have something useful :)

TODO (which might we should work on in another PR)
- [x] Enable to pass more arguments to profiler
  - ~Currently, we can configure the output-type of the async-profiler while other options are hard-coded or default.~
  - ~It would be nice if we can pass more configuration to async-profiler~
  - We can pass the arguments to async-profiler e.g. `nativeLinkProfiling events=cpu,interval=10000000,threads flamegraph`
  - see: https://github.com/tanishiking/scala-native/blob/03fb96a1743dce6f9b64188a74e8e0ac1150e594/project/MyScalaNativePlugin.scala#L78-L83
- **Better argument parser?**
  - current implementation doesn't have fancy argument parser, it just passes the given arguments to async-profiler. This is because this implementation is just only for the internal (scala-native/scala-native) use, and I didn't see it worth implementing a nicer argument parser for that.
  - If we'd like to have better argument parser, I'd like to parse it using `scopt` or something, but leaving it for now.
- [x] (optional) Better profiler state management
  - I found we hit this path https://github.com/jvm-profiling-tools/ap-loader/blob/c1c2d917c4368008230ad962089c72484ad1cc67/src/main/java/one/profiler/AsyncProfilerLoader.java#L716-L727 when we run `nativeLinkProfiling' reload` and re-run `nativeLinkProfiling`
  - This is because there's the restriction that a native library can only be loaded in one class loader. It seems old sbt session doesn't free the old classloader 🤔 Workaround is to restart sbt.
  - Maybe we can workaround it in better way, but I'm leaving it to just warn `please restart sbt` since I couldn't find a way + it's a lower priority it seems.

### DIFF
--- a/project/MyScalaNativePlugin.scala
+++ b/project/MyScalaNativePlugin.scala
@@ -23,9 +23,6 @@ object MyScalaNativePlugin extends AutoPlugin {
   final val isGeneratingForIDE =
     env.getOrElse("METALS_ENABLED", "false").toBoolean
 
-  final val enableProfiler =
-    env.getOrElse("ENABLE_PROFILER", "false").toBoolean
-
   final val enableExperimentalCompiler = {
     val ExperimentalCompilerEnv = "ENABLE_EXPERIMENTAL_COMPILER"
     val enabled = env.contains(ExperimentalCompilerEnv)
@@ -73,17 +70,14 @@ object MyScalaNativePlugin extends AutoPlugin {
 
     val args = spaceDelimited("<arg>").parsed
 
-    val commands = args.headOption match {
-      case None =>
-        throw new IllegalArgumentException(
-          "usage: nativeLinkProfiling <commands> <output-type>\n" +
-            "<commands>: `,` delimited arguments for async-profiler. refer https://github.com/async-profiler/async-profiler/blob/49d08fd068f81f1c952320c4bd082d991e09db97/src/arguments.cpp#L65-L113 \n" +
-            "<output-type>: text|collapsed|flamegraph|tree (default: flamegraph) \n" +
-            "e.g. `nativeLinkProfiling events=cpu,interval=10000000,threads"
-        )
-      case Some(value) =>
-        value
-    }
+    val commands = args.headOption.getOrElse(
+      throw new IllegalArgumentException(
+        "usage: nativeLinkProfiling <commands> <output-type>\n" +
+          "<commands>: `,` delimited arguments for async-profiler. refer https://github.com/async-profiler/async-profiler/blob/49d08fd068f81f1c952320c4bd082d991e09db97/src/arguments.cpp#L65-L113 \n" +
+          "<output-type>: text|collapsed|flamegraph|tree (default: flamegraph) \n" +
+          "e.g. `nativeLinkProfiling events=cpu,interval=10000000,threads"
+      )
+    )
 
     val outputType = (for {
       input <- args.tail.headOption.toRight(
@@ -106,8 +100,8 @@ object MyScalaNativePlugin extends AutoPlugin {
         } catch {
           case ex: IllegalStateException => {
             logger.warn(
-              s"Couldn't load async-prfiler, restart sbt to workaround the problem. " +
-                "This is usually caused because old sbt's classloader loaded the async-profiler DLL. \n" +
+              s"Couldn't load async-profiler, restart sbt to workaround the problem. " +
+                "This is usually caused because the previous sbt's classloader loaded the async-profiler DLL. \n" +
                 ex.getMessage()
             )
             throw ex
@@ -116,7 +110,7 @@ object MyScalaNativePlugin extends AutoPlugin {
         }
       else {
         logger.warn(
-          "Couldn't load async-prfiler for the current OS, architecture or glibc is unavailable. " +
+          "Couldn't load async-profiler for the current OS, architecture or glibc is unavailable. " +
             "Profiling will not be available." +
             "See the supported platforms https://github.com/jvm-profiling-tools/ap-loader#supported-platforms"
         )
@@ -128,11 +122,11 @@ object MyScalaNativePlugin extends AutoPlugin {
         (crossTarget.value / s"$module-profile.${outputType.extension}").toString
       profilerOpt match {
         case Some(profiler) =>
-          logger.info(
-            s"[async-profiler] starting profiler with commands: start,$commands"
-          )
-          profiler.execute(s"start,$commands")
           Def.task {
+            logger.info(
+              s"[async-profiler] starting profiler with commands: start,$commands"
+            )
+            profiler.execute(s"start,$commands")
             nativeLink.value
           } andFinally {
             logger.info(s"[async-profiler] stop profiler, output to ${out}")
@@ -161,7 +155,6 @@ object MyScalaNativePlugin extends AutoPlugin {
             .getOrElse(nc.multithreadingSupport)
         )
     },
-    // nativeLinkProfiling := nativeLinkProfilingImpl.tag(NativeTags.Link).value,
     scalacOptions ++= {
       // Link source maps to GitHub sources
       val isSnapshot = nativeVersion.endsWith("-SNAPSHOT")
--- a/project/MyScalaNativePlugin.scala
+++ b/project/MyScalaNativePlugin.scala
@@ -11,10 +11,7 @@ import complete.DefaultParsers._
 
 import one.profiler.AsyncProfilerLoader
 import one.profiler.AsyncProfiler
-import build.OutputType.Text
-import build.OutputType.Collapsed
-import build.OutputType.Flamegraph
-import build.OutputType.Tree
+import build.OutputType._
 
 object MyScalaNativePlugin extends AutoPlugin {
   override def requires: Plugins = ScalaNativePlugin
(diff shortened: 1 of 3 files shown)


sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -167,11 +169,25 @@ object ScalaNativePluginInternal {
...
+          Build
+            .buildCached(config)
+            .map(_.toFile())
+        }
+      } finally {
+        profiler.foreach { p =>
+          config.logger.info("[async-profiler] stop profiler")
+          p.execute("stop")
+          p.execute("flamegraph,file=profile.html")
+        }
```
WojciechMazur, line 190: By putting it in `ScalaNativePluginInternal` it would be included in all sbt-plugins. Because the `AsyncProfilerLoader` would not be available in other sbt builds, it would fail to start the sbt instance. 
Instead we can put it only in `MyScalaNativePlugin` by overriding `nativeLink` task there. What's more we can put it in some dedicated higher-order functions, allowing us to wrap any part of code we'd like to profile. 
So then it might look like: 
```scala
nativeLink := withAsyncProfiling(flamegraphFile="profile.html", interval=10000){
  nativeLink.value
 }
```

WojciechMazur: Yes, `MyScalaNativePlugin` is only internal override of `ScalaNativePlugin` with implementation defined in `ScalaNativePluginInternal`. `ScalaNativePlugin` is the public part which should be limited in terms of library dependencies and ammount of custom tasks/complexity, becouse that's the part we would need to preserve source/binary compatibility in next versions. 

Sorry if I've introduced any confusion. My initial though was that adding the profiling in `SNPluginInternal` would not be a good idea for 2 reasons: would introduce additional library dependencies and might add redundant overhead. 
Because of that we'd probably want to limit when profiling can be enabled. For our internal builds we can all dependencies as we wish, so we can add the implementation to MyScalaNativePlugin, and that's where we should focus. 
However, there might be some cases when we'd like to profile some 3rd party project. For that we can have 2 solutions: 
 1. Introduce an sbt plugin which would depend on `ScalaNativePlugin` similary to `MyScalaNativePlugin` and would introduce profiling tasks. I think that currently that might be too much 
 2. A simpler solution would be to structure profiling integration in such a way, that we would be able to easily copy-paste the integration to other project for ad-hoc integration. From what you've presented we should be able to limit the required code to 1 or 2 functions/tasks. We only need to make sure to not touch any internals, and instead work on the public API of `ScalaNativePlugin` If documented well we should be able to easily able profiling in custom projects this way. I think that for now that would be the best solution.

tanishiking: Thank you for the detailed explanation!
It seems we had a bit different views on that:

- I thought we'd like to add an integration with AsyncProfiler to public sbt-plugin so that all scala-native users can easily run profiling on any projects.
- While you'd like to focus on profiling scala-native running for our own projects

I agree that providing profiling feature to public scala-native plugin is too much (that's why I though it might be enough to running async-profiler on benchmark).
Let's focus on `MyScalaNativePlugin` for now 👍  I guess we can somehow find a good design to decouple the profiler with task implementation which is easy to imitate in other projects.

project/MyScalaNativePlugin.scala
```diff
@@ -55,6 +70,53 @@ object MyScalaNativePlugin extends AutoPlugin {
...
+        )
+        None
+      }
+
+      val module = moduleName.value
+      val out = (crossTarget.value / s"$module-profile.${outputType.extension}").toString
+      profilerOpt match {
+        case Some(profiler) =>
+          logger.info(s"[async-profiler] starting profiler: $profiler")
+          profiler.execute("start,event=cpu,interval=100000")
```
tanishiking, line 107: Hard coding some options for now.

project/MyScalaNativePlugin.scala
```diff
@@ -55,6 +67,83 @@ object MyScalaNativePlugin extends AutoPlugin {
...
+
+    val profilerOpt: Option[AsyncProfiler] =
+      if (AsyncProfilerLoader.isSupported())
+        try {
+          Some(AsyncProfilerLoader.load())
+        } catch {
+          case ex: IllegalStateException => {
+            logger.warn(
+              s"Couldn't load async-prfiler, restart sbt to workaround the problem. " +
+                "This is usually caused because old sbt's classloader loaded the async-profiler DLL. \n" +
```
ekrich, line 110:
```suggestion
              s"Couldn't load async-profiler, restart sbt to workaround the problem. " +
                "This is usually caused because the old sbt's classloader loaded the async-profiler DLL. \n" +
```

ekrich: Maybe the `previous` sbt classloader?

project/MyScalaNativePlugin.scala
```diff
@@ -55,6 +67,83 @@ object MyScalaNativePlugin extends AutoPlugin {
...
+                ex.getMessage()
+            )
+            throw ex
+          }
+          case e: Throwable => throw e
+        }
+      else {
+        logger.warn(
+          "Couldn't load async-prfiler for the current OS, architecture or glibc is unavailable. " +
```
ekrich, line 119:
```suggestion
          "Couldn't load async-profiler for the current OS, architecture or glibc is unavailable. " +
```

ekrich (changes requested):

project/MyScalaNativePlugin.scala
```diff
@@ -3,17 +3,29 @@ package build
...
+  lazy val nativeLinkProfiling =
+    inputKey[File]("Running nativeLink with AsyncProfiler.")
+
   // see: https://github.com/scalameta/metals/blob/0176a491cd209a09852ab33f99fd7de639e8e2dd/metals/src/main/scala/scala/meta/internal/builds/BloopInstall.scala#L81
   final val isGeneratingForIDE =
     env.getOrElse("METALS_ENABLED", "false").toBoolean
 
+  final val enableProfiler =
+    env.getOrElse("ENABLE_PROFILER", "false").toBoolean
+
```
WojciechMazur, line 28: Unused val

project/MyScalaNativePlugin.scala
```diff
@@ -55,6 +67,83 @@ object MyScalaNativePlugin extends AutoPlugin {
...
+      case None =>
+        throw new IllegalArgumentException(
+          "usage: nativeLinkProfiling <commands> <output-type>\n" +
+            "<commands>: `,` delimited arguments for async-profiler. refer https://github.com/async-profiler/async-profiler/blob/49d08fd068f81f1c952320c4bd082d991e09db97/src/arguments.cpp#L65-L113 \n" +
+            "<output-type>: text|collapsed|flamegraph|tree (default: flamegraph) \n" +
+            "e.g. `nativeLinkProfiling events=cpu,interval=10000000,threads"
+        )
+      case Some(value) =>
+        value
+    }
```
WojciechMazur, line 86: How about a simple `getOrElse`?

project/MyScalaNativePlugin.scala
```diff
@@ -55,6 +67,83 @@ object MyScalaNativePlugin extends AutoPlugin {
...
+        } catch {
+          case ex: IllegalStateException => {
+            logger.warn(
+              s"Couldn't load async-prfiler, restart sbt to workaround the problem. " +
+                "This is usually caused because old sbt's classloader loaded the async-profiler DLL. \n" +
+                ex.getMessage()
+            )
+            throw ex
+          }
+          case e: Throwable => throw e
```
WojciechMazur, line 115: Since we only rethrow here this case can be skipped

project/MyScalaNativePlugin.scala
```diff
@@ -55,6 +67,83 @@ object MyScalaNativePlugin extends AutoPlugin {
...
+
+      val module = moduleName.value
+      val out =
+        (crossTarget.value / s"$module-profile.${outputType.extension}").toString
+      profilerOpt match {
+        case Some(profiler) =>
+          logger.info(
+            s"[async-profiler] starting profiler with commands: start,$commands"
+          )
+          profiler.execute(s"start,$commands")
```
WojciechMazur, line 134: This line probably should be in `Def.task` body

project/MyScalaNativePlugin.scala
```diff
@@ -72,6 +161,7 @@ object MyScalaNativePlugin extends AutoPlugin {
             .getOrElse(nc.multithreadingSupport)
         )
     },
+    // nativeLinkProfiling := nativeLinkProfilingImpl.tag(NativeTags.Link).value,
```
WojciechMazur, line 164: Commented out code

WojciechMazur (commented):
Overall looks good, but there is some code to cleanup before merging

project/MyScalaNativePlugin.scala
tanishiking: The above branch catches only an `IllegalStateException` and the plugin sometimes throw other exceptions

WojciechMazur (approved):

PR #3604, opened 2023-11-16
When generating top-level extern methods check its annotations for `link`/`define`

### LINKED ISSUE #3603: `@link("lib")` is ignored with top-level `@extern` methods
While
```scala
@extern
@link("unit")
object externs {
  def nxt_unit_init(init: Ptr[Byte]): Ptr[Byte] = extern
}
```
does add `-lunit` to the linking options, the equivalent top-level definition doesn't add linking options:
```scala
@extern
@link("unit")
def nxt_unit_init(init: Ptr[Byte]): Ptr[Byte] = extern
```

WojciechMazur:
Fixes #3603

### DIFF
--- a/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenStat.scala
+++ b/nscplugin/src/main/scala-2/scala/scalanative/nscplugin/NirGenStat.scala
@@ -940,27 +940,47 @@ trait NirGenStat[G <: nsc.Global with Singleton] { self: NirGenPhase[G] =>
           }
       }
     }
+    private def genMethodAttrs(
+        sym: Symbol,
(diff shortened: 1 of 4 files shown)


armanbilge (commented):
Would we benefit from similar changes in Scala 2? We don't have top-level defs but wouldn't this allow individuals methods to be annotated `@extern`/`@define` within an `object` or `package object`?

WojciechMazur:
> Would we benefit from similar changes in Scala 2? We don't have top-level defs but wouldn't this allow individuals methods to be annotated @extern/@define within an object or package object? 

Good idea, I've ported the change to Scala2 compiler plugin

PR #3606, opened 2023-11-21
Fix typo on VEOL @name annotation in termios module

kapunga:
The spelling is correct in `termios.c` so there is a mismatch. (See here: [termios.c](https://github.com/scala-native/scala-native/blob/main/posixlib/src/main/resources/scala-native/termios.c#L16) )


tanishiking (approved):
LGTM, nice catch!

PR #3607, opened 2023-11-22
Document NIR value definitions

kyouko-taiga:
This PR simply adds documentation and uniformize code style. Special attention should be given to the docs of `Val.Virtual` and `Val.ClassOf` because I am not sure I got things right.

### DIFF
--- a/nir/src/main/scala/scala/scalanative/nir/Vals.scala
+++ b/nir/src/main/scala/scala/scalanative/nir/Vals.scala
@@ -4,121 +4,174 @@ package nir
 import java.lang.Float.floatToRawIntBits
 import java.lang.Double.doubleToRawLongBits
 
+/** A NIR value. */
 sealed abstract class Val {
-  final def ty: Type = this match {
-    case Val.Null                 => Type.Null
-    case Val.Zero(ty)             => ty
-    case Val.True | Val.False     => Type.Bool
-    case Val.Size(_)              => Type.Size
-    case Val.Char(_)              => Type.Char
-    case Val.Byte(_)              => Type.Byte
-    case Val.Short(_)             => Type.Short
-    case Val.Int(_)               => Type.Int
-    case Val.Long(_)              => Type.Long
-    case Val.Float(_)             => Type.Float
-    case Val.Double(_)            => Type.Double
-    case Val.StructValue(vals)    => Type.StructValue(vals.map(_.ty))
-    case Val.ArrayValue(ty, vals) => Type.ArrayValue(ty, vals.length)
-    case v: Val.ByteString        => Type.ArrayValue(Type.Byte, v.byteCount)
-    case Val.Local(_, ty)         => ty
-    case Val.Global(_, ty)        => ty
 
-    case Val.Unit     => Type.Unit
-    case Val.Const(_) => Type.Ptr
+  /** The type of the value. */
+  final def ty: Type = this match {
+    case Val.Null =>
+      Type.Null
+    case Val.Zero(ty) =>
+      ty
+    case Val.True | Val.False =>
+      Type.Bool
+    case Val.Size(_) =>
+      Type.Size
+    case Val.Char(_) =>
+      Type.Char
+    case Val.Byte(_) =>
+      Type.Byte
+    case Val.Short(_) =>
+      Type.Short
+    case Val.Int(_) =>
+      Type.Int
+    case Val.Long(_) =>
+      Type.Long
+    case Val.Float(_) =>
+      Type.Float
+    case Val.Double(_) =>
+      Type.Double
+    case Val.StructValue(vals) =>
+      Type.StructValue(vals.map(_.ty))
+    case Val.ArrayValue(ty, vals) =>
+      Type.ArrayValue(ty, vals.length)
+    case v: Val.ByteString =>
+      Type.ArrayValue(Type.Byte, v.byteCount)
+    case Val.Local(_, ty) =>
+      ty
+    case Val.Global(_, ty) =>
+      ty
+    case Val.Unit =>
+      Type.Unit
+    case Val.Const(_) =>
+      Type.Ptr
     case Val.String(_) =>
       Type.Ref(Rt.String.name, exact = true, nullable = false)
-    case Val.Virtual(_) => Type.Virtual
-    case Val.ClassOf(n) => Rt.Class
+    case Val.Virtual(_) =>
+      Type.Virtual
+    case Val.ClassOf(n) =>
+      Rt.Class
   }
 
+  /** A textual representation of `this`. */
   final def show: String = nir.Show(this)
 
+  /** `true` iff `this` represents a virtual value. */
   final def isVirtual: Boolean =
     this.isInstanceOf[Val.Virtual]
 
+  /** `true` iff `this` is in canonical form. */
   final def isCanonical: Boolean = this match {
     case Val.True | Val.False =>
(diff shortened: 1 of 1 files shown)


nir/src/main/scala/scala/scalanative/nir/Vals.scala
```diff
@@ -4,121 +4,174 @@ package nir
...
-    case Val.Double(_)            => Type.Double
-    case Val.StructValue(vals)    => Type.StructValue(vals.map(_.ty))
-    case Val.ArrayValue(ty, vals) => Type.ArrayValue(ty, vals.length)
-    case v: Val.ByteString        => Type.ArrayValue(Type.Byte, v.byteCount)
-    case Val.Local(_, ty)         => ty
-    case Val.Global(_, ty)        => ty
-
-    case Val.Unit     => Type.Unit
-    case Val.Const(_) => Type.Ptr
+    case Val.Null =>
```
ekrich, line 12: Is this what `scalafmt` does?

kyouko-taiga: scalamft adds inconsistent alignment when there are spaces in the case list. Further, aligning cases creates unnecessary changes in commits every time something changes in one specific case handler.

JD557:
Out of curiosity, did you check for performance regressions?

I believe `case Foo(_) =>` can be a bit slower than `case _: Foo`, as it adds an extra call to `Foo.unapply`allong with few extra calls ([example](https://godbolt.org/z/dqEfP3nhW)).

I think this is usually fine, but it might be worth it to double check.

[Example](https://godbolt.org/z/dqEfP3nhW):

`case Add(_) => true` generates

```
       2: aload_2
       3: instanceof    #9                  // class Foo$Add
       6: ifeq          30
       9: getstatic     #63                 // Field Foo$Add$.MODULE$:LFoo$Add$;
      12: aload_2
      13: checkcast     #9                  // class Foo$Add
      16: invokevirtual #67                 // Method Foo$Add$.unapply:(LFoo$Add;)LFoo$Add;
      19: astore_3
      20: aload_3
      21: invokevirtual #71                 // Method Foo$Add._1:()I
      24: istore        4
      26: iconst_1
      27: goto          129
```

while `case _: Add =>` generates just

```
       2: aload_2
       3: instanceof    #9                  // class Foo$Add
       6: ifeq          13
       9: iconst_1
      10: goto          55
```

kyouko-taiga:
> Out of curiosity, did you check for performance regressions?

I didn't because I don't know how. I'm happy to run these checks if someone can point me to a way to run them.

nir/src/main/scala/scala/scalanative/nir/Vals.scala
kyouko-taiga: Case in point: https://github.com/scala-native/scala-native/blob/f4b9078c6c96dade366518f037d8f47048d2a4f3/nir/src/main/scala/scala/scalanative/nir/Ops.scala#L11-L48

This function looks messy and scary to maintain. Having a more uniform way to format the match-expression that is resilient to changes is desirable, IMO. Tab-alignments of case handlers can look nice sometimes but they are also very high maintenance.

nir/src/main/scala/scala/scalanative/nir/Vals.scala
```diff
@@ -183,18 +262,37 @@ object Val {
...
   final case class Local(id: nir.Local, valty: nir.Type) extends Val
+
+  /** A symbol. */
   final case class Global(name: nir.Global, valty: nir.Type) extends Val
 
-  // high-level
+  /** The unit value. */
   case object Unit extends Val
+
+  /** A constant. */
```
WojciechMazur, line 286: Important information here is fact that `value` would be deduplicated in codegen and that the `Const` should be treated as pointer to a constant at the low level. It's not true that `Const` would work the same as literal constant - for these a `ByteString`, `Zero`, `Int`, `Long`, `Float`,  `StructValue`, `ArrayValue` etc can be used. These kind of vlaues are typically stored as a `value`

nir/src/main/scala/scala/scalanative/nir/Vals.scala
```diff
@@ -183,18 +262,37 @@ object Val {
...
   final case class Const(value: Val) extends Val
+
+  /** A character string. */
```
WojciechMazur, line 289: `Val.String` translates directly to object of `java.lang.String`, and is stored as a global constant. This allows it to contain a UTF-16 characters. For C-like strings (ASCII,UTF8) use ByteString.

nir/src/main/scala/scala/scalanative/nir/Vals.scala
```diff
@@ -183,18 +262,37 @@ object Val {
...
   final case class String(value: java.lang.String) extends Val
+
+  /** A virtual value. */
```
WojciechMazur, line 292: Note: Virtual values exists only in optimizer. They are not serializable, and would never be emitted by compiler plugin.

nir/src/main/scala/scala/scalanative/nir/Vals.scala
```diff
@@ -183,18 +262,37 @@ object Val {
...
   final case class Virtual(key: scala.Long) extends Val
+
+  /** A class symbol. */
```
WojciechMazur, line 295: A reference to `java.lang.Class[_]` of given symbol (`name`). These are emitted as global variables in codegen. `ClassOf` instances are used to deduplicate `Class` instances - there should be only 1 instance per type.

Side note: Even though they're emiited as global variables now, these could be constants. However, when adding multithreading and object monitors we needed capability of editing one of it's fields (lockWord) containing ObjectMonitor object or a bit set of lock word.

WojciechMazur (changes requested):
Thank for for these changes. Mostly looks good, although I have a few notes which should be applied to avoid misunderstandings.

nir/src/main/scala/scala/scalanative/nir/Vals.scala
WojciechMazur: Personally I'm more into the more compact style when most of match cases are single-liners - for me it seems way more readable and organised, however I understand the point about maintenance. I'm fine with both approaches, especially since this part of code is rarely changed.

WojciechMazur:
> > Out of curiosity, did you check for performance regressions?

>I didn't because I don't know how. I'm happy to run these checks if someone can point me to a way to run them.

We have some benchmarks, but these are rather a macro benchmarks testing whole stage (linking, optimizer, codegen), rather then micro benchmars dedicated to single functionally. Anyway these are defined in `tools-benchmarks` using the JMH framework.
These can be run using `toolsBenchmarks3/Jmh/run` to run all of them or 
`toolsBenchmarks3/Jmh/run scala.scalanative.benchmarks.LinkerBench` to run just a single benchmark. 

I don't think that difference would be significant, especially after JIT compilation by the JVM, but if possible let's stick to the most simple solution as these methods might be called a lot in optimizer.

kyouko-taiga:
> I don't think that difference would be significant, especially after JIT compilation by the JVM, but if possible let's stick to the most simple solution as these methods might be called a lot in optimizer.

I strongly believe that at this point clarity is far more important than speed. AFAIK, we have still to identify exactly what our performance bottlenecks are, and so we should not be overly concerned about low-level optimization tricks unless we can demonstrate their impact.

WojciechMazur (approved):

Issue #3610, opened 2023-11-27
What are the semantics of `SyncAttrs.isVolatile`?


kyouko-taiga:
[LLVM](https://llvm.org/docs/LangRef.html#volatile-memory-accesses) describes "volatile memory accesses" as follows (emphasis mine):

> Certain memory accesses, such as [load](https://llvm.org/docs/LangRef.html#i-load)’s, [store](https://llvm.org/docs/LangRef.html#i-store)’s, and [llvm.memcpy](https://llvm.org/docs/LangRef.html#int-memcpy)’s may be marked volatile. The optimizers must not change the number of volatile operations or change their order of execution relative to other volatile operations. The optimizers may change the order of volatile operations relative to non-volatile operations. **This is not Java’s “volatile” and has no cross-thread synchronization behavior.**

That lead me to wonder what were the semantics of `SyncAttrs.isVolatile`.

WojciechMazur:
`SyncAttrs.isVolatile` maps directly to `LLVM IR` volatile attribute for `load`/ `store` instructions. The JVM volatile behaviour is archived in other measures described below.  Current usages of LLVM `volatile` for load/store operations seams to be redundant and I believe it's usages could be removed. 

Every Scala field marked with `@volatile` would be marked with `Attr.Volatile`, and every `final` field would be marked with `Attr.Final` in compiler plugin when generating NIR. Field load/store operations for fields marked with these attributes would use LLVM IR `atomic load/store` with one of available memory orders - SeqCst for volatile fields, Monotonic for final and unordered for other fields (without any annotation). These should be sufficient to mostly comply with Java Memory Model.

armanbilge:
> Monotonic for final ... These should be sufficient to mostly comply with Java Memory Model.

Hmm, why is it that `monotonic` correspond to `final` fields? Here's what LLVM says:

> If an address is written monotonic-ally by one thread, and other threads monotonic-ally read that address repeatedly, the other threads must eventually see the write.

That doesn't seem strong enough to guarantee that all threads will _always_ see the value of a `final` field _after_ it has been initialized.

WojciechMazur:
> That doesn't seem strong enough to guarantee that all threads will always see the value of a final field after it has been initialized. 

Right, I forget about other aspect. If class has final fields then at the end of constructor call we do introduce a release fence. 
[...]

PR #3611, opened 2023-11-27
Document 'nir.MemoryOrder'

kyouko-taiga:
This PR simply adds documentation on memory order constraints.

Note that a note specifying the semantics of `SyncAttrs.isVolatile` would probably help clarifying the intent of this property. See #3610


WojciechMazur (commented):
Looks good to me, I've made a PR removing probably redundant `SyncAttrs` in #3616 so the comments on isVolatile would not be really important when it's merged.

WojciechMazur (approved):

PR #3615, opened 2023-12-04
Try fix Windows CI, replace no longer existing `Choco-Install`script

WojciechMazur:


ekrich:
Just wondering which LLVM this is installing. I had problems trying to compile with Windows using the Visual Studio tools. I think LLVM downloaded worked but I couldn't get either the 32bit or the 64bit to work with VS tools/libs.

WojciechMazur:
> Just wondering which LLVM this is installing.  

We're currently testing 17.0.6 (default, latest) and 16.0.6. All previous versions <15.x.x are not compiling on Windows Github Actions - they frequently update the minimal version without any further notice.

ekrich:
I was using 16.0.5 locally and it seemed to work.

PR #3658, opened 2024-01-08
fix: Fix deadlocks in CommixGC mutatorThreads list acquire and improve stability

### LINKED ISSUE #3334: Modification of linked list GC roots is not thread-safe
https://github.com/scala-native/scala-native/pull/3039#discussion_r1228328481 

> There is a risk that if two threads try to concurrently remove items from the singly linked list, one of them will access freed memory on this line.
Suppose the singly linked list has the following nodes: head -> A -> B -> C -> D;
Thread 1 wants to remove C. It calls GC_Roots_RemoveByRange. current is now pointing to C and prev is pointing to B. Thread 1 is just about to execute line 30, but context switching happened.
Then thread 2 starts executing. It wants to remove B. It called GC_Roots_RemoveByRange and it quickly finished. Now the linked list looks like: head -> A -> C -> D, while B is freed.
Then thread 1 wakes up, and continues to execute line 30. Now prev already points to freed memory. The atomic compare-exchange doesn't help because no matter whether it is successful or not, some freed memory have already been overwritten.
I guess the "custom GC roots" feature is for internal use for the ScalaNative runtime. In this case, I think it will not be called very often. It should be OK to use mutex for synchronisation.

WojciechMazur:
Fixes #3334

### DIFF
--- a/javalib/src/main/scala/java/lang/ref/WeakReferenceRegistry.scala
+++ b/javalib/src/main/scala/java/lang/ref/WeakReferenceRegistry.scala
@@ -75,13 +75,13 @@ private[java] object WeakReferenceRegistry {
       var head = weakRefsHead
       val expected = stackalloc[WeakReference[_]]()
       !expected = null
-      if (atomic_compare_exchange_weak(weakRefsHeadPtr, expected, weakRef)) ()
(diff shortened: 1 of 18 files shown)


ekrich:
Not sure if you remember but I added the sbt command test-sandbox-gc which can help make sure all the GCs will compile and run. You could also add others too if needed.

PR #3667, opened 2024-01-13
improvement: Microoptimizations to toolchains and runtime

WojciechMazur:
- Use ArraySeq as default when deserializing NIR sequences
- Disable some computions in optimizers when source level dubbing is disabled
- Use common thread_local implementation hinting usage of local-exec TLS mode.


tools/src/main/scala/scala/scalanative/interflow/MergeProcessor.scala
```diff
@@ -150,7 +151,7 @@ final class MergeProcessor(
           def escapes(addr: Addr): Boolean =
             states.exists(_.hasEscaped(addr))
 
-          states.head.heap.keys
+          states.head.heap.keys.iterator
```
JD557, line 154: Wouldn't it be preferable to call `keysIterator` directly?

```suggestion
          states.head.heap.keysIterator
```

PR #3699, opened 2024-01-20
Emit `acquire` fence before loading `final` field

armanbilge:
In https://github.com/typelevel/cats-effect/issues/3899#issuecomment-1898911499 @durban brought to my attention that the current implementation of `final` Field Semantics when loading a `final` field is not correct.

Specifically, we must issue an `aquire` fence _prior_ to loading the `final` field. This prevents the load of the field from being re-ordered before the load of the object reference (e.g. due to a compiler optimization that attempts an early, speculative load). Such a re-ordering could cause a pre-initialized value for the field to be observed.

Unfortunately I'm still not confident that we will have a correct implementation of `final` Field Semantics after this change. However, I do believe that this change brings us closer to that.

On the other end of the spectrum, I am hopeful that we might eventually be able to remove this fence entirely. Ideally, the dependency of the load of the `final` field on the load of the object reference would enforce the ordering. I don't have sufficient imagination to see how the compiler could generate speculative code that violates this ordering. If we can be confident that neither LLVM nor supported hardware will perform such a re-ordering, then it could be safe to remove this fence.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
@@ -594,10 +594,15 @@ object Lower {
           throw new LinkingException(s"Metadata for field '$name' not found")
       }
 
+      // No explicit memory order for load of final field,
+      // all final fields are loaded after a acquire fence
       val memoryOrder =
         if (field.attrs.isVolatile) nir.MemoryOrder.SeqCst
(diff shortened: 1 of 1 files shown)


WojciechMazur (approved):

WojciechMazur:
Don't worry about any failure in the CI. These are related to the bugs in the GC and I'm trying to fix them

WojciechMazur (approved):

Issue #3700, opened 2024-01-21
Compiler crash with opaque types, inline, varargs, unsigned


keynmol:
Error (you can see full stack trace in this run of Github actions: https://github.com/indoorvivants/sn-bindgen-examples/actions/runs/7588621095/job/20671553894?pr=116#step:9:122):

```
Exception in thread "main" java.lang.ClassCastException: class dotty.tools.dotc.core.Symbols$Symbol cannot be cast to class dotty.tools.dotc.core.Symbols$ClassSymbol (dotty.tools.dotc.core.Symbols$Symbol and dotty.tools.dotc.core.Symbols$ClassSymbol are in unnamed module of loader 'app')
        at dotty.tools.dotc.core.Symbols$Symbol.asClass(Symbols.scala:178)
        at scala.scalanative.nscplugin.NirGenType.isPrimitiveValueType(NirGenType.scala:85)
        at scala.scalanative.nscplugin.NirGenType.isPrimitiveValueType$(NirGenType.scala:20)
        at scala.scalanative.nscplugin.NirCodeGen.isPrimitiveValueType(NirCodeGen.scala:18)
        at scala.scalanative.nscplugin.NirGenExpr$ExprBuffer.ensureUnboxed(NirGenExpr.scala:1883)
        at scala.scalanative.nscplugin.NirGenExpr$ExprBuffer.genArg$1(NirGenExpr.scala:1503)
        at scala.scalanative.nscplugin.NirGenExpr$ExprBuffer.genExternMethodArgs$$anonfun$2$$anonfun$1(NirGenExpr.scala:1534)
        at scala.collection.immutable.List.foreach(List.scala:333)
        at scala.scalanative.nscplugin.NirGenExpr$ExprBuffer.genExternMethodArgs$$anonfun$2(NirGenExpr.scala:1547)
        at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:575)
        at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:573)
        at scala.collection.AbstractIterable.foreach(Iterable.scala:933)
```

I believe this is a minimisation with scala-cli (`scala-cli compile test.scala --server=false`):

```scala
//> using platform native
//> using nativeVersion 0.4.17
//> using scala 3.3.1

import scalanative.libc.stdio
import scalanative.unsafe.*
import scalanative.unsigned.*

opaque type cmark_event_type = CUnsignedInt
object cmark_event_type:
[...]

PR #3712, opened 2024-01-26
fix: compiler crash on primitive type checking against opaque types

tanishiking:
fix https://github.com/scala-native/scala-native/issues/3700

It is caused because in `isPrimitiveValueType` against opaque type, `t.widenDealias.typeSymbol` isn't `ClassSymbol` but a `TypeSymbol` for `cmark_event_type`. This commit adds an additional check that is the type symbol `ClassSymbol` before casting.


scripted-tests/scala3/opaque-type/src/main/scala/Hello.scala
```diff
@@ -0,0 +1,13 @@
...
+
+opaque type cmark_event_type = CUnsignedInt
+object cmark_event_type:
+  inline def define(inline a: Long): cmark_event_type = a.toUInt
+  val CMARK_EVENT_NONE = define(0)
+  
+@main def hello(): Unit =
+  val evtype = cmark_event_type.CMARK_EVENT_NONE
+  stdio.printf(c"bla: %s, hello: %d", evtype)
+
```
WojciechMazur, line 13: Instead of expensive scripted tests, let's define it as quick unit test in nscplugin, for example in this test https://github.com/scala-native/scala-native/blob/8df8dce73d8bcbdf076f75fb0b05c45bc4a84f6c/nscplugin/src/test/scala-3/scala/scalanative/NIRCompilerTest3.scala#L12-L13 

We only need to ensure that is compiles, we don't require linking

tanishiking: Thanks! done in https://github.com/scala-native/scala-native/pull/3712/commits/1a3dbcc4dd232ede14397f789dea94a87673fafa
(I added clib.cp to `Test/  nscplugin / javaOptions` since I couldn't find a way to reproduce the compilation crash without using `clib.stdio` 🤔

WojciechMazur: I think you might have problems with that, becouse there are 2 definitions of printf. One is defined as extension here: https://github.com/scala-native/scala-native/blob/ba2bd19e0ccc1d3ec6db97a6249aa397a4f7e308/clib/src/main/scala/scala/scalanative/libc/package.scala#L5-L8 it's a legacy support from before introduction of better varargs in extern  
The second one is standard extern binding https://github.com/scala-native/scala-native/blob/ba2bd19e0ccc1d3ec6db97a6249aa397a4f7e308/clib/src/main/scala/scala/scalanative/libc/stdio.scala#L587-L589 

I belive it could have been coused by the clash of the two. It would be great if we could get rid of the dependency. I'd like to remove legacy version in the future.

tanishiking: Thanks @WojciechMazur!
I'm not sure the duplication of `printf` is the problem, I could reproduce it without `clib` https://github.com/scala-native/scala-native/pull/3712/commits/74fbdddc1e59697f5a76a524b3a4107ba812513c

It seems like we hit the `ensureUnboxed` -> `isPrimitiveType` when the argument is for extern method's varargs (e.g. `printf`'s `vargs`)
https://github.com/scala-native/scala-native/blob/74fbdddc1e59697f5a76a524b3a4107ba812513c/nscplugin/src/main/scala-3/scala/scalanative/nscplugin/NirGenExpr.scala#L1722-L1723 

With something like `def test(t: Any*) = extern`, and if I pass the opaque typed value to the `test` method, we could break the compiler :+1:

PR #3713, opened 2024-01-27
fix: Improve stability of Immix/Commix GCs

WojciechMazur:
This PR contains multiple changes related to improving GC stability or fixing minor issues found when debugging. 

Since merge of https://github.com/scala-native/scala-native/pull/3681 every single CI job fails when using Scala 2.  - t was strange because mentioned PR have not touched GC at all, instead if was adding only additional methods and related to them tests. I've seen that reverting tests layout to original one fixed partially some of the runs. Because of that now ByteBuffer tests are generated from template resolved using source generator - we use special markers to exclude parts of the code that requires some version of JDK. It fixed, some of the issues, but not all of them. 

Additional changes: 
- Mark conservatively boxed pointers - these can contain pointers to heap allocated objects, especially in code used for atomic operations with `stdatomic.h` 
- Improved scanning of registers - introduce code using either dedicated x86 / x86_64 assembly to extract register values for scanning or use setjmp or other platforms. 
- Inline stack top approximations and state switching in mutator threads 
- Use read instead of write operation in yieldpoint traps, move yieldpoint traps code generation in AbstractCodeGen due to lack of volatile load operation in NIR. 
- Make mutatorThreads list atomic pointer, improve synchronization of it's sate
- Exit with non-zero value on Windows in GCThreads upon illegal state - errno was always zero, generate more cotext about error 
- Ensure freshly created thread waits until end of GC collection 
- Synchronise Allocator/LargeAllocator in Immix with changes done in Commix GC


WojciechMazur:
There still seems to be some issues, but overall it should give a better results then before. 
TODO in the follow ups: 
* SIGILL in MacOS - seems to be the only system with this defect. Maybe try Boehm GC approach of Darwin specific threads synchronization 
* Commix deadlock - no idea where it cames from. Once I was able to find one when Allocator was waiting infinitely long for Sweeper_IsSweepDone, however, there the sweep was done already and all gc threads were suspended. No idea how it could have happend

PR #3719, opened 2024-01-30
feautre: Relaxed memory model for final fields, strict semantics with `@safePublish` annotation

### LINKED ISSUE #3673: Consider having a configurable memory model
Currently following Java Memory Model imposes a significant performance regression, ~10-15%. Every `val` class member coming from scalac compiler would be treated as Java `final`. eg. `class (val x: Int){val y: AnyRef = ???}` or `case class(x: Int){val y: AnyRef}` requires us to introduce `load atomic acquire` for every load, and introduce `fence release` after the end of constructor. 
Not every application would require such strong guarantees. For these we might introduce a `relaxed` memory model which would reduce amount of synchronisation, as opposed to current `strict` memory model. The memory model should be selectable in the `NativeConfig`. 
Alternative would to annotate specific classess or modules/packages/libraries(?, how?) to use a dedicated memory model.

WojciechMazur:
resolves #3673  
Final fields are now by default unsynchronized. We introduce a `@safePublish` annotation to restore Java final fields semantics

TODO: 
* [x] - Introduce a compile time config (not linking time) that library authors can use to automatically apply the "@safePublish" annotation to all vals within an entire compilation unit. This would essentially be an escape hatch.
* [x] - Introduce a linking time config that applications can use to restore final Field Semantics for all vals, in their own code and library NIR. Another escape hatch.

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Lower.scala
@@ -46,6 +46,7 @@ object Lower {
     private val fresh = new util.ScopedVar[nir.Fresh]
     private val unwindHandler = new util.ScopedVar[Option[nir.Local]]
     private val currentDefn = new util.ScopedVar[nir.Defn.Define]
+    private val acquiredFinalFieldsOf = new util.ScopedVar[mutable.Set[nir.Val]]
     private val nullGuardedVals = mutable.Set.empty[nir.Val]
     private def currentDefnRetType = {
       val nir.Type.Function(_, ret) = currentDefn.get.ty
@@ -95,7 +96,8 @@ object Lower {
         val nir.Type.Function(_, ty) = defn.ty
         ScopedVar.scoped(
           fresh := nir.Fresh(defn.insts),
-          currentDefn := defn
+          currentDefn := defn,
+          acquiredFinalFieldsOf := mutable.Set.empty
         ) {
           try super.onDefn(defn)
           finally nullGuardedVals.clear()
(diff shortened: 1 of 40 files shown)


tools/src/main/scala/scala/scalanative/codegen/Lower.scala
```diff
@@ -601,7 +603,11 @@ object Lower {
         else nir.MemoryOrder.Unordered
 
       // Acquire memory fence before loading a final field
-      if (field.attrs.isFinal) buf.fence(nir.MemoryOrder.Acquire)
+      // It's required only on the first access to obj fields
+      if (field.attrs.finalWithSafePublish && acquiredFinalFieldsOf.get.add(
+            obj
+          ))
+        buf.fence(nir.MemoryOrder.Acquire)
```
armanbilge, line 610: What is this doing exactly? It's required on the first access _per thread_.

WojciechMazur: It's a very simple optimisation, for cases when we want to load multiple fields from the same object. This way we have at most 1 fence per object within a single method. Previously we're introducing a fence for each load, although now as I'm thinking about it it's bugged and should be reverted - we might add fence in only one conditional branch of code, leaving other one vulnerable.

WojciechMazur:
@armanbilge  I'm going to merge it into the main before the end of your review to unblock some of other work, however fell free to comment here, and I will definitely address any found issues.

tools/src/main/scala/scala/scalanative/codegen/Lower.scala
armanbilge: Oops, good catch.

I think we should always emit NIR with all the barriers. Then we can implement an NIR => NIR optimization phase that attempts to "coalesce" multiple barriers. The JVM implements these sorts of optimizations.

I am trying to find more information about this, but here is an example:
https://bugs.openjdk.org/browse/JDK-8032218

WojciechMazur: Yes, I was thinking about this, and also with coalescing of synchronized blocks. With these it should be easier, although probably we should create a dedicated NIR instruction for them to make it easier. Moving generation of fences from lowering to NIR generation would be required to correctly optimize them, but it would also harder to identify possibly redundant fences when using relaxed memory model - we'd need to perform a look ahead to check that. Anyway that's one important change to do

PR #3721, opened 2024-01-30
Fix misleading Heap_Grow logs

JD557:
I believe the log message here is misleading, as `Heap_Grow` is also called by the large allocator: https://github.com/scala-native/scala-native/blob/main/nativelib/src/main/resources/scala-native/gc/immix/LargeAllocator.c#L231

(I might have misunderstood the concept of small heap, though)


WojciechMazur (approved):

WojciechMazur:
Thank you for this fix, there is no such thing as a small or big heap, there's just a heap, so that might fix the confusion. Probably the original intention was to inform about `growing insufficiently small heap due to high memory usage`

PR #3736, opened 2024-02-08
refactor/fix: Use object offsets instead of field indexes for precise object scanning

WojciechMazur:
Immix and Commix scan objects precisely based on array of indexes containing possible references. Previously this list was populate by indexes of fields, now it's a list of offsets starting from object address (also known as start of object header) to make the code more readable. 

It also can possibly some issues on 32-bit CPU architectures, see code comment below. 
As an additional improvement used when debugging we're adding an utility to extract Class name as wide-string allowing to show it in debug information from the GC (best effort, no guarantess about String object layout)

### DIFF
--- a/tools/src/main/scala/scala/scalanative/codegen/Generate.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Generate.scala
@@ -548,7 +548,7 @@ object Generate {
                 .index(weakRefReferantField)
                 .ensuring(
                   _ > 0,
-                  "Runtime implementaiton error, no \"_gc_modified_referent\" field in java.lang.ref.WeakReference"
+                  "Runtime implementation error, no \"_gc_modified_referent\" field in java.lang.ref.WeakReference"
                 )
               val gcModifiedFieldReferentOffset = layout.layout
                 .tys(gcModifiedFieldReferentIdx)
--- a/tools/src/main/scala/scala/scalanative/codegen/Generate.scala
+++ b/tools/src/main/scala/scala/scalanative/codegen/Generate.scala
@@ -4,6 +4,7 @@ package codegen
 import scala.collection.mutable
 import scala.scalanative.linker.{
   Class,
+  Field,
   ScopeInfo,
   Unavailable,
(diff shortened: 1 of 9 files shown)


tools/src/main/scala/scala/scalanative/codegen/MemoryLayout.scala
```diff
@@ -21,23 +21,22 @@ final case class MemoryLayout(
...
       tys.collect {
         // offset in words without object header
         case MemoryLayout.PositionedType(
               _: RefKind | // normal or alligned reference field
               StructValue((_: RefKind) :: ArrayValue(nir.Type.Byte, _) :: Nil),
               offset
             ) =>
-          nir.Val.Long(
-            (offset - sizeOfHeader) / MemoryLayout.BYTES_IN_LONG
-          )
```
WojciechMazur, line 37: This way on calculating reference field indexes could have lead to runtime issues in 32-bit runtime. It would lead to having double indexes (eg. offset=16 & offset=20 would point to the same index) and might omit scanning some of misaligned fields.

kubukoz:
> we're adding an utility to extract Class name as wide-string allowing to show it in debug information from the GC

<3

tools/src/main/scala/scala/scalanative/codegen/Generate.scala
```diff
@@ -523,28 +524,42 @@ object Generate {
...
+          .headOption
+          .fold((-1, -1, -1)) {
+            case (weakRef, weakRefReferantField) =>
+              // if WeakReferences are being compiled and therefore supported
+              val layout = meta.layout(weakRef)
+              val gcModifiedFieldReferentIdx = layout
+                .index(weakRefReferantField)
+                .ensuring(
+                  _ > 0,
+                  "Runtime implementaiton error, no \"_gc_modified_referent\" field in java.lang.ref.WeakReference"
```
kubukoz, line 551: **issue:** typo

```suggestion
                  "Runtime implementation error, no \"_gc_modified_referent\" field in java.lang.ref.WeakReference"
```

PR #3765, opened 2024-02-15
fix[gc]: Make sure conservative marking of ranges uses alligned pointers

WojciechMazur:
Unaligned pointer reads might lead to undefined behaviour, ensure to align them. 
Also disable sanitization of delimited continuations functions


nativelib/src/main/resources/scala-native/gc/immix/Marker.c
```diff
@@ -128,7 +128,12 @@ NO_SANITIZE static void Marker_markRange(Heap *heap, Stack *stack,
                                          word_t **from, word_t **to) {
     assert(from != NULL);
     assert(to != NULL);
-    for (word_t **current = from; current <= to; current += 1) {
+    // Align start address
+    const intptr_t alignmentMask = ~(sizeof(word_t) - 1);
+    word_t **alignedFrom = (word_t **)((intptr_t)from & alignmentMask);
+    // Align end address to be optionally 1 higher when unalligned
```
ekrich, line 134:
```suggestion
    // Align end address to be optionally 1 higher when unaligned
```

PR #3766, opened 2024-02-15
fix: Multiple runtime fixes/improvments

WojciechMazur:
* Fix order of runtime.Array.copy check for len=0
* Use java.lang.StringBuilder instead of collection.mutable.StringBuilder whenerver possible
* Remove Mac version check in time_nano.c and time.c - fails in MacOS 14 ARM GithubActions runners

### DIFF
--- a/javalib/src/main/scala/java/io/DataInputStream.scala
+++ b/javalib/src/main/scala/java/io/DataInputStream.scala
@@ -1,6 +1,7 @@
 package java.io
 
 import java.nio.ByteBuffer
+import java.{lang => jl}
 
 import scala.annotation.tailrec
 
@@ -124,7 +125,7 @@ class DataInputStream(in: InputStream)
     var v = in.read()
     if (v == -1) null
     else {
-      val builder = new java.lang.StringBuilder
+      val builder = new jl.StringBuilder
       var c = v.toChar
       while (v != -1 && c != '\n' && c != '\r') {
         builder.append(c)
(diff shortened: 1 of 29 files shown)


javalib/src/main/scala/java/io/DataInputStream.scala
```diff
@@ -124,7 +124,7 @@ class DataInputStream(in: InputStream)
     var v = in.read()
     if (v == -1) null
     else {
-      val builder = new StringBuilder
+      val builder = new java.lang.StringBuilder
```
ekrich, line 127: I always just alias `java.lang` -> `jl` and `java.util` -> `ju`. I think that is pretty nice unless it is adding extra symbols or something.

nativelib/src/main/scala/scala/scalanative/regex/MachineInput.scala
```diff
@@ -192,7 +192,9 @@ object MachineInput {
       hayStack match {
         case hayStack: String =>
           hayStack.indexOf(needle, pos)
-        case hayStack: StringBuilder =>
+        case hayStack: java.lang.StringBuilder =>
+          hayStack.indexOf(needle, pos)
+        case hayStack: collection.mutable.StringBuilder =>
```
ekrich, line 197: You want both here?

ekrich (commented):
Interesting that the macro check doesn't work anymore.

nativelib/src/main/scala/scala/scalanative/regex/MachineInput.scala
WojciechMazur: Yes, I don't want to change previous semantics, but rather adapt it to using java.lang.StringBuilder by default

PR #3790, opened 2024-02-26
[fix] - Fix spurious GC memory overrides

WojciechMazur:
This PR fixes problems with GC allocations leading to segmantation fault or ClassCastExceptions. They're coused by invalid marking of objects in Immix/Commix leading to reusing memory occupied by still reachable objects. The issue was the easiest to spot in `ConcurrentHashMap.testIteratorsAreWeaklyConsistent` test which was used to create lots of allocations in a loop. 
During debugging we've spot some additional issue which were fixed: 
- Under non-optimized builds we've observed illegal states, eg. Array with length/stride = 0, coused by triggered GC collection/yieldpoint before setting these 2 type informations - it was fixed by introducing a dedicated `GC.array_alloc` used to set length/stride inside GC.
- In release modes (when clang compilation with `-ON` where N > 0) is used) were leading to runtime errors (seg fault/ ClassCastException), but it worked correctly when using all optimization techniques in Interflow while keeping compilation with clang usng `-O0`. It was caused by incorrect scanning of registers - `jmpbuf` used on Arm64 was not storing pointers (method arguments? passed via registers) in word (8-bytes) alignment. To fix this we've modified `Marker_markRange` function to take explicit stride. We still use defualt word-size alignment for faster scanning.  
- In Immix added assertions for valid allocations - ported from Commix

Additional fixes to issues found during debugging: 
- SourceCodeCache now resolves relative sources in a sorted order  starting with the shortes paths - it was messing up empty package sources
- ObjectMonitors were previously copied to clones of objects leading to undefined behaviour, it was fixed by explicitly zeroing object-monitors upon clone in `java.lang.Object` and `java.lang.Array` base implementaitons

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/commix/Heap.c
+++ b/nativelib/src/main/resources/scala-native/gc/commix/Heap.c
@@ -219,9 +219,14 @@ void Heap_Init(Heap *heap, size_t minHeapSize, size_t maxHeapSize) {
 }
 
 void Heap_Collect(Heap *heap) {
+    MutatorThread *mutatorThread = currentMutatorThread;
+    // GC collect triggered during StopTheWorld in interruptible thread might
+    // lead to deadlock It's fine to interrupt thread if it's done before
+    // allocating memory
+    bool wasInterruptible = MutatorThread_setInterruptible(mutatorThread, true);
 #ifdef SCALANATIVE_MULTITHREADING_ENABLED
     if (!Synchronizer_acquire())
-        return;
+        goto done;
     while (!Sweeper_IsSweepDone(heap)) {
         // Unlock mutator threads list to allow registration of new threads
         // WriteLock has higher priority then ReadLock - it does NOT wait until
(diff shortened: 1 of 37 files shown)


nativelib/src/main/resources/scala-native/gc/commix/Heap.c
```diff
@@ -219,9 +219,14 @@ void Heap_Init(Heap *heap, size_t minHeapSize, size_t maxHeapSize) {
...
 void Heap_Collect(Heap *heap) {
+    MutatorThread *mutatorThread = currentMutatorThread;
+    // GC collect triggered during StopTheWorld in interruptible thread might
+    // lead to deadlock It's fine to interrupt thread if it's done before
+    // allocating memory
+    bool wasInterruptible = MutatorThread_setInterruptible(mutatorThread, true);
 #ifdef SCALANATIVE_MULTITHREADING_ENABLED
     if (!Synchronizer_acquire())
-        return;
+        goto done;
```
ekrich, line 229: Maybe just repeating the following with a `return;` is better unless you want a really long `else` or something?
```c
MutatorThread_setInterruptible(mutatorThread, wasInterruptible);
```

PR #3811, opened 2024-03-05
refactor: More explicit `NativeConfig.multithreading` setting

WojciechMazur:
* Rename `multithreadingSupport` to `multithreading` 
* Make `NativeThread.multithreading` `Option[Boolean]`, defaulting to `None`. When not defined assume multithreading is enabled, but detect if threads are used. If not, relink with disabled multithreading to improve performance.
* Remove control env variable `SCALANATIVE_DISABLE_UNUSED_MULTITHREADING`

### DIFF
--- a/.github/workflows/run-tests-linux-multiarch.yml
+++ b/.github/workflows/run-tests-linux-multiarch.yml
@@ -166,7 +166,7 @@ jobs:
                 .withCheck(true)
                 .withCheckFatalWarnings(true)
                 .withTargetTriple(sys.env.get("CROSS_TRIPLE"))
-                .withMultithreadingSupport(${{ matrix.gc != 'commix' }})
+                .withMultithreading(${{ matrix.gc != 'commix' }})
                 .withCompileOptions(_ ++ sysRoot ++ List("-fuse-ld=lld"))
                 .withLinkingOptions(_ ++ sysRoot ++ List("-fuse-ld=lld", "-latomic"))
             }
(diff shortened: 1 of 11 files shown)


WojciechMazur:
@armanbilge AFAIR you didn't like the previous detection, do you think that making it more explicit and not controlled by env variables might be beneficial here? I wanted to have a good balance between performance and development experience. 
Now if setting can be empty the behaviour shall not be so much suprising.

armanbilge:
> AFAIR you didn't like the previous detection

Hmm, sorry, not sure if I remember 😅 I remember talking about ENV variables in https://github.com/scala-native/scala-native/pull/3517#discussion_r1340356053 and in https://github.com/scala-native/scala-native/pull/2800#issuecomment-1229641785.

Still, I agree this is better.

> If not, relink with disabled multithreading to improve performance.

Sorry if I missed it, is that in this PR? Does it log a warning to the user? Linking twice seems expensive.

WojciechMazur:
>> If not, relink with disabled multithreading to improve performance.

> Sorry if I missed it, is that in this PR? Does it log a warning to the user? Linking twice seems expensive. 

This mechanism was already added, only thing that changed in this PR was the control over when to perform detection of threads usage. However, maybe I've used wrong names here - we detect usage of threads after initial class loading (loading NIR from files + its post processing). 
In general cost of class-loading from disk it relatively low, especially when compared to possible performance gains. At that point we have both warmed up JVM and cached NIR read from disk. 
First classloading of our tests takes ~4s and the whole linking process ~97s. In case of simple sandbox apps, first load is ~1s, and second one with disabled multithreading ~0.5s

armanbilge:
That makes more sense. Thanks for clarifying!

PR #3812, opened 2024-03-05
refactor: Restrict access to tools and nir types

WojciechMazur:
Makes most of toolchain internal definitions package private and update docs to point users to NativeConfig API

### DIFF
--- a/docs/user/sbt.md
+++ b/docs/user/sbt.md
@@ -120,15 +120,13 @@ nativeConfig ~= { c =>
 | 0.1    | `publish`               | `Unit`          | Similar to standard publish with addition of NIR (1)                          |
 | 0.1    | `nativeLink`            | `File`          | Link NIR and generate native binary                                           |
 | 0.4.0  | `nativeConfig`          | `NativeConfig`  | Configuration of the Scala Native plugin                                      |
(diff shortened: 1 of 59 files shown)


docs/user/sbt.md
```diff
@@ -126,25 +119,16 @@ nativeConfig ~= { c =>
...
+For the details of available `NativeConfig` options see [API](https://javadoc.io/doc/org.scala-native/tools_3/latest/scala/scalanative/build/NativeConfig.html)
 
-1.  See [Publishing](#publishing) and [Cross
-    compilation](#cross-compilation) for details.
+1.  See [Publishing](#publishing) and [Cross compilation](#cross-compilation) for details.
 2.  See [Compilation modes](#compilation-modes) for details.
 3.  See [Garbage collectors](#garbage-collectors) for details.
-4.  See [Link-Time Optimization (LTO)](#link-time-optimization-lto) for
-    details.
+4.  See [Link-Time Optimization (LTO)](#link-time-optimization-lto) for details.
```
JD557, line 131: I think points 2, 3 and 4 are not referenced anywhere anymore, so leaving this here (at least formatted like this) could be a bit confusing.

PR #3815, opened 2024-03-07
fix: Execute `WeakReference` post-cleanup callbacks only in `java.lang.Thread`

WojciechMazur:
Executing callbacks in GC created thread might have lead to undefined behaviour due to lack of `Thread.currentThread()` which might have been called by handler.


WojciechMazur:
There shall be green CI light yet again. 
This PR should fix some of the remaining spurious CI issues, especially SIGILL found on MacOS in release mode

Issue #2483, opened 2021-12-24
`sbt test` fails on subproject without any tests when it depends on module which is using `scalatests` for tests and `sbt-crossproject`


catap:
I'm using `scalatests` and it isn't implemented at all subproject in my main repository where I'm using scala-native. And have quite a complicated build. It includes `sbt-crossproject` and main code is builds for different paltforms and different scala version.

One of subproject is using 2.13 and haven't used any tests. But it depends on a project which uses (a lot of) tests.

Usually I run tests as `sbt +test` and after  updating from `0.4.0` to `0.4.2` I've discovered that this project had fails. It fails with symptoms:
1. `-out` binary consume 100% of CPU and never ends;
2. It fails when sbt tries to run tests like that:
```
[error] Fatal signal 11 caught
[error]   at scala.scalanative.testinterface.signalhandling.SignalConfig$$$Lambda$1.(Unknown Source)
[error]   at <none>._sigtramp(Unknown Source)
[error]   at <none>._ZN9libunwind22CompactUnwinder_x86_64INS_17LocalAddressSpaceEE32stepWithCompactEncodingFramelessEjyRS1_RNS_16Registers_x86_64Eb(Unknown Source)
[error]   at <none>._ZN9libunwind12UnwindCursorINS_17LocalAddressSpaceENS_16Registers_x86_64EE4stepEv(Unknown Source)
[error]   at java.lang.StackTrace$.currentStackTrace(Unknown Source)
```

(2) doesn't produce any issues and sbt continue as is. From antoher hand (1) exists after sbt is crashed  and I've discovered this bug because my laptop was very noisy when it just rests on the table and should do nothing.

The simplest reproducer is:
```
import sbtcrossproject.CrossPlugin.autoImport.crossProject

ThisBuild / scalaVersion := "2.12.15"
crossScalaVersions := Seq()

lazy val bugCore = crossProject(NativePlatform)
  .in(file("bugCore"))
  .settings(
    libraryDependencies ++= Seq(
      "org.scalatest" %%% "scalatest" % "3.2.10" % Test,
    )
  )

lazy val bug = crossProject(NativePlatform)
  .in(file("bug"))
  .dependsOn(bugCore)
  .nativeSettings(
    nativeMode := "release-fast",
    nativeLTO := "thin"
  )
```

[...]

PR #3821, opened 2024-03-08
Support shutdown hooks with signals

ekrich:
Now you can use Ctrl-C or `kill -s SIGTERM <pid>` and shutdown hooks will be called and the process will exit gracefully.

Here is the `sandbox` I used - not sure how to test.
```scala
object Test {
    println("Shutdown test - use Ctrl-C or send SIGTERM")
    // Runtime
    //   .getRuntime()
    //   .addShutdownHook(new Thread() {
    //     override def run(): Unit = messageThread()
    //   })
    // or this
    sys.addShutdownHook(messageThread())
    while (true) {
      try Thread.sleep(1000)
      catch {
        case e: InterruptedException => {
          println("Interrupted")
        }
      }
    }
  }
  def messageThread() = (0 to 5).foreach { i =>
    println("Shutting down")
    Thread.sleep(500)
  }
}
```

### DIFF
--- a/javalib/src/main/scala/java/lang/Runtime.scala
+++ b/javalib/src/main/scala/java/lang/Runtime.scala
@@ -20,6 +20,9 @@ class Runtime private () {
     stdlib.atexit(() => Runtime.getRuntime().runHooks())
   }
 
+  // https://docs.oracle.com/en/java/javase/21/docs/specs/man/java.html
+  // Currently, we use C lib signals so SIGHUP is not covered for POSIX platforms.
+
   lazy val setupSignalHandler = {
     signal.signal(signal.SIGINT, handleSignal(_))
     signal.signal(signal.SIGTERM, handleSignal(_))
--- a/javalib/src/main/scala/java/lang/Runtime.scala
+++ b/javalib/src/main/scala/java/lang/Runtime.scala
@@ -54,7 +54,7 @@ class Runtime private () {
       .asInstanceOf[Array[Thread]]
       .sorted(Ordering.by[Thread, Int](-_.getPriority()))
     hooks.foreach { t =>
-      t.setUncaughtExceptionHandler(ShutdownHookUncoughExceptionHandler)
+      t.setUncaughtExceptionHandler(ShutdownHookUncaughtExceptionHandler)
     }
     // JDK specifies that hooks might run in any order.
     // However, for Scala Native it might be beneficial to support partial ordering
@@ -83,7 +83,7 @@ class Runtime private () {
         try t.run()
         catch {
           case ex: Throwable =>
-            ShutdownHookUncoughExceptionHandler.uncaughtException(t, ex)
+            ShutdownHookUncaughtExceptionHandler.uncaughtException(t, ex)
         }
       }
   }
@@ -123,7 +123,7 @@ class Runtime private () {
     exec(Array(cmd), envp, dir)
 }
 
-private object ShutdownHookUncoughExceptionHandler
+private object ShutdownHookUncaughtExceptionHandler
     extends Thread.UncaughtExceptionHandler {
   def uncaughtException(t: Thread, e: Throwable): Unit = {
     System.err.println(s"Shutdown hook $t failed, reason: $e")
--- a/javalib/src/main/scala/java/lang/Runtime.scala
+++ b/javalib/src/main/scala/java/lang/Runtime.scala
@@ -27,6 +27,7 @@ class Runtime private () {
 
   private def handleSignal(sig: CInt): Unit = {
     Runtime.getRuntime().runHooks()
+    exit(0)
   }
 
   private def ensureCanModify(hook: Thread): Unit = if (shutdownStarted) {
@@ -89,9 +90,11 @@ class Runtime private () {
   private def runHooks() = {
     import scala.scalanative.meta.LinktimeInfo.isMultithreadingEnabled
     hooks.synchronized {
-      shutdownStarted = true
-      if (isMultithreadingEnabled) runHooksConcurrent()
-      else runHooksSequential()
+      if (!shutdownStarted) {
+        shutdownStarted = true
+        if (isMultithreadingEnabled) runHooksConcurrent()
+        else runHooksSequential()
+      }
     }
   }
 
--- a/javalib/src/main/scala/java/lang/Runtime.scala
+++ b/javalib/src/main/scala/java/lang/Runtime.scala
@@ -3,6 +3,7 @@ package java.lang
 import java.io.File
 import java.util.{Set => juSet}
 import java.util.Comparator
+import scala.scalanative.libc.signal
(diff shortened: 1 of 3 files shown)


javalib/src/main/scala/java/lang/Runtime.scala
```diff
@@ -67,7 +79,7 @@ class Runtime private () {
       .asInstanceOf[Array[Thread]]
       .sorted(Ordering.by[Thread, Int](-_.getPriority()))
       .foreach { t =>
-        try t.run()
+        try t.start()
```
WojciechMazur, line 82: We should revert this change, `runHooksSequential` is run only in the singlethreaded mode

javalib/src/main/scala/java/lang/Runtime.scala
```diff
@@ -81,6 +93,7 @@ class Runtime private () {
       if (isMultithreadingEnabled) runHooksConcurrent()
       else runHooksSequential()
     }
+    exit(0)
```
WojciechMazur, line 96: What if the shutdown was started by unhandled thread or signal? In such case we would report successful exit during program failure

WojciechMazur (changes requested):
Overall LGTM, only 2 things I don't think are correct

javalib/src/main/scala/java/lang/Runtime.scala
ekrich: Ok, I didn't understand that one because you normally call `start()` on a thread.

javalib/src/main/scala/java/lang/Runtime.scala
ekrich: This one I put in because it was hanging after the shutdown hooks ran and I had no idea how to solve that.

ekrich:
This seems to work now. Avoid race without calling shutdown hooks twice and call `exit` after handling shutdown hooks. Not sure if this is 100% correct but it seems ok.

ekrich:
I'll need to fix this but not sure what value to put.
```
[error] Test scala.scalanative.linker.MinimalRequiredSymbolsTest.multithreading failed: java.lang.AssertionError: 
[error] Found more symbols then expected, config={debugMetadata=false, multithreading=true, targetTriple=x86_64-unknown-unknown}:
[error] Expected at most: {types=1050, members=6600, total=7650}
[error] Found:            {types=1028, members=6640, total=7668}
[error] Diff:             {types=-22, members=40, total=18}
```

javalib/src/main/scala/java/lang/Runtime.scala
```diff
@@ -18,6 +19,17 @@ class Runtime private () {
...
     stdlib.atexit(() => Runtime.getRuntime().runHooks())
   }
+
+  lazy val setupSignalHandler = {
+    signal.signal(signal.SIGINT, handleSignal(_))
+    signal.signal(signal.SIGTERM, handleSignal(_))
+  }
+
+  private def handleSignal(sig: CInt): Unit = {
+    Runtime.getRuntime().runHooks()
```
WojciechMazur, line 29: I think it won't work when we'll have multiple threads running. Starting `runHooks` would also start `scala.scanative.runtime.JoinNonDaemonThreads` which would wait for all thread to finish. We should at least remove this shutdown hook. 
Also signal handler might be run by any of the threads. Which might affect its execution.

WojciechMazur (changes requested):
For the tools tests it's fine to just update the values to the number of found symbols/classes + alignment to `% 50` so small changes won't affect it much. Maybe we should remove this test. 

Also, this use cases seems to be good candidate for scripted test. We probably should be able to create a custom sbt task to start compiled program (hopefully using multitple threads running in the lop) and create a task to run `Process.destroy` - I guess it should be sending the SIGINT / SIGTERM signal

catap:
Can it help with https://github.com/scala-native/scala-native/issues/2483 ?

javalib/src/main/scala/java/lang/Runtime.scala
```diff
@@ -18,6 +19,17 @@ class Runtime private () {
   lazy val setupAtExitHandler = {
     stdlib.atexit(() => Runtime.getRuntime().runHooks())
   }
+
+  lazy val setupSignalHandler = {
+    signal.signal(signal.SIGINT, handleSignal(_))
+    signal.signal(signal.SIGTERM, handleSignal(_))
+  }
```
catap, line 26:
```suggestion
  lazy val setupSignalHandler = {
    signal.signal(signal.SIGHUP, handleSignal(_))
    signal.signal(signal.SIGINT, handleSignal(_))
    signal.signal(signal.SIGQUIT, handleSignal(_))
    signal.signal(signal.SIGTERM, handleSignal(_))
  }
```
let's add a few more expected reason to quit

ekrich: As far as I understand "The JVM runs shutdown hooks only in case of normal terminations." - what this seems to mean is Ctrl-C `SIGINT` and `SIGTERM`. This is also in the current Test comments.

catap: > The JVM catches signals to implement shutdown hooks for unexpected termination. The JVM uses SIGHUP, SIGINT, and SIGTERM to initiate the running of shutdown hooks.
> Applications embedding the JVM frequently need to trap signals such as SIGINT or SIGTERM, which can lead to interference with the JVM signal handlers. The -Xrs option is available to address this issue. When -Xrs is used, the signal masks for SIGINT, SIGTERM, SIGHUP, and SIGQUIT aren't changed by the JVM, and signal handlers for these signals aren't installed.

https://docs.oracle.com/en/java/javase/21/docs/specs/man/java.html

ekrich: Ok, I'll add `SIGHUP`.

ekrich: On second thought, we need to try and get this to work properly this way first as we are using `clib` and `SIGHUP` is not a member. I second pass could use `posixlib` and something special for Windows to make it JVM compliant.

catap: Keep in mind that `SIGHUP` is a signal which is send when SSH connection is hand up due network issue.

Without it, shutdown hook at CLI utilite won't work on this case and it's show stoper for real life use case I guess.

ekrich: No problem but we have nothing now and to make it cross platform is more complex so it might be better to have something now.

WojciechMazur (approved):
LGTM for the single threaded execution, I have prepared a variant handling the multithreaded case including handling on the GC side. I'll add the follow up PR soon

ekrich:
Thanks, I was looking into the multi-threaded version and it was quite a bit more involved.

PR #3823, opened 2024-03-09
Add `pledge` and `unveil`

catap:
`pledge` is using to restrict system operations, and `unveil` to unveil parts of a restricted filesystem view.

Both of them available at OpenBSD for years. Anyway some work to port it to Linux is onging but it far from the end, and the best status is Justine's work to port it into cosmopolitan libc.

### DIFF
--- a/posixlib/src/main/scala/scala/scalanative/posix/unistd.scala
+++ b/posixlib/src/main/scala/scala/scalanative/posix/unistd.scala
@@ -216,6 +216,10 @@ object unistd {
 
   @blocking def write(fildes: CInt, buf: CVoidPtr, nbyte: CSize): CInt = extern
 
+  def pledge(promises: CString, execpromises: CString): CInt = extern
+
+  def unveil(path: CString, permissions: CString): CInt = extern
+
 // Symbolic constants
 
   // NULL, see POSIX stddef


posixlib/src/main/resources/scala-native/unistd.c
```diff
@@ -34,6 +34,24 @@
...
 #endif // __FreeBSD__
 
+// seems that pledge and unveil is OpenBSD,
+// some work to port it to Linux is onging but it far from the end,
+// and the best status is Justine's work to port it into cosmopolitan.
+#ifndef __OpenBSD_
+
+#include <errno.h>
+
+int pledge(const char *promises, const char *execpromises) {
```
WojciechMazur, line 44: We should not try to redefine it here under the original name, instead let's define it as `scalanative_pledge` and redirect Scala to it by adding `@name("scalanative_pledge")` annotation. Same for `unveil` function

WojciechMazur (changes requested):

catap:
After some thinking I'd like to remove C-level placeholder. Let linker fail if user tries to use this function on unsupported OS.

This is much better and user friendly behaviour that return undocummented error via errno.

I'll update PR shortly

WojciechMazur (approved):

WojciechMazur (approved):

PR #3831, opened 2024-03-11
Expose scheduler policy

catap:


posixlib/src/main/scala/scala/scalanative/posix/pthread.scala
```diff
@@ -393,4 +393,21 @@ object pthread {
...
+  def SCHED_RR: CInt = extern
+
+  @name("scalanative_sched_batch")
+  def SCHED_BATCH: CInt = extern
+
+  @name("scalanative_sched_idle")
+  def SCHED_IDLE: CInt = extern
+
+  @name("scalanative_sched_deadline")
+  def SCHED_DEADLINE: CInt = extern
```
WojciechMazur, line 412: The non-POSIX values should be documented, especially since they can lead to runtime failures https://stackoverflow.com/questions/43167963/setting-scheduling-policy-sched-idle-or-sched-batch-fails-with-einval-22

A simple scaladoc comment `/** Not defined in POSIX standard, might lead to runtime errors */`  for each of them should be enough

WojciechMazur (changes requested):

posixlib/src/main/scala/scala/scalanative/posix/pthread.scala
catap: Fair enough. I also moved this constant into the rigth place, near `sched_get_priority_max` where it should be.

posixlib/src/main/scala/scala/scalanative/posix/sched.scala
```diff
@@ -41,6 +41,30 @@ object sched {
...
+  @name("scalanative_sched_other")
+  def SCHED_OTHER: CInt = extern
+
+  /** Not defined in POSIX standard, might lead to runtime errors */
+  @name("scalanative_sched_fifo")
+  def SCHED_FIFO: CInt = extern
+
+  /** Not defined in POSIX standard, might lead to runtime errors */
+  @name("scalanative_sched_rr")
+  def SCHED_RR: CInt = extern
```
WojciechMazur, line 54: This 3 are mentioned in the GNU Posix so the comment is incorrect: 
https://man7.org/linux/man-pages/man3/pthread_attr_setschedpolicy.3.html  

> The supported values for policy are SCHED_FIFO, SCHED_RR, and
       SCHED_OTHER, with the semantics described in [sched(7)](https://man7.org/linux/man-pages/man7/sched.7.html).

catap: Seems they are inside POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_getschedparam.html

Thus, it contains one more optional scheduler policy `SCHED_SPORADIC` which I also included.

WojciechMazur (approved):

PR #3841, opened 2024-03-12
Allow to run github action tests manually

catap:
Right now for a contributor the only way to test his work on different platform is to create PR.

Some changes can be too early to create PR, or a kind of test.

So, here I allow to run test actions via github UI on desired branch against user's worker quota.

An example of usage:
image

### DIFF
--- a/.github/workflows/run-tests-linux-multiarch.yml
+++ b/.github/workflows/run-tests-linux-multiarch.yml
@@ -5,6 +5,7 @@ on:
   schedule:
     # Every day at 2 AM UTC
     - cron: "0 2 * * *"
+  workflow_dispatch:
 concurrency:
   group: linux-multiarch-${{ github.head_ref }}-${{ github.event_name }}
   cancel-in-progress: true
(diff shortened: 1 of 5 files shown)


.github/workflows/run-tests-linux-multiarch.yml
```diff
@@ -13,7 +14,6 @@ jobs:
   # Build testing image that would be used to build and run against different platforms
   # Currently only Linux x64 is tested
   build-image:
-    if: "github.event_name == 'pull_request' || ((github.event_name == 'schedule' || github.event_name == 'workflow_call') && github.repository == 'scala-native/scala-native')"
```
WojciechMazur, line 16: Removing this would trigger scheduled builds in all forks of the project, we need to keep it verbose unfortunetlly so that the nightly builds which enable some additional tests would not be started outside the main repo 
`github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' || (...)`

```

Comment (WojciechMazur) [reply .github/workflows/run-tests-linux-multiarch.yml:16]: In fact they're still triggered but are stopped quickly: https://github.com/WojciechMazur/scala-native/actions?query=event%3Aschedule

Comment (catap) [reply .github/workflows/run-tests-linux-multiarch.yml:16]: @WojciechMazur do you mind to reduce this if to:
```
    if: "!(github.event_name == 'schedule' && github.repository != 'scala-native/scala-native')"
```
that should make it easy to support and enable schedule build only for main repository.

but current way probably cleaner and much easy to understand and support.

WojciechMazur: sure, looks good

catap: @WojciechMazur after some thought I decided to move forward with old way. Suggested way isn't clear and may lead to future errors when someone tries to change it.

WojciechMazur (approved):

PR #3844, opened 2024-03-14
Expose API to run one iteration of event loop

catap:
This API should be used to mix Scala-Native's event loop with external one like `select`, `poll`, etc or a wraper from `libub` or `libev` and so on.

Right now such integration requires quite ugly and very frigiel hack which grand access to the queue via `RawPtr`.

We had a few PRs with attemt to implement in, for example see: https://github.com/scala-native/scala-native/pull/1852

So, here I made things quite trivial and without any refactoring.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/package.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/package.scala
@@ -119,6 +119,11 @@ package object runtime {
    */
   @noinline def loop(): Unit = ExecutionContext.loop()
 
+  /** Run the runtime's event loop once and return true when internal queue has
+   *  more Runnable to run.
+   */
+  @noinline def loopRunOnce(): Boolean = ExecutionContext.loopRunOnce()
+
   /** Called by the generated code in case of division by zero. */
   @noinline
   private[scalanative] def throwDivisionByZero(): Nothing =
--- a/nativelib/src/main/scala/scala/scalanative/runtime/ExecutionContext.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/ExecutionContext.scala
@@ -14,8 +14,8 @@ object ExecutionContext {
 
   private val queue: ListBuffer[Runnable] = new ListBuffer
 
-  private[runtime] def loop(): Unit = {
-    while (queue.nonEmpty) {
+  private[runtime] def loopRunOnce(): Boolean = {
+    if (queue.nonEmpty) {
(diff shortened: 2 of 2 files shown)


nativelib/src/main/scala/scala/scalanative/runtime/package.scala
```diff
@@ -119,6 +119,11 @@ package object runtime {
    */
   @noinline def loop(): Unit = ExecutionContext.loop()
 
+  /** Run the runtime's event loop once and return true when internal queue has
+   *  more Runnable to run.
+   */
+  @noinline def loopRunOnce(): Boolean = ExecutionContext.loopRunOnce()
```
WojciechMazur, line 125: Can we use some more meaningful name? Maybe some out of 
`dequeueAndExecute`, `runNext`, `execute`, `interleave` or maybe something else. `loopRunOnce` doesn't seem to be a good name. 
Also would be useful to also port the tests from the original PR. Especially to ensure we don't use multithreaded execution context by mistake.

WojciechMazur, line 125: Personally I think we should not introduce new methods to `runtime` package especially since QueueExecutionContext would now probably be a niche in context of having a real multithreading. If so, it would be better to add this methods to `runtime.ExecutionContext`.

WojciechMazur, line 125: I guess we might need also to revisit #3144

armanbilge: Yes, see my comment on that PR. Any changes that we make to these APIs should be considering that if Scala stdlib introduces a `Scheduler` API then single-threaded Scala Native will be able to implement an event loop that also supports timers / asynchronous sleeping.

- https://github.com/scala-native/scala-native/pull/3144#discussion_r1097757732

catap: @WojciechMazur idea of this PR is kind of shortcut for #3144 which is much better, but discussion for this API was here for years.

So, I've tried minimal hack to avoid very ugly code to play with raw-ptr.

If you plan to move forward your PR I'd like to close this one, but I not sure that reimplement it here is the right move.

Rename some method, do not expose it more than needed -- good point, but I suggest to keep it as simple and small as we can.

catap:
I close it because the right PR was merged

PR #3849, opened 2024-03-23
improvement: Limit amount of compiled code, restrict glue layer only referenced files.

WojciechMazur:
By using `@define` annotation we can limit amount of glue layer that needs to be compiled - we add an entry for potentially optional bindings and guard the glue-layer with `#if defined(x)` preprocesor conditions. This should allow to lower number of compilation errors in non-posix compliant systems, especially on 32-bit architectures.

The need of controlling the amount of C glue layer files on non-compatible runtimes was already needed in the WASM/WASI prototype https://github.com/WojciechMazur/scala-native/tree/scala-wasm and when targeting Playdate devices (arm v7, limited C/POSIX stdlib) https://github.com/kubukoz/scala-native/pull/1 

Based on these changes we're removing usages of `stdatomic` in minimal single-threaded execution

### DIFF
--- a/clib/src/main/resources/scala-native/complex.c
+++ b/clib/src/main/resources/scala-native/complex.c
@@ -1,4 +1,4 @@
-#if defined(__SCALANATIVE_C_COMPLEX)
+#if defined(SCALANATIVE_COMPILE_ALWAYS) || defined(__SCALANATIVE_C_COMPLEX)
 #include <complex.h>
 
 #if defined(_WIN32)
--- a/clib/src/main/resources/scala-native/complex.c
+++ b/clib/src/main/resources/scala-native/complex.c
@@ -1,3 +1,4 @@
+#if defined(__SCANATIVE_C_COMPLEX)
 #include <complex.h>
 
 #if defined(_WIN32)
@@ -201,3 +202,4 @@ float scalanative_crealf(float snfc[2]) { return crealf(toFloatComplex(snfc)); }
 double scalanative_creal(double sndc[2]) {
     return creal(toDoubleComplex(sndc));
 }
+#endif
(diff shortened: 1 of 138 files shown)


ekrich:
One idea I had was perhaps we could have different plugin versions or allow the plugin to remove the `posixlib` dependency for example on certain platforms.

clib/src/main/resources/scala-native/complex.c
```diff
@@ -1,3 +1,4 @@
+#if defined(__SCANATIVE_C_COMPLEX)
```
ekrich, line 1:
```suggestion
#if defined(__SCALANATIVE_C_COMPLEX)
```

clib/src/main/scala/scala/scalanative/libc/complex.scala
```diff
@@ -23,7 +23,9 @@ import scalanative.unsafe._
  */
 @extern object complex extends complex
 
-@extern private[scalanative] trait complex {
+@extern
+@define("__SCANATIVE_C_COMPLEX")
```
ekrich, line 27:
```suggestion
@define("__SCALANATIVE_C_COMPLEX")
```

ekrich (commented):
Just this one spelling issue.

clib/src/main/resources/scala-native/complex.c
ekrich: Looks like a duplicate?

WojciechMazur: Thank you, I don't think it's duplicate, there is one type in preprocessor guards and one in @define annotation

WojciechMazur:
> One idea I had was perhaps we could have different plugin versions or allow the plugin to remove the posixlib dependency for example on certain platforms. 

I think it might quite a complex feature to make it correct. Also some of posixlib definitions might be used on Windows, because some of them might be compatible, eg. sockets. In general control of what to be compiled shall be more stable then removing possibly unused ones.

kubukoz:
OMG how did I miss this before? This is amazing!

PR #3852, opened 2024-03-28
fix: Prevent deadlocks  when mixing threads with queue based execution context

WojciechMazur:
* Prevent deadlocks or pre-early exit when mixing threads with queue based execution context
* NativeThread.registry.aliveThreads now returns iterator instead of array
* Use thread-safe queue in QueueExecutionContext - there is always 1 consumer, but there can be multiple producers

Can help with resolving #3848 - the scaltest framework snapshot might have been adding new tasks to the queue.


ekrich:
@WojciechMazur This is very nice. What is the purpose of naming a class `package.state.scala`? Not familiar with that convention / choice.

ekrich:
Thank you for the explanation. I hadn't seen that naming before and the info about `package object` and access control is good info too. Fine tuned design is an art and I am not really at that level.

PR #3855, opened 2024-03-28
Use lower case file names on Windows

avdv:
This PR changes header names in include directives and names of library names in lib pragmas to use lower case.

The reason to do this is that the header files and libraries are provided in lower case when installing a cross compiler (e.g. zig cc or llvm-mingw) on Linux targeting Windows.

Currently, as a work-around I added `-isystem src/main/c` to the compile options, with the `src/main/c` directory containing many header files looking like this:
```Windows.h
// Windows.h
#include <windows.h>
```

- Convert Windows header names to lower case
- Include windows.h after declaring `WIN32_LEAN_AND_MEAN`
- Use lower case Windows library names

### DIFF
--- a/javalib/src/main/resources/scala-native/time_millis.c
+++ b/javalib/src/main/resources/scala-native/time_millis.c
@@ -1,7 +1,7 @@
 #include <time.h>
 #if defined(_WIN32)
 #define WIN32_LEAN_AND_MEAN
-#include <Windows.h>
+#include <windows.h>
 #include "win_freq.h"
 #else
 #include <stdio.h>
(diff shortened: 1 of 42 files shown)


ekrich (approved):
Super great that you are doing this important work. Do you think we should change these based on best practices? Not sure which is preferred but maybe they should be consistent.
```c
// from
#ifdef _WIN32
// to
#if defined(_WIN32)
```
I think this looks good if CI passes.

WojciechMazur (approved):

avdv:
> Do you think we should change these based on best practices? Not sure which is preferred but maybe they should be consistent.
> 
> ```c
> // from
> #ifdef _WIN32
> // to
> #if defined(_WIN32)
> ```

For simple cases I often prefer using `ifdef`. Although in general, I think consistency is nice but if this is changed it should be enforced by some linting rules so you don't introduce these inconsistencies in the future again.

PR #3860, opened 2024-04-02
fix: Deadlocks in singlethreaded execution context

### LINKED ISSUE #3859: Deadlock in future without multithreading
A trivial and quite expected code can be deadlocked if build it without multithreadin.

I use  a06f46dcb9c5aa2cbb52aff298bd551f71dcfbe2 as local root and had changed definition of sandbox module to disable multithreading like this:
```diff
diff --git a/project/Build.scala b/project/Build.scala
index 26be043c2..b204cf8b7 100644
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -671,7 +671,11 @@ object Build {
       .enablePlugins(MyScalaNativePlugin)
       .withNativeCompilerPlugin
       .withJUnitPlugin
-      .settings(noJavaReleaseSettings)
+      .settings(nativeConfig ~= { c =>
+          import scala.scalanative.build._
+          c.withMultithreading(false)
+        },
+        noJavaReleaseSettings)
       .dependsOn(scalalib, javalib, testInterface % "test")
 
 // Testing infrastructure ------------------------------------------------
```

Now, I made a trivial example of code
```scala
import scala.util._

import scala.concurrent._
import scala.concurrent.duration._

object Test {
  implicit val ec: ExecutionContext = ExecutionContext.global

    Await.result(Future.successful(1) map (_ + 1), Duration.Inf)
  }
}
```
which is deadlocked. But when multithreading is enabled, it works.

If I remove part with `map (_ + 1)` it works as well without multithreading.

WojciechMazur:
Fixes #3859 
Based on work in #3853, kept as draft until deciding the new API for Native ExecutionContext

The only changes related to this fix are in the commit 516444b03f14f2ebc546b2753ef9b523fc674bb2

### DIFF
--- a/javalib/src/main/scala/scala/scalanative/runtime/Proxy.scala
+++ b/javalib/src/main/scala/scala/scalanative/runtime/Proxy.scala
@@ -1,6 +1,8 @@
-package scala.scalanative.runtime
+package scala.scalanative
+package runtime
 
 import scala.scalanative.annotation.alwaysinline
+import scala.concurrent.duration.FiniteDuration
 
 object Proxy {
   @alwaysinline
@@ -22,4 +24,10 @@ object Proxy {
 
   def disableGracefullShutdown(): Unit =
     MainThreadShutdownContext.gracefully = false
+
+  def stealWork(maxSteals: Int): Unit =
+    concurrent.NativeExecutionContext.queueInternal.stealWork(maxSteals)
+  def stealWork(timeout: FiniteDuration): Unit =
+    concurrent.NativeExecutionContext.queueInternal.stealWork(timeout)
+
 }
--- a/nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
@@ -12,6 +12,8 @@ import scala.annotation.nowarn
 
 import java.util.concurrent.ConcurrentHashMap
 import java.{util => ju}
+import scala.scalanative.concurrent.NativeExecutionContext
+import scala.concurrent.duration._
 
 trait NativeThread {
   import NativeThread._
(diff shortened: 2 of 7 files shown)


catap:
So, no need for user to make their own interface after all :)

nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
```diff
@@ -42,16 +44,22 @@ trait NativeThread {
...
   final def park(): Unit =
-    park(0, isAbsolute = false)
+    if (isMultithreadingEnabled) park(0, isAbsolute = false)
+    else NativeExecutionContext.queueInternal.helpComplete()
 
   @alwaysinline
   final def parkNanos(nanos: Long): Unit = if (nanos > 0) {
-    park(nanos, isAbsolute = false)
+    if (isMultithreadingEnabled) park(nanos, isAbsolute = false)
+    else NativeExecutionContext.queueInternal.stealWork(nanos.nanos)
```
JD557, line 53: I'm not sure if this works as expected, as `stealWork(FiniteDuration)` rounds down to the millisecond, so in a lot of situations this won't yield.

I guess this matches what's said on the javadoc, but I also see that the `WindowsThread` uses `millis.max(1).toUInt`... I don't know, maybe this is OK. 🤷

catap: If we use JDK as reference implementation I’d like to point that small delay is a kind of edge case.

see: https://kirill.korins.ky/articles/sleeping-at-jvm-thread-for-a-very-short-period/ where I kept notes on that investigation

javalib/src/main/scala/scala/scalanative/runtime/Proxy.scala
```diff
@@ -1,6 +1,8 @@
-package scala.scalanative.runtime
+package scala.scalanative
+package runtime
 
 import scala.scalanative.annotation.alwaysinline
+import scala.concurrent.duration.FiniteDuration
 
 object Proxy {
```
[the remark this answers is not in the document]
WojciechMazur: Right, this one should be fixed, but I'll do it in follow up PR. It's actually more dangerous since it's defined in javalib which does not emit Bytecode

Issue #3076, opened 2023-01-07
Windows fails to link sandbox, Visual Studio 2022


ekrich:
This is a placeholder issue. It would be nice to nice easy instructions to install Windows tools.

Previous working setup is described in the Scala Native docs using Visual Studio Community 2019 and LLVM vendored version 12.0.1.

Current library setup is Visual Studio Build Tools 2022, version 17.4.3

Scala Native HEAD checked out.

Details:
Edition	Windows 10 Enterprise
Version	22H2
Installed on	‎5/‎26/‎2022
OS build	19045.2364
Experience	Windows Feature Experience Pack 120.2212.4190.0

Windows installed Clang/LLVM
```shell
> clang --version
clang version 15.0.1
Target: i686-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\Llvm\bin
```
```scala
[info] Checking intermediate code (475 ms)
[info] Dumping intermediate code (linked) (1301 ms)
[info] Discovered 627 classes and 4191 methods
[info] Optimizing (debug mode) (2267 ms)
[info] Checking intermediate code (109 ms)
[info] Dumping intermediate code (optimized) (550 ms)
[info] Dumping intermediate code (lowered) (834 ms)
[info] Generating intermediate code (3338 ms)
[info] Produced 36 files
[info] Compiling to native code (59069 ms)
[info] channels.ll.o : error LNK2019: unresolved external symbol _WriteFile referenced in function __SM33java.nio.channels.FileChannelImplD5writeLAb_iiuEO
[info]   Hint on symbols that are defined and could potentially match:
[info]     _WriteFile@20
[info] file.ll.o : error LNK2019: unresolved external symbol _GetLastError referenced in function __SM31java.nio.file.WindowsException$D6onPathL16java.lang.StringL19java.io.IOExceptionEO
[info]   Hint on symbols that are defined and could potentially match:
[info]     _GetLastError@0
[info] windows.ll.o : error LNK2019: unresolved external symbol _GetStdHandle referenced in function __SM40scala.scalanative.windows.ConsoleApiExt$D5stdInL28scala.scalanative.unsafe.PtrEO
[info]   Hint on symbols that are defined and could potentially match:
[info]     _GetStdHandle@4
[...]

PR #3882, opened 2024-04-17
Experiment using clang-cl with MSVC

ekrich:
This just hacks the tools but could be a configuration when having a target triple `env` of `msvc`.

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/shared/ScalaNativeGC.h
+++ b/nativelib/src/main/resources/scala-native/gc/shared/ScalaNativeGC.h
@@ -12,12 +12,13 @@
 #pragma comment(lib, "kernel32.lib")
 #include <windows.h>
 typedef DWORD ThreadRoutineReturnType;
+typedef ThreadRoutineReturnType (WINAPI *ThreadStartRoutine)(void *);
 #else
 #include <pthread.h>
 typedef void *ThreadRoutineReturnType;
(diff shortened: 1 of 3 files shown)


ekrich:
See https://clang.llvm.org/docs/UsersManual.html#clang-cl

Also see This https://github.com/scala-native/scala-native/issues/3076

Repeated here as well.

I tried hacking the build to compile using `clang-cl` which seems to be recommended for MSVC but I end up with the following. See PR branch https://github.com/scala-native/scala-native/pull/3882

```scala
[info] Dumping intermediate code (lowered) (1159 ms)
[error] C:/Users/<user>/workspace/scala-native/sandbox/.2.13/target/scala-2.13/native/dependencies/nativelib_native0.5.0-SNAPSHOT_2.13-0.5.0-SNAPSHOT-0/scala-native/eh.cpp(19,37): error: cannot use 'throw' with exceptions disabled        
[error]    19 | void scalanative_throw(void *obj) { throw scalanative::ExceptionWrapper(obj); }
[error]       |                                     ^
[error] 1 error generated.
[info] Produced 40 LLVM IR files
[info] Generating intermediate code (with debug metadata) (225945 ms)
[info] Compiling to native code (205032 ms)
[info] Total (245084 ms)
```

ekrich:
This experiment is complete - it is too far off the beaten path and I have a potential solution for Windows 32bit linking now.

PR #3884, opened 2024-04-17
ThreadUtil: don't use thread APIs if multithreading&commix aren't in use

### LINKED ISSUE #3877: (playdate support) ThreadUtil: missing APIs
Ref: https://github.com/scala-native/scala-native/discussions/3876

In `nativelib/src/main/resources/scala-native/gc/shared/ThreadUtil.c`, there's some usage of APIs that aren't available on Playdate (`pthread`):

```
ld: libroot.a(dependencies_nativelib_native0.5_3-0_scala-native_gc_shared_ThreadUtil.c.o): in function `mutex_init':
(...)nativelib_native0.5_3-0/scala-native/gc/shared/ThreadUtil.c:56:(.text.mutex_init+0x14): undefined reference to `pthread_mutexattr_init'
ld: (...)nativelib_native0.5_3-0/scala-native/gc/shared/ThreadUtil.c:57:(.text.mutex_init+0x1c): undefined reference to `pthread_mutexattr_settype'
ld: (...)nativelib_native0.5_3-0/scala-native/gc/shared/ThreadUtil.c:58:(.text.mutex_init+0x24): undefined reference to `pthread_mutex_init'
ld: libroot.a(dependencies_nativelib_native0.5_3-0_scala-native_gc_shared_ThreadUtil.c.o): in function `mutex_lock':
(...)nativelib_native0.5_3-0/scala-native/gc/shared/ThreadUtil.c:67:(.text.mutex_lock+0x2): undefined reference to `pthread_mutex_lock'
ld: libroot.a(dependencies_nativelib_native0.5_3-0_scala-native_gc_shared_ThreadUtil.c.o): in function `mutex_unlock':
(...)nativelib_native0.5_3-0/scala-native/gc/shared/ThreadUtil.c:85:(.text.mutex_unlock+0x2): undefined reference to `pthread_mutex_unlock'
```

Cleaned up further:

```
undefined reference to `pthread_mutexattr_init'
undefined reference to `pthread_mutexattr_settype'
undefined reference to `pthread_mutex_init'
undefined reference to `pthread_mutex_lock'
undefined reference to `pthread_mutex_unlock'
```

What can we do about those? Would it make sense to hardcode `true` if multithreading is disabled (as I've been doing so far)?

kubukoz:
Closes #3877

### DIFF
--- a/nativelib/src/main/resources/scala-native/gc/shared/ThreadUtil.c
+++ b/nativelib/src/main/resources/scala-native/gc/shared/ThreadUtil.c
@@ -2,6 +2,20 @@
 #include <stdio.h>
 #include <limits.h>
 
+#if !(defined SCALANATIVE_GC_COMMIX ||                                         \
+      defined SCALANATIVE_MULTITHREADING_ENABLED)
+
+// If neither multithreading nor Commix GC is enabled, there's no concurrency
+// so we don't need to involve actual mutexes.
+
+INLINE
+bool mutex_init(mutex_t *ref) { return true; }
+
+INLINE bool mutex_lock(mutex_t *ref) { return true; }
+
+INLINE bool mutex_unlock(mutex_t *ref) { return true; }
+#else
+
 INLINE
 bool thread_create(thread_t *ref, routine_fn routine, void *data) {
 #ifdef _WIN32
@@ -160,3 +174,6 @@ bool rwlock_unlockRead(rwlock_t *ref) {
     return pthread_rwlock_unlock(ref) == 0;
 #endif
 }
+
+#endif
+//(defined SCALANATIVE_GC_COMMIX || defined SCALANATIVE_MULTITHREADING_ENABLED)


nativelib/src/main/resources/scala-native/gc/shared/ThreadUtil.c
```diff
@@ -2,6 +2,20 @@
...
+// If neither multithreading nor Commix GC is enabled, there's no concurrency
+// so we don't need to involve actual mutexes.
+
+INLINE
+bool mutex_init(mutex_t *ref) { return true; }
+
+INLINE bool mutex_lock(mutex_t *ref) { return true; }
+
+INLINE bool mutex_unlock(mutex_t *ref) { return true; }
+#else
```
kubukoz, line 17: **question:** I only needed these three so I implemented only these. Perhaps we should do this on a per-function basis?

nativelib/src/main/resources/scala-native/gc/shared/ThreadUtil.c
```diff
@@ -2,6 +2,20 @@
 #include <stdio.h>
 #include <limits.h>
 
+#if !(defined SCALANATIVE_GC_COMMIX ||                                         \
+      defined SCALANATIVE_MULTITHREADING_ENABLED)
```
kubukoz, line 6: **thought:** it'd be good to have a single definition that we can check, e.g. "has pthread" - does something like this exist already?

kubukoz: also, is this fine or should I flip the boolean (concurrent case first, fallback second)?

ekrich: By default, all the `shared` code should be compiled so something is not correct 100%. The `SCALANATIVE_GC_COMMIX` is defined only when selecting Commix so that code will get compiled and similar for other GCs. Does this help at all?

ekrich: It is used by all the internal GCs and also single threaded runtime because no guard against.

kubukoz: hmm yeah. What I did here was inspired by [Wojciech's comment on the ticket](https://github.com/scala-native/scala-native/issues/3877#issuecomment-2060536437), but I see that this might not be the best way to do it because it's not really "shared" anymore.

PR #3908, opened 2024-05-07
Respect LLVM_BIN when invoking llvm-config

keynmol:
This prevents polluting compilation/linking flags with whatever different LLVM you have installed globally.

I noticed this because in bindgen my globally installed llvm-config was adding old flags to the NativeConfig.


ekrich (approved):
LGTM - thanks. Looks like the failure is unrelated.

WojciechMazur (approved):

PR #3953, opened 2024-06-08
nix flake defined development environment

coreyoconnor:
Provides a `nix develop` environment that should everything necessary to build, test and performance measure.  

This largely just replaces `scripts/scala-native.nix` with a nix flake. 

This does not define scala-native as an installable package. This only defines a `devShell` for `nix develop`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,81 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1667395993,
+        "narHash": "sha256-nuEHfE/LcWyuSWnS8t12N1wc105Qtau+/OdUAjtQ0rA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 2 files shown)


coreyoconnor:
I (just now) ran across https://github.com/typelevel/typelevel-nix

Which has a better approach for setting up a development environment. `sbt-derivation` is more for packaging applications. Switching to that.

Can still use `sbt-derivation` later if needed.

ekrich:
@coreyoconnor Maybe some info in the docs for Nix aficionados would be good? Docs, always get the least attention.

coreyoconnor:
> @coreyoconnor Maybe some info in the docs for Nix aficionados would be good? Docs, always get the least attention.

Just closed this one to open a PR from a different branch :) https://github.com/scala-native/scala-native/pull/4318

I could have re-directed this one but got lazy on the command line.

That said: General docs on how to use scala native with nix would be nice. Along with developer docs. I'll add a note about that to the other PR.

PR #3955, opened 2024-06-10
Reproducing #3956 (infinite loop in ThreadLocal)

durban:
See #3956.


WojciechMazur:
Thank you for reproducer, fixed in #3966

PR #3966, opened 2024-06-16
fix[javalib]: Fix calculation of initial maximal load of ThreadLocal.Values table

### LINKED ISSUE #3956: ThreadLocal.Values#getAfterMiss can get into an infinite loop
See #3955 for reproduction. Run e.g., `tests2_13/testOnly org.scalanative.testsuite.javalib.lang.ThreadLocalTest` in sbt, and the test added in that PR gets into an infinite loop.

WojciechMazur:
Was incorrectly using the doubled size as a base thus preventing growing the table entry over the INITIAL_SIZE=16 
Fixes #3956

### DIFF
--- a/javalib/src/main/scala/java/lang/ThreadLocal.scala
+++ b/javalib/src/main/scala/java/lang/ThreadLocal.scala
@@ -45,11 +45,11 @@ object ThreadLocal {
 
     /** Size must always be a power of 2.
      */
-    private val INITIAL_SIZE = 16
+    private final val INITIAL_SIZE = 16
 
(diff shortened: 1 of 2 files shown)


sideeffffect:
Can we expect a release with this fix anytime soon? :pray: It would unblock at least one PR https://github.com/circe/circe/pull/2254

WojciechMazur:
I plan to make a release this week, although now I'm mostly focused on the quick delivery of backports to Scala LTS. These tend to have the highest priority right now as we're already behind the schedule a lot.

PR #3975, opened 2024-06-22
Only include delimcc on platforms that support continuations

### LINKED ISSUE #3885: (playdate support) delimcc.c is compiled even if Continuations isn't used
To reproduce:

1. Add this line to delimcc.c, just to make it clear that the code is being included
```diff
diff --git a/nativelib/src/main/resources/scala-native/delimcc.c b/nativelib/src/main/resources/scala-native/delimcc.c
index bd00eb6aa..dd0597d1e 100644
--- a/nativelib/src/main/resources/scala-native/delimcc.c
+++ b/nativelib/src/main/resources/scala-native/delimcc.c
@@ -1,4 +1,6 @@
 #if defined(SCALANATIVE_COMPILE_ALWAYS) || defined(__SCALANATIVE_DELIMCC)
+
+#error "this shouldn't happen!"
 #include "delimcc.h"
 #include <stddef.h>
 #include <stdio.h>

```
2. Publish a snapshot
3. Try to link an app/library, even a "hello world" or a no-op main.

Output:

```
[info] Linking (multithreadingEnabled=true, disable if not used) (322 ms)
[info] Discovered 907 classes and 5623 methods after classloading
[info] Checking intermediate code (quick) (4 ms)
[info] Multithreading was not explicitly enabled - initial class loading has not detected any usage of system threads. Multithreading support will be disabled to improve performance.
[info] Linking (multithreadingEnabled=false) (277 ms)
[info] Discovered 774 classes and 4690 methods after classloading
[info] Checking intermediate code (quick) (2 ms)
[info] Discovered 749 classes and 3668 methods after optimization
[info] Optimizing (debug mode) (156 ms)
[info] Produced 43 LLVM IR files
[info] Generating intermediate code (413 ms)
[error] /Users/kubukoz/projects/demos/target/scala-3.3.3/native/dependencies/nativelib_native0.5_3-0/scala-native/delimcc.c:3:2: error: "this shouldn't happen!"
[error] #error "this shouldn't happen!"
[error]  ^
[error] 1 error generated.
[info] Compiling to native code (1524 ms)
[error] stack trace is suppressed; run last Compile / nativeLink for the full output
[error] stack trace is suppressed; run last Compile / nativeLink for the full output
[error] stack trace is suppressed; run last Compile / nativeLink for the full output
[error] stack trace is suppressed; run last Compile / nativeLink for the full output
[error] stack trace is suppressed; run last Compile / nativeLink for the full output
[error] stack trace is suppressed; run last Compile / nativeLink for the full output
[info] Total (2502 ms)
[error] Failed to compile /Users/kubukoz/projects/demos/target/scala-3.3.3/native/dependencies/nativelib_native0.5_3-0/scala-native/delimcc.c
[error] (Compile / nativeLink) Failed to compile /Users/kubukoz/projects/demos/target/scala-3.3.3/native/dependencies/nativelib_native0.5_3-0/scala-native/delimcc.c
```

Example app setup (requires a publish of `0.5.1-SNAPSHOT`): https://github.com/kubukoz/demos/compare/sn-continuations-repro?expand=1

JD557:
Fix #3885 by only including delimcc on platforms that support continuations.

In turn, this will only initialize continuations in platforms that support them.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -45,6 +45,10 @@ object LinktimeInfo {
   )
   def isWeakReferenceSupported: Boolean = resolved
 
+  @resolvedAtLinktime()
+  def isContinuationsSupported: Boolean =
+    isLinux || isMac || isFreeBSD || isOpenBSD || isNetBSD
+
   @resolvedAtLinktime(
     "scala.scalanative.meta.linktimeinfo.isMultithreadingEnabled"
   )
--- a/nativelib/src/main/scala-3/scala/scalanative/runtime/Continuations.scala
+++ b/nativelib/src/main/scala-3/scala/scalanative/runtime/Continuations.scala
@@ -6,7 +6,7 @@ import scala.scalanative.runtime.Intrinsics.*
 import scala.collection.mutable
 
 import scala.util.Try
-import scala.scalanative.meta.LinktimeInfo.isWindows
+import scala.scalanative.meta.LinktimeInfo.isContinuationsSupported
 
 object Continuations:
   import Impl.*
@@ -24,7 +24,8 @@ object Continuations:
    *  We want our implementation to use the allocations done by
    *  `Continuation.alloc`, so that the GC is aware of the stack fragment.
    */
-  Impl.init(CFuncPtr2.fromScalaFunction(allocateBlob))
+  if isContinuationsSupported then
+    Impl.init(CFuncPtr2.fromScalaFunction(allocateBlob))
 
   /** Marks the given body as suspendable with a `BoundaryLabel` that `suspend`
    *  can refer to. Forwards the return value of `body`, or the return value of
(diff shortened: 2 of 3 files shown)


nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
```diff
@@ -45,6 +45,10 @@ object LinktimeInfo {
   )
   def isWeakReferenceSupported: Boolean = resolved
 
+  @resolvedAtLinktime()
+  def isContinuationsSupported: Boolean =
```
JD557, line 49: Not sure if this should be `areContinuationsSupported`... but then shouldn't the above check also be `areWeakReferencesSupported`?

Also, hopefully the list of platforms here is correct. I assumed so because previously only Windows was excluded.

WojciechMazur: The plural variant seems more correct, but we won't replace the already existing `isWeakReferenceSupported` for compatibility reasons. So going with either singular (for keeping convention) or plural (for correctness) is fine with me

WojciechMazur (approved):

Issue #4018, opened 2024-08-09
Platform runtime metrics


iRevive:
I'm a maintainer of the [otel4s](https://github.com/typelevel/otel4s) project. otel4s provides the cross-platform implementation of the OpenTelemetry specification. Hence, application metrics can be collected while using the Scala Native platform.

Currently, I'm exploring options to collect platform-specific metrics.

JVM [runtime metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/jvm-metrics.md) are the following:
- CPU: `jvm.cpu.time`, `jvm.cpu.count`, `jvm.cpu.recent_utilization`
- Classes:  `jvm.class.loaded`, `jvm.class.unloaded`, `jvm.class.count`
- GC: `jvm.gc.duration` - duration per GC action type
- Memory: `jvm.memory.used`, `jvm.memory.committed` - usage by type: heap / non-heap
- Thread: `jvm.thread.count` - count per state

------

I'm new to the Scala Native codebase, but I briefly checked what can be done:

### CPU

Nothing there yet.

### Classes

I guess this metric is irrelevant in the Scala Native world.

### GC

Both [Immix](https://github.com/scala-native/scala-native/blob/v0.5.4/nativelib/src/main/resources/scala-native/gc/immix/Stats.h) and [Commix](https://github.com/scala-native/scala-native/blob/v0.5.4/nativelib/src/main/resources/scala-native/gc/commix/Stats.h) have `Stats` classes to write metrics to the file. I guess it can be extended to some degree to expose the metrics.

### Memory

Both [Immix](https://github.com/scala-native/scala-native/blob/v0.5.4/nativelib/src/main/resources/scala-native/gc/immix/Heap.h) and [Commix](https://github.com/scala-native/scala-native/blob/v0.5.4/nativelib/src/main/resources/scala-native/gc/commix/Heap.h) have `heapSize` and `maxHeapSize` which can be utilized. 

Zero ideas about non-heap memory. 

### Thread

No idea, to be honest.

-----

OpenTelemetry Java relies on the `*MXBeans` (e.g. [MemoryMXBean](https://docs.oracle.com/javase/8/docs/api/java/lang/management/MemoryMXBean.html)) from the `java.lang.management` package for metrics collections.
Would providing a minimal implementation for the `java.lang.management` package make sense? 
For example, `MemoryMXBean` can use the Immix or Commix interface to provide memory usage details. 

[...]

PR #4021, opened 2024-08-13
feature [nativelib]: implement `MemoryMXBean` and `ManagementFactory`

iRevive:
The first step towards https://github.com/scala-native/scala-native/issues/4018.

### Scope

1) Implement `java.lang.management.MemoryMXBean`
2) Implement `java.lang.management.MemoryUsage`
3) Implement `java.lang.management.ManagementFactory.getMemoryMXBean()`

### DIFF
--- a/javalib/src/main/scala/java/lang/management/ManagementFactory.scala
+++ b/javalib/src/main/scala/java/lang/management/ManagementFactory.scala
@@ -1,0 +1,20 @@
+package java.lang.management
+
+object ManagementFactory {
+
+  private lazy val MemoryBean = MemoryMXBean()
+
+  /** Returns the memory-specific bean.
+   *
+   *  @example
+   *    {{{
+   *  val memoryBean = getMemoryMXBean()
+   *  println(s"current heap: $${memoryBean.getHeapMemoryUsage().getUsed()}")
+   *
+   *  val list = List.fill(Short.MaxValue)(0) // allocate memory
+   *  println(s"current heap: $${memoryBean.getHeapMemoryUsage().getUsed()}")
+   *    }}}
+   */
+  def getMemoryMXBean(): MemoryMXBean = MemoryBean
+
+}
--- a/javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
+++ b/javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
@@ -7,9 +7,6 @@ trait MemoryMXBean {
   /** Returns the current runtime memory usage.
    *
    *  @note
-   *    the committed memory is always -1
-   *
-   *  @note
    *    the memory usage values may differ between invocations
    */
   def getHeapMemoryUsage(): MemoryUsage
@@ -27,13 +24,24 @@ object MemoryMXBean {
     new Impl
 
   private class Impl extends MemoryMXBean {
-    def getHeapMemoryUsage(): MemoryUsage =
-      new MemoryUsage(
-        GC.getInitHeapSize().toLong,
-        GC.getUsedHeapSize().toLong,
-        -1L,
-        GC.getMaxHeapSize().toLong
-      )
+
+    /** The committed memory = used memory.
+     *
+     *  By @WojciechMazur: This one might be more difficult to track. On Windows
+     *  we do manually commit memory provided by VirtualAlloc, but in the case
+     *  of Unix we typically only reserve a maximal chunk of memory using mmap
+     *  and committing is done automatically by the system on the first usage of
+     *  every page. Access to this memory is only limited by Immix/Commix Heap
+     *  blocks limit. I guess the best choice would be to fill it with usedHeap
+     *  which is still correct regarding Javadocs: committed will always be
+     *  greater than or equal to used.
+     */
+    def getHeapMemoryUsage(): MemoryUsage = {
+      val init = GC.getInitHeapSize().toLong
+      val used = GC.getUsedHeapSize().toLong
+      val max = GC.getMaxHeapSize().toLong
+      new MemoryUsage(init, used, used, max)
+    }
 
     def gc(): Unit =
       System.gc()
--- a/javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
+++ b/javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
@@ -1,0 +1,42 @@
+package java.lang.management
+
+import scala.scalanative.runtime.GC
+
+trait MemoryMXBean {
+
+  /** Returns the current runtime memory usage.
+   *
+   *  @note
+   *    the committed memory is always -1
+   *
+   *  @note
+   *    the memory usage values may differ between invocations
+   */
+  def getHeapMemoryUsage(): MemoryUsage
+
+  /** Runs the garbage collector.
+   *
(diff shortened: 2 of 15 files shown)


javalib/src/main/scala/java/lang/management/ManagementFactory.scala
```diff
@@ -0,0 +1,20 @@
...
+   *  @example
+   *    {{{
+   *  val memoryBean = getMemoryMXBean()
+   *  println(s"current heap: ${memoryBean.getHeapMemoryUsage().getUsed()}")
+   *
+   *  val list = List.fill(Short.MaxValue)(0) // allocate memory
+   *  println(s"current heap: ${memoryBean.getHeapMemoryUsage().getUsed()}")
+   *    }}}
+   */
+  def getMemoryMXBean(): MemoryMXBean = MemoryBean
```
iRevive, line 18: OpenJDK implementation - [ManagementFactory.java](https://github.com/openjdk/jdk/blob/master/src/java.management/share/classes/java/lang/management/ManagementFactory.java#L345-L347).

javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
```diff
@@ -0,0 +1,42 @@
+package java.lang.management
+
+import scala.scalanative.runtime.GC
+
+trait MemoryMXBean {
```
iRevive, line 5: OpenJDK implementation - [MemoryMXBean.java](https://github.com/openjdk/jdk/blob/master/src/java.management/share/classes/java/lang/management/MemoryMXBean.java).

A few differences: 
1) The trait doesn't extend `PlatformManagedObject` 
2) The following methods aren't implemented: `getObjectPendingFinalizationCount`, `getNonHeapMemoryUsage`, `isVerbose`, `setVerbose`

javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  private[management] def apply(): MemoryMXBean =
+    new Impl
+
+  private class Impl extends MemoryMXBean {
+    def getHeapMemoryUsage(): MemoryUsage =
+      new MemoryUsage(
+        GC.getInitHeapSize().toLong,
+        GC.getUsedHeapSize().toLong,
+        -1L,
```
iRevive, line 34: I didn't find a simple way to get (calculate) the committed memory across all garbage collectors.

nativelib/src/main/resources/scala-native/gc/none/gc.c
```diff
@@ -52,6 +52,8 @@ size_t scalanative_GC_get_max_heapsize() {
     return Parse_Env_Or_Default("GC_MAXIMUM_HEAP_SIZE", getMemorySize());
 }
 
+size_t scalanative_GC_get_used_heapsize() { return -1L; }
```
iRevive, line 55: Is there a way to get the total allocated memory? 
I guess we can add a static variable and explicitly increment it on allocations. But is there a better way perhaps?

nativelib/src/main/resources/scala-native/gc/boehm/gc.c
```diff
@@ -61,6 +61,16 @@ size_t scalanative_GC_get_max_heapsize() {
...
 }
 
+size_t scalanative_GC_get_used_heapsize() {
+    struct GC_prof_stats_s *stats =
+        (struct GC_prof_stats_s *)malloc(sizeof(struct GC_prof_stats_s));
+    GC_get_prof_stats(stats, sizeof(struct GC_prof_stats_s));
+    size_t heap_sz = stats->heapsize_full;
+    size_t unmapped_bytes = stats->unmapped_bytes;
+    free(stats);
+    return heap_sz - unmapped_bytes;
```
iRevive, line 71: Hm. Could we use `GC_get_heap_size()` instead?

javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
keynmol: @iRevive open jdk is GPL licensed, you cannot use its sources as a reference.

ekrich: Yes, just use the Javadoc as a specification. You can also write tests that can run against the JVM and Native just to make sure the values make sense in general.

javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
WojciechMazur: This one might be more difficult to track. On Windows we do manually commit memory provided by VirtualAlloc, but in the case of Unix we typically only reserve a maximal chunk of memory using mmap and committing is done automatically by the system on the first usage of every page. Access to this memory is only limited by Immix/Commix Heap blocks limit. 
I guess the best choice would be to fill it with usedHeap which is still correct regarding Javadocs: `committed will always be greater than or equal to used`

nativelib/src/main/resources/scala-native/gc/boehm/gc.c
WojciechMazur: Probably it's thread-safe alternative might be better to use https://github.com/ivmai/bdwgc/blob/c3c44d85081044707eb8fa4f0db39e1fe6834853/include/gc/gc.h#L841-L850 
However, current solutions seems to be fine.

WojciechMazur, line 71: You might consider allocating the struct on the stack
```suggestion
    struct GC_prof_stats_s stats = {};
    GC_get_prof_stats(&stats, sizeof(struct GC_prof_stats_s));
    size_t heap_sz = stats.heapsize_full;
    size_t unmapped_bytes = stats.unmapped_bytes;
    return heap_sz - unmapped_bytes;
```

nativelib/src/main/resources/scala-native/gc/none/gc.c
WojciechMazur: Allocated memory is not tracked in none GC (it works by always allocating until there is no memory left) Adding a static variable to track allocations should do the job

javalib/src/main/scala/java/lang/management/MemoryMXBean.scala
iRevive: Thanks for the explanations! Let's fill committed memory with the amount of used heap.

nativelib/src/main/resources/scala-native/gc/none/gc.c
iRevive: I added the static `TOTAL_ALLOCATED` variable to track the allocated memory.

nativelib/src/main/resources/scala-native/gc/boehm/gc.c
iRevive: Done, thanks.

iRevive:
Are tests flaky, or did I break something?

WojciechMazur:
> Are tests flaky, or did I break something?

One of the recently added tests was probably flaky. I was hopefully fixed on main right already, but is not applicable here without the rebase.

WojciechMazur (approved):

PR #4028, opened 2024-08-18
feature [nativelib]: add `runtimeVersion` to the `LinktimeInfo`

iRevive:
### Motivation

The `runtimeVersion` is a handy attribute to add to the metrics. For example, otel4s can use this property while providing [runtime attributes](https://github.com/typelevel/otel4s/blob/ebb319a02dc418054c3b265965e8a7ecf8fe4acf/sdk/common/native/src/main/scala/org/typelevel/otel4s/sdk/resource/ProcessRuntimeDetectorPlatform.scala#L37).

### Why don't use a `BuildInfo` plugin to get the Scala Native version? 

The compiler and runtime versions may differ. For example, the `otel4s` library can be compiled with Scala Native `0.5.1`, while the executable (application) will be compiled with `0.5.3`.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -60,6 +60,11 @@ object LinktimeInfo {
   )
   def contendedPaddingWidth: Int = resolved
 
+  @resolvedAtLinktime(
+    "scala.scalanative.meta.linktimeinfo.runtimeVersion"
+  )
+  def runtimeVersion: String = resolved
+
   object target {
     @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.target.arch")
     def arch: String = resolved
--- a/unit-tests/native/src/test/scala/scala/scalanative/meta/LinktimeInfoTest.scala
+++ b/unit-tests/native/src/test/scala/scala/scalanative/meta/LinktimeInfoTest.scala
@@ -1,5 +1,6 @@
 package scala.scalanative.meta
 
+import scala.scalanative.buildinfo.ScalaNativeBuildInfo
 import scala.scalanative.runtime.Platform
 
 import org.junit.Test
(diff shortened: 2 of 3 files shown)


unit-tests/native/src/test/scala/scala/scalanative/meta/LinktimeInfoTest.scala
```diff
@@ -11,6 +12,10 @@ class LinktimeInfoTest {
     assertEquals(LinktimeInfo.debugMode, !LinktimeInfo.releaseMode)
   }
 
+  @Test def testVersion(): Unit = {
+    assertEquals(LinktimeInfo.runtimeVersion, ScalaNativeBuildInfo.version)
```
iRevive, line 16: The test assumes that `nir.Versions.current` and `ScalaNativeBuildInfo.version` are always in sync. Is that correct?

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
```diff
@@ -60,6 +60,11 @@ object LinktimeInfo {
   )
   def contendedPaddingWidth: Int = resolved
 
+  @resolvedAtLinktime(
+    "scala.scalanative.meta.linktimeinfo.runtimeVersion"
+  )
+  def runtimeVersion: String = resolved
```
iRevive, line 66: Maybe `version` or `scalaNativeVersion` would be more suitable?

unit-tests/native/src/test/scala/scala/scalanative/meta/LinktimeInfoTest.scala
WojciechMazur: Yes, it's correct

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
WojciechMazur: I think that current wording is good enough.

WojciechMazur (approved):

PR #4034, opened 2024-08-20
feature [nativelib]: partially implement `RuntimeMXBean`

iRevive:
A follow-up to the #4018. 

### Scope

- Extend `scalanative.runtime`: add `uptime` and `startTime`, record process start time
- Partially implement `java.lang.management.RuntimeMXBean`
- Implement `java.lang.management.ManagementFactory.getRuntimeMXBean()`

### DIFF
--- a/javalib/src/main/scala/java/lang/management/RuntimeMXBean.scala
+++ b/javalib/src/main/scala/java/lang/management/RuntimeMXBean.scala
@@ -1,0 +1,131 @@
+package java.lang.management
+
+trait RuntimeMXBean {
+
+  /** Returns the process ID of the running virtual machine (process).
+   *
+   *  @since 10
+   */
+  def getPid(): Long
+
+  /** Returns the name of the running virtual machine (process).
+   */
+  def getName(): String
+
+  /** Returns the name of the virtual machine implementation.
+   *
+   *  The equivalent of `sys.props("java.vm.name")`.
+   */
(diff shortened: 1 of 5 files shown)


javalib/src/main/scala/java/lang/management/RuntimeMXBean.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    def getPid(): Long =
+      unistd.getpid()
+
+    def getName(): String = {
+      val pid = getPid()
+
+      // InetAddress.getLocalHost returns IP (e.g. /127.0.0.1) while we need the name
+      val hostname = {
+        val hostNameLength = 256.toUInt // SUSv2 255 + 1 for terminal NUL
```
ekrich, line 83: This should use POSIX `{HOST_NAME_MAX}` defined as `_SC_HOST_NAME_MAX` in `unistd.h`. There could be constants, defines, or config items that might not be defined as you go forward.

iRevive: Good point, thanks. I replaced the arbitrary value with `unistd._SC_HOST_NAME_MAX.toCSize.`

WojciechMazur (approved):

WojciechMazur (approved):
LGTM, only requires a rebase after merging other changes

PR #4035, opened 2024-08-21
feature [windowslib]: add `GetProcessTimes` to the `ProcessThreadsApi`

iRevive:
### Motivation

[GetProcessTimes](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getprocesstimes) lets us get the process CPU time on the Win platform. 

### Usage 

```scala
import scala.scalanative.windows.MinWinBaseApi.FileTimeStruct
import scala.scalanative.windows.MinWinBaseApiOps._
import scala.scalanative.windows.ProcessThreadsApi
    
val creationTime = stackalloc[FileTimeStruct]()
val exitTime = stackalloc[FileTimeStruct]()
val kernelTime = stackalloc[FileTimeStruct]()
val userTime = stackalloc[FileTimeStruct]()

val success = ProcessThreadsApi.GetProcessTimes(
  ProcessThreadsApi.GetCurrentProcess(),
  creationTime,
  exitTime,
  kernelTime,
  userTime
)

if (success) {
  val totalTime = kernelTime.fileTime + userTime.fileTime
  val nanos = totalTime.toLong * FileTimeOps.EpochInterval
  println(s"Current process CPU time: ${nanos}")
} else {
  println("Err")
}
```

### Questions

1) Where is the best place to add tests? I didn't find many Windows-specific tests. I guess I can implement tests similar to [ResourceTest](https://github.com/scala-native/scala-native/blob/v0.5.5/unit-tests/native/src/test/scala/org/scalanative/testsuite/posixlib/sys/ResourceTest.scala#L154-L183).


WojciechMazur:
> Where is the best place to add tests? I didn't find many Windows-specific tests. I guess I can implement tests similar to [ResourceTest](https://github.com/scala-native/scala-native/blob/v0.5.5/unit-tests/native/src/test/scala/org/scalanative/testsuite/posixlib/sys/ResourceTest.scala#L154-L183). 

Yes, you can use the similar concept. You can create a new package dedicated for windows tests at `unit-tests/native/src/test/scala/org/scalanative/testsuite/windowslib` (we use 1 dedicated project for runtime tests). 
Then when defining a test wraps it's body in `if (scalanative.meta.LinktimeInfo.isWindows) { ... }` - it would create a special linktime-resolve if statement that would be erased to `(): Unit` on Linux, and the body of `then` part only when building on Windows. This way when testing on Unix we would not try to link against non existing Windows API methods.

iRevive:
@WojciechMazur thanks for the hints. I added tests.

Issue #4040, opened 2024-08-28
Slow performance of Game of Life benchmark


WojciechMazur:
We've received the reports on poor performance when running Game of Life simulation https://github.com/scala/scala3/issues/21432#issuecomment-2314717328 
We should reproduce and profile the code to find out what are the current bottlenecks

JD557:
I was playing a bit with this and I think the bottlenecks are in the `ArrayBuffer` and in either string hashing or string interpolation.

Using an `Array[Cell]` instead of an `ArrayBuffer[Cell]` and storing the cells as a `Map[(Int, Int), Cell]` instead of a `Map[String, Cell]` seems to have quite some impact on Native, but not on the JVM.

Benchmark results
And here are the benchmark results:

<details><summary>Baseline</summary>
<pre>
JVM: #1500 - World Tick (L: 0,364; A: 0,558) - Rendering (L: 0,262; A: 0,349)
Native: #1500 - World Tick (L: 0.866; A: 1.876) - Rendering (L: 1.103; A: 1.646)
</pre>
</details>
<details><summary>With Array (improves World Tick)</summary>
<pre>
JVM: #1500 - World Tick (L: 0,328; A: 0,503) - Rendering (L: 0,254; A: 0,344)
Native: #1500 - World Tick (L: 0.434; A: 1.024) - Rendering (L: 1.167; A: 1.822)
</pre>
</details>
<details><summary>With (Int, Int) keys (Improves World Tick and Rendering)</summary>
<pre>
JVM: #1500 - World Tick (L: 0,389; A: 0,551) - Rendering (L: 0,274; A: 0,352)
Native: #1500 - World Tick (L: 0.400; A: 0.693) - Rendering (L: 0.555; A: 0.716)
</pre>
</details>

This is my diff:

```diff
diff --git a/scala/Play.scala b/scala/Play.scala
index 675fd06..81a13d4 100644
--- a/scala/Play.scala
+++ b/scala/Play.scala
@@ -10,7 +10,7 @@ object Play {
       height = World_Height,
     )
 
-    val minimal = System.getenv("MINIMAL") != null
+    val minimal = true//System.getenv("MINIMAL") != null
 
     if (!minimal) {
       println(world.render)
diff --git a/scala/World.scala b/scala/World.scala
index 2e65f82..c079743 100644
--- a/scala/World.scala
+++ b/scala/World.scala
@@ -11,7 +11,7 @@ class World(
   class LocationOccupied extends Exception { }
[...]

PR #4048, opened 2024-09-15
Optimize Integer.toString

JD557:
Related to #4040

Uses a simpler implementation of `Integer#toString`, which was quite faster on my benchmarks.
Maybe there's still room for optimization... The old implementation was quite complex, so I was a bit surprised that I was able to get faster results with a naive solution 🤷 

Benchmark code:
```scala
object Main {
  val numIter = 10000000

    println("Sanity Checks:")
    println(s"${0} / ${-123} / ${123} / ${-123456} / ${123456} / ${Int.MinValue} / ${Int.MaxValue}")

    locally {
      val start = System.currentTimeMillis()
      var res = 0
      var i = 0
      while (i < numIter) {
        res = res + s"$res".##
        i = i + 1
      }
      println("RES: " + res)
      println("Time: " + (System.currentTimeMillis() - start))
    }
  }
}
```

Results on 0.5.5:
```
Sanity Checks:
0 / -123 / 123 / -123456 / 123456 / -2147483648 / 2147483647
RES: -1548773655
Time: 1625
```

Results on 0.5.6-SNAPSHOT:
```
Sanity Checks:
0 / -123 / 123 / -123456 / 123456 / -2147483648 / 2147483647
RES: -1548773655
Time: 766
```

### DIFF
--- a/javalib/src/main/scala/java/lang/Integer.scala
+++ b/javalib/src/main/scala/java/lang/Integer.scala
@@ -418,9 +418,9 @@ object Integer {
   }
 
   def toString(i: scala.Int): String = {
-    if (i == 0) {"0"}
-    else if (i == java.lang.Integer.MIN_VALUE) {"-2147483648"}
-    else if (i == java.lang.Integer.MAX_VALUE) {"2147483647"}
+    if (i == 0) { "0" }
+    else if (i == java.lang.Integer.MIN_VALUE) { "-2147483648" }
+    else if (i == java.lang.Integer.MAX_VALUE) { "2147483647" }
     else {
       val negative = i < 0
       val bufferSize = if (i < 1000 && i > -1000) 4 else 11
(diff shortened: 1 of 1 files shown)


plokhotnyuk:
To speed up divisions by 10 you can use:
```scala
val next = (positiveValue * 429496730L >> 32).toInt
```

More efficient algorithms that require some lookup table: https://jk-jeon.github.io/posts/2022/02/jeaiii-algorithm/

JD557:
I avoided those tricks for now (the original code also implemented multiplication by 10 with 2 bitshifts and a sum).

In part because I wanted to keep the code simple, but also because my computer was acting out with OOM errors when building scala native, so it was quite hard to test changes.

However, shouldn't LLVM be able to take care of those well known cases? I haven't checked the byte code and won't be able to look into this today, though.

WojciechMazur (approved):
LGTM

PR #4051, opened 2024-09-18
[improve/nativelib] Better detection of `LinktimeInfo.isMac`

WojciechMazur:
Support alternative 'macosx` OS naming convention. Handle `unknown` vendor for custom builds

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -21,7 +21,7 @@ object LinktimeInfo {
 
   @resolvedAtLinktime
   def isMac: Boolean =
-    (target.vendor == "apple" || target.vendor == "unknown") &&
+    target.vendor == "apple" &&
       (target.os == "darwin" || target.os == "macosx")
 
   @resolvedAtLinktime
--- a/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
+++ b/nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
@@ -20,7 +20,9 @@ object LinktimeInfo {
   def isLinux: Boolean = target.os == "linux"
 
   @resolvedAtLinktime
-  def isMac: Boolean = target.vendor == "apple" && target.os == "darwin"
+  def isMac: Boolean =
+    (target.vendor == "apple" || target.vendor == "unknown") &&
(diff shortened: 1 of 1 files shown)


tanishiking (commented):
I'm wondering what motivates this change?
It seems that clang computes the more specific  target triple https://github.com/shiika-lang/shiika/issues/281, but the vendor should be still apple(?)

WojciechMazur:
> I'm wondering what motivates this change? It seems that clang computes the more specific target triple [shiika-lang/shiika#281](https://github.com/shiika-lang/shiika/issues/281), but the vendor should be still apple(?)

`os == macosx` was introduced to handle cases when on some devices it's used instead of darwin. There was an issue with that recently reported on the Discord channel. 

The `vendor == unknown` is my additional initiative. The purpose was to handle cases when using a custom target triple without setting `vendor == apple`. I'm not sure if it would be really useful

tanishiking:
Thanks, okay this conversation :eyes:

> Reid — 昨日 10:32
clang --version
Homebrew clang version 18.1.8
Target: arm64-apple-darwin23.6.0
Thread model: posix
InstalledDir: /opt/homebrew/opt/llvm/bin
However, based on documentation, I have set my target in the sbt build to arm64-apple-macosx11.0.0 - should I change it to the darwin23.6.0 suffix?
Did LLVM change the target name (macosx versus darwin)? 
https://discord.com/channels/632150470000902164/635668881951686686/1285775392644534302

> Wojciech Mazur — 昨日 17:14
> Thanks, based on that I've spotted 1 bug. I'll fix it soon.  Changing it back to darwing should make it work as a workaround. 
In general if your not cross-compiling for some different target you can leave the target triple empty, it would use the default one.


It seems like the he set the target triple, as described in [the document](https://github.com/scala-native/scala-native/blob/f5d7c5d419c11f04bb01eb11f17e34e8fa0ec3d3/docs/user/sbt.md?plain=1#L222). I think we should update the documentation to use `arm64-apple-darwin23.2.0` or something?

Accepting `target.os == "macosx"` looks good to me :+1:, since LLVM does the same (https://github.com/llvm/llvm-project/blob/ed8f78827895050442f544edef2933a60d4a7935/llvm/include/llvm/TargetParser/Triple.h#L520-L524). However, I'm not sure if it's a good idea to accept `unknown` as the vendor 🤔 (FYI `rustup target list` doesn't seem to support the `unknown` vendor for Darwin/iOS.)

tanishiking (approved):
LGTM 👏

PR #4055, opened 2024-09-19
Improve interflow inliner log messages

JD557:
Some minor nitpicks. I noticed that some checks were missing from the inliner logs, which made the message a bit confusing ("could not inline because:... chirps")

I also added some sizes to the callee/caller too big checks, as otherwise it's impossible to know if a change improved that or not.


WojciechMazur (approved):

WojciechMazur:
LGTM, only requires a lint fix. Please run `./scripts/scalafmt` and commit changes

JD557:
Sorry, I was on holiday and now I accidentally pushed some extra files.
I'll fix this in a jiffy.

PR #4066, opened 2024-09-29
feature [nativelib]: partially implement `GarbageCollectorMXBean`

iRevive:
A follow-up to the #4018.

Scope:
1) Extend `LinktimeInfo`: add `garbageCollector`
2) Extend `GC`: add `getStatsCollectionTotal` and `getStatsCollectionDurationTotal` 
3) Partially implement `GarbageCollectorMXBean`
4) Move `Time.c` and `Time.h` to `gc/shared`

_____

There is some overlap with the existing `Stats` functionality in `Immix` and `Commix` collectors. 
Since the primary goal of `Stats` is writing stats to the file, I decided to add new variables instead.


WojciechMazur (changes requested):
Overall looks good to me, however currently, collection of stats might not be reliable. `scalanative_GC_collect` method is mostly called by `System.gc()`. GC allocator however, can call Heap_colllect directly, thus skip part where we collect stats

nativelib/src/main/resources/scala-native/gc/commix/CommixGC.c
```diff
@@ -114,6 +114,14 @@ size_t scalanative_GC_get_max_heapsize() {
 
 size_t scalanative_GC_get_used_heapsize() { return Heap_getMemoryUsed(&heap); }
 
+size_t scalanative_GC_stats_collection_total() {
+    return heap.gcStats.collectionTotal;
+}
+
+size_t scalanative_GC_stats_collection_duration_total() {
```
ekrich, line 121: As I recall, our standard was to make these helper methods in C style with camel_case and lowercase letters.

nativelib/src/main/resources/scala-native/gc/immix/ImmixGC.c
```diff
@@ -11,12 +11,18 @@
...
 #include "immix_commix/Synchronizer.h"
 #endif
 #include "MutatorThread.h"
 #include <stdatomic.h>
 
+// The total (accumulated) number of GC runs
+static size_t GC_STATS_COLLECTION_TOTAL = 0L;
+// The total (accumulated) elapsed time in nanos of GC runs
+static size_t GC_STATS_COLLECTION_DURATION_TOTAL = 0L;
+
```
ekrich, line 25: Have you considered make these part of perhaps a shared `jmx.c` unit so they can be shared across GCs instead of duplicated? That might leave a space for future enhancements to go.

nativelib/src/main/resources/scala-native/gc/commix/Heap.c
```diff
@@ -1,4 +1,4 @@
-#include "immix_commix/utils/Time.h"
+#include "shared/Time.h"
```
ekrich, line 1: This is an excellent change so it can be used on all the GCs.

ekrich (changes requested):
Looks good.

I think we had changed all the helper functions to use lowercase across the codebase at one time all starting with `scalanative_` for namespacing concerns. Not sure when the `_GC_` got added. We have a variety of different styles in the C codebase but that was one we tried to standardize.

iRevive:
> I think we had changed all the helper functions to use lowercase across the codebase at one time all starting with `scalanative_` for namespacing concerns. Not sure when the `_GC_` got added. We have a variety of different styles in the C codebase but that was one we tried to standardize.

Relatively recently https://github.com/scala-native/scala-native/pull/3652.

nativelib/src/main/resources/scala-native/gc/immix/ImmixGC.c
```diff
@@ -74,7 +80,14 @@ INLINE void *scalanative_GC_alloc_array(Rtti *info, size_t length,
...
 }
 
-INLINE void scalanative_GC_collect() { Heap_Collect(&heap, &stack); }
+INLINE void scalanative_GC_collect() {
+    size_t start_ns = Time_current_nanos();
+    Heap_Collect(&heap, &stack);
+    size_t end_ns = Time_current_nanos();
+
+    GC_STATS_COLLECTION_TOTAL++;
+    GC_STATS_COLLECTION_DURATION_TOTAL += (end_ns - start_ns);
```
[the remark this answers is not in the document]
iRevive: Indeed. I moved measurements to the `Heap_Collect`.

nativelib/src/main/resources/scala-native/gc/commix/CommixGC.c
iRevive: I can make a follow-up PR to change `scalanative_GC_x` to `scalanative_gc_x` across all GCs.

nativelib/src/main/resources/scala-native/gc/immix/ImmixGC.c
iRevive: Good point, thanks for the suggestion. I added `jmx.h` and `jmx.c`.

nativelib/src/main/resources/scala-native/gc/commix/CommixGC.c
ekrich: I see that @WojciechMazur  changed that to match Boehm and we also use Boehm env vars so that is a good exception I think.

WojciechMazur (approved):

PR #4085, opened 2024-11-26
Replace single TraitDispatchTable with class specific itables to reduce size of binaries

WojciechMazur:
Replaces generation of big, single `TraitDispatchTable` that was growing a quadratic pace. Instead, not every class has an attached list of ITables.  I also makes the `class_has_trait` and `trait_has_trait` tables obsolete.  

In case of `tests3/Test` the binary size has dropped from ~59MB to ~42MB  
In most benchmarks the execution times seems to be not affected, with some exceptions: 
- GCBenchmarks - speedup from 76ms to 57ms
- Mandelbrot - speedup from 97ms to 85ms
- CDBenchmark - speedup from 14ms to 12ms
- Deltablue - slowdown from 0.197ms to to 0.209 ms (~5%)
- KMeans - slowdown from 32.5ms to 33.9ms

The size of small binaries might slightly increase - the cumulative size of all 16 benchmark binaries has risen from 32MB to 33MB 

* Every class defines 2 additional fields - itablesCount (`int32`) and itables (`{traitId: int32, methods: void*}`
* We introduce a step to build a graph of traits relativity. We introduce coloring of the graph so that 2 traits implemented by the same class need to have different colour 
* Each trait has an assigned unique id (int32): where the most significant bit is a trait marker (for quick is interface check) and 5 least significant bits represent the color 
* We try to construct a minimal perfect hash table for `itable` of size <0..32> so that there are no collosions based on `traitId & (size - 1)` 
* When resolving the function at runtime we and we can guarantee that all implementations of trait are using fast itables then we use a quick lookup of function to call `cls.itables[traitId & (size -1 )].methods[methodId]` 
* Otherwise on slow path we perfect a binary search to find a itable so that `itable[idx].traitId == traitId` 
* We use similar approach for quick `isIsntanceOf` check based on `traitId` associated with each `ITableEntry` 

Additional changes: 
* Add additional `interfacesCount` and `interfacecs` fields to `java.lang.Class` layout - these are used to implement `Class.getIterfaces()`  and `Class.isAssignable`/`Class.isInstance` checks
* Add additional `superclass` field to `java.lang.Class` to implement `Class.getSuperclass`  
* Minor optimizations / readability improvements to `java.lang.Class` implementation


nativelib/src/main/resources/scala-native/gc/immix_commix/headers/ObjectHeader.h
```diff
@@ -33,20 +33,31 @@ extern const int __boxed_ptr_id;
...
 #ifdef USES_LOCKWORD
         word_t *lockWord;
 #endif
         int32_t id;
-        int32_t tid;
+        int32_t interfacesCount;
+        struct BaseRtti *interfaces;
         StringObject *name;
     } rt;
+    // Fields defined only for instantinable types (classes)
```
ekrich, line 53:
```suggestion
    // Fields defined only for instantiable types (classes)
```

PR #4092, opened 2024-12-02
fix [tools] Enhance build error handling for a build module named native

ekrich:
As discussed in Discord, when a project is named `native` on UNIX type systems the executable will be named `native` which will cause a exception thrown as shown below but it does not stop the build and the executable named `native` is put into the `native` `workDir` directory rather than in the same directory as the `workDir`.

This PR catches the exception now, and if the potential artifact will be named the same as the `workDir` by using `exists()` method then a `BuildException` with a message is output with a solution for the user to avoid the conflict. Any other exception is output with some context included. The following is an example although I am not sure why the message is printed twice (sbt?):

New Outcome:
```scala
...
[info] Linking with [pthread, dl]
[info] Total (3511 ms)
[error] Executable build module or `baseName` is named 'native'
[error] which conflicts with the compiler `workDir`.
[error] Please rename the build module or
[error] use `withBaseName` to rename the executable.
[error] Cause: java.nio.file.DirectoryNotEmptyException: /Users/eric/workspace/scala-native-bugs/native-dir/native/target/scala-3.3.4/native
[error] (native / Compile / nativeLink) Executable build module or `baseName` is named 'native'
[error] which conflicts with the compiler `workDir`.
[error] Please rename the build module or
[error] use `withBaseName` to rename the executable.
[error] Cause: java.nio.file.DirectoryNotEmptyException: /Users/eric/workspace/scala-native-bugs/native-dir/native/target/scala-3.3.4/native
[error] Total time: 4 s, completed Dec 2, 2024, 9:55:47 AM
```

Original outcome:
```scala
sbt:native-dir> native/run
[info] Linking (multithreadingEnabled=true, disable if not used) (929 ms)
[info] Discovered 884 classes and 5396 methods after classloading
[info] Checking intermediate code (quick) (48 ms)
[info] Multithreading was not explicitly enabled - initial class loading has not detected any usage of system threads. Multithreading support will be disabled to improve performance.
[info] Linking (multithreadingEnabled=false) (337 ms)
[info] Discovered 495 classes and 2485 methods after classloading
[info] Checking intermediate code (quick) (8 ms)
[info] Discovered 474 classes and 1898 methods after optimization
[info] Optimizing (debug mode) (534 ms)
[info] Produced 44 LLVM IR files
[info] Generating intermediate code (643 ms)
[info] Compiling to native code (2316 ms)
[info] Linking with [pthread, dl]
[info] Total (4687 ms)
[error] stack trace is suppressed; run last native / Compile / nativeLink for the full output
[error] (native / Compile / nativeLink) java.nio.file.DirectoryNotEmptyException: /Users/eric/workspace/scala-native-bugs/native-dir/native/target/scala-3.3.4/native
[error] Total time: 5 s, completed Dec 2, 2024, 7:42:55 AM
sbt:native-dir> clean
[success] Total time: 1 s, completed Dec 2, 2024, 8:27:35 AM
```

### DIFF
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -173,12 +173,39 @@ private[scalanative] object LLVM {
       case Success(_) =>
     }
 
+  /** This function allows a project to have multiple `main` files by copying
+   *  the one selected to the same parent directory as the `workDir` which is by
+   *  default named `native`. Since the directory is named `native`, having a
+   *  project named `native` will by default produce a executable named `native`
+   *  which will throw an exception since the copy command uses
+   *  REPLACE_EXISTING.
+   *
+   *  Having a project or `baseName` named `native` conflict with the build.
+   */
   private def copyOutput(config: Config, buildPath: Path) = {
     val outPath = config.artifactPath
-    config.compilerConfig.buildTarget match {
-      case BuildTarget.Application =>
-        Files.copy(buildPath, outPath, StandardCopyOption.REPLACE_EXISTING)
-      case _: BuildTarget.Library => outPath
+    try {
+      config.compilerConfig.buildTarget match {
+        case BuildTarget.Application =>
+          Files.copy(buildPath, outPath, StandardCopyOption.REPLACE_EXISTING)
+        case _: BuildTarget.Library => outPath
+      }
+    } catch {
+      case t: Throwable =>
+        val msg =
+          if (outPath.toFile().exists())
+            s"""Executable build module or `baseName` is named 'native'
+                |which conflicts with the compiler `workDir`.
+                |Please rename the build module or
+                |use `withBaseName` to rename the executable.
+                |Cause: ${t}""".stripMargin
+          else
+            s"""Unexpected exception:
+                |Build path: $buildPath
+                |Artifact: $outPath
+                |Cause: ${t}""".stripMargin
+
+        throw new BuildException(msg)
     }
   }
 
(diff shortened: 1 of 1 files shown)


tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -173,12 +173,39 @@ private[scalanative] object LLVM {
...
-        Files.copy(buildPath, outPath, StandardCopyOption.REPLACE_EXISTING)
-      case _: BuildTarget.Library => outPath
+    try {
+      config.compilerConfig.buildTarget match {
+        case BuildTarget.Application =>
+          Files.copy(buildPath, outPath, StandardCopyOption.REPLACE_EXISTING)
+        case _: BuildTarget.Library => outPath
+      }
+    } catch {
+      case t: Throwable =>
```
WojciechMazur, line 194: Typically we should not catch Throwable. Consider using Exception or IOException instead

WojciechMazur: Probably going with IOException would be the best choice, there is no other Exception that `Files.copy` can throw. We also consider only having `case ex: IOExcpeiton if outPath.toFile().exists() => ` handler. In case of unexpected exception it might be better to pass it through so at least we'll get a stack trace (but it's unlikely to catch anything else here)

ekrich: Agree that this is a much better solution - this case we know but others are probably completely out of our control.

Besides, this looks nicer.

WojciechMazur (approved):

PR #4094, opened 2024-12-04
Get function name from DWARF when available

lolgab:
`unw_get_proc_name` doesn't work on static binaries. This PR uses DWARF to get the procedure names at runtime when they are available.
It uses `DW_AT_linkage_name` which contains mangled names for functions. The `main` function has an ugly `_SM2__C4main` name which is meaningless. I can add a special case for that (or maybe we can consider removing the generated main function from the stack traces altogether, so we don't need to think about it

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Backtrace.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Backtrace.scala
@@ -6,6 +6,7 @@ import scala.scalanative.runtime.dwarf.DWARF
 import scala.scalanative.runtime.dwarf.DWARF.DIE
 import scala.scalanative.runtime.dwarf.DWARF.CompileUnit
 
+import scala.scalanative.unsafe.CString
 import scala.scalanative.unsafe.Tag
 import scala.scalanative.unsafe.Zone
 import scala.scalanative.unsigned.UInt
@@ -35,7 +36,8 @@ private[runtime] object Backtrace {
       lowPC: Long,
       highPC: Long,
       line: Int,
-      filenameAt: Option[UInt]
+      filenameAt: Option[UInt],
+      linkageNameAt: Option[UInt]
   )
 
   private val MACHO_MAGIC = "cffaedfe"
@@ -44,9 +46,9 @@ private[runtime] object Backtrace {
   private val cache: AbstractMap[String, Option[DwarfInfo]] =
     if (isMultithreadingEnabled) new ConcurrentHashMap
     else new HashMap
-  case class Position(filename: String, line: Int)
+  case class Position(linkageName: CString, filename: String, line: Int)
   object Position {
-    final val empty = Position(null, 0)
+    final val empty = Position(null, null, 0)
   }
 
   def decodePosition(pc: Long): Position = {
@@ -80,10 +82,12 @@ private[runtime] object Backtrace {
     val address = pc - info.offset
     val position = for {
       subprogram <- search(info.subprograms, address)
-      at <- subprogram.filenameAt
+      filenameAt <- subprogram.filenameAt
+      linkageNameAt <- subprogram.linkageNameAt
     } yield {
-      val filename = info.strings.read(at)
-      Position(filename, subprogram.line + 1) // line number in DWARF is 0-based
+      val filename = info.strings.read(filenameAt)
+      val linkageName = info.strings.buf.asInstanceOf[ByteArray].at(linkageNameAt.toInt)
+      Position(linkageName, filename, subprogram.line + 1) // line number in DWARF is 0-based
     }
     position.getOrElse(Position.empty)
   }
@@ -137,7 +141,7 @@ private[runtime] object Backtrace {
             line <- die.getLine
             low <- die.getLowPC
             high <- die.getHighPC(low)
-          } yield SubprogramDIE(low, high, line, filenameAt)
+          } yield SubprogramDIE(low, high, line, filenameAt, die.getLinkageName)
         } else if (die.is(DWARF.Tag.DW_TAG_compile_unit)) {
           // Debug Information Entries (DIE) in DWARF has a tree structure, and
           // the DIEs after the Compile Unit DIE belongs to that compile unit (file in Scala)
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Throwables.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Throwables.scala
@@ -79,25 +79,42 @@ object StackTrace {
       cursor: CVoidPtr,
       ip: CUnsignedLong
   ): StackTraceElement = {
-    val nameMax = 1024
-    val name = fromRawPtr[CChar](
-      calloc(
-        Intrinsics.castIntToRawSizeUnsigned(nameMax),
-        Intrinsics.sizeOf[CChar]
-      )
-    )
-    val offset = fromRawPtr[Long](Intrinsics.stackalloc[Long]())
+    val hasDebugInfo =
+      LinktimeInfo.isMac && LinktimeInfo.sourceLevelDebuging.generateFunctionSourcePositions
 
-    unwind.get_proc_name(cursor, name, nameMax.toUSize, offset)
-
-    // Make sure the name is definitely 0-terminated.
-    // Unmangler is going to use strlen on this name and it's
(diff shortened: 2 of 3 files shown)


tanishiking:
Wow, awesome 🚀 
Demangling the `DW_AT_linkage_name` attribute seems like a solid solution. 
Because `DW_AT_name` represents the name as it appears in the source file, while `DW_AT_linkage_name` is intended for referencing symbols across compilation unit boundaries. This means it should contain the full path to the symbol. :+1:

Regarding ELF, there's a WIP impl here https://github.com/scala-native/scala-native/pull/4061

- [x] ELF parsing is done
- [x] I guess DWARF5 parsing is ... working at least with our sandbox binary (but not sure it covers the most specs)
- [ ] not yet: we need to obtain the ASLR offset from `/proc/<pid>/maps` (but it should work with `-nopie` option) 
  - see: https://github.com/boostorg/stacktrace/pull/138

tanishiking (commented):
Wow, awesome 🚀 
Demangling the `DW_AT_linkage_name` attribute seems like a solid solution. 
Because `DW_AT_name` represents the name as it appears in the source file, while `DW_AT_linkage_name` is intended for referencing symbols across compilation unit boundaries, which means it should contain the full path to the symbol. :+1:

Regarding ELF, there's a WIP impl here https://github.com/scala-native/scala-native/pull/4061, if you're interested in working on that, feel free to taking it over :)

- [x] ELF parsing is done
- [x] I guess DWARF5 parsing is ... working at least with our sandbox binary (but not sure it covers the most specs)
  - It should be robust when failed to parse DWARF
- [ ] not yet: we need to obtain the ASLR offset from `/proc/<pid>/maps` (but it should work with `-nopie` option) 
  - https://github.com/boostorg/stacktrace/pull/138
  - https://github.com/rust-lang/backtrace-rs/blob/26ab4afa46e889f2f4c81e5f5022f61569f02d16/src/symbolize/gimli/parse_running_mmaps_unix.rs
  - I think this is worth separating a PR

lolgab:
Stack traces work on static binaries :tada: 
screenshot

Many thanks to @tanishiking and @keynmol who did 99% of the work. I just connected the dots :)

nativelib/src/main/scala/scala/scalanative/runtime/Throwables.scala
```diff
@@ -75,31 +75,60 @@ object StackTrace {
...
+      // linkageName has an extra "_" that we don't want in stack traces
+      def isScalaNativeMangledName =
+        position.linkageName(0) == '_'.toByte &&
+          position.linkageName(1) == '_'.toByte &&
+          position.linkageName(2) == 'S'.toByte
+      val name =
+        if (isScalaNativeMangledName)
+          // skip first `_`
+          position.linkageName + 1
+        else position.linkageName
```
lolgab, line 101: An alternative could be to prepend `_` to all names. But I haven't investigated the implications.

lolgab:
Optimization results in debug mode.

## Before
immagine

## After
immagine

ekrich:
In regards to optimization are you referring to the 450ms to 107ms or the better naming?

lolgab:
> In regards to optimization are you referring to the 450ms to 107ms or the better naming?

I'm referring to the improved speed.
The naming was also fixed since the `linakageName` debug data did miss a `_` prefix, which is expected by the scala-native-demangler.

lolgab:
With efficient `skip` (`release-fast` mode):
immagine

lolgab:
### With efficient `BinaryFile.position`:

immagine

lolgab:
I don't understand why in some tests even though I used the option to force DWARFV4, the toolchaon is generating DWARFV5 which I haven't supported in this PR.
I think it's important to pin a DWARF version otherwise when v6 becomes the new default, that would break Scala Native, even if we included v5 in the runtime right away.
@WojciechMazur @tanishiking @keynmol Any theory why crossplatform tests generate Dwarf v5?

tanishiking:
Ah, I think those DWARF5 debug information likely comes from `libc` or other object files. Even if we build our project with `-gdwarf-4` to generate DWARF4 debug information, we cannot change object files like libc that already have DWARF5 debug information attached.

Possible solutions would be:
  
- Implementing a DWARF5 parser.  
- Stripping the debug information from those object files using `--strip-debug` or something? However, I am not sure how to strip debug information from specific object files only.

lolgab:
@tanishiking The object files coming from the c files we compile also have the `-gdwarf-4`, so they have the right dwarf version. Do we include any other object files?
It is strange that it happens only on some tests.

lolgab:
After latest optimizations (mainly coming from https://github.com/scala-native/scala-native/pull/4094/commits/a53659389f1c907a91863fcf63da7591e2934844 and https://github.com/scala-native/scala-native/pull/4094/commits/151aa70e0e209d6b89eba78c5b1cb47245ceac88):
immagine

lolgab:
@tanishiking I went with the second option of skipping the DWARFv5 data.
I fixed the header parsing using your PR and then I skip the v5 units with:

```scala
bf.seek(header.unit_offset)
skipBytes(header.unit_length)
```

nativelib/src/main/scala/scala/scalanative/runtime/Backtrace.scala
```diff
@@ -15,86 +19,60 @@ import scala.annotation.tailrec
...
+        // Subtract the offset to match the pc address from libunwind (runtime) and address in debug info (compile/link time).
+        val address = pc - info.offset
+        val position = for {
+          subprogram <- search(info.subprograms, address)
+          filenameAt <- subprogram.filenameAt
+          linkageNameAt <- subprogram.linkageNameAt
+        } yield {
+          val filename = info.strings.read(filenameAt)
+          val linkageName =
+            info.strings.buf.asInstanceOf[ByteArray].at(linkageNameAt.toInt)
```
WojciechMazur: Not very important but with `import scala.scalanative.unsafe.UnsafeRichArray` we can use extension instead of explicit cast.  If we're sure `linkageNameAt.toInt` would always be in valid bounds, we can even use `atUnsafe` to skip redundant bound check. 

```suggestion
            info.strings.buf.at(linkageNameAt.toInt)
```

nativelib/src/main/scala/scala/scalanative/runtime/Throwables.scala
```diff
@@ -75,31 +75,62 @@ object StackTrace {
...
       else Backtrace.Position.empty
-    try StackTraceElement(name, position)
-    finally free(name)
+
+    def withNameFromDWARF() = {
+      // linkageName has an extra "_" that we don't want in stack traces
+      def isScalaNativeMangledName =
+        position.linkageName(0) == '_'.toByte &&
+          position.linkageName(1) == '_'.toByte &&
+          position.linkageName(2) == 'S'.toByte
```
WojciechMazur, line 98: We could consider usage of strncmp 
```suggestion
      def isScalaNativeMangledName = strncmp(position.linkageName, c"__SM", 4.toCSize) == 0
```
However, it would require us to add binding in `runtime.fii`

nativelib/src/main/scala/scala/scalanative/runtime/ffi.scala
```diff
@@ -20,6 +20,7 @@ object ffi {
   def wcslen(str: CWideString): CSize = extern
   def strncpy(dest: CString, src: CString, count: RawSize): CString = extern
   def strcpy(dest: CString, src: CString): CString = extern
+  def strcmp(str1: CString, str2: CString): Int = extern
```
WojciechMazur, line 23: It's probably unused

WojciechMazur (commented):
Great work. I have some comments, but I'll wait for Rikito's review as he has more knowledge about debug symbols 

Almost 4k less symbols on with enabled debug info is also a great win!

nativelib/src/main/scala/scala/scalanative/runtime/Throwables.scala
lolgab: Done ✅

nativelib/src/main/scala/scala/scalanative/runtime/ffi.scala
lolgab: Replaced with `strncmp`

PR #4106, opened 2024-12-12
Always inline when code is annotated with `@alwaysinline`

lolgab:
Currently, when a condition to not inline is present, then no code is inlined, even if it is annotated with `@alwaysinline`

Other changes:
- Fix debug logger to log when `hasUnwind` is `true`


WojciechMazur (approved):
I wonder if there can be some negative side effects of this change, but probably it could only affect if literally every method would have alwaysinline annotation. I guess, it should be fine

PR #4110, opened 2024-12-14
Reuse `hasUnwind` lazy val in Opt.scala

lolgab:


lolgab:
@WojciechMazur Can you rerun the CI here? It seems flakiness, I the two definitions should be the same, I don't see why this could break.

WojciechMazur (approved):

PR #4114, opened 2024-12-15
Fix signaturs of java.lang.{Math,Runtime} methods

WojciechMazur:
These were interpreted as java.lang.{Int, Double, Boolean} instead of primitive types


keynmol:
@WojciechMazur would the incorrect return type affect method visibility?
IIRC lack of removeShutdownHook was brought up a couple of times - is it because the signature was incorrect?

WojciechMazur:
Yes, that imght have to be the reason why it was not reachable. If the signature does not match then when classloading we would report a missing symbol.

PR #4122, opened 2024-12-20
Implement minimal C++-like runtime to support exceptions without linking libc++

lolgab:
Exceptions work as before but we link using `clang` and not using `clang++`, so we only require `libc`.

The `eh.c` code was adapted from https://github.com/s17b2-voroneckij/infos-user

### Before
immagine

### After
immagine

## Questions

> About the process monitor, the first generation still uses the c++ standard library, which we don't want to link anymore.
What should we do about it?
Can we start requiring a new enough version of the Linux kernel and always use gen2? Or should we maybe add a config to link C++ (I prefer not, since we always wanted to stop linking c++).

`process_monitor.cpp` was converted from C++ to C and using two linked lists instead of the C++ standard library map.
With this change the gen1 and gen2 work as before.

> Should we add `-fno-exceptions` etc. only to `nativelib` code?
I'm thinking about allowing to link with c++ stdlib and c++ exceptions in custom `resources/scala-native` code. Should we support that?
I don't know how much we support in `resources/scala-native` cpp files.

Currently, we add this flag only if `config.usingCPPExceptions`. We should probably use the old flags when `config.isCPPRuntimeEnabled` is `true`

## Semantic changes

- After this PR, exceptions are not C++ exceptions anymore, so it is undefined behavior to throw uncaught exceptions from Scala Native code to C++. Exceptions should be handled and returned to exported functions as error codes.
- Stop supporting/testing throwing Scala exceptions in `@exported` functions

## Other changes

- Add `-lm` to link `libm` (before was added automatically by clang++)
- Start using `-Wl,--as-needed` linker flag when supported to avoid linking libraries we don't use (namely `libm`)
- Use literal constants in `RyuFloat` to avoid depending on `libm`
- Add `@scala.scalanative.unsafe.linkCppRuntime` annotation to link using `clang++` and link the cpp runtime. The annotation needs to be added to `@extern object`s that call C++ functions using the c++ runtime. When the annotation is missing from the `ReachabilityAnalysis.Result` linking is performed with `clang` and c++ files are compiled with `-fno-exceptions`. Even when present the Scala Native exceptions use the new runtime, so can't be catched from C++ code.
- Convert `nir.Attrs` `case class` to a regular class and implement the binary compatibility shims.

## Improvements

As a side effect of not depending on the system ABI (which is not guaranteed to be the same we tested upon), exceptions work correctly on lima on Mac aarch64 + Ubuntu 24.10, which wasn't able to catch exceptions on the baseline.

### Before (Ubuntu 24.10 on lima aarch64, Mac M2)
```
$ sandbox/.2.13/target/scala-2.13/sandbox
terminate called recursively
Aborted (core dumped)
```

### After (Ubuntu 24.10 on lima aarch64, Mac M2)
```
$ sandbox/.2.13/target/scala-2.13/sandbox
Exception in thread "main"java.lang.Exception: BOOM!!!
	at java.lang.Throwable.fillInStackTrace(Throwables.scala:58)
	at <none>.main(unknown:2)
	at <none>.(Unknown Source)
```


ekrich:
This is very cool - assuming that Windows could be done but avoiding it for now?

lolgab:
@ekrich Thank you!
About Windows, I haven't tried Windows at all but I assume it can be supported. I'm not a Windows user so it will need someone else to try to port the changes to Windows.
Probably windows-gnu shouldn't be hard to port.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
```diff
@@ -54,41 +49,21 @@ private[codegen] class UnixCompat(codegen: AbstractCodeGen)
...
-    str("br ")
+    line(s"$exc = call $ptrT $scalaNativeCatch($ptrT $r0)")
+    line("br ")
     codegen.genNext(next)
     unindent()
-
-    line(s"$excfail:")
-    indent()
-    line(s"resume $excRecTy $rec")
-    unindent()
```
lolgab, line 82: Since we catch all exceptions, this block can never happen.
Catching any exception that is not a Scala Native exception is undefined behavior

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -85,8 +85,14 @@ private[scalanative] object LLVM {
       multithreadingEnabled ++ allowTargetOverrrides
     }
     val exceptionsHandling = {
-      val opt = if (isCpp) List("-fcxx-exceptions") else Nil
-      List("-fexceptions", "-funwind-tables") ::: opt
+      val targetSpecific = if (config.targetsWindows) {
+        val opt = if (isCpp) List("-fcxx-exceptions") else Nil
+        List("-fexceptions", "-funwind-tables") ++ opt
+      } else {
+        if (isCpp) List("-fno-rtti", "-fno-exceptions", "-funwind-tables")
```
lolgab, line 92: Should this be applied only to libunwind (or `nativelib`) code? Otherwise it is impossible to use c++ exceptions in c++ code in `scala-native` resources.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
```diff
@@ -54,41 +49,21 @@ private[codegen] class UnixCompat(codegen: AbstractCodeGen)
     indent()
     line(s"$rec = $landingpad")
     line(s"$r0 = extractvalue $excRecTy $rec, 0")
-    line(s"$r1 = extractvalue $excRecTy $rec, 1")
-    line(s"$id = $typeid")
-    line(s"$cmp = icmp eq i32 $r1, $id")
-    line(s"br i1 $cmp, label %$excsucc, label %$excfail")
```
lolgab, line 60: We don't do any type check. We assume any exception is a java.lang.Throwable and we handle it.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
```diff
@@ -54,41 +49,21 @@ private[codegen] class UnixCompat(codegen: AbstractCodeGen)
...
-      line(s"$exc = load ptr, ptr $w2")
-    } else {
-      line(s"$w1 = bitcast i8* $w0 to i8**")
-      line(s"$w2 = getelementptr i8*, i8** $w1, i32 1")
-      line(s"$exc = load i8*, i8** $w2")
-    }
-    line(s"call void $endCatch()")
-    str("br ")
+    line(s"$exc = call $ptrT $scalaNativeCatch($ptrT $r0)")
+    line("br ")
```
lolgab, line 58: Instead of upacking the ExceptionWrapper c++ class object, we simply call `scalanative_catch` which receives a `_UnwindException` and `return` the pointer to the Scala Native (java.lang.Throwable) exception object.
We assign `exc` to it and handle it.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
```diff
@@ -11,19 +11,15 @@ private[codegen] class UnixCompat(codegen: AbstractCodeGen)
...
-  val ehWrapperTy = "@_ZTIN11scalanative16ExceptionWrapperE"
   val excRecTy = s"{ $ptrT, i32 }"
-  val beginCatch = "@__cxa_begin_catch"
-  val endCatch = "@__cxa_end_catch"
+  val scalaNativeCatch = "@scalanative_catch"
   val catchSig =
-    if (useOpaquePointers) s"$ptrT $ehWrapperTy"
-    else s"i8* bitcast ({ i8*, i8*, i8* }* $ehWrapperTy to i8*)"
+    if (useOpaquePointers) s"$ptrT null"
+    else s"i8* null"
```
lolgab, line 18: `null` is the llvm way to catch all.

lolgab:
@WojciechMazur Can you read the questions in the description and give me hints on how to handle them? I don't know what's best to do with the process_monitor gen1, which can't be supported without the c++ stdlib.
Also, I left some more questions as Github review comments in the code.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
```diff
@@ -137,6 +137,11 @@ sealed trait NativeConfig {
     }
   }
 
+  /** Shall enable the use of C++ runtime. Useful if you want to compile c++
+   *  code in your Scala Native application.
+   */
+  def useCPPRuntime: Boolean
```
lolgab, line 143: This is actually a lie because windows can't function without the cpp runtime at the moment, so should I rename this? Do you have suggestions?

ekrich: Default to true on Windows?

javalib/src/main/resources/scala-native/process_monitor.cpp
```diff
@@ -1,3 +1,4 @@
+#if defined(SCALANATIVE_CPP_RUNTIME_ENABLED)
```
ekrich, line 1: I guess I don't understand. Doesn't this have to also be converted to C? Hard to tell based on C preprocessor format. Perhaps we should go to a 2 space indented style for conditional compilation.

lolgab: This is the implementation we use on Windows. It should be c++ because we use the c++ exceptions there.
The `ifdef` is using the format used on the other c++ files.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
lolgab: The current code defaults to `false` and communicates the "intention" to use the c++ runtime. But on Config.scala this is added with `|| config.targetsWindows`.
But I accept suggestions on how to improve the config.

javalib/src/main/resources/scala-native/process_monitor.cpp
ekrich: Yeah, I set up the clang-format so don't have any setup to speak of.

lolgab: I meant the choice between
```c++
#ifdef FOO
```

and

```c++
#if defined(FOO)
```

ekrich: The second one is preferred for more complicated expressions or arguments to `defined`.

javalib/src/main/scala/java/lang/process/UnixProcess.scala
```diff
@@ -6,7 +6,9 @@ private[lang] abstract class UnixProcess extends GenericProcess {}
 
 object UnixProcess {
   def apply(builder: ProcessBuilder): Process = {
-    val useGen2 = if (LinktimeInfo.is32BitPlatform) {
+    val useGen2 = if (!LinktimeInfo.isCPPRuntimeEnabled) {
+      true
+    } else if (LinktimeInfo.is32BitPlatform) {
```
lolgab, line 11: Now, it would be nice to be able to link-time fail if you don't use the cpp runtime on a 32 bit machine.
Is it possible? We could ask the user to enable the cpp runtime if they want to use `java.lang.process.Process` on that platform.

ekrich: I looked at the process code and it was just the collections (Map) that was C++. It seems that could be replaced to make that pure C too.

lolgab: Then I have to check if I can migrate that to c as well.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -80,13 +80,27 @@ private[scalanative] object LLVM {
         if (config.compilerConfig.multithreadingSupport)
           Seq("-DSCALANATIVE_MULTITHREADING_ENABLED")
         else Nil
+      val cppRuntimeEnabled =
+        if (config.isCPPRuntimeEnabled)
+          Seq("-DSCALANATIVE_CPP_RUNTIME_ENABLED")
+        else Nil
+      val usingCPPExceptions =
+        if (config.usingCPPExceptions)
+          Seq("-DSCALANATIVE_USING_CPP_EXCEPTIONS")
```
lolgab, line 89: Suggestions accepted for this define name

tools/src/main/scala/scala/scalanative/build/LLVM.scala
lolgab: Now that there is the config boolean to check if we are using the cpp runtime, we can use it here to selectively enable exceptions or not. To be discussed what flags we want.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -221,11 +235,22 @@ private[scalanative] object LLVM {
         if (config.targetsWindows) Seq("dbghelp")
         else if (config.targetsOpenBSD || config.targetsNetBSD)
           Seq("pthread")
-        else Seq("pthread", "dl")
+        else Seq("pthread", "dl", "m")
```
lolgab, line 238: This is an artifact of linking with `clang` instead of `clang++` on linux, probably `libm` was a dependency of `libc++` and we were importing it by accident.

javalib/src/main/resources/scala-native/process_monitor.c
```diff
@@ -0,0 +1,200 @@
...
+
+typedef struct FinishedProc {
+    int pid;
+    int result;
+    struct FinishedProc *next;
+} FinishedProc;
+
+static pthread_mutex_t shared_mutex;
+static WaitingProc *waiting_procs = NULL;
+static FinishedProc *finished_procs = NULL;
```
lolgab, line 46: I changed the map with a linked list. We have linear access, now but we don't need to implement the hash map.

javalib/src/main/scala/java/lang/process/UnixProcess.scala
lolgab: Converted process_monitor to C, using simple linked lists

lolgab:
Ready for first round of review @WojciechMazur

tools/src/main/scala/scala/scalanative/build/LLVM.scala
lolgab: Probably the easiest thing to do is to change the if to `if (config.isCPPRuntimeEnabled)` so we enable exceptions to all cpp code if the C++ runtime is enabled.

lolgab:
Can you rerun the job in https://github.com/scala-native/scala-native/actions/runs/12526993166/job/34940115020?pr=4122
It seems a flaky test to me

WojciechMazur:
Great work! Sorry I was not here to help with questions during last days. I've only taken a short look at changes so far, but I'll try to come back with full review by the end of this year

javalib/src/main/scala/java/lang/process/UnixProcess.scala
ekrich: I had poked around and found this which has a liberal license - https://github.com/troydhanson/uthash

tools/src/main/scala/scala/scalanative/build/LLVM.scala
ekrich: I think the two settings together make 4 options but some of the options don't make since with each other. e.g. if you take CPPExceptions then CPPRuntime needs to be true as well. Maybe they need to be a setting pair with some check of validity?

I also think that if we are not making a CPPRuntime then we should disallow compiling C++ in `scala-native` resources as because then we must link with C++.

It is unclear to me if a C++ lib has a C interface as with something like the C API for `libtensorflow` then would the runtime be C but then has a dependency on a C++ lib. Also, what if we try and link that statically. You might be able to test the default dynamic using this just to make sure - <https://github.com/ekrich/stensorflow.g8>

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
WojciechMazur, line 143: I'm not really a fun of this setting to be honest.

At first it might be misleading - actually it's only controlling which program should we use for linking `clang` or `clang++`. If so, then maybe we should introduce `alternativeLinker: Option[Path]` which would be to used if defined, altough it might be problematic in usage becouse of hard-coded settings we enforce, so probably that's not a good idea to follow.

From the user perspective I'd prefer to not need any additional configuration, only because upstream library I use is using .cpp files for the internals. So we should have some way of automatic detection when c++ stdlib is really needed.  
One idea how to achive that would be checking if we're compiling any .cpp files if so we'd use clang++ to link and clang otherwise. There are 2 issues however. First problem is we're still defining .cpp sources in core Scala Native project - we ship libunwind which contains C++ sources, we can consider removing it in the follow up PR, a quick experiment on my machine proved there is no issues with that.  
The second problem is that we can find effectively empty .cpp files becouse their content can be erased by pre-processor - in fact that's what we frequently use. We could potentially inspect output `.o` files with `llvm-nm` to check if it contains at least 1 symbol. Still, we cannot guarantee this binary would be reachable, so it maybe not be reliable. 

From what I've observed the only difference between using `clang` vs `clang++` for linking `.o` files is addition of `-lc++`. Both are delegating to `ld` under the hood using the same options. 
Based on that we can take one more direction - let's put the responsability on the author of the glue code using c++. As an author of library you can either put `@link("c++")` or create `resources/scala-native.properties` files with `nir.link.names=c++` to ensure the c++ stdlib is linked. 
We can take advantage of this fact even further and detect if `LLVM.prepareLinkCommand -> links variable` contains `c++` entry and use it to control if `clang` or `clang++` should be used. 

I think that the last solution might be the most reliable in the long term. We would be able to combine our about  using any c++ features internally by toolchain (.eg`-fcxx-exceptions`, targeting Windows) with source metadata from users (`@link`). 
If going with this solutions we should probably add documentation section how user/library author can enable c++ support.

WojciechMazur (commented):
Overall LGTM, but I've commented some ideas how we can make the C++ suport more idomatic if it would be required by used libraries.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
WojciechMazur: The only problem with this solution is fact that it might not be inlined, but we 'd need a benchmark to measure the impact in a code that is using exceptions heavily. We probably should try to inline it as it was done previously. Probably it would be easiest to implement if we'd get rid of `malloc`/`free` calls as well (mentioned above), but probably we can do it in the follow up PR.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
```diff
@@ -11,19 +11,15 @@ private[codegen] class UnixCompat(codegen: AbstractCodeGen)
...
   val landingpad =
     s"landingpad $excRecTy catch $catchSig"
-  val typeid =
-    s"call i32 @llvm.eh.typeid.for($catchSig)"
 
-  protected val osPersonalityType: String = "@__gxx_personality_v0"
+  protected val osPersonalityType: String = "@scalanative_personality"
```
WojciechMazur, line 22: I wonder if we should keep the original variant of `UnixCompat` for exception handling and either old or new variant depending on `Config.usingCPPExceptions`. This would also allow us to keep old behaviour of being able to catch C++ exception from code that is calling Scala Native library (probably)

lolgab: In my opinion, allowing to catch Scala Native exceptions from C++ is a feature nobody uses and it can't really be used, since the wrapped exception is a `void *` that has no meaning in C++, you can't get the class, the error message, etc. (you can only if you play with pointer arithmetics, and the object structure is private to the Scala Native runtime).
We could do that, but I don't see value in it, and having the code running in both modes it's simpler and would lead to fewer surprises when using `c++` code.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
lolgab: > There are 2 issues however. First problem is we're still defining .cpp sources in core Scala Native project - we ship libunwind which contains C++ sources, we can consider removing it in the follow up PR, a quick experiment on my machine proved there is no issues with that.

I'm not sure I understand what is the problem with libunwind. libunwind itself uses the `-fno-rtti -fno-exceptions -funwind-tables` flags to compile its c++ with not reliance on the c++ runtime.

lolgab: About implicitly adding libc++ based on having a single c++ seems too intrusive. I would prefer a manual way if possible. For sure I like that `clang++` can be replaced by the `lc++` flag, I think I can do that already, which would simplify the discussion.

tools/src/main/scala/scala/scalanative/build/Config.scala
```diff
@@ -196,6 +196,12 @@ sealed trait Config {
           .getOrElse(compilerConfig.mode.isInstanceOf[Mode.Release])
       case _ => false
     }
+
+  private[scalanative] def isCPPRuntimeEnabled: Boolean =
+    targetsWindows || compilerConfig.useCPPRuntime
+
+  private[scalanative] def usingCPPExceptions: Boolean =
+    targetsWindows
```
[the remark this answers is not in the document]
lolgab: Nice idea! But currently we set `-fcxx-exceptions` based on `usingCPPExceptions` so we need to make it not recursive.

We could check for the `-fcxx-exceptions` and use it as a way to know if we need to enable or not exceptions.
One problem I see is that `-fcxx-exceptions` in our code is used only for C++ files and not for C files and we don't support user-defined c++-only or c-only compile options. So if a user does:
```scala
nativeConfig ~= { c => c.withCompileOptions(c.compileOptions ++ Seq("-fcxx-exceptions")) }
```
It gets applied to all files, also to C files.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
lolgab: But the calls to `beginCatch` and `endCatch` were not inlined already, and those were to unpack the C++ exception object which we don't have now. It would be lovely to remove the `ExceptionWrapper` struct and add some reserved bytes inside `Throwable` (maybe a bunch of `Long`s) that could be used by the C code to store the `_Unwind_Exception`.
Since the exceptions are not in the hot path I would leave this to a follow-up PR.

lolgab:
Tested `-lc++` but it's not that easy (I remember having had problems, but I didn't remember the exact problem).
I believe `clang++` adds the correct flags based on the platform:

```
info] LINK : fatal error LNK1104: cannot open file 'c++.lib'
[error] clang: error: linker command failed with exit code 1104 (use -v to see invocation)
[info] Total (16633 ms)
[error] Failed to link D:\a\scala-native\scala-native\sandbox\.2.12\target\scala-2.12\native\Test.exe
[error] (sandbox2_12 / Compile / nativeLink) Failed to link D:\a\scala-native\scala-native\sandbox\.2.12\target\scala-2.12\native\Test.exe
[error] Total time: 104 s (01:44), completed Dec 31, 2024 7:08:17 PM

```

nativelib/src/main/scala/scala/scalanative/meta/LinktimeInfo.scala
```diff
@@ -96,4 +96,7 @@ object LinktimeInfo {
     def generateFunctionSourcePositions: Boolean = resolved
 
   }
+
+  @resolvedAtLinktime("scala.scalanative.meta.linktimeinfo.isCPPRuntimeEnabled")
+  def isCPPRuntimeEnabled: Boolean = resolved
```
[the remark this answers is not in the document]
lolgab: Done ✅

lolgab:
I reverted the change to not use `clang++` which doesn't work.
We need `isCPPRuntimeEnabled`, maybe derived from the `compileOptions`, or something else, but we do need it.

tools/src/main/scala/scala/scalanative/codegen/llvm/compat/os/UnixCompat.scala
ekrich: I am a little confused on the idea of catching C++ exceptions at all. I see three use cases, but maybe there are more.

1. The user puts C++ glue code but needs have `extern C` to allow Scala Native to call it.
2. A library uses C++ but must have `extern C` to allow Scala Native to call it.
3. Create a Scala Native library. It must have exposed C functions to call it.

I don't see any reason to support C++ exceptions for the above because we only interact with the C ABI and thus standard C error handling techniques would be used. I don't think throwing a C++ exception across those boundaries makes sense.

As I recall, using C++ and `libunwind` was just used as the easiest way to provide for 1) Java/Scala exceptions and 2) stack traces. I could be off base but if we can support these two features and the cases above in a portable manner then we should be good. Detecting to use C++ to link is the trick but would that mean that a `.cpp` file was encountered? - I think that was mentioned that they can't be empty for the conditional compilation case?

We would just keep the status quo for Windows or would it be possible the support PE or other binary formats that might come up similar to the Dwarf support? Hopefully this discussion is useful and not completely off base.

lolgab: One use case would be to provide a catch all clause in C++ when calling a Scala Native library.
It's an advanced use case which can be also solved by returning a Either-like data structure in C.
As I said earlier, I think this is more a bug than a feature, I think the best behavior is just to crash.

ekrich: Calling a C ABI and expecting a C++ exception is a bit much. If I was calling from C I wouldn't expect to get a C++ exception so I think that would be a crash. I agree.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
WojciechMazur: I had no idea the C++ runtime might be distributed under different names depending on platform, that's a shame. 

Having just a `useCPPRuntime` in NativeConfig would work, but again it would require final users to know know that their library dependency is using C++. 
However, I think we should still try to detect if C++ runtime is actually needed. I've made some experiments on that below based on the 2nd approach mentioned above: 
- if available use `llvm-nm` / `nm` to inspect .o files for usage of C++ exceptions or stdlib 
- if `llvm-nm` is not available check if there are any C++ files being compiled - we test if after executing preprocessor the input .cpp source is non empty 
`nm` / `llvm-nm` should be often available on Unix machines, and for Windows we always require C++ so the check is not needed, so that should not be a problem

This way we should cover most of the cases where C++ runtime might be required. 

Based on that I wonder if we should keep `NativeConfig.useCPPRuntime` at all. It might be useful to override the detection mechanism, but probably it should not be `Boolean` but rather `Option[Boolean]`. If empty it should mean that we let the Scala Native toolchain to decide / detect if it's needed. That's what we're using for `multithreading`. 
We always can add it later if needed, but once added it would remain there for a long time.

<details> 

```diff
diff --git a/tools/src/main/scala/scala/scalanative/build/Build.scala b/tools/src/main/scala/scala/scalanative/build/Build.scala
index 4c05496b2..f6f6608fa 100644
--- a/tools/src/main/scala/scala/scalanative/build/Build.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Build.scala
@@ -150,6 +150,7 @@ object Build {
             )
           )
         }
+        .map(_.flatten)
 
       /* Finds all the libraries on the classpath that contain native
        * code and then compiles them.
diff --git a/tools/src/main/scala/scala/scalanative/build/Config.scala b/tools/src/main/scala/scala/scalanative/build/Config.scala
index 0f15b83ff..d3bbe7ac5 100644
--- a/tools/src/main/scala/scala/scalanative/build/Config.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Config.scala
@@ -197,10 +197,10 @@ sealed trait Config {
       case _ => false
     }
 
-  private[scalanative] def isCPPRuntimeEnabled: Boolean =
-    targetsWindows || compilerConfig.useCPPRuntime
+  private[scalanative] lazy val requiresCppRuntime: Boolean =
+    usingCPPExceptions || compilerConfig.useCPPRuntime
 
-  private[scalanative] def usingCPPExceptions: Boolean =
+  private[scalanative] lazy val usingCPPExceptions: Boolean =
     targetsWindows
 }
 
diff --git a/tools/src/main/scala/scala/scalanative/build/Discover.scala b/tools/src/main/scala/scala/scalanative/build/Discover.scala
index fc873e34b..647befaee 100644
--- a/tools/src/main/scala/scala/scalanative/build/Discover.scala
+++ b/tools/src/main/scala/scala/scalanative/build/Discover.scala
@@ -52,6 +52,11 @@ object Discover {
     path
   }
 
+  private[scalanative] lazy val `llvm-nm`: Option[Path] =
+    tryDiscover("llvm-nm", "LLVM_BIN")
+      .orElse(tryDiscover("nm"))
+      .toOption
+
   /** Use llvm-config binary on the path or via LLVM_BIN env var */
   private lazy val llvmConfigCLI: String =
     tryDiscover("llvm-config", "LLVM_BIN")
diff --git a/tools/src/main/scala/scala/scalanative/build/LLVM.scala b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
index be0ddf04f..a6d080673 100644
--- a/tools/src/main/scala/scala/scalanative/build/LLVM.scala
+++ b/tools/src/main/scala/scala/scalanative/build/LLVM.scala
@@ -1,7 +1,7 @@
 package scala.scalanative
 package build
 
-import java.io.{File, PrintWriter}
+import java.io.{File, PrintWriter, IOException}
 import java.nio.file.{Files, Path, Paths, StandardCopyOption}
 import scala.sys.process._
 import scala.scalanative.build.IO.RichPath
@@ -11,7 +11,6 @@ import scala.scalanative.nir.Attr.Link
 import scala.concurrent._
 import scala.util.Failure
 import scala.util.Success
-import _root_.java.io.IOException
 
 /** Internal utilities to interact with LLVM command-line tools. */
 private[scalanative] object LLVM {
@@ -32,14 +31,15 @@ private[scalanative] object LLVM {
    *
    *  @param config
    *    The configuration of the toolchain.
-   *  @param paths
-   *    The directory paths containing native files to compile.
+   *  @param path
+   *    The path of C/C++/LLVM IR file to compile.
    *  @return
-   *    The paths of the `.o` files.
+   *    The Some[java.nio.file.Path] of the `.o` file or None if the input
+   *    source is effectivelly empty
    */
   def compile(config: Config, path: Path)(implicit
       ec: ExecutionContext
-  ): Future[Path] = {
+  ): Future[Option[Path]] = {
     implicit val _config: Config = config
 
     val inpath = path.abs
@@ -47,13 +47,26 @@ private[scalanative] object LLVM {
     val objPath = Paths.get(outpath)
     // compile if out of date or no object file
     if (needsCompiling(path, objPath)) compileFile(path, objPath)
-    else Future.successful(objPath)
+    else Future.successful(Some(objPath))
+  }
+
+  @scala.annotation.nowarn("cat=deprecation")
+  private def isEmptyAfterPreprocessing(
+      inpath: String,
+      compiler: String,
+      flags: Seq[String]
+  ): Boolean = {
+    try
+      s"$compiler -E -P -c $inpath ${flags.mkString(" ")}"
+        .lineStream(Logger.toProcessLogger(Logger.nullLogger))
+        .forall(_.trim().isEmpty)
+    catch { case _: Throwable => false }
   }
 
   private def compileFile(srcPath: Path, objPath: Path)(implicit
       config: Config,
       ec: ExecutionContext
-  ): Future[Path] = Future {
+  ): Future[Option[Path]] = Future {
     val inpath = srcPath.abs
     val outpath = objPath.abs
     val isCpp = inpath.endsWith(cppExt)
@@ -112,18 +125,23 @@ private[scalanative] object LLVM {
         configFlags ++ Seq("-fvisibility=hidden", opt) ++
         Seq("-fomit-frame-pointer") ++
         config.compileOptions
-    val compilec: Seq[String] =
-      Seq(compiler, "-c", inpath, "-o", outpath) ++ flags
 
-    // compile
-    config.logger.running(compilec)
-    val result = Process(compilec, workDir.toFile) !
-      Logger.toProcessLogger(config.logger)
-    if (result != 0) {
-      throw new BuildException(s"Failed to compile ${inpath}")
-    }
+    if (!isLl &&
+        isEmptyAfterPreprocessing(inpath, compiler, flags)) None
+    else {
+      val compilec: Seq[String] =
+        Seq(compiler, "-c", inpath, "-o", outpath) ++ flags
+
+      // compile
+      config.logger.running(compilec)
+      val result = Process(compilec, workDir.toFile) !
+        Logger.toProcessLogger(config.logger)
+      if (result != 0) {
+        throw new BuildException(s"Failed to compile ${inpath}")
+      }
 
-    objPath
+      Some(objPath)
+    }
   }
 
   /** Links a collection of `.ll.o` files and the `.o` files from the
@@ -215,6 +233,27 @@ private[scalanative] object LLVM {
     }
   }
 
+  @scala.annotation.nowarn("cat=deprecation")
+  private def requiresCppRuntime(
+      objectsPaths: Seq[Path]
+  )(implicit config: Config): Boolean = {
+    def hasCppSources = objectsPaths.exists(_.toString().endsWith(".cpp.o"))
+    def usesCppRuntime = Discover.`llvm-nm`.map(_.abs).fold(true) { nm =>
+      objectsPaths.exists { path =>
+        try
+          s"$nm -C $path"
+            .lineStream(Logger.toProcessLogger(Logger.nullLogger))
+            .exists { symbol =>
+              // 00000000 T __cxa_allocate_exception - example output
+              symbol.contains(" std::") || // stdlib
+              symbol.contains(" __cxa_") // exceptions
+            }
+        catch { case _: Throwable => false }
+      }
+    }
+    config.requiresCppRuntime || (hasCppSources && usesCppRuntime)
+  }
+
   private def prepareLinkCommand(
       objectsPaths: Seq[Path],
       analysis: ReachabilityAnalysis.Result
@@ -223,6 +262,7 @@ private[scalanative] object LLVM {
     val links = {
       val srclinks = analysis.links.map(_.name)
       val gclinks = config.gc.links
+
       // We need extra linking dependencies for:
       // * libdl for our vendored libunwind implementation.
       // * libpthread for process APIs and parallel garbage collection.
@@ -308,10 +348,10 @@ private[scalanative] object LLVM {
       finally pw.close()
     }
 
-    val compiler =
-      if (config.isCPPRuntimeEnabled) config.clangPP.abs else config.clang.abs
-
-    val command = Seq(compiler, s"@${configFile.getAbsolutePath()}")
+    val linker =
+      if (requiresCppRuntime(objectsPaths)) config.clangPP
+      else config.clang
+    val command = Seq(linker.abs, s"@${configFile.getAbsolutePath()}")
     config.logger.running(command)
     Process(command, config.workDir.toFile())
   }
diff --git a/tools/src/main/scala/scala/scalanative/build/NativeLib.scala b/tools/src/main/scala/scala/scalanative/build/NativeLib.scala
index 6e7098454..11d227881 100644
--- a/tools/src/main/scala/scala/scalanative/build/NativeLib.scala
+++ b/tools/src/main/scala/scala/scalanative/build/NativeLib.scala
@@ -45,9 +45,11 @@ private[scalanative] object NativeLib {
     val destPath = unpackNativeCode(nativeLib)
     val paths = findNativePaths(destPath)
     val projConfig = configureNativeLibrary(config, analysis, destPath)
-    Future.sequence {
-      paths.map(LLVM.compile(projConfig, _))
-    }
+    Future
+      .sequence {
+        paths.map(LLVM.compile(projConfig, _))
+      }
+      .map(_.flatten)
   }
 
   /** Update the project configuration if a project `Descriptor` is present.
diff --git a/tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala b/tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala
index 0aea0df73..6af82f3f5 100644
--- a/tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala
+++ b/tools/src/main/scala/scala/scalanative/linker/LinktimeValueResolver.scala
@@ -23,7 +23,6 @@ private[linker] trait LinktimeValueResolver { self: Reach =>
       s"$linktimeInfo.debugMode" -> (conf.mode == Mode.debug),
       s"$linktimeInfo.releaseMode" -> (conf.mode == Mode.releaseFast || conf.mode == Mode.releaseFull || conf.mode == Mode.releaseSize),
       s"$linktimeInfo.isMultithreadingEnabled" -> conf.multithreadingSupport,
-      s"$linktimeInfo.isCPPRuntimeEnabled" -> config.isCPPRuntimeEnabled,
       s"$linktimeInfo.isWeakReferenceSupported" -> {
         conf.gc == GC.Immix ||
         conf.gc == GC.Commix

```

<details>

tools/src/main/scala/scala/scalanative/build/Config.scala
WojciechMazur, line 204: Maybe instead let's call it `requireCppRuntime` so that we can actually detect if the Cxx is needed, we can then also deduplicate some detections. 

```suggestion
    private[scalanative] lazy val requiresCppRuntime: Boolean =
    usingCPPExceptions || compilerConfig.useCPPRuntime

  private[scalanative] def usingCPPExceptions: Boolean =
    targetsWindows
```

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
lolgab: Checking if there are non-empty c++ sources is wrong because not all c++ sources require linking the c++ runtime. With this approach deleting or not having the `llvm-nm` binary creates the "spooky action at a distance" if linking the c++ runtime which is completely unrelated.
What about adding a new annotation like `@link("name")` but specifically to enable the c++ runtime?
This way it's the Scala code you link together deciding if you link it or not.
It could be `@linkCPPStdlib` or something along those lines.
Also, you get dead-code elimination for free, since if you don't use the c++ function, you don't visit the annotation and you end up not linking using `clang++`.
It should be faster as well since you don't need to execute external programs or analyze object files.
Let me know what think about this other approach.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
lolgab: Not all C++ sources require the C++ runtime.
libunwind has C++ sources that don't require any libstdc++ to work. So disallowing c++ compilation when you don't have the c++ runtime is wrong.

tools/src/main/scala/scala/scalanative/build/NativeConfig.scala
WojciechMazur: The dedicated annotation might be a good idea, maybe let's try to add it. 
Probably we can either use the existing mechanism for `@link` and filter out and handle special links early or add new attributes in NIR.

lolgab: @WojciechMazur I added the annotation. I'm ready for another review round 👍

nativelib/src/main/scala/scala/scalanative/unsafe/linkCppRuntime.scala
```diff
@@ -0,0 +1,7 @@
+package scala.scalanative
+package unsafe
+
+/** An annotation that is used to link the c++ runtime using clang++ instead of
+ *  clang.
+ */
```
lolgab, line 6: This scaladoc needs to be improved

nir/src/main/scala/scala/scalanative/nir/serialization/BinarySerializer.scala
```diff
@@ -195,6 +195,7 @@ final class BinarySerializer(channel: WritableByteChannel) {
       case Attr.Stub               => putTag(T.StubAttr)
       case Attr.Extern(isBlocking) => putTag(T.ExternAttr); putBool(isBlocking)
       case Attr.Link(s)            => putTag(T.LinkAttr); putString(s)
+      case Attr.LinkCppRuntime     => putTag(T.LinkCppRuntimeAttr)
```
WojciechMazur, line 198: Becouse we're adding new attribute we need to update NIR version in `nir/src/main/scala/scala/scalanative/nir/Versions.scala` 
The change is backward binary compatible so we need to change it from 6.10 to 6.11

WojciechMazur (changes requested):
Great job, overall looks good to me, there is only 1 remaining change to versioning of the NIR format

nir/src/main/scala/scala/scalanative/nir/serialization/BinarySerializer.scala
lolgab: ✅  Done

WojciechMazur (approved):

PR #4139, opened 2025-01-01
Convert boxing runtime calls into NIR box/unbox instructions

lolgab:
The NIR box/unbox instructions can be eliminated by the optimizer. When the Scala compiler generates scala.runtime.BoxesRunTime those are treated like any other method calls and so the optimizer can't remove them.
When evaluating blocks we convert those calls into `nir.Op.Box`, `nir.Op.Unbox` instructions, which can be later optimized away by interflow.

Fixes https://github.com/scala-native/scala-native/issues/4132
Fixes https://github.com/scala-native/scala-native/issues/4137


He-Pin:
A question is ,does this optimization lives in Scala 3's backend too?

lolgab:
> A question is ,does this optimization lives in Scala 3's backend too?

I'm not sure if you can do this in the `nscplugin`. In case we can, we need to implement it twice for Scala 3 and Scala 2 and it will be applied only to libraries compiled with Scala Native published after the fix.
Now it's implemented in the optimizer which applies to all versions of Scala and works also for all libraries published with Scala Native `0.5.0+`.
I'll leave it to the reviewer to decide where to put the optimization.

He-Pin:
I mean write a MiniPhase in Scala3 compiler，both are trees 🌳 transformation

tools/src/main/scala/scala/scalanative/interflow/Eval.scala
```diff
@@ -150,6 +150,7 @@ private[interflow] trait Eval { self: Interflow =>
     def bailOut =
       throw BailOut("can't eval op: " + op.show)
     op match {
+      case BoxesRunTimeCall(value) => value
```
WojciechMazur, line 153: If possible we should try to move it `tools/src/main/scala/scala/scalanative/interflow/Intrinsics.scala` - that's the place where we apply special transformation for some well know functions. 
Otherwise looks good to me

lolgab: Done ✅

WojciechMazur: Sorry if I was not accurate enough and could have introduced confusion here. The `Intrinsics` defines a list of intrinsic methods that we can are able handle and their appropriate handlers. Few lines below  in the `case nir.Op.Call` we run 
```scala
        emeth match {
          case nir.Val.Global(name: nir.Global.Member, _)
              if intrinsics.contains(name) =>
            intrinsic(sig, name, args).getOrElse {
              nonIntrinsic
            }
          case _ =>
            nonIntrinsic
        }
```
By my comment I meant that we should keep Eval as it it, but instead register new intrinsics and handle them in `Intrinsics::intrinsic` function. If that's not an issue we should keep intrinsic handlers together so that it's easier to maintain it in the future

lolgab: After some work (handling the new calls the very same way is not easy because we use both `scala.runtime.RunTimeBoxes` and `scala.runtime.RunTimeBoxes$` and sometimes the type is precise , like `java.lang.Integer`, sometimes is `java.lang.Object`), so I removed the `val intrinsics: Set[nir.Global]` and made intrinsic a total function that runs on every Global.
Please check the result.
I had to rewrite this piece of code multiple times because of those details.

tools/src/main/scala/scala/scalanative/interflow/Intrinsics.scala
```diff
@@ -170,13 +172,36 @@ private[interflow] trait Intrinsics { self: Interflow =>
...
+              className + '$'
+            }
+
+            val boxIntrinsic = boxIntrinsicType
+              .get((companionObjectName, methodName))
+              .map(tpe => eval(nir.Op.Box(tpe, rawArgs.last)))
+
+            val unboxIntrinsic = unboxIntrinsicType
+              .get((companionObjectName, methodName))
+              .map(tpe => eval(nir.Op.Unbox(tpe, rawArgs.last)))
```
WojciechMazur, line 200: We can make it lazy
```suggestion
            def unboxIntrinsic = unboxIntrinsicType
              .get((companionObjectName, methodName))
              .map(tpe => eval(nir.Op.Unbox(tpe, rawArgs.last)))
```

WojciechMazur (commented):
LGTM, just one comment about making some computiations lazy

tools/src/main/scala/scala/scalanative/interflow/Intrinsics.scala
lolgab: Ended up inlining the two values.
Let me know if you like it.

WojciechMazur (approved):

PR #4142, opened 2025-01-03
Dead-code eliminate unused boxes runtime methods

lolgab:
`scala.scalanative.runtime.Boxes` and `scala.runtime.BoxesRunTime` were kept regardless of usage.
This treats them as any other object by reaching their methods when a `nir.Op.Box` or `nir.Op.Unbox` operation happens.


lolgab:
It is harder than I thought since we materialize `Boxes` calls after the first Reach phase (sometimes after the second) and checking for `nir.Op.Box/Unbox` is not enough to cover all the cases.

PR #4153, opened 2025-01-07
Detect and handle StackOverflowErrors

WojciechMazur:
Adds guard pages to detect and handle StackOverflowErrors. 

We introduce an additional regions of memory of all created threads equal to 4 * page size to detect access near the end of the stack. We also store information about stack size, stack bottom/top for each of the threads. Based on these we setup 2 guard pages at the end of each stacks regions - these have protected access and would raise signal on access. 
Inside the dedicated signal handler we check if SIGSEGV (or SIGBUS on MacOS) is refering to stack memory, especially the 2 guards. 
We cannot throw from signal handler thus we introduce a special logic on how to handle stack overflow:

When 1st guard is reached we temporarily remove protection for this part of stack and continue execution. We setup an event that would be polled in some well known location (e.g. on call to `Thread.currentNativeThread`). These would trigger throwing stack overflow error outside the signal handler stack. 

Because 1st guard handler is based on polling it might not be triggered. That's why we also define a 2nd page guard. After reaching it would remove protection for rest of the stack and would try to modify the execution context of current thread so that it lands in the function throwing StackOverflowError when it exits the signal handler. However, these modification is unsafe (registers / ucontext manipulation) and might cause libunwind to not be able to throw exception.  If no platform specific logic is defined we continue exececion with hope to reach polling event. 

If code is compiled with `-omit-frame-pointers` or non zero optimizer flag the frame pointers would not be produced. Becouse of that libunwind would try to fallback to other techniques to find out handler for exception being throwing, e.g. it might use DWARF info / compact headers. These we cannot modify, and libunwind would fail to unwind the stack. In such case it would lead to crash of the binary with special message about exception and its stacktrace that we failed to throw. This behaviour even though not perfect is overall an improvement becouse we can actually get information that crash was due to StackOverflow and not other kind of undefined behaviour. 

To setup a restart mechanism for StackOverflow we add an `onCatchHandler` in `scalanative.runtime.Throwable` that would be executed in `eh.c` when exception is caught. We would use it to reset stack guards so that these protect the regions of memory close to the end of the stack. During the process new StackOverflowErrors might be triggered if we have not yet unwinded deep enough - these would be silently swallowed. 

TODO: 
 - [x] - Windows support
 
 Other changes required to implement this feature: 
 - Always reach exported definitions - it would allow to reference Scala exported function to be called from C in Application build mode. 
 - Introduce scala.scalanative.runtime.Throwable from which java.lang.Throwable inherits. It allows us to define custom, Scala Native specific behaviour 
 - Adapt eh.c to print stack trace if exception cannot be thrown 
 - Allow to override default stack size of created thread 
 - Set default stack size of created threads to 1MB (same as JVM) 
 - Allow to override default stack size for created threads using `SCALANATIVE_THREAD_SIZE` env variable
 - Allow to define handler to be called when exception in being cought. 
 - Preallocate memory for _Unwind_exception - it's a part of runtime.Throwable layout
 - Update shipped LibUnwind version 17.0.1 -> 19.1.6 
 - Ensure BlobArray and it's methods are reachable


nativelib/src/main/resources/scala-native/stackOverflowGuards.c
```diff
@@ -1,18 +1,21 @@
-#ifndef _WIN32
+#ifndef defined(__linux__)
```
ekrich, line 1: Did you want `#if !defined(__linux__)` ?

nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
```diff
@@ -36,7 +36,7 @@ static void Synchronizer_WaitForResumption(MutatorThread *selfThread);
 // internal flag, better for debuggin, but slower
 #ifdef SCALANATIVE_GC_USE_YIELDPOINT_TRAPS
 #include "shared/YieldPointTrap.h"
-#include "immix_commix/StackTrace.h"
+#include "../../StackTrace.h"
```
ekrich, line 39: In `resources/scala-native.properties` you can add `, .` the the end of the includes and then you can just use `"StackTrace.h"` anywhere in `nativelib`. We never had the need for any truly shared files up to this point. You would also need to add the following to `compile-flags.txt` so `clangd` would pick it up.
```
-I
.
```

ekrich: I tested it out just to make sure of the ins and outs.

ekrich: > You would also need to add the following to `compile-flags.txt` so clangd would pick it up.
```
-I
.
```
@WojciechMazur Are you using `clangd` in VSCode or another editor? It compiles all the code based on the platform you are on and allow code naviagation etc. so very helpful.

WojciechMazur: Yes, I'm using it regularly. I think we could also think about auto dumping currently used flags to the CWD of the project so these could be automatically updated (propably controlled with env variable or system property set in build.sbt)

nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
```diff
@@ -72,15 +77,67 @@ trait NativeThread {
...
 
+private object ThreadStackSize {
+  final val Minimal = 64 * 1024
+  final val JVMDefault = 1024 * 1024 // 1MB is JVM default
+
+  // Additional stack size to compenstate memory for internals
+  private def extraThreadStackSize: Long = StackOverflowGuards.size
+
+  private val overrideDefaultThreadSize: Option[Long] = {
+    System.getenv("SCALANATIVE_THREAD_SIZE") match {
```
lolgab, line 88: `SCALANATIVE_THREAD_SIZE` is a too generic name. I think you should also mention the stack, like `SCALANATIVE_THREAD_STACK_SIZE`

nativelib/src/main/scala/scala/scalanative/runtime/NativeThread.scala
```diff
@@ -72,15 +77,67 @@ trait NativeThread {
...
+      case null => None
+      case default =>
+        val numberPart = default.takeWhile(_.isDigit)
+        val multiplier = default.stripPrefix(numberPart).toLowerCase() match {
+          case ""         => 1
+          case "k" | "kb" => 1024
+          case "m" | "mb" => 1024 * 1024
+          case other =>
+            System.err.println(
+              s"Invalid setting for SCALANATIVE_THREAD_SIZE env variable would be ignored: $other"
```
ekrich, line 98: Looks like we missed the name change here `SCALANATIVE_THREAD_STACK_SIZE`.

PR #4154, opened 2025-01-08
Better encoding for `unsafe.Nat` to be optimized away

lolgab:
The previous implementation overloaded `Tag` for `Nat` implementing a trait `NatTag` which added a `def toInt: Int` method. The fact that Tag.DigitN both extended `NatTag` and received `NatTag` made its `def toInt: Int` method recursive in the terms of interflow. Interflow was unable to optimize the dynamic dispatch away and the `NatN` classes were materialized even in release-full mode. A new encoding is proposed which adds a new `abstract class NatValue` which has a single `def value: Int` method.
A second `abstract class NatBaseValue` is used to avoid the recursion for `Digit`s. Moreover, instead of creating classes for Digit, which Interflow can't optimize away, we implement the `toInt` on the fly via an anonimous instance which uses `implicitly[NatValue[N]].value` to get the value for the single digits.
Also, instead of using a `var` we inline the sum of powers, which, together with the other techniques used, allows interflow to inline the final value as a int.

```scala
println(implicitly[Tag[CArray[Int, Nat.Digit5[Nat._1, Nat._2, Nat._5, Nat._8, Nat._2]]]].size)
```

now generates the literal result value `50328` (`12582 * 4`) in the NIR:
```
1: %3000007 = box[@"T17java.lang.Integer"] int 50328
```

The result is backward binary compatible since all previous definitions are maintained and marked as `@deprecated` which changed to not have the `implicit` keyword, so the Scala compiler will not choose them anymore.

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/unsafe/CArray.scala
+++ b/nativelib/src/main/scala/scala/scalanative/unsafe/CArray.scala
@@ -37,6 +37,11 @@ final class CArray[T, N <: Nat] private[scalanative] (
     ptr(idx) = value
   }
 
+  @alwaysinline def length(implicit natValue: NatValue[N]): Int = {
+    natValue.value
+  }
+
+  @deprecated("Use overload which takes natValue: NatValue[N]", since = "0.5.7")
   @alwaysinline def length(implicit tag: Tag[N]): Int = {
     tag.asInstanceOf[Tag.NatTag].toInt
   }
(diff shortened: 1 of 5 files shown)


ekrich:
Is this one failing on a test or can a test be added?

lolgab:
@WojciechMazur You are right, this encoding doesn't inline better than the other, the only advantage is it being non-recursive, but if you are going to disable `@alwaysinline` again for recursive functions then there is little point in introducing a new encoding and deprecating the other.
I got confused during my tests.

PR #4160, opened 2025-01-10
Fix missing unwind symbols when linking with lld

lolgab:
Fixes https://github.com/scala-native/scala-native/issues/4159


tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -298,10 +298,22 @@ private[scalanative] object LLVM {
       buildTargetLinkOpts ++ flto ++ debugFlags ++ platformFlags ++ linkNameFlags ++ output ++ sanitizer ++ target
     }
     val paths = objectsPaths.map(_.abs)
+    val (lldStartOptions, lldEndOptions) =
+      if (config.linkingOptions.contains("-fuse-ld=lld")) {
```
WojciechMazur, line 302: `-fuse-ld=lld` is also set internally in same cases on Windows.  We should check flags insteads

tools/src/main/scala/scala/scalanative/build/LLVM.scala
```diff
@@ -298,10 +298,22 @@ private[scalanative] object LLVM {
...
+        (List("-Wl,--start-lib"), List("-Wl,--end-lib"))
+      } else {
+        (Nil, Nil)
+      }
     // it's a fix for passing too many file paths to the clang compiler,
     // If too many packages are compiled and the platform is windows, windows
     // terminal doesn't support too many characters, which will cause an error.
-    val llvmLinkInfo = flags ++ paths ++ linkopts
+    val llvmLinkInfo =
+      flags ++ lldStartOptions ++ paths ++ lldEndOptions ++ linkopts
```
WojciechMazur, line 316: I agree that alternative approach in #4161 might be too eager, but probably we can adopt one change added there --start-lib to make it a bit simpler and reduce ammount of allocations 

https://github.com/scala-native/scala-native/blob/092334a9d87f488ba4e93a89a759eb94155a447b/tools/src/main/scala/scala/scalanative/build/LLVM.scala#L308-L325

tools/src/main/scala/scala/scalanative/build/LLVM.scala
lolgab: I guess this is a problem we have in general if we use that linker, so we probably can add it in all cases where we see the flag. But I can add only when the user adds it manually if you prefer it.

keynmol: Is it just lld? Would that affect Mold as well? Situation would be different if this is a sign of a fundamental issue, not a peculiarity of some particular linker.

tools/src/main/scala/scala/scalanative/build/LLVM.scala
lolgab: I've used your suggestion and extended to `links`.
Ready for another round of review @WojciechMazur

WojciechMazur (approved):

PR #4165, opened 2025-01-11
Add Scala 2.13.16 to the build

WojciechMazur:


SethTisue:
is any further work needed or is Native good to go for 2.13.16?

I tried a hello world program with `scala run -S 2.13.16 --platform native S.scala` and it worked :+1:

PR #4189, opened 2025-01-29
Remove the unwind handler for operations that can never throw exception.

WojciechMazur:
Remove the unwind handler for operations that can never throw exception.. Would allow to remove no longer needed unwind handlers


ekrich:
This is a really good idea 😄

PR #4191, opened 2025-01-31
Fix unwinding when throw is directly before catch 

### LINKED ISSUE #4190: New exceptions can skip finally block in LTO mode
Let's consider this snippet used when testing StackOverflowError correctness

```scala
package scala.scalanative.runtime

object Test {
  def main(args: scala.Array[String]): Unit = {
    val devNull = new java.io.PrintStream(
      new java.io.File("/dev/null")
    )

    @scala.noinline
    @scala.scalanative.annotation.nooptimize 
    def stackoverflow(): Int = {
      // for faster stack overlfow, 
      // with too big stack allocation devNull.println would not be entered before throwing StackOverflowError
      val buf = scala.scalanative.unsafe.stackalloc[Byte](512) 
      devNull.println(".")
      buf.toInt + stackoverflow()
    }
      try stackoverflow()
      catch { case ex: StackOverflowError => println(ex) }
      finally {
        println(scalanative.runtime.getMonitor(devNull.asInstanceOf[scalanative.runtime._Object]).show)
      }
      assert(!Thread.holdsLock(devNull))
  }
}
```
Inside the test we're mostly interested when `devNull.println()` triggers `StackOverflowError` - it's important becouse internally it enters `devNull` monitor twice: 
- in `PrintStream.println(String)` 
- in `PrintStream.printString(String)` - private method

During any normal executed the monitors are correctly closed.  
However, when LTO is enabled there is 1 non-released monitor at the end of execution. It does not happend when switching to C++ based exception handling

WojciechMazur:
Fixes #4190 

During LTO we were able to get into scenarios where we're throwing exception for which the catch handler is defined as a next instruction. Such cases were not handled before, but are handled by C++ exceptions handling. 

The most impacting change include additional condition for `pcOffset == landingPad` meaning that if we're inside handler we should use it. 

This fixes problems where try-finally block executing `object.exitMonitor` routine was skipped leading to deadlocks

### DIFF
--- a/nativelib/src/main/scala/scala/scalanative/runtime/monitor/ObjectMonitor.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/monitor/ObjectMonitor.scala
@@ -41,8 +41,9 @@ private[monitor] class ObjectMonitor() {
    *  become successor thread.
    */
   @volatile private var enterQueue: WaiterNode = _
-  
-  override def toString(): String = s"ObjectMonitor(owner=$ownerThread),waiting=$waiting,recursion=$recursion)"
+
+  override def toString(): String =
+    s"ObjectMonitor(owner=$ownerThread),waiting=$waiting,recursion=$recursion)"
 
   /** Ring list of waiting threads. Access limited by the modification lock.
    *  Upon InEnterQueue the wait zone threads would enqueue to the queue, and
--- a/nativelib/src/main/scala/scala/scalanative/runtime/monitor/ObjectMonitor.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/monitor/ObjectMonitor.scala
@@ -41,6 +41,8 @@ private[monitor] class ObjectMonitor() {
    *  become successor thread.
    */
   @volatile private var enterQueue: WaiterNode = _
+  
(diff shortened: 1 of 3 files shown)


nativelib/src/main/scala/scala/scalanative/runtime/monitor/ObjectMonitor.scala
```diff
@@ -42,6 +42,9 @@ private[monitor] class ObjectMonitor() {
    */
   @volatile private var enterQueue: WaiterNode = _
 
+  override def toString(): String =
+    s"ObjectMonitor(owner=$ownerThread),waiting=$waiting,recursion=$recursion)"
```
lolgab, line 46: Default toString don't return the parameter names but usually just `ClassName(param1,param2,param3)`

lolgab:
I see you changed the code reading LSDA to read some DWARF data. Is DWARF always available? How does it work?

WojciechMazur:
> I see you changed the code reading LSDA to read some DWARF data. Is DWARF always available? How does it work?

Not really, so far I have not observed them locally, but I've added the checks based on LLVM cxx_personality logic 
https://github.com/llvm/llvm-project/blob/77d3f8a92564b533a3c60a8c8e0657c38fd88ba1/libcxxabi/src/cxa_personality.cpp#L663-L665 

In theory at least lpBase can be used to have a correct landing pad but beside that I seems to not improve much

Issue #1513, opened 2019-03-27
Interflow: optimize away exceptional control-flow


densh:
Currently interflow doesn't make any attempts at optimization of the exceptional control-flow. This means that a few common Scala patterns can not be partial-escape-analyzed away, for example:

1. Non-local returns are compiled as exceptions in some cases.
2. `scala.util.Try`.
3. `scala.util.control.Breaks`.

While those are not guaranteed to be zero cost it would be nice to support them as well.

PR #4194, opened 2025-01-31
Optimize throwing exception to local handler as jump

### LINKED ISSUE #4190: New exceptions can skip finally block in LTO mode

### LINKED ISSUE #4195: Throw to local catch handler might be skipped under LTO
> Self contained minimization: 
> 
> ```scala
> @main def Test() =  try {
>       println("in 1 try")
>       try {
>         println("in 2 try")
>         try {
>           println("in 3 try")
>           try throw new RuntimeException()
>           catch { case ex: java.io.IOException => println(ex) }
>         } finally {
>           println("finally 3")
>         }
>       } finally {
>         println("finally 2")
>       }
>     } catch {
>       case ex: java.lang.Throwable => println(s"catch 1: $ex")
>     } finally {
>       println("finally 1")
>     }
> ```
> 
> On JVM, in current SN without LTO, or with LTO using C++ exceptions: 
> ```scala
> in 1 try
> in 2 try
> in 3 try
> finally 3
> finally 2
> catch 1: java.lang.RuntimeException
> finally 1
> ```
> 
> With LTO either not cought at all (handled by top-level handled generated for `main`) or skips 2 finally blocks: 
> 
> ```scala
> in 1 try
> in 2 try
> in 3 try
> catch 1: java.lang.RuntimeException
> finally 1
> ``` 

 _Originally posted by @WojciechMazur in [#4190](https://github.com/scala-native/scala-native/issues/4190#issuecomment-2626968760)_

WojciechMazur:
Allows to skip unwinding process and jump directly to handler if it's defined in the same scope. 

Fixes #4195 on the CodeGen side by eliminating most of unwinding calls that were failing after LTO

### DIFF
--- a/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
+++ b/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
@@ -728,7 +728,7 @@ class IssuesTest {
   @Test def issue4194(): Unit = {
     var tryCounter = 0
     var finallyCounter = 0
-    var cought = false
+    var caught = false
     try {
       tryCounter += 1
       try {
--- a/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
+++ b/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
@@ -738,20 +738,20 @@ class IssuesTest {
           try throw new RuntimeException()
           catch {
             case ex: java.io.IOException => // exception unrelated to throw one
-              fail("Should not be cought")
+              fail("Should not be caught")
           } finally {
             finallyCounter += 1
           }
         } finally {
           finallyCounter += 1
         }
-      } catch { case ex: java.lang.Throwable => cought = true }
+      } catch { case ex: java.lang.Throwable => caught = true }
     } finally {
       finallyCounter += 1
     }
     assertEquals("some finally block was skipped", 3, finallyCounter)
     assertEquals(tryCounter, finallyCounter)
-    assertTrue("exception not cought", cought)
+    assertTrue("exception not caught", caught)
   }
 
   // Based on Scala 2.13.16 fix in delambdafy https://github.com/scala/scala/pull/10831
--- a/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
+++ b/unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
@@ -725,6 +725,35 @@ class IssuesTest {
     assertEquals(result, "b")
   }
 
+  @Test def issue4194(): Unit = {
+    var tryCounter = 0
+    var finallyCounter = 0
+    var cought = false
+    try {
+      tryCounter += 1
+      try {
+        tryCounter += 1
+        try {
+          tryCounter += 1
+          try throw new RuntimeException()
+          catch {
+            case ex: java.io.IOException => // exception unrelated to throw one
+              fail("Should not be cought")
+          } finally {
+            finallyCounter += 1
+          }
+        } finally {
+          finallyCounter += 1
+        }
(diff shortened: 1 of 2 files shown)


unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
```diff
@@ -725,6 +725,35 @@ class IssuesTest {
     assertEquals(result, "b")
   }
 
+  @Test def issue4194(): Unit = {
+    var tryCounter = 0
+    var finallyCounter = 0
+    var cought = false
```
ekrich, line 731:
```suggestion
    var caught = false
```

unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
```diff
@@ -725,6 +725,35 @@ class IssuesTest {
...
+    try {
+      tryCounter += 1
+      try {
+        tryCounter += 1
+        try {
+          tryCounter += 1
+          try throw new RuntimeException()
+          catch {
+            case ex: java.io.IOException => // exception unrelated to throw one
+              fail("Should not be cought")
```
ekrich, line 741:
```suggestion
              fail("Should not be caught")
```

unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
```diff
@@ -725,6 +725,35 @@ class IssuesTest {
...
+          } finally {
+            finallyCounter += 1
+          }
+        } finally {
+          finallyCounter += 1
+        }
+      } catch { case ex: java.lang.Throwable => cought = true }
```
ekrich, line 748:
```suggestion
      } catch { case ex: java.lang.Throwable => caught = true }
```

unit-tests/native/src/test/scala/scala/scalanative/IssuesTest.scala
```diff
@@ -725,6 +725,35 @@ class IssuesTest {
...
+    } finally {
+      finallyCounter += 1
+    }
+    assertEquals("some finally block was skipped", 3, finallyCounter)
+    assertEquals(tryCounter, finallyCounter)
+    assertTrue("exception not cought", cought)
```
ekrich, line 754:
```suggestion
    assertTrue("exception not caught", caught)
```

ekrich (commented):
Change spelling - think I got them all.

WojciechMazur:
In the future this backend optimization should be replaced with more general approach in #1513

PR #4242, opened 2025-02-24
Upgrade scalafmt to 3.9.4

ekrich:
There are so changes so I will add them here.

https://gist.github.com/ekrich/bcd8e364d1b79857507a9ce58431daec

Super unscientific but native seems faster here.
```sh
scripts/scalafmt --test  146.37s user 4.84s system 692% cpu 21.836 total
scalafmt --test  43.02s user 30.60s system 330% cpu 22.282 total
```

### DIFF
--- a/auxlib/src/main/scala-3/scala/runtime/function/JProcedure.scala
+++ b/auxlib/src/main/scala-3/scala/runtime/function/JProcedure.scala
@@ -6,6 +6,13 @@
 //
 // See nativelib runtime/Arrays.scala.gyb for details.
 
+/* scalafmt: {
+     binPack.defnSite = always
+     binPack.callSite = always
+     newlines.configStyle.fallBack.prefer = false
+   }
+ */
+
 package scala.runtime.function
 
 import scala.runtime.BoxedUnit
@@ -93,25 +100,11 @@ trait JProcedure8[T1, T2, T3, T4, T5, T6, T7, T8]
     extends scala.Function8[T1, T2, T3, T4, T5, T6, T7, T8, Object]
     with java.io.Serializable {
   def applyVoid(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8
   ): Unit
 
   def apply(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8
   ): Object = {
     applyVoid(t1, t2, t3, t4, t5, t6, t7, t8)
     return BoxedUnit.UNIT
@@ -121,27 +114,11 @@ trait JProcedure9[T1, T2, T3, T4, T5, T6, T7, T8, T9]
     extends scala.Function9[T1, T2, T3, T4, T5, T6, T7, T8, T9, Object]
     with java.io.Serializable {
   def applyVoid(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8,
-      t9: T9
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8, t9: T9
   ): Unit
 
   def apply(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8,
-      t9: T9
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8, t9: T9
   ): Object = {
     applyVoid(t1, t2, t3, t4, t5, t6, t7, t8, t9)
     return BoxedUnit.UNIT
@@ -151,28 +128,12 @@ trait JProcedure10[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10]
     extends scala.Function10[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, Object]
     with java.io.Serializable {
   def applyVoid(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8,
-      t9: T9,
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8, t9: T9,
       t10: T10
   ): Unit
 
   def apply(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8,
-      t9: T9,
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8, t9: T9,
       t10: T10
   ): Object = {
     applyVoid(t1, t2, t3, t4, t5, t6, t7, t8, t9, t10)
@@ -181,46 +142,17 @@ trait JProcedure10[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10]
 }
 trait JProcedure11[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11]
     extends scala.Function11[
-      T1,
-      T2,
-      T3,
-      T4,
-      T5,
-      T6,
-      T7,
-      T8,
-      T9,
-      T10,
-      T11,
-      Object
+      T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, Object
     ]
     with java.io.Serializable {
   def applyVoid(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8,
-      t9: T9,
-      t10: T10,
-      t11: T11
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8, t9: T9,
+      t10: T10, t11: T11
   ): Unit
 
   def apply(
-      t1: T1,
-      t2: T2,
-      t3: T3,
-      t4: T4,
-      t5: T5,
-      t6: T6,
-      t7: T7,
-      t8: T8,
-      t9: T9,
-      t10: T10,
-      t11: T11
+      t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8: T8, t9: T9,
+      t10: T10, t11: T11
   ): Object = {
     applyVoid(t1, t2, t3, t4, t5, t6, t7, t8, t9, t10, t11)
     return BoxedUnit.UNIT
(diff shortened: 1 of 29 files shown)


LeeTibbert:
> Super unscientific but native seems faster here.

Given some clues on the scalafmt Issue, I too was able to get SN scalafmt up and running.
When I did a --check-only of an already formatted scratch SN repository tree, the SN 'real' (wall clock elapsed)
time was about 42% of the JVM time.  

It appears that the SN scalafmt binary packaging & distribution is still "In Process". When that settles
so that we have a reproducible src origin, what do you think about switching to using the SN binary?
Or do you want a couple of us to use it privately for a few weeks before going whole hog?

Perhaps the existing `scripts/scalafmt` could be renamed to `scripts/scalafmt-JVM` and a new
`scripts/scalafmt` created which first checked if the SN image existed and, if present, used it?
Some of the CHECK_MODIFIED_ONLY logic would probably need to be kept to maintain
swap-in/swap-out compatibility.

Thank you for running the new JVM scalafmt 4.9.1 on the SN repository and finding all
those changes. 

 Looks to me like the new version is stripping off some trailing blanks/spaces
(the way you had always thought it did).  In a number of places I could not quite figure out
why and how it was changing the line.

ekrich:
I was really just seeing if there were any major objections to the changes. If we setup Scala Steward, the PR would include the changes and an entry in `.git-blame-ignore-revs`. 

There is a change to the `implicit Zone` placement and a few non binpacked items plus the space Lee mentioned otherwise it seems workable to me. In `sconfig` I had changes for `3.7.0` and `3.8.2` so we are seeing the `3.8.2` changes here.

If there are no objections I can add a commit for the changes and a `.git-blame-ignore-revs` change with the SHA.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
```diff
@@ -1351,10 +1355,12 @@ object ConcurrentHashMap {
...
-          baseSize,
-          { baseLimit = h; h },
-          f,
-          { est >>>= 1; est },
+          baseSize, {
+            baseLimit = h; h
+          },
+          f, {
+            est >>>= 1; est
+          },
```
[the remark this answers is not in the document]
tgodzik: Not sure, maybe something around newlines would help? https://scalameta.org/scalafmt/docs/configuration.html#newlinessource

ekrich:
I can look into this - I wasn't super excited about this formatting either.
The rest was really just trailing space.

javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala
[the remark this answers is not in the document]
kitbellew: scalameta/scalafmt#4851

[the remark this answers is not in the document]
ekrich: Thank you so much - for the record formatting has been very stable with just minor changes.

ekrich:
The changes I added to the files are things I saw or adjustments to fix things that `scalafmt` will leave alone.

- no space added after `/*` for inline comments but after so I added a space after for consistency and scalafmt leaves alone.
- number and letter sequences now are indented so the `j.` was considered one of these so I added a word to avoid the formatting
- now scalafmt will not unwrap `{ condition }` and will allow folding after infix (operator) like `&&` in `if` and `while` so I adjust a site that was unwrapped.

ekrich:
Update: Looks like we will have to wait for `3.9.4` as the indent has been changed - maybe due to some other fixes they have done for us.

LeeTibbert:
I am tidying up a moderate whack of code for a PR and I think I am seeing something
which might affect a mass conversion of the SN codebase and be hard to find.

Using the very fast SN scalafmt 3.9.4 and 3.9.3, I am seeing what I believe to be a glitch
in indenting (well, outdenting) /* */ comment blocks.

@kitbellew Is this me for am I really seeing a scalafmt glitch?  
Thank you, and my apology if the glitch is on my part. 

I think we have chased regex expression is this area sometime in the
past year (or does the memory of pain persist long after the anniversary?).

```
  /* JVM 23 and probably earlier uses the "bound must be greater than origin"   
   * message if this argument check fails.                                      
   * This is appropriate in the (bound <= offset) but at least two              
   * conditions lead to counter-factual nonsense:                               
   *   rng.nextDouble(Double.NEGATIVE_INFINITY, 5.0)  // bound is > origin      
   *   rng.nextDouble(1.0 , Double.POSITIVE_INFINITY) // bound is > origin      
   *                                                                            
     * Cases where either or both arguments are Double.NaN give the             
   * same message instead of reporting NaN. Technically correct but             
   * less than helpful.                                                         
   *                                                                            
   * These two methods handle NaNs invisibly. Recall                            
   * jl.Double.isFinite(scala.NaN) == false.                                    
   */
```

kitbellew:
@LeeTibbert are you running formatting on this specific codebase, with the same configuration file? and the bug appears only using native version only, not jvm?

ekrich:
@kitbellew The changes are looking good to me. Most of the changes are training spaces followed by some other formatting that was not really correct before. I am still seeing a few lines longer that 80 but they can be hand wrapped if needed.

ekrich:
I am in the process of making some manual changes to clean up a few things.

javalib/src/main/scala/java/util/jar/JarFile.scala
```diff
@@ -125,7 +125,7 @@ class JarFile(file: File, verify: Boolean, mode: Int)
     val in = super.getInputStream(ze)
     if (in == null) {
       null
-    } else if (/*verifier == null || */ ze.getSize() == -1) {
+    } else if ( /* verifier == null || */ ze.getSize() == -1) {
```
ekrich, line 128: Not sure why this is in the code. @LeeTibbert any ideas if this should be removed in another PR?

ekrich:
@WojciechMazur This should be ready for review now. I think the changes are very positive including trailing spaces and some indent differences and especially `if` can be wrapped after logicals `&&` or `||`  making them more readable especially with embedded blocks e.g. `{ foo = bar; foo }` can be kept inline unless they are super long. We will need an entry after merge in the `.git-blame-ignore-revs` file.

Oops, one errant character so one more commit.

auxlib/src/main/scala-3/scala/runtime/function/JProcedure.scala
kitbellew, line 1: specifically for this file, you could add, somewhere at the beginning, this:
```
/* scalafmt: {
     binPack.defnSite = always
     binPack.callSite = always
     newlines.configStyle.fallBack.prefer = false
   }
 */
```

kitbellew (commented):
looks as expected. one comment, to improve readability of one specific file.

auxlib/src/main/scala-3/scala/runtime/function/JProcedure.scala
ekrich: Thanks for the idea. It makes it much less linear. We initially tried to be as default as possible but if the team likes this idea we could do it after this PR - https://gist.github.com/ekrich/daaf86c4da0e6889094ae37e12f3c8ce

kitbellew: i wouldn't do this everywhere, only in this file, as it contains a bunch of virtually identical and very simple code (all arguments are the same, all types are parametric, there's no complexity). in other files, that's not the case.

ekrich: Sorry, I didn't read that correctly. That might be good for the other generated file with bunches of type parameters.

auxlib/src/main/scala-3/scala/runtime/function/JProcedure.scala.gyb
kitbellew, line 1: if this is not a `.scala` file, it wouldn't be formatted anyway.

ekrich: The header gets copied if it ever gets generated again.

LeeTibbert:
@kitbellew 

I've been using the Scala Native scalafmt pretty heavily and it is very fast. Good work.
It might be my imagination, but I think scalafmt 3.9.4 is blink or two faster than 3.9.3.
No measurement numbers though 

> are you running formatting on this specific codebase
Same codebase direct from daily pulls from GitHub, with the exception of my current
Work-In-Progress in `java.util.random`.  I am using the linux gnu scalafmt artifact.
 
>  with the same configuration file? 

Only change is to the 'version' line in order to use scalafmt 3.9.4

> and the bug appears only using native version only, not jvm?

Good question, I noticed it this past week, because I was running the SN scalafmt
experiment and had my eyes looking for quirks.

I am within days of submitting a SN PR. One of my final steps will be to run the JVM
variant of scalafmt on the candidate files. I'll look for how the JVM variant handles
the errant asterisk.

I may be misremembering prior (< 3.9) scalafmt behavior.

If this issue exists, it is IMHO, tiny and, if other people are not seeing it, not hold up this PR.

LeeTibbert:
> The header gets copied if it ever gets generated again

I have two new javalib `Arrays.scala` .gyb files waiting to submit, so any
improvements to the current .gyb practice should get multiplied. 

The existing .gyb files that I have been studying generate Scala code which explicitly
turns formatting OFF.

javalib/src/main/scala/java/util/jar/JarFile.scala
ekrich: @LeeTibbert Any ideas?

LeeTibbert:
@ekrich @kitbellew 

Perhaps I should create a formal SN issue for this discussion.  Let me sketch some 
concerns here, so that I/we can begin to formulate an intelligible discussion.

I've been successfully using scalafmt-for-SN pretty intensively since it became available.
Human nature being what it is, the speed up going from scalafmt-JVM to scalafmt-SN is
noticible; the slowdown going back from scalafmt-SN to scalafmt-JVM is marked.

I think the next steps are to make scalafmt-SN available to first to SN devos and then to CI.

kitbellew, what is the eventual strategy for installing scalafmt-SN on a system?   I understand
that the scalafmt-SN bits have not even cooled down yet and that we are in early days.

I've been using the 'manually-copy-down-the-file-and-make-sure-you-have-the-right-one'
method, along with a personal SN 'scritps/scalafmt-SN' file.  Fine for early and bootstrap development
but not for CI.   

I can conceive of several ~~hack~~ temporary ways forward for devo-only use until there is a formal,
multi-platform scalafmt-SN. I'm  interested if there are SN or scalafmt plans in motion or if either or both of you
 have ideas.  


Thank you for your time.

kitbellew:
@LeeTibbert i am not aware of any formal plans. however, @tgodzik is the driving force behind native formatter.

javalib/src/main/scala/java/util/jar/JarFile.scala
LeeTibbert: @ekrich  

TL;DR  

I suggest accepting the new line 128, stepping smartly, and whistling  past any existing nonsense 
in 'JarFile.scala'.  

I usually shoot for a higher standard, but here a minimal change does no worse
than existing behavior and preserves possibilities of more beneficial change elsewhere.

Longer story:

IMO, any change to 'JarFile.scala' is exceeding likely to be economic, especially when
contrasted to other uses of a good devo's time & energy. 

`java. util.jar` relies upon 'java.util.zip'.  I have only briefly scanned the 'jar' code,
but I have a good idea of the defect density of the 'zip' code.  Actually, I had to look
at that code recently for some 'ZipEntryTest' work.  Not all the defects there are
due to changes since the code was written.  

Now this may sound like Fear, Uncertainty, & Doubt, but is the vivid memory of many
hours of pain past, and, yes, pain yet to be.

LeeTibbert:
kitbellew 

Thank you for the timely & helpful info.   I think I well co-ordinate with @ekrich and
sketch out a plan at the SN end, including contacting todzik directly.  When I surface
from my current PR, I'll look at the scalafmt github make myself current.  Events
flash by.

javalib/src/main/scala/java/util/jar/JarFile.scala
ekrich: Ok, that is fine. I just ran across it and wondered.

ekrich:
@LeeTibbert This ticket is just about updating the Scala Format version. I think another ticket is in order if we plan to use the SN built version instead of the JVM version.

tgodzik:
btw. what is currently working best is running Scala CLI `scala format` command which will use native image for older versions and Scala native version for the latest versions. It will download the correct version and run it on it's own.

There is one small issue that it will create .scala-build directory with copied .scalafmt.conf, which we should be able to fix.

WojciechMazur (approved):
Let's wait a couple of days before merging this PR until some other big changes in the queue are merged, eg #4267 and related methods, just to reduce possible source code conflicts after upgrade

PR #4250, opened 2025-02-26
Reduce number of wrappers when fetching aliveThreads

JD557:
Changes `NativeThread.Registry.aliveThreads` into two methods: `NativeThread.Registry.aliveThreadsIterator` and `NativeThread.Registry.numAliveThreads`, to try to reduce the number of wrappers allocated when operating on this.

Follow up to the profile in https://profiler.firefox.com/public/xj9vrexr9t1yz0egx06sty9zhsamvphrq57t4y8/calltree/?globalTrackOrder=0w2&hiddenGlobalTracks=12&hiddenLocalTracksByPid=97759-1w7&symbolServer=http%3A%2F%2F127.0.0.1%3A3000%2F0qqvz038sfbj7vinzp5a09rrfw5wsv0m5fdb253&thread=0&v=10

I see quite a few calls to allocations in `pollNonDaemonThreads`. This is called in a `while(true)` loop, so I guess it's expected that this method is called a lot.

The original method looked like:
```scala
NativeThread.Registry.aliveThreads.iterator
      .map(_.thread)
      .filter { thread =>
        (thread ne shutdownThread) && !thread.isDaemon() &&
        thread.isAlive()
      }
```

Since `aliveThreads` is a Scala wrapped over a Java collection (a [`JIterableWrapper`](https://github.com/scala/scala/blob/347341babac6a5e75c61c25fdf8b0417ad7d2188/src/library/scala/collection/convert/JavaCollectionWrappers.scala#L83-L95)) where most methods are implemented on top of [`iterator`](https://github.com/scala/scala/blob/347341babac6a5e75c61c25fdf8b0417ad7d2188/src/library/scala/collection/convert/JavaCollectionWrappers.scala#L87) this will:
- Create a Java iterator
- Create a Scala iterator wrapper
- Create an extra iterator for the map
- Create an extra iterator for the filter

Only to check an `hasNext` later.

This changes it to
```scala
def pollNonDaemonThreads = NativeThread.Registry.aliveThreadsIterator.exists { aliveThread =>
        val thread = aliveThread.thread
        (thread ne shutdownThread) &&
        !thread.isDaemon() &&
        thread.isAlive()
      }
```

Which should not allocate any object in the heap after the call to `NativeThread.Registry.aliveThreadsIterator`

Other parts of the code were also changed for similar reasons.

I did not check if some of those allocations were already optimized away by Interflow, but based on the profiler I don't think that's always the case.

### DIFF
--- a/javalib/src/main/scala/java/lang/ThreadGroup.scala
+++ b/javalib/src/main/scala/java/lang/ThreadGroup.scala
@@ -77,7 +77,7 @@ class ThreadGroup(
   def isDestroyed(): scala.Boolean = false
 
   def activeCount(): Int = {
-    NativeThread.Registry.aliveThreads
+    NativeThread.Registry.aliveThreadsIterator
       .count { nativeThread =>
         val group = nativeThread.thread.getThreadGroup()
         this.parentOf(group)
@@ -109,7 +109,7 @@ class ThreadGroup(
     if (out == null) throw new NullPointerException()
     if (out.length == 0) 0
     else {
-      val aliveThreads = NativeThread.Registry.aliveThreads.toArray
+      val aliveThreads = NativeThread.Registry.aliveThreadsIterator.toArray
       @tailrec def loop(idx: Int, included: Int): Int =
         if (idx == aliveThreads.length || included == out.length) included
         else {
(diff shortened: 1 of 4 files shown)


ekrich:
Are you aware of `javalib` `java.util.ScalaOps`? It brings Scala niceness while using Java collections. In general we discourage use Scala collections etc. in the `javalib` so this may help you.

JD557:
Good point. Removed the Scalalib dependency in [3ab9824](https://github.com/scala-native/scala-native/pull/4250/commits/3ab98240d7e12704e5fba565e10313eb00e19ae3)

WojciechMazur:
The glorious Microsoft has again upgraded the minimal version of LLVM required to compile our glue layer with its API...eh. 
The fix is on the way in #4251 PR would require a rebase to resolved failing CI builds

WojciechMazur (approved):

PR #4301, opened 2025-04-10
Fix `isPositionIndependentBinary`

durban:
As far as I can tell, the `isPositionIndependentBinary` information is parsed from the ELF header, but then it's just hardcoded to `true`. This PR fixes that.

This seems to fix the incorrect exception stacktraces I've encountered locally (as I've mentioned in typelevel/cats-effect/pull/4362).


durban:
CI failed due to check-cla, but I've "signed" the CLA since then.

durban:
Well, something else also failed... :man_shrugging:

WojciechMazur (approved):

PR #4310, opened 2025-04-19
Make error messages more consistent

JD557:
I noticed that the error messages were a bit inconsistent... some used `ScalaNative Fatal Error:` others `Scala Native:`, others `ScalaNative ::`...

Normalized them into two types:
- `ScalaNative:`
- `ScalaNative Fatal Error:`

I actually suspect that they are all fatal errors, but I don't have enough domain knowledge to know.

When possible (i.e. when the error messages are constructed with `fprintf`), I use a constant for the prefix. This should not only guarantee that things stay consistent, but might also drop the binary sizes a tiny bit.

Incidentally, this PR also fixes a few typos in the error messages.

### DIFF
--- a/nativelib/src/main/resources/scala-native/constants.h
+++ b/nativelib/src/main/resources/scala-native/constants.h
@@ -1,7 +1,7 @@
 #ifndef CONSTANTS_H
 #define CONSTANTS_H
 
-char* snErrorPrefix;
-char* snFatalErrorPrefix;
+char *snErrorPrefix;
+char *snFatalErrorPrefix;
 
 #endif // CONSTANTS_H
--- a/nativelib/src/main/resources/scala-native/constants.h
+++ b/nativelib/src/main/resources/scala-native/constants.h
@@ -1,0 +1,7 @@
+#ifndef CONSTANTS_H
+#define CONSTANTS_H
+
+char* snErrorPrefix;
+char* snFatalErrorPrefix;
+
+#endif // CONSTANTS_H
--- a/nativelib/src/main/resources/scala-native/constants.h
+++ b/nativelib/src/main/resources/scala-native/constants.h
@@ -1,7 +1,7 @@
 #ifndef CONSTANTS_H
 #define CONSTANTS_H
 
-char *snErrorPrefix;
-char *snFatalErrorPrefix;
+extern const char *const snErrorPrefix;
+extern const char *const snFatalErrorPrefix;
 
 #endif // CONSTANTS_H
--- a/nativelib/src/main/resources/scala-native/constants.h
+++ b/nativelib/src/main/resources/scala-native/constants.h
@@ -1,6 +1,9 @@
 #ifndef CONSTANTS_H
 #define CONSTANTS_H
 
+#define SN_ERROR_MSG(msg) "ScalaNative: " msg
+#define SN_FATAL_ERROR_MSG(msg) "ScalaNative Fatal Error: " msg
+
 extern const char *const snErrorPrefix;
 extern const char *const snFatalErrorPrefix;
 
--- a/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
+++ b/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
@@ -1,6 +1,7 @@
 #if defined(SCALANATIVE_MULTITHREADING_ENABLED) &&                             \
     defined(SCALANATIVE_GC_COMMIX)
 
+#include "../constants.h"
 #include "immix_commix/Synchronizer.h"
 #include "shared/ScalaNativeGC.h"
 #include <stdio.h>
@@ -106,8 +107,9 @@ static void SafepointTrapHandler(int signal, siginfo_t *siginfo, void *uap) {
     }
 
     fprintf(stderr,
-            "ScalaNative :: Unhandled signal %d triggered when accessing "
+            "%s Unhandled signal %d triggered when accessing "
             "memory address %p, code=%d\n\n",
+            snErrorPrefix,
             signal, siginfo->si_addr, siginfo->si_code);
     StackTrace_PrintStackTrace();
     abort();
--- a/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
+++ b/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
@@ -1,7 +1,7 @@
 #if defined(SCALANATIVE_MULTITHREADING_ENABLED) &&                             \
     defined(SCALANATIVE_GC_COMMIX)
 
-#include "constants.h"
+#include "string_constants.h"
 #include "immix_commix/Synchronizer.h"
 #include "shared/ScalaNativeGC.h"
 #include <stdio.h>
--- a/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
+++ b/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
@@ -109,8 +109,7 @@ static void SafepointTrapHandler(int signal, siginfo_t *siginfo, void *uap) {
     fprintf(stderr,
             "%s Unhandled signal %d triggered when accessing "
             "memory address %p, code=%d\n\n",
-            snErrorPrefix,
-            signal, siginfo->si_addr, siginfo->si_code);
+            snErrorPrefix, signal, siginfo->si_addr, siginfo->si_code);
     StackTrace_PrintStackTrace();
     abort();
 }
--- a/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
+++ b/nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
@@ -1,7 +1,7 @@
 #if defined(SCALANATIVE_MULTITHREADING_ENABLED) &&                             \
     defined(SCALANATIVE_GC_COMMIX)
 
-#include "../constants.h"
+#include "constants.h"
 #include "immix_commix/Synchronizer.h"
 #include "shared/ScalaNativeGC.h"
 #include <stdio.h>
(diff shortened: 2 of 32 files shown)


nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
```diff
@@ -1,6 +1,7 @@
 #if defined(SCALANATIVE_MULTITHREADING_ENABLED) &&                             \
     defined(SCALANATIVE_GC_COMMIX)
 
+#include "../constants.h"
```
ekrich, line 4: This can be just `"constants.h"`. If you look at `compile_flags.txt` you can that both `.` and `gc` are added to the include path. This of course is for using `clangd` (LSP) - See https://scala-native.org/en/stable/contrib/ides.html#setup-for-clangd

The real include path sent to the Scala Native compiler (tools) is in the `scala_native.properties` in each project with native code - `resources/scala_native` dir. The default is just to have `.` in the include path but others can be added via that special properties file. This was added because `nativelib` had some special rules and we had a lot of confusing relative paths in the includes. Hope this helps.

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
```diff
@@ -78,25 +80,27 @@ static struct sigaction *resolvePreviousSignalHandler(int sig) {
...
+                snErrorPrefix, sig);
         exit(sig);
     }
 }
 
 static void protectStackGuardPage() {
     if (mprotect(currentThreadInfo.stackGuardPage, resolvePageSize(),
                  PROT_NONE) == -1) {
-        perror("ScalaNative Fatal Error :: StackOverflowHandler guard "
+        perror("ScalaNative Fatal Error: "
```
ekrich, line 93: Should this be the shared constant too?

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
```diff
@@ -78,25 +80,27 @@ static struct sigaction *resolvePreviousSignalHandler(int sig) {
...
+               "StackOverflowHandler guard "
                "protection failed");
         exit(EXIT_FAILURE);
     }
 }
 static void unprotectStackGuardPage() {
     if (mprotect(currentThreadInfo.stackGuardPage, resolvePageSize(),
                  PROT_READ | PROT_WRITE) == -1) {
-        perror("ScalaNative Fatal Error :: StackOverflowHandler guard "
+        perror("ScalaNative Fatal Error: "
```
ekrich, line 102: Again here?

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
```diff
@@ -252,13 +256,13 @@ static void setupSignalHandlerAltstack() {
         (size_t)alignToPageStart((void *)SIG_HANDLER_STACK_SIZE);
     handlerStack.ss_sp = malloc(handlerStack.ss_size);
     if (handlerStack.ss_sp == NULL) {
-        perror("Scala Native: StackOverflowGuards failed to allocate alternate "
+        perror("ScalaNative: StackOverflowGuards failed to allocate alternate "
```
ekrich, line 259: And the next 5 too?

nativelib/src/main/scala/scala/scalanative/runtime/package.scala
```diff
@@ -67,7 +67,7 @@ package object runtime {
     val mainThread = Thread.currentThread()
     if (mainThread == null) {
       ffi.printf(
-        c"Scala Native Fatal Error: failed to initialize main java.lang.Thread\n"
+        c"ScalaNative Fatal Error: failed to initialize main java.lang.Thread\n"
```
ekrich, line 70: This one here would need a standard helper function binding if you want to remain 100% DRY.

ekrich (changes requested):
Thanks for working on this. Just a few suggestions and hopefully helpful hints.

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
JD557: Probably... but `perror` doesn't appear to support string interpolation like `fprintf`, so I skipped over those. :grimacing:

nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
JD557: Oh, indeed  I had the impression that I got that wrong (after all, shouldn0t it be `../../`?), but it worked, so I ignored it.
Will change it.

nativelib/src/main/resources/scala-native/constants.h
```diff
@@ -0,0 +1,7 @@
+#ifndef CONSTANTS_H
+#define CONSTANTS_H
+
+extern const char *const snErrorPrefix;
```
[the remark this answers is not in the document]
JD557: I think if I use `#defines` it will inline the strings in the code, so the (arguably small) storage savings will be gone.

nativelib/src/main/resources/scala-native/gc/commix/Synchronizer.c
JD557: Changed in [c191bc0](https://github.com/scala-native/scala-native/pull/4310/commits/c191bc0c1cfa1ce9a5ef0586d0421f20267185f0)

nativelib/src/main/resources/scala-native/constants.h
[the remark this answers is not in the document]
JD557: Added some macros anyway in [15cfff3](https://github.com/scala-native/scala-native/pull/4310/commits/15cfff33b7cc030553c3c3eddbe4fad62d87de87), so that they can be used with methods like `perror`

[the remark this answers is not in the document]
JD557: Added some macros anyway in [15cfff3](https://github.com/scala-native/scala-native/pull/4310/commits/15cfff33b7cc030553c3c3eddbe4fad62d87de87), so that they can be used with methods like `perror`

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
JD557: Fixed in https://github.com/scala-native/scala-native/pull/4310/commits/15cfff33b7cc030553c3c3eddbe4fad62d87de87 by using a macro instead.

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
JD557: Fixed in https://github.com/scala-native/scala-native/pull/4310/commits/15cfff33b7cc030553c3c3eddbe4fad62d87de87.

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
JD557: Fixed in https://github.com/scala-native/scala-native/pull/4310/commits/15cfff33b7cc030553c3c3eddbe4fad62d87de87.

nativelib/src/main/resources/scala-native/constants.h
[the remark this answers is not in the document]
ekrich: Yes, I thought about that as well. I think the defines only occur once via the `#ifndef;#define` in the first compilation unit otherwise you would have the constant defined twice.

nativelib/src/main/resources/scala-native/gc/immix/Synchronizer.c
```diff
@@ -1,5 +1,6 @@
 #if defined(SCALANATIVE_MULTITHREADING_ENABLED) && defined(SCALANATIVE_GC_IMMIX)
 
+#include "../constants.h"
```
ekrich, line 3: Missed this one? Did you check out the `clangd`? I use it with VSCode and the plugin and it compiles everything as you are working on it. Very helpful.

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
```diff
@@ -78,26 +80,26 @@ static struct sigaction *resolvePreviousSignalHandler(int sig) {
...
         exit(sig);
     }
 }
 
 static void protectStackGuardPage() {
     if (mprotect(currentThreadInfo.stackGuardPage, resolvePageSize(),
                  PROT_NONE) == -1) {
-        perror("ScalaNative Fatal Error :: StackOverflowHandler guard "
-               "protection failed");
+        perror(SN_FATAL_ERROR_MSG("StackOverflowHandler guard "
```
ekrich, line 93: If the goal is executable size, these are expanded by the preprocessor so each one will be laid down inline. It does the reuse with no computational overhead with no space saving. The only other alternative is to use `strcat` and handle the memory hit but since this for an error and exit? then that might be the best bet.

nativelib/src/main/resources/scala-native/gc/immix/Synchronizer.c
JD557: > Missed this one? Did you check out the clangd?

Oh, I didn't have that enabled, thanks.
Although even after enabling the LSP, it didn't point out anything, so I might have something wrong 😅 

Fixed in [fb292e9](https://github.com/scala-native/scala-native/pull/4310/commits/fb292e9dbfca87be933542b19acbe87adc1c2aec)

nativelib/src/main/scala/scala/scalanative/runtime/package.scala
JD557: Added a binding in [bde61cf](https://github.com/scala-native/scala-native/pull/4310/commits/bde61cf2c3cf432b22ef51b552577a7864949c36)

nativelib/src/main/resources/scala-native/stackOverflowGuards.c
JD557: To be honest, I'm not sure what's the best way to deal with this.

I started this PR for the space savings, although those are pretty minimal, and I do have mixed feelings about adding some manual memory management on the error handling :/

So I decided to just take the small win here of making error messages consistent... I still like the `constants.c` to have some savings where it's easy and to be able to pass the prefixes to the Scala side.

WojciechMazur (approved):

PR #4315, opened 2025-04-26
chore: Avoid ConcurrentModificationException by using a CCHM.

He-Pin:
refs: https://github.com/scala-native/scala-native/issues/4265

### DIFF
--- a/test-interface-common/src/main/scala/scala/scalanative/testinterface/common/RPCCore.scala
+++ b/test-interface-common/src/main/scala/scala/scalanative/testinterface/common/RPCCore.scala
@@ -26,7 +26,8 @@ private[testinterface] abstract class RPCCore()(implicit ec: ExecutionContext) {
   import RPCCore._
 
   /** Pending calls. */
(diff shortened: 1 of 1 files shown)


WojciechMazur (approved):
Good catch, Scala.js does also use CCHM for these 2 https://github.com/scala-js/scala-js/blob/b53a090206eb72be2c94893394297503106d0f04/test-common/src/main/scala/org/scalajs/testing/common/RPCCore.scala#L42-L43 

Probably, when making the port of test-infrastructure to Scala Native CCHM was not implemented in Scala Native, leading to usage of HashMap instead

fwbrasil:
@WojciechMazur do you have an estimate of when we'd have a release with this fix? Thanks!

WojciechMazur:
> @WojciechMazur do you have an estimate of when we'd have a release with this fix? Thanks!

I'll try to fix some remaining issues introduced or reported  since last release and possible low hangings fruits in the upcoming days. I will make release by the end of this week.

fwbrasil:
awesome! thank you  🙏

PR #4326, opened 2025-05-10
[tools] Add `Build.buildCachedAwait` in toolchain`

### LINKED ISSUE #4323: Expose a way to call `Build.buildCached()` synchronously.
To [integrate](https://github.com/dubinsky/scalajs-gradle) Scala Native linker with Gradle, I used Scala Native sbt plugin as inspiration. Thank you for leading the way!

I call the linker in a way very similar to the way Scala Native sbt plugin [does](https://github.com/scala-native/scala-native/blob/main/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L201):

```scala
  def buildCachedSync(
    config: Config,
    trace: Throwable => Unit
  ): Path =
    SharedScope { implicit sharedScope: Scope =>
      await(trace) { implicit ec: ExecutionContext =>
        Build.buildCached(config)
      }
    }
```

One difference is the use of the non-sbt-specific throwable logging; another is the removal of `.map(_.toFile())` - I call `.toFile` on the resulting `Path` instead.

To make this code fragment possible, I had to copy from `ScalaNativePluginInternal` into my code the `SharedScope` object and the `await()` method (and make it non-sbt-dependent). For obvious reasons, such copying is not ideal. It would be nice if instead this functionality was exposed as a method `Build.buildCachedSync()` - or, if that is not feasible, if its building blocks (`SharedScope` and `await()`) were moved from the `sbt-scala-native` module to `tools` or `util` and exposed there.

Would you consider exposing this functionality to simplify integration with tooling?

Thank you!

WojciechMazur:
Resolves #4323

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -24,7 +24,6 @@ import scala.scalanative.build.Platform
 import sjsonnew.BasicJsonProtocol._
 import java.nio.file.{Files, Path}
 import java.lang.Runtime
-import java.util.concurrent.Executors
 import java.util.concurrent.locks.ReentrantLock
 import sbt.librarymanagement.{
   DependencyResolution,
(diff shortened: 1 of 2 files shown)


ekrich:
@lolgab If you using this in Mill, will this help?

PR #4329, opened 2025-05-10
Make `nativeConfig` a setting key again and restore`nativeLink/artifactParh`

### LINKED ISSUE #4146: sbt setting `nativeLink / artifactPath` is no longer set
It used to be set in 0.4.x.

https://github.com/scala-native/scala-native/blob/8ce3b765e20ff1b8c7f59c286787161ad6809800/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L114-L117

This setting is useful for reading the location of the binary without having to evaluate a task.

WojciechMazur:
Fixes #4146 

AFAIR `nativeConfig` was changed to `taskKey` only to workaround the issue with duplicate definitions in the 0.4.x where we needed a way to populate `nativeMode`,`nativeGC`, etc. into `nativeConfig` and intitialize the former using the later. 

 This cyclic dependency does no longer exist so we can switch to using settingKey again. It allows us to restore `nativeLink/artifactPath`. 
 We do the same for `nativeLinkRelease{Full,Fast}` commands and simply the `nativeLinkImpl` so that in only takes varying inputs.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -243,6 +243,13 @@ object ScalaNativePluginInternal {
       }
       .tag(NativeTags.Link)
       .value,
+    nativeLink / artifactPath := build.Config.empty
+      .withBaseDir(crossTarget.value.toPath())
+      .withModuleName(moduleName.value)
+      .withCompilerConfig(nativeConfig.value)
+      .withTestConfig(testConfig)
+      .artifactPath
+      .toFile,
     nativeLink := Def
       .task {
         val sbtLogger = streams.value.log
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -51,6 +51,9 @@ object ScalaNativePluginInternal {
   val nativeWarnOldJVM =
     taskKey[Unit]("Warn if JVM 7 or older is used.")
 
+  val baseBuildConfig =
+    settingKey[build.Config]("Base configuration for nativeLink commands")
+
     lazy val scalaNativeDependencySettings: Seq[Setting[_]] = {
       val organization = "org.scala-native"
       val nativeStandardLibraries =
@@ -142,28 +145,25 @@ object ScalaNativePluginInternal {
   )
 
   private def nativeLinkImpl(
-      nativeConfig: NativeConfig,
-      sbtLogger: sbt.Logger,
-      baseDir: Path,
-      moduleName: String,
+      baseConfig: build.Config,
       mainClass: Option[String],
-      testConfig: Boolean,
-      classpath: Seq[Path],
-      sourcesClassPath: Seq[Path],
-      nativeLogger: build.Logger
+      classpath: Classpath,
+      nativeLogger: build.Logger,
+      dependencyResolution: DependencyResolution,
+      externalClassPath: Classpath
   ) = {
-
-    val config =
-      build.Config.empty
-        .withLogger(nativeLogger)
-        .withClassPath(classpath)
-        .withSourcesClassPath(sourcesClassPath)
-        .withBaseDir(baseDir)
-        .withModuleName(moduleName)
-        .withMainClass(mainClass)
-        .withTestConfig(testConfig)
-        .withCompilerConfig(nativeConfig)
-
+    val config = baseConfig
+      .withMainClass(mainClass)
+      .withLogger(nativeLogger)
+      .withClassPath(classpath.map(_.data.toPath))
+      .withSourcesClassPath(
+        resolveSourcesClassPath(
+          baseConfig.compilerConfig.sourceLevelDebuggingConfig.enabled,
+          dependencyResolution,
+          externalClassPath,
+          nativeLogger
+        )
+      )
     interceptBuildException {
       SharedScope { implicit sharedScope: Scope =>
         Build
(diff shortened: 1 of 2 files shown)


WojciechMazur:
@armanbilge @lolgab this change might cause problems in your projects. I don't have strong opinion if this one should wait til next minor or can be introduced now. I does break compatibility since now nativeConfig needs to depend only on other settings. If you both agree we can ship it now, otherwise it can wait.

lolgab:
Thank you for tagging me. No problems for SNUnit since the sbt plugin is not production ready yet, I have time to work around this change.
👍

keynmol:
I also use task dependencies in sn-vcpkg: https://github.com/indoorvivants/sn-vcpkg/blob/main/modules/sbt-vcpkg-native-plugin/src/main/scala/com/indoorvivants/vcpkg/VcpkgNativePlugin.scala#L110-L123

In this case vcpkg dependency resolution and pkg-config invocations cannot possibly be settings (I think?) so I don't know how else I could implement this pattern.

Any ideas? It'd be a real shame to lose this option.

WojciechMazur:
> In this case vcpkg dependency resolution and pkg-config invocations cannot possibly be settings (I think?) 

Seems to be doable, but it seems strange to me oto see `vcpkgInstall` defined as task. Anyway let's give this proposal a bit of time for discussions

```diff
diff --git a/build.sbt b/build.sbt
index fef2d70..882155c 100644
--- a/build.sbt
+++ b/build.sbt
@@ -45,7 +45,7 @@ val V = new {
 
   val scribe = "3.15.2"
 
-  val scalaNative = "0.4.17"
+  val scalaNative = "0.5.8-SNAPSHOT"
 
   val circe = "0.14.13"
 
diff --git a/example/project/plugins.sbt b/example/project/plugins.sbt
index 0ea315f..5c32308 100644
--- a/example/project/plugins.sbt
+++ b/example/project/plugins.sbt
@@ -5,7 +5,7 @@ val VcpkgVersion =
   sys.env.getOrElse("SBT_VCPKG_VERSION", "0.0.21")
 
 addSbtPlugin("com.indoorvivants" % "bindgen-sbt-plugin" % BindgenVersion)
-addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.5.7")
+addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.5.8-SNAPSHOT")
 addSbtPlugin("com.indoorvivants.vcpkg" % "sbt-vcpkg-native" % VcpkgVersion)
 
 resolvers += Resolver.sonatypeRepo("snapshots")
diff --git a/modules/sbt-vcpkg-native-plugin/src/main/scala/com/indoorvivants/vcpkg/VcpkgNativePlugin.scala b/modules/sbt-vcpkg-native-plugin/src/main/scala/com/indoorvivants/vcpkg/VcpkgNativePlugin.scala
index 4bd2145..075aa61 100644
--- a/modules/sbt-vcpkg-native-plugin/src/main/scala/com/indoorvivants/vcpkg/VcpkgNativePlugin.scala
+++ b/modules/sbt-vcpkg-native-plugin/src/main/scala/com/indoorvivants/vcpkg/VcpkgNativePlugin.scala
@@ -22,9 +22,9 @@ object VcpkgNativePlugin extends AutoPlugin with vcpkg.VcpkgPluginNativeImpl {
 
   object autoImport {
     val vcpkgNativeLinking =
-      taskKey[Seq[String]]("Configured linking flags for vcpkg packges")
+      settingKey[Seq[String]]("Configured linking flags for vcpkg packges")
     val vcpkgNativeCompilation =
-      taskKey[Seq[String]]("Configured compilation flags for vcpkg packages")
+      settingKey[Seq[String]]("Configured compilation flags for vcpkg packages")
     val vcpkgNativeConfig = settingKey[vcpkg.VcpkgNativeConfig](
       "Configuration object for VcpkgNativePlugin"
     )
diff --git a/modules/sbt-vcpkg-plugin/src/main/scala/com/indoorvivants/vcpkg/sbt/VcpkgPlugin.scala b/modules/sbt-vcpkg-plugin/src/main/scala/com/indoorvivants/vcpkg/sbt/VcpkgPlugin.scala
index da80478..a19a3c9 100644
--- a/modules/sbt-vcpkg-plugin/src/main/scala/com/indoorvivants/vcpkg/sbt/VcpkgPlugin.scala
+++ b/modules/sbt-vcpkg-plugin/src/main/scala/com/indoorvivants/vcpkg/sbt/VcpkgPlugin.scala
@@ -23,26 +23,26 @@ object VcpkgPlugin extends AutoPlugin with vcpkg.VcpkgPluginImpl {
     val vcpkgDependencies =
       settingKey[VcpkgDependencies]("List of vcpkg dependencies")
 
-    val vcpkgInstall = taskKey[Map[vcpkg.Dependency, vcpkg.FilesInfo]](
+    val vcpkgInstall = settingKey[Map[vcpkg.Dependency, vcpkg.FilesInfo]](
       "Invoke Vcpkg and attempt to install packages"
     )
 
-    val vcpkgRootInit = taskKey[vcpkg.VcpkgRootInit](
+    val vcpkgRootInit = settingKey[vcpkg.VcpkgRootInit](
       "Specification of the location where vcpkg will be bootstrapped and where the packages will be installed"
     )
-    val vcpkgRoot = taskKey[File]("Root of vcpkg installation")
+    val vcpkgRoot = settingKey[File]("Root of vcpkg installation")
     val vcpkgAllowBootstrap =
-      taskKey[Boolean]("Allow bootstrapping root location")
+      settingKey[Boolean]("Allow bootstrapping root location")
 
-    val vcpkgInstallDir = taskKey[File]("Where to put installed packages")
+    val vcpkgInstallDir = settingKey[File]("Where to put installed packages")
 
-    val vcpkgBinary = taskKey[File]("Path to vcpkg binary")
+    val vcpkgBinary = settingKey[File]("Path to vcpkg binary")
 
-    val vcpkgManager = taskKey[Vcpkg](
+    val vcpkgManager = settingKey[Vcpkg](
       "Value of `Vcpkg` type - a thin wrapper around the `vcpkg` CLI, already bootstrapped"
     )
 
-    val vcpkgConfigurator = taskKey[vcpkg.VcpkgConfigurator](
+    val vcpkgConfigurator = settingKey[vcpkg.VcpkgConfigurator](
       "`VcpkgConfigurator` instance, which provides access to PkgConfig (a thin wrapper over pre-configured pkg-config) and some methods to approximate compilation/linking arguments for installed packages"
     )
 
@@ -117,7 +117,7 @@ object VcpkgPlugin extends AutoPlugin with vcpkg.VcpkgPluginImpl {
     vcpkgInstall := {
       val manager = vcpkgManager.value
       val logger = sbtLogger(sLog.value)
-      val cacheFactory = streams.value.cacheStoreFactory
+      val cacheFactory = sbt.util.CacheStoreFactory(target.value / "vcpkg")
       val dependencies = vcpkgDependencies.value
 
       val cachedInstall = Cache.cached(

```

ekrich:
I think you remember a bit better than I but I remember that we needed to construct the `Build/NativeConfig` lazily and I believe it was because of not being able to change the settings later in a build or command line. As a result `nativeLink/artifactPath` was removed because the only way to get the value was to run all the initialization which meant there was a ton of work going on just to get the `Build` internal calculated value of artifact path. Your explanation of recursive issue is better.

My main objective at the time was to have the `Build` do all the work and logic and have the plugin just use the build to allow for easier integration for Mill and perhaps a Native build app. I was also looking at all the work the plugin was doing just to deal with all the projects in Scala Native (it was slow). The other thing was allowing the user to specify the executable name and maybe some other features as well. The `artifactPath` was not compatible but if it is now then great. Another thing was querying the build in `sbt` using `show` or `inspect` was not able to show the true picture.

Hope this helps.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -51,53 +51,57 @@ object ScalaNativePluginInternal {
   val nativeWarnOldJVM =
     taskKey[Unit]("Warn if JVM 7 or older is used.")
 
-    lazy val scalaNativeDependencySettings: Seq[Setting[_]] = {
-      val organization = "org.scala-native"
-      val nativeStandardLibraries =
-        Seq("nativelib", "clib", "posixlib", "windowslib", "javalib", "auxlib")
+  val baseBuildConfig = settingKey[build.Config](
+    "Base configuration for nativeLink commands calcullated base on setting keys. Internal, not visible to users"
```
ekrich, line 55:
```suggestion
    "Base configuration for nativeLink commands calculated base on setting keys. Internal, not visible to users"
```

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -181,134 +182,116 @@ object ScalaNativePluginInternal {
...
+
+    Seq(
+      scalacOptions ++= {
+        if (isGeneratingForIDE) None
+        else
+          Some(
+            s"-P:scalanative:positionRelativizationPaths:${sourceDirectories.value.map(_.getAbsolutePath()).mkString(";")}"
+          )
+      },
+      // Base build config is populate using settings, it should be enought to calculate artifactPath
```
ekrich, line 202:
```suggestion
      // Base build config is populate using settings, it should be enough to calculate artifactPath
```

ekrich (changes requested):
Looks good to me overall - have not tried it.

The title also needs to say `artifactPath`.

PR #4376, opened 2025-06-19
[ci] Test macos extended tests using LLVM 19 (bug in LLVM 15)

WojciechMazur:
[test-extended]


He-Pin:
Kotlin 2.2 will bump to llvm 19 by default.

PR #4402, opened 2025-07-18
Enable clangd configurations for developers

ekrich:
Before, we only had some documentation and `.gitignore` entries for the `compile_flags.txt` files. This means that we cannot share a setup that works or improve it as needed and developers that are not knowledgeable about this capability can not take advantage of it.

Initially, this was setup when developing some C functions for time which needed WIndow's support. `clangd` recognizes / sets the basic platform defines so other platforms code are grayed out. It helps you code in C showing hints, errors etc.

Some have suggested we could generate these files which is possible but maybe better for a future improvement. One disadvantage is the code would have these file bundled in jar but that seems to not be a problem.


nativelib/src/main/resources/scala-native/compile_flags.txt
```diff
@@ -0,0 +1,24 @@
+# nativelib
+# macOS arm
+-I
+/opt/homebrew/include
+# gc
+-I
+gc/
```
WojciechMazur: These can be concatenated to `-Igc/` to make it more readable and easy to comment out if needed.

docs/contrib/ides.md
```diff
@@ -52,37 +52,18 @@ development using C and C++.
...
--DDEBUG_ASSERT
-# end GC
-```
+`clangd` works well but only has certain defines for your platform for
+example. Thus `clangd` can work out of the box for simple setups which is
+probably fine for most projects that just need `C` glue code for Windows, Linux or macOS.
+
+If you have native code included in your own project, you can copy one of these
+files that are located in the `src/main/resources/scala-native` directory of the
+projects listed above. Then you can modify the configuration as needed.
```
WojciechMazur: The compile_flags.txt should not be placed in the `resources` - our internal development setups should not leak to jars. 
If we want to have `compile_flags.txt` let's define one in the top-level - it's good enough for development, and makes them easier to find, here's example of my `compile_flags.txt` placed in root directory: 

```
-DSCALANATIVE_GC_IMMIX
-D__SCALANATIVE_DELIMCC
-Inativelib/src/main/resources/scala-native
-Inativelib/src/main/resources/scala-native/gc/
-Inativelib/src/main/resources/scala-native/gc/immix_commix
-DSCALANATIVE_MULTITHREADING_ENABLED
```

nativelib/src/main/resources/scala-native/compile_flags.txt
ekrich: Maybe in the past it was different (or for `.`) but that is good to know.

docs/contrib/ides.md
ekrich: Sure, I will put together a combined file.

ekrich: I looked at all the projects and this is working pretty well. Maybe others might find improvements once it is being used.

ekrich: I will have to fix the docs before merge.

WojciechMazur: LGTM, let's just update the docs. I think only this part is now invalid.

ekrich: This should be ready to go now.

ekrich: Done.

PR #4490, opened 2025-09-30
Add support for Scala 2.13.17

WojciechMazur:
Adds Scala 2.13.17 support and makes it a default version used in the project 
Also backports changes related to standard library tests from #4485: 
- always emit constructor calls as static method calls (fixes UsingTest)
- adjust `UsingTest` to the pre Scala 2.13.17 behaviour and the new one (split into 2 tests with version checked at runtime)

### DIFF
--- a/unit-tests/shared/src/test/scala-2.13+/scala/util/UsingTestOld.scala
+++ b/unit-tests/shared/src/test/scala-2.13+/scala/util/UsingTestOld.scala
@@ -1,0 +1,1023 @@
+// Test ported from scala-collection-compat repo, to test regression in Scala 3
+/*
+ * Scala (https://www.scala-lang.org)
+ *
+ * Copyright EPFL and Lightbend, Inc.
+ *
+ * Licensed under Apache License 2.0
+ * (http://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.util
+
+import org.junit.{Test, Assume, BeforeClass}
+import org.junit.Assert._
+import scala.reflect.ClassTag
+import scala.runtime.NonLocalReturnControl
+
+@deprecated(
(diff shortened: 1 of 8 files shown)


unit-tests/shared/src/test/scala-2.13+/scala/util/UsingTestOld.scala
```diff
@@ -865,7 +865,29 @@ class UsingTest {
   }
 }
 
-object UsingTest {
+object UsingTestOld {
+  @BeforeClass def checkRuntime(): Unit = {
+    import org.scalanative.testsuite.utils.Platform
+    // Implementation has changed in Scala 2.13.17, backported to 3.8.0
```
ekrich, line 871: Should this say "backported from"??

unit-tests/shared/src/test/scala-2.13+/scala/util/UsingTestOld.scala
```diff
@@ -865,7 +865,29 @@ class UsingTest {
...
+    // scalalib_3 is built against scalalib_2.13 using the last Scala version
+    // instead of the one defined in natural scala3_library_3 dependenies
+    def hasCompliantScalaVersion =
+      Platform.scalaVersion
+        .split('.')
+        .take(3)
+        .map(_.takeWhile(_.isDigit)) // becouse of versions like "3.4.0-RC1"
```
ekrich, line 878:
```suggestion
.map(_.takeWhile(_.isDigit)) // because of versions like "3.4.0-RC1"
```

ekrich (commented):
Just a couple of nits. Is it getting close to the time we can drop 2.12?

unit-tests/shared/src/test/scala-2.13+/scala/util/UsingTestOld.scala
He-Pin: Oops

He-Pin:
For 2.12, maybe drop support for it once scala js did that？

PR #4494, opened 2025-10-02
nir.Versions: allow pre-release suffix to use `+`

kitbellew:
`+` is standard separator (in fact, native uses 2.13.16+0.5.8 itself). Helps with #4488.


kitbellew:

WojciechMazur (approved):

PR #4495, opened 2025-10-02
Plugin: add nativeOrgName, clean up nativeVersion

kitbellew:
Helps with #4488.

### DIFF
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativeCrossVersion.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativeCrossVersion.scala
@@ -25,11 +25,14 @@ object ScalaNativeCrossVersion {
     }
   }
 
+  private def scalaNativePrefixFromBinary(binaryVersion: String): String =
+    "native" + binaryVersion
+
+  private[scalanative] def scalaNativePrefix: String =
+    scalaNativePrefixFromBinary(currentBinaryVersion)
+
   def scalaNativeMapped(cross: CrossVersion): CrossVersion =
-    crossVersionAddPlatformPart(
-      cross,
-      "native" + nir.Versions.currentBinaryVersion
-    )
+    crossVersionAddPlatformPart(cross, scalaNativePrefix)
 
   val binary: CrossVersion = scalaNativeMapped(CrossVersion.binary)
 
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
@@ -1,8 +1,6 @@
 package scala.scalanative
 package sbtplugin
 
-import ScalaNativePluginInternal._
-
 import sbt._
 
 object ScalaNativePlugin extends AutoPlugin {
@@ -12,6 +10,7 @@ object ScalaNativePlugin extends AutoPlugin {
 
     val ScalaNativeCrossVersion = sbtplugin.ScalaNativeCrossVersion
 
+    private[sbtplugin] val nativeOrgName = "org.scala-native"
     val nativeVersion = nir.Versions.current
 
     def scalalibVersion(scalaVersion: String, nativeVersion: String): String =
--- a/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
+++ b/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
@@ -52,50 +52,42 @@ object ScalaNativePluginInternal {
     taskKey[Unit]("Warn if JVM 7 or older is used.")
 
     lazy val scalaNativeDependencySettings: Seq[Setting[_]] = {
-      val organization = "org.scala-native"
       val nativeStandardLibraries =
         Seq("nativelib", "clib", "posixlib", "windowslib", "javalib", "auxlib")
 
       Seq(
-        libraryDependencies ++= Seq(
-          organization %%% "test-interface" % nativeVersion % Test
-        ),
-        libraryDependencies += CrossVersion
-          .partialVersion(scalaVersion.value)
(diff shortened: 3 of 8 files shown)


kitbellew:

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
```diff
@@ -12,6 +10,7 @@ object ScalaNativePlugin extends AutoPlugin {
 
     val ScalaNativeCrossVersion = sbtplugin.ScalaNativeCrossVersion
 
+    val nativeOrgName = "org.scala-native"
```
WojciechMazur: There's no need to make it public. All vals defined here would be exposed to sbt plugin users (and typically autoimported)

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -52,50 +52,36 @@ object ScalaNativePluginInternal {
...
-              organization %%% "scala3lib" % scalalibVersion(
-                scalaVersion.value,
-                nativeVersion
-              )
-          },
-        libraryDependencies ++= nativeStandardLibraries.map(
-          organization %%% _ % nativeVersion
-        ),
+        libraryDependencies ++= {
+          val org = nativeOrgName
```
WojciechMazur, line 60: Let's decouple a current dependency between our build definition and the sbt plugin we publish. 
Let's simply inline `nativeOrgName` in exactly 2 places - here and in the ThisBuild / version definition. All other usages refer to these two.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
```diff
@@ -52,50 +52,36 @@ object ScalaNativePluginInternal {
...
           nativeStandardLibraries.map { lib =>
-            val scalaBinVersion =
-              if (scalaVersion.value.startsWith("3.")) "2.13"
-              else "3"
-            ExclusionRule()
-              .withOrganization(organization)
-              .withName(
-                s"${lib}_native${ScalaNativeCrossVersion.currentBinaryVersion}_${scalaBinVersion}"
-              )
+            exclRule.withName(s"${lib}_${suffix}")
```
WojciechMazur: If possible let's split the suffix into Scala Native and Scala binary parts, when looking at interpolated string it's clear that we're looking on platform specific library

WojciechMazur: Also, when looking at it it seems that we unnecessary exclude Scala 2.13 dependencies in Scala 2.12 - it would never be needed. If you want you an fix it as well in here. So potentially match on CrossVersion.partialVersion would allow to cleanly split these things up.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativeCrossVersion.scala
```diff
@@ -25,11 +25,14 @@ object ScalaNativeCrossVersion {
     }
   }
 
+  private def scalaNativePartFromBinary(binaryVersion: String): String =
```
WojciechMazur:
```suggestion
  private def scalaNativePrefixForBinary(binaryVersion: String): String =
    "native" + binaryVersion

  private[scalanative] val currentScalaNativeBinaryPrefix: String =
    scalaNativePartFromBinary(currentBinaryVersion)

def scalaNativeMapped(cross: CrossVersion): CrossVersion =
    crossVersionAddPlatformPart(cross, scalaNativePart)
```

kitbellew: done.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala
kitbellew: made package-private.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
kitbellew: done.

sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala
kitbellew: done (split suffix).

don't understand the second comment. we exclude 2.13 in a 3.x build, and exclude 3.x otherwise. should 3.x be excluded for 2.13 only, and for 2.12 excludeDependencies should be empty?

WojciechMazur: Yeah, so the exclusions are here only because of cross compilation between 2.13 and 3. 
scalalib_3 depends on scalalib_2.13 (same as scala3-library_3 depends on scala-library for 2.13 on the JVM (it would change in 3.8 however) 
It was needed to workaround `conflicting binary version suffixes` errors emitted by sbt when it detects that library depends (directly or transitively) on 2 different binary versions (typically _2.13 and _3). 
That's why for Scala 3 we exclude 2.13 artifacts introduced by scalalib_2.13 (javalib, nativelib, etc) 
The other way around was probably set due to some bug we found, where some Scala 2.13 project introduced transitive dependency to Scala 3 artefacts (which is possible, see CrossVersion.for2_13use3 and CrossVersion.for3use2_13) which brings the unexpected artefacts for `_3` 

It's not possible to cross compile Scala 2.12 with other versions so conflicting dependencies can never be brought in.

kitbellew: done (exclusion rules).

WojciechMazur (approved):

PR #4496, opened 2025-10-02
SBT: enable implicits to add settings and plugins

kitbellew:
Add MyScalaNativePlugin into .withNativeCompilerPlugin. Define .withBuildInfo and .withCommonTools as well. Helps with #4488.


kitbellew:

PR #4497, opened 2025-10-02
MyScalaNativePlugin: move to scala.scalanative pkg

kitbellew:
That way, we'll be able to use package-private code in it. Helps with #4488.

### DIFF
--- a/project/MultiScalaProject.scala
+++ b/project/MultiScalaProject.scala
@@ -4,7 +4,7 @@ import sbt._
 import Keys._
 import Def.SettingsDefinition
 import scala.language.implicitConversions
-import MyScalaNativePlugin.{ideScalaVersion, enableExperimentalCompiler}
+import scala.scalanative.build.MyScalaNativePlugin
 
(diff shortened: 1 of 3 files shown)


kitbellew:

kitbellew:
> I don't really thing this change should be introduced. There are reasons for clear separation between build definitions and actual code. Distinct namespaces is one of the first good barriers for that. If something needs to cross it then we should as ourself is it really necessary. Package `scala.scalanative.build` is used for the toolchain build (linker, config, api)

If we can't use `nativeBuildVersion` (see https://github.com/scala-native/scala-native/pull/4488#discussion_r2396289722), then this change is not necessary.

But in that case I don't know how to solve #4487.

kitbellew:
> If we can't use `nativeBuildVersion` (see [#4488 (comment)](https://github.com/scala-native/scala-native/pull/4488#discussion_r2396289722)), then this change is not necessary.

abandoning.

> But in that case I don't know how to solve #4487.

apparently, i do.

PR #4498, opened 2025-10-02
StackTraceElementTest: more assertions

kitbellew:


WojciechMazur:
The current failure araises during JVM execution of tests - before we execute unit tests natively, for tests that are shared between JVM and Scala Native to ensure that the test themself are compatible with JVM.  
See the `testsJVMXXX` vs `testsXXX` projects

tools/src/main/scala/scala/scalanative/codegen/llvm/CodeGen.scala
```diff
@@ -2,23 +2,18 @@ package scala.scalanative
...
 import scala.scalanative.io.VirtualDirectory
 import scala.scalanative.build
 import scala.scalanative.build.IO
 import scala.scalanative.linker.ReachabilityAnalysis
 import scala.scalanative.util.{Scope, partitionBy, procs}
-import java.nio.file.StandardCopyOption
-
-import scala.scalanative.codegen.{Metadata => CodeGenMetadata}
-import scala.concurrent._
+import scala.scalanative.codegen.Metadata as CodeGenMetadata
```
ekrich: Nice :-) and the next line too.

kitbellew: why is scalafmt configured with scala3 dialect here? most CI builds are 2.12 or 2.13, and this accidental rewrite ended up breaking my tests :)

ekrich: Gosh, I guess when it changed we switched to scala3 and use scala213 for Scala 2 code only. What did it break? - I haven't had that problem.

kitbellew: you haven't had this problem because you didn't try to rewrite to scala3 syntax. and intellij tries it :) also, scala2 code now assumes significant indentation and refuses to indent stuff correctly.

ekrich: You are welcome to change it. You are the expert so I'm sure nobody will object.

PR #4502, opened 2025-10-03
[docs] Add entires latest stable/snapshot version 

WojciechMazur:
During generation of docs, we fetch for `last-published` tag and recreate the snapshot version info based on it
We add the table with stable and latest snapshot version


docs/conf.py
```diff
@@ -20,14 +25,57 @@
...
-from datetime import date
-
 def generateScalaNativeCurrentYear():
     ## encode/decode ugliness is to handle both python 2.7 and 3.n
     return str(date.today().year).encode("utf-8").decode("utf-8")
 
+def lastSnapshotVersion():
+    # Ensures we're inside a repo (raises if not)
+    _ensure_repo()
+    # Recosntructs version based on the last-published tag
```
kitbellew, line 35: typo

docs/conf.py
```diff
@@ -20,14 +25,57 @@
...
 
-from datetime import date
-
 def generateScalaNativeCurrentYear():
     ## encode/decode ugliness is to handle both python 2.7 and 3.n
     return str(date.today().year).encode("utf-8").decode("utf-8")
 
+def lastSnapshotVersion():
+    # Ensures we're inside a repo (raises if not)
+    _ensure_repo()
```
kitbellew, line 34: tag_commit_sha also does rev-parse, why call git and throw away result?

docs/conf.py
```diff
@@ -20,14 +25,57 @@
...
+    # Recosntructs version based on the last-published tag
+    tagName="last-published"
```
kitbellew, line 36: do you also need to fetch the tag first, if you do shallow checkout? and if not shallow, then you don't need to hardcode previous release below, you can "git describe" it.

PR #4506, opened 2025-10-04
git-blame-ignore-revs: use correct commit refs

kitbellew:
If a change in this file is part of a pull request which also contains a reformatting commit, that pull request may only be MERGED, never REBASED or SQUASHED.

Otherwise, the commit reference added to this file will become invalid.

This change is to fix all references which had been correct when they were submitted for review but rendered incorrect upon non-MERGE.


kitbellew:
fyi @ekrich @WojciechMazur

WojciechMazur (approved):

PR #4507, opened 2025-10-04
scalafmt: adjust dialects (scala212 as default)

kitbellew:
Also, remove excluded paths and re-order `.scalafmt.conf`.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -14,6 +14,9 @@ docstrings {
 }
 
 fileOverride {
+  # default is 2.12, see under runner
+  "glob:**/scripts/*" = scala3
+  "glob:**/src/**/require-scala3*/**" = scala3
   "lang:scala-next" = scala3future
 }
 
@@ -51,7 +54,7 @@ rewrite {
 
 runner {
   # Default runner.dialect is deprecated now, needs to be explicitly set
-  dialect = scala3
+  dialect = scala212
   # Added for CI error via --test option
   fatalWarnings = true
   # default value is 10,000
--- a/javalib/src/main/scala/java/lang/ProcessBuilder.scala
+++ b/javalib/src/main/scala/java/lang/ProcessBuilder.scala
@@ -151,9 +151,7 @@ object ProcessBuilder {
       override def file(): File = redirectFile
 
       override def toString =
-        s"Redirect.$tpe${
-            if (redirectFile != null) s": ${redirectFile}" else ""
-          }"
+        s"Redirect.$tpe${if (redirectFile != null) s": ${redirectFile}" else ""}"
     }
 
     val INHERIT: Redirect = new RedirectImpl(Type.INHERIT, null)
--- a/nativelib/src/main/scala/scala/scalanative/runtime/Intrinsics.scala
+++ b/nativelib/src/main/scala/scala/scalanative/runtime/Intrinsics.scala
@@ -125,7 +125,7 @@ object Intrinsics {
   def elemRawPtr(rawptr: RawPtr, offset: RawSize): RawPtr =
     intrinsic
 
-    /** Intrinsified computation of derived raw pointer. */
+  /** Intrinsified computation of derived raw pointer. */
   def elemRawPtr(rawptr: RawPtr, offset: Int): RawPtr =
     intrinsic
 
--- a/nativelib/src/main/scala/scala/scalanative/unsafe/Zone.scala
+++ b/nativelib/src/main/scala/scala/scalanative/unsafe/Zone.scala
@@ -19,7 +19,7 @@ trait Zone {
   def alloc(size: Int): Ptr[Byte] =
     alloc(unsignedOf(castIntToRawSizeUnsigned(size)))
 
-    /** Allocates memory of given size. */
+  /** Allocates memory of given size. */
   def alloc(size: UInt): Ptr[Byte] =
     alloc(size.toUSize)
 
--- a/project/MyScalaNativePlugin.scala
+++ b/project/MyScalaNativePlugin.scala
@@ -117,25 +117,25 @@ object MyScalaNativePlugin extends AutoPlugin {
         None
       }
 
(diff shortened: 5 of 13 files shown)


kitbellew:
fyi @ekrich @WojciechMazur

kitbellew:
also, i haven't added an entry to git-blame-ignore-revs, for two reasons:
- it will conflict with #4506 until that PR is merged
- if the repo doesn't allow merge commits, the appropriate entry should be added after this PR is merged (and actual commit refs are known)

.scalafmt.conf
```diff
@@ -4,40 +4,55 @@
...
+  tokens = {
+    "⇒": "=>"
+    "→": "->"
+    "←": "<-"
   }
 }
+
+runner {
+  # Default runner.dialect is deprecated now, needs to be explicitly set
+  dialect = scala212
```
WojciechMazur, line 57: Scala 2.12 would likely be dropped in next major release. If we really need to change dialect it should be scala213
As soon as 2.12 is dropped we might switch to using `-Xsource:3-cross` in the Scala 2.13 codebase

.scalafmt.conf
```diff
@@ -4,40 +4,55 @@
...
+  style = AsteriskSpace
 }
+
 fileOverride {
-  "glob:**/src/**/scala-2*/**" {
-     runner.dialect = scala213
+  # default is 2.12, see under runner
+  ".sbt" = scala3
+  "glob:**/{project,scripts}/*" = scala3
+  "glob:**/src/**/require-scala3*/**" = scala3
```
WojciechMazur, line 19: This one exists only in tests, there should an additonal override for `scala-3` subdirectories containing only Scala 3 specific code

.scalafmt.conf
```diff
@@ -4,40 +4,55 @@
...
+
+docstrings {
+  style = AsteriskSpace
 }
+
 fileOverride {
-  "glob:**/src/**/scala-2*/**" {
-     runner.dialect = scala213
+  # default is 2.12, see under runner
+  ".sbt" = scala3
```
WojciechMazur: This one does not feel correct, it should probably be `sbt1`, but probably it should be also picked automatically by file extension (needs to be checked)

.scalafmt.conf
```diff
@@ -4,40 +4,55 @@
...
+  "glob:**/{project,scripts}/*" = scala3
```
WojciechMazur: It's correct for `scripts`, but not for `project` which is compiled using Scala 2.12

.scalafmt.conf
```diff
@@ -4,40 +4,55 @@
...
+  "glob:**/src/**/require-scala3*/**" = scala3
+  "lang:scala-next" = scala3future
```
WojciechMazur, line 20: I haven't knew about `lang` selector, how does it work? Can you point me to some docs?

scalalib/overrides-2.12/scala/collection/immutable/VM.scala
```diff
@@ -5,6 +5,7 @@ import scala.scalanative.libc.stdatomic.memory_order._
 
 // Backport from scala.runtime moved into s.c.immutable and made package
 // private to avoid the need for MiMa allowlisting.
-/* private[immutable] */ object VM {
+/* private[immutable] */
```
WojciechMazur: Please add an exclusion rule for the whole`scalalib` directory - nothing in here should be formatted to reduce the diff when comparing our overrides with the original Scala standard library sources 

We'd also need to revert all changes made in this directory

WojciechMazur (changes requested):

.scalafmt.conf
kitbellew: when it's dropped, _then_ we change the dialect. otherwise, might have a problem.

if you use `-Xsource:3`, then the dialect is `Scala212Source3` or `Scala213Source3`.

.scalafmt.conf
kitbellew: if `.sbt` and `project/*` are both compiled using 2.12, then the default is sufficient.

however, `project/ScalaNativeBuildInfo.scala` uses new wildcard import syntax and seems to work... perhaps `Scala212Source3` is the more correct default dialect?

.scalafmt.conf
kitbellew: see below

.scalafmt.conf
kitbellew: `scala-3` and any `scala-2.13` are covered by `project.layout = StandardConvention`.

.scalafmt.conf
kitbellew: https://scalameta.org/scalafmt/docs/configuration.html#fileoverride (at the end, talks about shortcuts)

.scalafmt.conf
```diff
@@ -4,40 +4,57 @@
...
+newlines {
+  # This creates less of a diff but is not default
+  # but is more aligned with Scala.js syntax.
+  beforeCurlyLambdaParams = multilineWithCaseOnly
+}
+
+project {
+  git = true
+  excludePaths = [
+    "glob:**/scalalib/**"
```
kitbellew, line 40: done, put it back, reverted formatting changes there.

ekrich (commented):
Can't believe so many changes but before the format seemed not correct. Overall looks good.

.scalafmt.conf
He-Pin: I think we should merge this and once 2.12 is droppend ,just update the related setting here.

He-Pin: @WojciechMazur The next major release of Scala Native 1.0.0 or 0.6.x?

WojciechMazur: The plan is to make 1.0 so we can finally have a proper semantic versioning

kitbellew:
the only changes remaining are from the code which under scala3 was interpreted as having significant indentation.

PR #4517, opened 2025-10-10
[scalalib] Fix scalalib patching for 3.3.1

WojciechMazur:
Fixes scala3lib patches for 3.3.1 and ensure these are now tests in standard (non-scheduled) CI by optimizing their execution

### DIFF
--- a/scripts/scalalib-patch-tool.scala
+++ b/scripts/scalalib-patch-tool.scala
@@ -38,7 +38,9 @@ def main(
 
   implicit val wd: os.Path = pwd
 
-  val sourcesDir = pwd / "scalalib" / "target" / "scalaSources" / scalaVersion
+  val sourcesDir = pwd / scalalibProject(scalaVersion) /
+    "target" / "scalaSources" / scalaVersion
+
   val overridesDirPath: os.Path =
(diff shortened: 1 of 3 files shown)


scripts/scalalib-patch-tool.scala
```diff
@@ -199,6 +201,14 @@ def fileToLines(path: os.Path) = {
   list
 }
 
+def scalalibProject(scalaVersion: String): String = scalaVersion
+  .split('.')
+  .take(2)
+  .map(_.toInt) match {
+  case Array(3, minor) if minor <= 7 => "scala3lib"
```
ekrich, line 208: Is this suppose to `minor >= 7`?

WojciechMazur: It's correct - for versions 3.0 ... 3.7 the sources are fetched by `scala3lib/fetchSources` which downloads scala3-library_3 sources. Since 3.8 and for Scala 2 all sources are in scala-library provided by `scalalib/fetchSources`

ekrich: Ok, I see I knew there was a change but wasn't sure which way.

PR #4524, opened 2025-10-17
Reproduce stacktrace bug

durban:


He-Pin:
I encounter this too in sjsonnet, it gives something like `unwind.getContextException`

PR #4534, opened 2025-10-19
scalafmt: rewrite with SortModifiers

kitbellew:
Use the style-guide sequence of ordering modifiers.

### DIFF
--- a/javalib/src/main/scala/java/lang/Thread.scala
+++ b/javalib/src/main/scala/java/lang/Thread.scala
@@ -634,7 +634,7 @@ object Thread {
 
   // Counter used to generate thread's ID, 0 resevered for main
   sealed abstract class Numbering {
-    final protected var cursor = 1L
+    protected final var cursor = 1L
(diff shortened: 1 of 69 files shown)


javalib/src/main/scala/java/lang/Thread.scala
```diff
@@ -634,7 +634,7 @@ object Thread {
 
   // Counter used to generate thread's ID, 0 resevered for main
   sealed abstract class Numbering {
-    final protected var cursor = 1L
+    protected final var cursor = 1L
```
He-Pin, line 637: Is there a style which `final` is the leading keyword?

kitbellew: you don't have to use a preset style, can write them all out explicitly:

https://scalameta.org/scalafmt/docs/configuration.html#sortmodifiers

here, the ordering from the [style guide](https://docs.scala-lang.org/style/declarations.html#modifiers) is used. as you can see, it puts `final` as... well, final.

PR #4541, opened 2025-10-22
UnixProcessGen1: DRY use UnixProcessGen2.forkChild

kitbellew:
The implementations of this legacy logic are identical, there's no need to maintain them separately, especially as we're trying to rectify a few issues applicable to that logic.

Helps with #4544.


kitbellew:
@LeeTibbert @WojciechMazur fyi.

WojciechMazur (approved):

PR #4545, opened 2025-10-22
UnixProcessGen2: re-use binaryPaths in spawnChild

kitbellew:
Remove the UnixPathIterator and associated logic which is duplicate.


kitbellew:
@LeeTibbert @WojciechMazur fyi

WojciechMazur (approved):

PR #4548, opened 2025-10-23
UnixProcess: don't keep pipe fd pairs as fields

kitbellew:
They are not allocated in managed memory so their values might not be available after the constructor returns. Therefore, let's extract the file descriptor identifiers while they are available.

Helps with #4544.

### DIFF
--- a/javalib/src/main/scala/java/lang/process/UnixProcess.scala
+++ b/javalib/src/main/scala/java/lang/process/UnixProcess.scala
@@ -20,17 +20,26 @@ private[process] abstract class UnixProcessHandle extends GenericProcessHandle {
 
 private[process] object UnixProcess {
   def apply(
+      stdin: FileDescriptor,
+      stdout: FileDescriptor,
+      stderr: FileDescriptor
+  )(handle: UnixProcessHandle): GenericProcess = new GenericProcess(handle) {
+    override protected def fdIn: FileDescriptor = stdin
+    override protected def fdOut: FileDescriptor = stdout
+    override protected def fdErr: FileDescriptor = stderr
+  }
+
+  def apply(
       handle: UnixProcessHandle,
       infds: Ptr[CInt],
       outfds: Ptr[CInt],
       errfds: Ptr[CInt]
-  ): GenericProcess = new GenericProcess(handle) {
-    override protected def fdIn = new FileDescriptor(!(infds + 1))
-    override protected def fdOut = new FileDescriptor(!outfds, readOnly = true)
-    override protected def fdErr =
-      if (null == errfds) new FileDescriptor()
-      else new FileDescriptor(!errfds, readOnly = true)
-  }
+  ): GenericProcess = apply(
+    new FileDescriptor(!(infds + 1)),
+    new FileDescriptor(!outfds, readOnly = true),
+    if (null == errfds) new FileDescriptor()
+    else new FileDescriptor(!errfds, readOnly = true)
+  )(handle)
 
   def apply(pb: ProcessBuilder): GenericProcess = {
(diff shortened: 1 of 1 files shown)


kitbellew:
@LeeTibbert @WojciechMazur don't know why it worked before... please review this after #4543 is merged, it adds one commit on top of that.

javalib/src/main/scala/java/lang/process/UnixProcess.scala
```diff
@@ -19,18 +19,27 @@ private[process] abstract class UnixProcessHandle extends GenericProcessHandle {
...
       infds: Ptr[CInt],
       outfds: Ptr[CInt],
       errfds: Ptr[CInt]
-  ): GenericProcess = new GenericProcess(handle) {
-    override protected def fdIn = new FileDescriptor(!(infds + 1))
-    override protected def fdOut = new FileDescriptor(!outfds, readOnly = true)
-    override protected def fdErr =
-      if (null == errfds) new FileDescriptor()
-      else new FileDescriptor(!errfds, readOnly = true)
-  }
```
WojciechMazur, line 33: The created anonymous class would not store Ptr, but already initialised file descriptor. We check it when inspecting generated Bytecode (it would be the same for inspecting NIR). Scala compiler is smart enough to not keep infds as clousures. 

Based on this simplified example (compiled with Scala 3.7): 
```scala
type CInt = Int
class FileDescriptor(val value: Int, val readOnly: Boolean = false)
class UnixProcessHandle(val value: Int)
abstract class GenericProcess(handle: UnixProcessHandle){
  protected def fdIn: FileDescriptor
  protected def fdOut: FileDescriptor
  protected def fdErr: FileDescriptor
}

object UnixProcess:
  def apply(
      handle: UnixProcessHandle,
      infds: CInt,
      outfds: CInt,
      errfds: CInt
  ): GenericProcess =  new GenericProcess(handle) {
    override protected def fdIn = new FileDescriptor(infds + 1)
    override protected def fdOut = new FileDescriptor(outfds, readOnly = true)
    override protected def fdErr = new FileDescriptor(errfds, readOnly = true)
  }
```
We'll receive

```java
public final class UnixProcess$$anon$1 extends GenericProcess {
   private final FileDescriptor stdin$2;
   private final FileDescriptor stdout$2;
   private final FileDescriptor stderr$2;

   public UnixProcess$$anon$1(final UnixProcessHandle handle$1, final FileDescriptor stdin$1, final FileDescriptor stdout$1, final FileDescriptor stderr$1) {
      super(handle$1);
      this.stdin$2 = stdin$1;
      this.stdout$2 = stdout$1;
      this.stderr$2 = stderr$1;
   }
```
and 
```java
public final class UnixProcess$ implements Serializable {
   public GenericProcess apply(final FileDescriptor stdin, final FileDescriptor stdout, final FileDescriptor stderr, final UnixProcessHandle handle) {
      return new UnixProcess$$anon$1(handle, stdin, stdout, stderr);
   }

   public GenericProcess apply(final UnixProcessHandle handle, final int infds, final int outfds, final int errfds) {
      return this.apply(new FileDescriptor(infds + 1, .MODULE$.$lessinit$greater$default$2()), new FileDescriptor(outfds, true), new FileDescriptor(errfds, true), handle);
   }
```

WojciechMazur: However, this refactor makes it a bit more explicit, and does not require reader to know what's happening behind the scenes, so it's still useful. Would also prevent from accessing raw file descriptor (pointer) in unexpected ways

kitbellew:
@WojciechMazur ready to re-run the single flaky windows test.

PR #4550, opened 2025-10-23
UnixProcessGen2: don't create unused pipes, close all unused descriptors

kitbellew:
In order to reduce the chance of deadlocks while reading from stdout and stderr, and possibly receiving a proper EOF, AI has suggested to do a thorough job ensuring no stray descriptors leak. Thus, let's not create pipes if we won't use them, let's close descriptors after copying them to std{in=0,out=1,err=2} etc.

- UnixProcessGen2: conditionally create all pipes
  -  They are only needed in case when redirect is Redirect.PIPE.
     Otherwise, the descriptors are leaking.
  - Previously, we were only somewhat conditionally creating stderr.
- UnixProcess: close child pipe ends in the ctor
- UnixProcessGen2: child closes all fds except 0,1,2
  - Despite the comment in the code, dup2 doesn't close the original file
    descriptor.
  - And if the child doesn't properly close some write file descriptors, it
    might be messing with the kernel pipe logic detecting EOF. (Reportedly,
    the kernel will only report EOF on a pipe if there are no open write
    descriptors and no data.)

### DIFF
--- a/javalib/src/main/scala/java/lang/process/UnixProcessGen2.scala
+++ b/javalib/src/main/scala/java/lang/process/UnixProcessGen2.scala
@@ -391,17 +391,15 @@ private[process] object UnixProcessGen2 {
             unistd.STDERR_FILENO
           )
 
-        // No sense closing stuff either active or already closed!
-        // dup2() will close() what is not INHERITed.
-        val parentFds = new ArrayList[CInt] // No Scala Collections in javalib
-        if (null ne infds)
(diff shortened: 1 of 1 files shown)


kitbellew:
@WojciechMazur fyi

kitbellew:
don't know why this sole test timed out, before the rebase it worked well

kitbellew:
There's a chance that the hang might be caused by #4564 since now NullInput is more frequently used.

kitbellew:
yes! it finally worked.

PR #4584, opened 2025-10-27
UnixProcessXxx: refactoring (waitpid, imports)

kitbellew:
Review the last two commits only. First commit from #4585, merge after it.

### DIFF
--- a/javalib/src/main/scala/java/lang/process/UnixProcess.scala
+++ b/javalib/src/main/scala/java/lang/process/UnixProcess.scala
@@ -1,8 +1,8 @@
 package java.lang.process
 
-import java.io.FileDescriptor
+import java.io.{FileDescriptor, IOException}
 
-import scala.scalanative.libc.{signal => csig}
+import scala.scalanative.libc.{LibcExt, signal => csig}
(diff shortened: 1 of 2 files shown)


clib/src/main/scala/scala/scalanative/libc/package.scala
```diff
@@ -2,6 +2,10 @@ package scala.scalanative
 import scalanative.unsafe._
 
 package object libc {
+
+  @inline def strError(err: CInt): String = fromCString(string.strerror(err))
+  @inline def strErrNo(): String = strError(errno.errno)
```
ekrich:
```suggestion
@inline def strError(errno: CInt): String = fromCString(string.strerror(err))
@inline def strError(): String = strError(errno.errno)
```

I think this might look a little better - also they probably should have scaladoc.

We typically don't do scaladoc for libc and posix as they have their own authoritive docs although thinking about it an @see might be appropriate at the top level object unless they have web tags at the function level. Just thinking out loud about this.

kitbellew:
@WojciechMazur @ekrich ready for review

